PR #249, opened 2013-06-25
Added Assertion Flow

tsov:
Here a simple implementation for an Assertion Flow for Doorkeeper.

This lets you define a method to authenticate users coming from 3rd Party applications like facebook, twitter, etc.

**Example:**

``` ruby
Doorkeeper.configure do
    resource_owner_from_assertion do
        facebook = URI.parse('https://graph.facebook.com/me?access_token=' + params[:assertion])
        response = Net::HTTP.get_response(facebook)
        user_data = JSON.parse(response.body)
        User.find_by_facebook_id(user_data['id'])
    end
end
```

### DIFF
--- a/spec/dummy/db/schema.rb
+++ b/spec/dummy/db/schema.rb
@@ -11,56 +11,56 @@
 #
 # It's strongly recommended to check this file into your version control system.
 
-ActiveRecord::Schema.define(:version => 20140422215603) do
+ActiveRecord::Schema.define(version: 20140422215603) do
 
-  create_table "oauth_access_grants", :force => true do |t|
-    t.integer  "resource_owner_id",                 :null => false
-    t.integer  "application_id",                    :null => false
-    t.string   "token",                             :null => false
-    t.integer  "expires_in",                        :null => false
-    t.string   "redirect_uri",      :limit => 2048, :null => false
-    t.datetime "created_at",                        :null => false
-    t.datetime "revoked_at"
-    t.string   "scopes"
+  create_table 'oauth_access_grants', force: true do |t|
+    t.integer  'resource_owner_id',                 null: false
+    t.integer  'application_id',                    null: false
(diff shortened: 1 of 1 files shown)


finalight:
hi, i also have the same issue with using the latest version of doorkeeper

any solutions for this issue??

StarWar:
How to issue token for a particular app in assertion flow?

Currently I'm using facebook token to issue access_token and it works without passing client_id and client_secret.
I want to issue token only if correct client_id and client_secret is provided.

adityamajeti:
Hi, I'm trying to get access token with following api call from postman getting an error:500  ArgumentError (wrong number of arguments (given 1, expected 3)):     
http://mydomain.com/oauth/token?client_id=<client_id>&client_secret=<client_secret>&grant_type=assertion&assertion=<assertion>

So, What params ? has to pass to get access_token

adityamajeti:
I'm using   gem 'doorkeeper-grants_assertion', git: 'https://github.com/Badiapp/doorkeeper-grants_assertion'   which has dependency  of doorkeeper  >= 4.0.0   if i'm using  doorkeeper  >= 4.0.0  i'm unable to generate access_token  if i use doorkeeper  >= 4.0.0.rc1  i can generate access_token  but unable to call assertion method in doorkeeper.rb

could any one suggest me which version of doorkeeper is best   and i'm using mongodb as my db

dagumak:
@nbulaj Nothing wrong with it, but I do have questions. In this example, the authentication just passes through and does not alter the database (ex, creation of a new user, etc...): https://github.com/doorkeeper-gem/doorkeeper-grants_assertion/wiki

If I were previously using Devise and password grant and if I wanted to add Google login then how would user record creation and everything else work? My initial assumption would be to check if the user exists by google_id then email on authentication before creating a user record.

PR #294, opened 2013-10-07
Changed generation behavior on Application creation

ianunruh:
The callbacks that generate client uid and secret for the Application model will now check if one is already provided. If so, it will not generate a new uid or secret.

The motivation for this was to make creating seed data easier. Now, I would have to either skip the callbacks before validation or create the Application and then update the uid/secret with the desired values.


mmzoo:
Thank you very much, this helps a lot!

PR #337, opened 2013-12-30
Dynamic request uri

ryanjones:
Some authentication libraries need a way to pass through redirect_uri parameters (basically a wildcard of the redirect_uri).

Here's an example:
http://msdn.microsoft.com/en-us/library/microsoft.aspnet.membership.openauth.openauth.requestauthentication(v=vs.100).aspx

The RequestAuthentication method will add querystring attributes like **provider** and **sid**. 

I've done a match from the start of the URL to prevent a parameter from containing the URL. I'm not sure of the ramifications of allowing in query parameters (I couldn't find much in the OAuth 2 spec), however it doesn't seem like there would be much harm.

### DIFF
--- a/spec/lib/config_spec.rb
+++ b/spec/lib/config_spec.rb
@@ -170,20 +170,20 @@ describe Doorkeeper, "configuration" do
       }
       expect(subject.realm).to eq("Example")
     end
+  end
 
-    it 'raises an exception when configuration is not set' do
-      old_config = Doorkeeper.configuration
-      Doorkeeper.module_eval do
-        @config = nil
-      end
+  it 'raises an exception when configuration is not set' do
+    old_config = Doorkeeper.configuration
+    Doorkeeper.module_eval do
+      @config = nil
+    end
 
-      expect do
-        Doorkeeper.configuration
-      end.to raise_error Doorkeeper::MissingConfiguration
+    expect do
+      Doorkeeper.configuration
+    end.to raise_error Doorkeeper::MissingConfiguration
 
-      Doorkeeper.module_eval do
-        @config = old_config
-      end
+    Doorkeeper.module_eval do
+      @config = old_config
     end
   end
 end


tute:
> Some authentication libraries need a way to pass through redirect_uri parameters (basically a wildcard of the redirect_uri).

@ryanjones I'm starting to think that this feature was actually not needed. In https://github.com/doorkeeper-gem/doorkeeper/pull/337/files#diff-c222d58d7900f89ad63a415b8fdd5940R18 `url.query` is cleaned before validation, effectively allowing any query parameters. Clients have the `state` OAuth parameter if they need state back after the flow. Why did you need the wildcard?

ryanjones:
@tute Let me do some digging and I'll get back to you. I know for sure without the tweak it wouldn't have worked for us. /cc @mvandenbeuken

tute:
@ryanjones @mvandenbeuken I've got as next task in my roadmap to remove this feature for security reasons, I'd like to know your use case so your app doesn't break with newer doorkeeper versions. Keep in touch if interested please!

ryanjones:
Hi @tute,

I did some digging. Essentially our MS provider is using this class: http://msdn.microsoft.com/en-us/library/microsoft.aspnet.membership.openauth.openauth.requestauthentication(v=vs.100).aspx so they pass along these attributes which we need to pass back. The wildcard (forcing the domain to start with the proper url) allowed us to do that. They don't use the state parameter unfortunately.

I was reading https://github.com/homakov/oauthsecurity specifically this:
"When you managed to add %23(#) in the end of redirect_uri. It will result in sending the code in the fragment Location: http://CLIENT/callback/../open_redirect?to=evil#&code=CODE"

For us, we're a closed OAuth provider. We're the ones who make the client id/secret id. So users wouldn't be able to create an application (and set the redirect uri). I don't think you'd ever want to turn on the wildcard feature for public access. 

I suspect this will break for people interacting with .NET sites that are using the above class http://msdn.microsoft.com/en-us/library/jj891080(v=vs.111).aspx I'm not sure how big of an audience that is though. It does look like it's the main oauth lib for .NET 4.0/4.5.

Thoughts on how we can work around this? I don't really want to write a gem extension (monkey patch it).

Thanks,
Ryan

/cc @mvandenbeuken

tute:
This sounds exactly like something that ought to be solved outside of the gem. You know what you are doing, you know the specific reasons why it will be ok to do it without introducing major risks, and you know where those checks happen in doorkeeper to override them.

I don't want to add an option that seems convenient at first, and lets less experienced programmers turn it on without understanding the risks. I prefer this be solved outside. A wiki page explaining this process would be extra-killer, so the next person with a similar need gets access to all this context we are discussing across different issues.

Thank you for your time and feedback.

PR #420, opened 2014-06-24
Move hidden fields from view to pre_authorization, for supporting other parameters.

takada-s:
Currently, authorizations/new view has hardwired `hidden_field_tags`, used to pass parameters, such as scope and state, between authN and authZ.

I also trying to support OpenID Connect on this Doorkeeper, so I want to pass a new parameter `nonce`.
Using my changes, we can easily pass the new parameter and can avoid overriding/repeating code.

### DIFF
--- a/lib/doorkeeper/oauth/pre_authorization.rb
+++ b/lib/doorkeeper/oauth/pre_authorization.rb
@@ -38,11 +38,11 @@ module Doorkeeper
 
       def hidden_fields
         {
-            client_id:     client.uid,
-            redirect_uri:  redirect_uri,
-            state:         state,
-            response_type: response_type,
-            scope:         scope,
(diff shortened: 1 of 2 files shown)


toupeira:
@tute follow-up to [my comment](https://github.com/doorkeeper-gem/doorkeeper/pull/897#issuecomment-256591623) in #897: I've now implemented nonces in doorkeeper-openid_connect by [extending certain Doorkeeper classes](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/compare/feat/nonce-tracking?expand=1) (switching to `prepend` so I can use `super`).

So since I'm already monkey-patching, the current solution in this PR seems acceptable to me, and I think overriding the view would only make things more complicated for end-users of my gem. Would you be willing to merge this as is?

PR #452, opened 2014-08-07
Allow revoked_at to be set in the future, for future expiry

radar:
Hello,

I'm working with IFTTT integration on a project and they've requested that a refresh token is not revoked immediately. I have no idea how to implement this in doorkeeper without setting `revoked_at` to a future time. I noticed that the `revoked?` method merely checks if that field is set, and not if that time has yet passed. This PR attempts to rectify that behaviour.

If you know of a better way to keep refresh tokens around, I would love to hear about it.

Thanks!

### DIFF
--- a/lib/doorkeeper/models/revocable.rb
+++ b/lib/doorkeeper/models/revocable.rb
@@ -6,7 +6,7 @@ module Doorkeeper
       end
 
       def revoked?
-        revoked_at.present?
+        !!(revoked_at && revoked_at <= DateTime.now)
       end
     end
   end
--- a/spec/lib/models/revocable_spec.rb
+++ b/spec/lib/models/revocable_spec.rb
@@ -18,9 +18,14 @@ describe 'Revocable' do
   end
 
   describe :revoked? do
-    it 'is revoked if :revoked_at is set' do
-      allow(subject).to receive(:revoked_at).and_return(double)
+    it 'is revoked if :revoked_at has passed' do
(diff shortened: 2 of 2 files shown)


francesle:
@radar @tute Is there a pull request for actually allowing developers to set `revoked_at` to a future time? I'm also using this gem to integrate with IFTTT and was wondering whether there's a way to achieve that with the current version...

tute:
@francesle see the options:

* access_token_expires_in
* custom_access_token_expires_in
* authorization_code_expires_in

francesle:
I'm actually referring to the refresh token. This pull request seems to be the most relevant: https://github.com/doorkeeper-gem/doorkeeper/pull/578

tute:
Right, that's WIP right now.

voltechs:
For anyone else still struggling with this, the trick for us (we're integrating with IFTTT as well) is to ensure you are using the `previous_refresh_token`. Make sure you have this column in your `oauth_access_tokens` table.

See (line 35)
https://github.com/doorkeeper-gem/doorkeeper/blob/bf36149a82693c92d7a3b4c3a3628ae3d4094dd1/lib/doorkeeper/oauth/refresh_token_request.rb#L30-L39
and (line 44)
https://github.com/doorkeeper-gem/doorkeeper/blob/07d155590512cdfed696683ccf222fbf30d47e6c/lib/doorkeeper/orm/active_record/access_token.rb#L43-L45

Issue #317, opened 2013-11-07
Support ability to limit scopes by application


jdoconnor:
It would be great if an application (or type of application) had a set of scopes that it was allowed to grant.  For instance, an internal application (like an admin tool) would be able to include a scope like 'manage_users' that you would never want to expose to a publicly available application (like one that you would expose to customers of your api).

I've done something like this in the past (not using oauth) and was thinking about adding this to doorkeeper in the form of another table (to minimize impact on existing installations)

```
oauth_application_id
allowed_scopes (string)
```

PR #469, opened 2014-08-22
Client application specific scopes

cpapazian:
This PR attempts to address https://github.com/doorkeeper-gem/doorkeeper/issues/317. We saw @tute's comments, but could not find a great way to implement this without either putting all of the logic into the authoriizations controller (doing an end run around the work that the `OAuth::PreAuthorizations` class does, or monkeypatching Doorkeeper a bit. So I created this PR in an attempt to implement this functionality in a way that fits with the rest of the Doorkeeper API.

It works by adding a `scopes` field to the `Doorkeeper::Application` model and validating the requested scopes against it during pre-authorization. When `enable_application_scopes` is set, the `Doorkeeper::Models::Scopes` mixin is added to `Doorkeeper::Application` and `validate_scopes` will check both the server/config scopes as well as the client application's scopes.

Still TODO: adding support to the admin pages and testing with ORM's besides ActiveRecord

In the meantime, I would greatly appreciate some feedback.

thanks,
-chris


seanarnold:
@tute @cpapazian Just stumbled upon this and it's saved me so much time. I thought I had to do this manually.

Is this documented anywhere? https://github.com/doorkeeper-gem/doorkeeper/wiki/Using-Scopes Only mentions creating and using scopes, but not limiting per Application. I can write something in the Wiki about it unless it's documented somewhere else?

tute:
> I can write something in the Wiki

Please do! Thank you. :)

seanarnold:
@tute Got around to adding this https://github.com/doorkeeper-gem/doorkeeper/wiki/Using-Scopes#limiting-scopes-per-application

Feel free to make your own changes.

PR #481, opened 2014-09-16
Customize token flow OAuth expirations with a config lambda

DavidHuie:
At NationBuilder, we assign a custom expiration to access tokens derived from token-based OAuth. We do this because our authorization code tokens tend to have very high expirations, while our token-based access tokens have very short ones. This pull request solves this problem for us by enabling the user to define an expiration on each app. Since we placed this configuration option inside of a lambda, it could also dynamically customized to vary in different ways.

-David

### DIFF
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -172,27 +172,29 @@ and that your `initialize_models!` method doesn't raise any errors.\n
 
     option :resource_owner_authenticator,
            as: :authenticate_resource_owner,
-           default: (lambda do |routes|
+           default: (lambda do |_routes|
              logger.warn(I18n.translate('doorkeeper.errors.messages.resource_owner_authenticator_not_configured'))
              nil
            end)
     option :admin_authenticator,
            as: :authenticate_admin,
-           default: ->(routes) {}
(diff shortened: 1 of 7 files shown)


tute:
@DavidHuie are you interested in working through this feature? Otherwise, I'd close this PR. Less options makes for more reliable software. :)

DavidHuie:
Sorry about that, @tute. I lost track of this over the holidays. Anyway, this is ready for another look over. I moved the code that was in the module into the Authorization::Token module. Thanks for your patience.

tute:
Thanks for your work David. I wonder why Travis is red. Is it coming from master?

DavidHuie:
@tute it appears so. The error is `superclass mismatch for class AccessGrant` from `lib/doorkeeper/orm/active_record/access_grant.rb:2`. I'm guessing that there's either a naming conflict or an error there.

tute:
I re run Travis in master and it's green, failure is related to this PR.

DavidHuie:
You were right, @tute. I didn't specify `orm` inside the configure block. Travis looks good now.

tute:
thanks!

PR #506, opened 2014-11-10
Throw a 401 on revoke if no auth header is found

### LINKED ISSUE #505: oauth/revoke returns 200 when unauthorized
If I call oauth/revoke with a valid token but exclude the Authorization header:
`curl -i -F token=<token> -X POST localhost:3000/oauth/revoke`

I receive a 200 but was expecting a 401 based on the docs:
https://github.com/doorkeeper-gem/doorkeeper/wiki/API-endpoint-descriptions-and-examples
"Returns 401 if authentication fails, otherwise 200 OK, even if token doesn't exist or has already been revoked."

The token doesn't get revoked in the DB as expected.

ryanjones:
This should fix #505

@tute I wasn't sure the best way to raise a 401 in doorkeeper. I did come across `Doorkeeper::OAuth::ErrorResponse`, but I wasn't sure if you wanted me to do something like...
`render json: {}, status: Doorkeeper::OAuth::ErrorResponse.new.status and return unless doorkeeper_token`

Let me know if you have a better idea/way to do this.

### DIFF
--- a/app/controllers/doorkeeper/tokens_controller.rb
+++ b/app/controllers/doorkeeper/tokens_controller.rb
@@ -1,6 +1,5 @@
 module Doorkeeper
   class TokensController < Doorkeeper::ApplicationMetalController
-
     def create
       response = strategy.authorize
       self.headers.merge! response.headers


mehhii:
@tute 
How is it possible that the wiki had reference about the 401 code and requiring the Authorization-header for revoke?

I updated https://github.com/doorkeeper-gem/doorkeeper/wiki/API-endpoint-descriptions-and-examples according to this PR.

Please change it back if this was incorrect.

tute:
@mehhii 
Thanks for your update! The wiki is community maintained, so it tends to be a step or two behind the codebase. Thankfully there's people like you who bring it back up to date.

PR #516, opened 2014-12-09
Add `protect_from_forgery` to turn on Rails' built-in protection

phillbaker:
Since the Doorkeeper controllers inherit from Doorkeeper::Application (which inherits directly from ActionController::Base) and not ApplicationController, they never call `protect_from_forgery`, which means that non-GET methods don’t validate CSRF tokens. Thus, it’s possible for an attacker to host a form on an arbitrary URL, and if a users is logged into a site that uses Doorkeeper visits the URL, the attacker can grant access to a application on that site.

### DIFF
--- a/app/controllers/doorkeeper/application_controller.rb
+++ b/app/controllers/doorkeeper/application_controller.rb
@@ -2,6 +2,12 @@ module Doorkeeper
   class ApplicationController < ActionController::Base
     include Helpers::Controller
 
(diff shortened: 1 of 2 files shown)


svoynow:
+1 on an option to skip this, particularly for the authorize route.
I think this breaks a reasonable use case, which is using the ```client_credentials``` grant type for a non-browser API client (i.e. a mobile app) that is rendering it's own login UI.

Issue #510, opened 2014-11-24
Improve error message for expired refresh token


thomasballinger:
When an expired access token is used in a request, the `:expired` message "The access token expired" is pretty helpful, however when an expired refresh token is used to request a new access token only the generic `:invalid_request` message "The request is missing a required parameter, includes an unsupported parameter value, or is otherwise malformed." message is used.

It would be nice to provide a more specific message for this case, particularly since using single-use refresh tokens is optional in the spec. Would there be interest in such a PR?

tute:
Fixed in https://github.com/doorkeeper-gem/doorkeeper/commit/722113f with the help of @whatasunnyday and @euraldius. Thanks also to @thomasballinger for noting it too!

PR #544, opened 2015-01-07
Failing specs for refresh tokens

whatasunnyday:
I tried to get these specs to pass but hit a wall. Here's a rough idea of my code

```ruby
module Doorkeeper
  module OAuth
    class RefreshTokenRequest
      include Validations
      include OAuth::RequestConcern
      include OAuth::Helpers

      validate :token_present, error: :invalid_request
      validate :token,        error: :invalid_grant
      validate :client,       error: :invalid_client
      validate :client_match, error: :invalid_grant
      validate :scope,        error: :invalid_scope
...
      def validate_token_present
        refresh_token.present?
      end
      
      def validate_token
        validate_token_present && !refresh_token.revoked?
      end
...
     end
  end
end

```

Depending on the order of the validate specs, a different test would fail. Any idea what the issue is?

Related PR: https://github.com/doorkeeper-gem/doorkeeper/issues/510.

### DIFF
--- a/spec/lib/oauth/refresh_token_request_spec.rb
+++ b/spec/lib/oauth/refresh_token_request_spec.rb
@@ -40,7 +40,7 @@ module Doorkeeper::OAuth
     it 'rejects revoked tokens' do
       refresh_token.revoke
       subject.validate
-      expect(subject.error).to eq(:invalid_request)
+      expect(subject.error).to eq(:invalid_grant)
     end
 
     it 'accepts expired tokens' do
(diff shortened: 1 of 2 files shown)


whatasunnyday:
Also, I couldn't find the specs to make expired/revoked access tokens return invalid grant.

tute:
Thanks for your work, @whatasunnyday! With the help of @euraldius we took your tests and deployed a fix in https://github.com/doorkeeper-gem/doorkeeper/commit/722113f.

PR #548, opened 2015-01-11
Update .travis.yml

maclover7:
Make it so Travis will use the live version of Rails 4.2.0. No need for the release candidates anymore.

I don't think this needs to be included in the changelog since this isn't being shipped to end users?

@tute I'm back! :wink:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ env:
   - rails=3.2.0
   - rails=4.0.0
   - rails=4.1.0
-  - rails=4.2.0.rc2
+  - rails=4.2.0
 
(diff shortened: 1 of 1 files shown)


tute:
Hey that's good!

I tend to be verbose on the CHANGELOG, so people can trust it's the curated `git log` it's meant to be. This is an internal enough change that we can indeed omit, I have no strong opinions on this one, your take.

Will wait for Travis green flag, and merge!

maclover7:
That's my second PR and merge of the day! :smile:

tute:
Thank you twice! :clap:

PR #549, opened 2015-01-11
Bootstrap 3.3.1

maclover7:
Upgraded bootstrap.min.css to 3.3.1 from 3.1


tute:
@maclover7 I appreciate your work, but if this PR is not fixing known bugs or adding behavior I prefer not to merge it. doorkeeper is better stable than cutting edge. :)

I'd suggest you other features to work on to take advantage of your productive day, but actually I have nothing in the roadmap right now! Happy to continue working in other projects anyway, just ping me and we can collaborate.

Thanks again!

PR #554, opened 2015-01-17
Remove Metal integration link from README.md

elskwid:
Removes unused link in README since the Metal integration information was modified and mostly removed in 977057fa40c3ca125.


tute:
Thanks!

Issue #541, opened 2015-01-05
Undefined method `attr_accessible


rjrferreira:
Hi,

I have this error:
NoMethodError: undefined method `attr_accessible' for Doorkeeper::AccessGrant:Class
                        AccessGrantMixin at C:/jruby-1.7.16/lib/ruby/gems/shared/gems/doorkeeper-2.0.1/lib/doorkeeper/models/access_grant_mixin.rb:1

Gemfile.lock:

(...)
doorkeeper (2.0.1)
      railties (>= 3.1)
protected_attributes (1.0.8)
rails (4.0.8)
(...)

Any idea?

jmguedes:
I have the same problem! 
Any idea to fix this?

tute:
`protected_attributes (1.0.8)` should provide `attr_accessible`. The failure comes form: https://github.com/doorkeeper-gem/doorkeeper/blob/master/lib/doorkeeper/models/access_grant_mixin.rb#L14-L16. Do you see anything wrong with that condition? Can you provide a failing test case or a failing app in GitHub?

Thanks for reporting.

rjrferreira:
The condition works fine.
"::Rails.version.to_i" returns 4 and "defined?(::ProtectedAttributes)" returns ProtectedAttributes defined!

Unfortunately i can't provide my failing app.

tute:
I'll need an app where to reproduce the problem in order to try solutions. Thanks for reporting.

rjrferreira:
Hi Tute,

Please follow this link:
https://github.com/rjrferreira/MyProject/tree/master/doorkeeper_test

tute:
Thanks, can reproduce it there. Will leave open until it is fixed.

tute:
For some reason in mongoid `protected_attributes` gem is not extending the models. If we add `include ActiveModel::MassAssignmentSecurity` in doorkeeper models/mixins, then it works (https://github.com/doorkeeper-gem/doorkeeper/blob/master/lib/doorkeeper/models/access_grant_mixin.rb#L14-L16).

I'd like to know, though, when exactly do we need to add it, and if we need to add it here in some cases or it's rather a PR for protected_attributes. cc @sgrif @rafaelfranca

rjrferreira:
Thanks Tute!

andremfaria:
I was with the same problem @tute, and your answer solved my problem! thanks

rafaelfranca:
[...]

PR #555, opened 2015-01-17
Use attr_accessible only if object responds to attr_accessible

knovoselic:
Fixed problem described in #541.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 2.2.0 (unreleased)
 
 - Remove `wildcard_redirect_url` option
-
+- [#541] Fixed `undefined method attr_accessible` problem on Rails4
 
 ## 2.1.0
 
--- a/lib/doorkeeper/models/access_grant_mixin.rb
+++ b/lib/doorkeeper/models/access_grant_mixin.rb
@@ -11,7 +11,7 @@ module Doorkeeper
     included do
       belongs_to :application, class_name: 'Doorkeeper::Application', inverse_of: :access_grants
 
-      if ::Rails.version.to_i < 4 || defined?(::ProtectedAttributes)
+      if ::Rails.version.to_i < 4 || respond_to?(:attr_accessible)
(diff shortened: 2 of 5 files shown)


tute:
Now I remember why this solution doesn't work, see failure:

```
/home/travis/build/doorkeeper-gem/doorkeeper/vendor/bundle/ruby/2.0.0/gems/activemodel-4.0.13/lib/active_model/deprecated_mass_assignment_security.rb:14:in `attr_accessible': `attr_accessible` is extracted out of Rails into a gem. Please use new recommended protection model for params(strong_parameters) or add `protected_attributes` to your Gemfile to use old one. (RuntimeError)
```

https://travis-ci.org/doorkeeper-gem/doorkeeper/jobs/47352188

Thank you for your work anyway.

knovoselic:
Hey @tute , I've pushed another commit which should fix this problem. Do I need to create another pull request or is this one going to get updated?
Can I somehow run tests using Travis on my forked repo?

tute:
Travis running, sorry I closed it too early. :)

tute:
This change though means that mongo ORMs won't have `attr_accessible`, as few people will see they need to add `include ActiveModel::MassAssignmentSecurity` to their models.

What I would love is not need parameters protection in the codebase, remove this lines, and anywhere we call `save` or the `update` flavors, inline each attribute to make sure this vulnerability doesn't exist.

That would be a refactor, I'm not sure if it would be wiser future-thinking anyways.

knovoselic:
Yeah, they won't have `attr_access`, but they will still be protected with strong parameters. Maybe a note in readme or a warning is enough? 
Another solution would be to manually include `ActiveModel::MassAssignmentSecurity` if it's defined in doorkeeper mongoid models. What do you think about that?

The best solution would be refactor, but that could take a lot of time and effort.

tute:
> Another solution would be to manually include `ActiveModel::MassAssignmentSecurity` if it's defined in mongo models. What do you think about that?

That's better.

PR #568, opened 2015-02-02
TokensController: Memoize strategy.authorize_response result

pjambet:
This allows easy access to the authorize response in a subclass of Doorkeeper::TokensController.
Without this change, I had to user the following code to access the authorize response:

```ruby
strategy.request.instance_variable_get('@response')
```

If you accept this PR, I'll update the commit to include the CHANGELOG entry, I simply didn't want to add that if you think this PR should never get merged.
I'm not 100% satisfied with the spec I added, but at least it gets the job done (It's red without the change, and green with the change), let me know if you want to improve it.

Another alternative I had in mind was to return `response` from `TokensController#create`, but I thought memoizing the result was a better solution.

Let me know if you have any questions.

### DIFF
--- a/app/controllers/doorkeeper/tokens_controller.rb
+++ b/app/controllers/doorkeeper/tokens_controller.rb
@@ -1,7 +1,7 @@
 module Doorkeeper
   class TokensController < Doorkeeper::ApplicationMetalController
     def create
-      response = strategy.authorize
+      response = authorize_response
       self.headers.merge! response.headers
       self.response_body = response.body.to_json
       self.status        = response.status
@@ -36,6 +36,10 @@ module Doorkeeper
 
     def strategy
       @strategy ||= server.token_request params[:grant_type]
+    end
+
+    def authorize_response
+      @authorize_response ||= strategy.authorize
     end
   end
 end
--- a/spec/controllers/tokens_controller_spec.rb
+++ b/spec/controllers/tokens_controller_spec.rb
@@ -40,4 +40,18 @@ describe Doorkeeper::TokensController do
       expect(response.status).to eq 200
     end
   end
+
+  describe 'authorize response memoization' do
+    it "memoizes the result of the authorization" do
+      strategy = double(:strategy, authorize: true)
+      expect(strategy).to receive(:authorize).once
+      allow(controller).to receive(:strategy) { strategy }
+
+      controller.stub(:create) do
+        controller.send :authorize_response
+        controller.send :authorize_response
+      end
+      post :create
+    end
+  end
 end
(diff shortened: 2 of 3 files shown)


app/controllers/doorkeeper/tokens_controller.rb
```diff
@@ -37,5 +37,9 @@ def revoke_token(token)
     def strategy
       @strategy ||= server.token_request params[:grant_type]
     end
+
+    def authorize_response
+      @authorize_response ||= strategy.authorize
```
tute: We don't need the `@authorize_response ||= ` memoization part.

spec/controllers/tokens_controller_spec.rb
```diff
@@ -40,4 +40,18 @@
...
+      expect(strategy).to receive(:authorize).once
+      allow(controller).to receive(:strategy) { strategy }
+
+      controller.stub(:create) do
+        controller.send :authorize_response
+        controller.send :authorize_response
+      end
+      post :create
+    end
+  end
```
tute: And if we don't memoize, we don't need this spec.

tute:
I'm happy with this extract method, if it makes working with doorkeeper easier for you. Your child controller could get the response calling the method: `authorize_response`. I'm fine with it not being memoized.

Please explain in the commit message and CHANGELOG that this allows for subclasses to get the response object to do whatever they need, with a concrete example if you have, and I'll merge.

Thanks!

app/controllers/doorkeeper/tokens_controller.rb
pjambet: Actually this is the main reason I sent this pull request, because `strategy.authorize` is not idempotent, if in a subclass of `TokensController` I need the response, I cannot call `strategy.authorize` again, therefore the need for the memoization.

tute: Got it. Please explain this in the commit message too. :)

tute: And when we have the CHANGELOG updated and a good commit message, I'll merge.

pjambet: Awesome, going to update this to make it more explicit, and add the CHANGELOG entry, that should be ready by tomorrow.

Thanks!

pjambet:
I updated the commit message and added a CHANGELOG entry. I tried to keep the CHANGELOG entry pretty short as the commit contains all the details.

Let me know if you want to change anything before merging this.

tute:
That's great, thank you!

pjambet:
Thanks for merging this! And kudos for the gem, it's been a pleasure to work with so far.

PR #571, opened 2015-02-04
Fix relation "oauth_applications" does not exist

dwilkie:
This issue typically occurs on a CI when `bundle exec rake db:schema:load` is executed. To reproduce it locally run:

```
bundle exec rake db:drop && bundle exec rake db:create && bundle exec rake db:schema:load
```

The error happens because the initializer `doorkeeper.rb` is run before the table has been created.


skorth:
Maybe same as https://github.com/doorkeeper-gem/doorkeeper/pull/572

tute:
Thank you!

PR #572, opened 2015-02-04
Rescue ActiveRecord::StatementInvalid on migration

skorth:
- configuration runs during initialization which causes generator to fail
- seems this rescue was removed during this commit c0e25987efda8ea38c9351f4cb12a120e15f1b48


tute:
@skorth can you confirm https://github.com/doorkeeper-gem/doorkeeper/pull/571 solves the same problem? In that case, I'd prefer to fix it with 571's conditional rather than a rescue.

skorth:
I can confirm. I tested both generators (install/migrations) on a Rails 3.2.x with #571.

tute:
Merge that PR, thank you very much!

PR #574, opened 2015-02-09
Remove unused update authorization route.

kaleworsley:
I may be missing something but it looks like there is no update action on the authorizations controller.

https://github.com/doorkeeper-gem/doorkeeper/blob/master/app/controllers/doorkeeper/authorizations_controller.rb


tute:
Thank you!

PR #575, opened 2015-02-10
Allow Refresh tokens to not revoke immediately on use

rsharrott:
This PR allows refresh tokens to not either revoke at some point in the future (if you want to allow some wiggle room for network failures, etc) set refresh_token_revoked_in to some number of seconds in the future.

Also allows for refresh tokens to not be revoked until an access token created with that refresh token is successfully used once. Set refresh_token_revoked_on_use to true.

Both are off by default.

I made this work with ActiveRecord and it is passing test but no work was done for Mongoid.


tute:
It's looking good. Can you please [rebase on top of latest master and squash your commits together](http://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history)?

Will need to take care of other ORMs before merge too.

rsharrott:
Opened a rebased PR here: https://github.com/doorkeeper-gem/doorkeeper/pull/578

PR #577, opened 2015-02-10
Change scopes for absent application/client to restrict to empty set

gb-timeplay:
Currently, if no application is authenticated (via the client_id/client_secret), the client is allowed to access any set of scopes defined for Doorkeeper.

I believe this is undesirable, and it would be better if no scopes were available if no application is authorized, since the available scopes are defined per application.

### DIFF
--- a/lib/doorkeeper/oauth/helpers/scope_checker.rb
+++ b/lib/doorkeeper/oauth/helpers/scope_checker.rb
@@ -2,16 +2,26 @@ module Doorkeeper
   module OAuth
     module Helpers
       module ScopeChecker
-        def self.valid?(scope_str, server_scopes, application_scopes = nil)
-          valid_scopes = if application_scopes.present?
-                           server_scopes & application_scopes
-                         else
-                           server_scopes
-                         end
+        def self.valid?(scope_str, scopes_source, application = nil)
+          if scope_str.present? && scope_str !~ /[\n|\r|\t]/
+            valid_scopes = if application
+                             if application.scopes.present?
+                               scopes_source.scopes & application.scopes
+                             else
+                               scopes_source.scopes
+                             end
+                           else
+                             if scopes_source.respond_to? :default_scopes
+                               scopes_source.default_scopes
+                             else
+                               scopes_source.scopes
+                             end
+                           end
 
-          scope_str.present? &&
-            scope_str !~ /[\n|\r|\t]/ &&
             valid_scopes.has_scopes?(OAuth::Scopes.from_string(scope_str))
+          else
+            false
+          end
         end
       end
     end
--- a/lib/doorkeeper/oauth/refresh_token_request.rb
+++ b/lib/doorkeeper/oauth/refresh_token_request.rb
@@ -67,7 +67,11 @@ module Doorkeeper
 
       def validate_scope
         if @original_scopes.present?
-          ScopeChecker.valid?(@original_scopes, refresh_token.scopes)
+          ScopeChecker.valid?(
+            @original_scopes,
+            refresh_token,
+            client
+          )
         else
           true
         end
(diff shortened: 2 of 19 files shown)


tute:
Sounds good to me. We need a regression test for this.
Thanks for your work!

tute:
We also need green tests: https://travis-ci.org/doorkeeper-gem/doorkeeper/builds/50280910

lib/doorkeeper/oauth/helpers/scope_checker.rb
```diff
@@ -6,7 +6,7 @@ def self.valid?(scope_str, server_scopes, application_scopes = nil)
           valid_scopes = if application_scopes.present?
                            server_scopes & application_scopes
                          else
-                           server_scopes
+                           OAuth::Scopes.new # Empty set
```
tute: We want existing applications with empty scopes to take into account only `server_scopes`, but this wouldn't allow that.

lib/doorkeeper/oauth/refresh_token_request.rb
```diff
@@ -67,7 +67,11 @@ def validate_client_match
 
       def validate_scope
         if @original_scopes.present?
-          ScopeChecker.valid?(@original_scopes, refresh_token.scopes)
+          ScopeChecker.valid?(
+              @original_scopes,
+              refresh_token.scopes,
+              @client.try(:scopes)
```
tute: You can use the `client` getter (no need of the at sign).
Also, indentation is weird (two additional spaces).

tute:
We have other places where we'd need this check:

* `AccessTokenMixin#scopes_match?`
* `PasswordAccessTokenRequest#validate_scopes`
* `ClientCredentialsRequest#validate_scopes`

This implies we'll need some refactor where this logic is written only once. Also, applications with empty scopes should be assumed to allow any scope.

tute:
ping @gb-timeplay

tute:
Thanks for your early work @gb-timeplay, let's not let it get stale!

gb-timeplay:
So I have a set of changes that I think is right. The only questionable part, to my mind, is that if you are re-using tokens and you request a token with a subset of the scopes of the original, the original will be returned including the extra scopes. This doesn't seem too bad to me, though.

tute:
We'll need to merge conflicts. Also, code is not ready for master yet, it's more complex than it should be.

Question: is this an issue anyway? If you'd like the list to be empty when there's no authenticate client in your provider, you can define empty lists for `default_scopes` and `optional_scopes`. Please correct me if I'm missing anything, otherwise in two weeks I'll close this PR as "not a problem".

Thanks!

gb-timeplay:
If you set empty lists for default_scopes and optional_scopes, then surely all applications will be reduced to the empty set of scopes? I'm not sure if something has changed recently, but I seem to recall the scope validator checking that the requested scopes were in the intersection of application scopes and server scopes.

tute:
That's right, my suggestion wouldn't work that way.

gb-timeplay:
OK, all I want is for an application/client to have be authenticated in order be able to access any optional scopes associated with it.

tute:
@jasl I think this might be an important fix to incorporate in doorkeeper, but the PR should we reworked. Would you like to take it?

jasl:
no problem.

jasl:
still have 3 cases not passed, need time to resolve.

jasl:
the PR broke a behaviour that the `Application` can specify scopes(but in our dashboard isn't show that yet), when `AccessToken` has more scopes than `Application`, the scopes of `AccessToken` will be set to the intersection of two scopes.

I think scopes of `AccessToken` should equal to `Application`'s, but the work should be done in generating AccessToken.

jasl:
And `AccessToken` should have at least 1 scope.

jasl:
I PR my work, you can review that @tute

tute:
Closing in favor of @jasl's PR.

PR #578, opened 2015-02-12

rsharrott:
This PR allows refresh tokens to not either revoke at some point in the future (if you want to allow some wiggle room for network failures, etc) set refresh_token_revoked_in to some number of seconds in the future.

Also allows for refresh tokens to not be revoked until an access token created with that refresh token is successfully used once. Set refresh_token_revoked_on_use to true.

Both are off by default.

I made this work with ActiveRecord and it is passing test but no work was done for Mongoid.

### DIFF
--- a/lib/doorkeeper/models/concerns/revocable.rb
+++ b/lib/doorkeeper/models/concerns/revocable.rb
@@ -3,10 +3,29 @@ module Doorkeeper
     module Revocable
       def revoke(clock = DateTime)
         update_attribute :revoked_at, clock.now
+        Doorkeeper.configuration.refresh_token_revoked_on_use
+      end
+
+      def revoke_in(time)
+        update_attribute :revoked_at, (DateTime.now + time)
       end
 
       def revoked?
         !!(revoked_at && revoked_at <= DateTime.now)
+      end
+
+      def revoke_previous_refresh_token!
+        return if previous_refresh_token.nil?
+        old_refresh_token = AccessToken.by_refresh_token(previous_refresh_token)
+        old_refresh_token = nil if old_refresh_token.nil? or old_refresh_token.revoked?
+        if old_refresh_token
+          if Doorkeeper.configuration.refresh_token_revoked_in
+            old_refresh_token.revoke
+          else
+            old_refresh_token.revoke_in(Doorkeeper.configuration.refresh_token_revoked_in)
+          end
+        end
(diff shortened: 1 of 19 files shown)


tute:
Thanks @rsharrott! We'll need some massaging of code and git history. Regarding the commit message (and history rewriting, in general), you can always force push the branch after doing this. This will include the discussion from the PR in sync with what you are updating, which is nice, no need to open new PRs.

Your PR title and description are good, the commit message should be similar! We could amend the commit to something along the following lines:

```
Refresh Token revocation adjustments

This change allows refresh tokens to be revoke at some point in the future (for if you want to allow room for network failures for example). The option `refresh_token_revoked_in` can be set to a number of seconds in the future.

The change allows a refresh tokens to not be revoked until an access token is successfully created through it. The option `refresh_token_revoked_on_use` can be set to `true` for this behavior to take effect.

Both options are off by default.
```

Will do a code review later. In general, taking into account @houndci comments is a :+1:.

Thanks for your good work!

lib/doorkeeper/models/concerns/revocable.rb
```diff
@@ -3,11 +3,30 @@ module Models
...
+        old_refresh_token = nil if old_refresh_token.nil? or old_refresh_token.revoked?
+        if old_refresh_token
+          if Doorkeeper.configuration.refresh_token_revoked_in
+            old_refresh_token.revoke
+          else
+            old_refresh_token.revoke_in(Doorkeeper.configuration.refresh_token_revoked_in)
+          end
+        end
+        update_attribute :previous_refresh_token, nil
+      end
```
tute: Suggested rewrite of this method:

```ruby
  old_refresh_token = AccessToken.by_refresh_token(previous_refresh_token)
    if configuration.refresh_token_revoked_in
      old_refresh_token.revoke
    else
      old_refresh_token.revoke_in(configuration.refresh_token_revoked_in)
    end
  end
  update_attribute :previous_refresh_token, nil
end

def configuration
  Doorkeeper.configuration
end
```

tute:
ping @rsharrott

tute:
ping @rsharrott?

tute:
Thanks for your early work @rsharrott, let's not let it get stale!

I'll close it for now for lack of activity, but I'll be happy to reopen for working on it again.

PR #582, opened 2015-02-23
Make the Authorization header case insensitive

delba:
Some oauth2 libraries are writing Authorization headers field names in lowercase and thus can't be used with applications using doorkeeper.

For instance, the token won't be extracted:
```
Authorization: bearer 4f7781e23a7695865bce77232fed0a0e5b3b20d9cd9c700a6b48a5ee05b4fa94
```

Moreover, HTTP header field names should be case-insensitive according to the [specs](http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - [#568] TokensController: Memoize strategy.authorize_response result to enable
     subclasses to use the response object.
 - [#571] Fix database initialization issues in some configurations.
+- [#582] The Authorization header fields are now case insensitive.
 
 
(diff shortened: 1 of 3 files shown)


tute:
Great PR, thanks! Can you please add the description of the PR into the commit message? If you can add also an entry to the Changelog it will be perfect.

Thank you!

delba:
@tute Done!

tute:
Thank you!

delba:
You're welcome! Thanks for the gem :v:

PR #586, opened 2015-02-26
Rails before 4 with strong_parameters gem

crabonature:
That check can be handy in situation if someone for example got Rails 3.2 app with strong_parameters gem to not generate attr_accessible in models. What you think?

### DIFF
--- a/lib/doorkeeper/models/access_grant_mixin.rb
+++ b/lib/doorkeeper/models/access_grant_mixin.rb
@@ -11,7 +11,7 @@ module Doorkeeper
     included do
       belongs_to :application, class_name: 'Doorkeeper::Application', inverse_of: :access_grants
 
-      if ::Rails.version.to_i < 4 || defined?(::ProtectedAttributes)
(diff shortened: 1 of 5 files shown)


tute:
Hey @crabonature, thanks for your work. This condition has proven to be particularly brittle.

Please check what I said in 2013 about the new condition:
https://github.com/doorkeeper-gem/doorkeeper/commit/3cf72ef0#commitcomment-3987063

If we decided that it's a good idea, the repeated logic that's becoming a bit more complex should be extracted to its own method. But I'm afraid when we fix one problem another will unveil. For the corner cases in which current condition is an issue, what would be a good workaround for an application? I'd rather see a wiki page with the solution than trying to fix all cases with conditionals.

Again, thanks for your work!

crabonature:
Ouch, my fault that didn't look at history before I've made changes. So I think there could be no need to merge this one.

tute:
Thanks for your time and input!

PR #588, opened 2015-02-27
Fixes scopes_match? bug that skipped authorization form in some cases

akkolvt:
There are cases when `AccessToken.find_or_create_for` doesn't work properly and authorization form is not shown to the user, even though scopes have been changed.

If a user has already been authorized once with say `public write` scopes and another authorization request was sent with `public` or `public email` scopes, then `scopes_match?` correctly determines that these scopes do not match and authorization form appears.

However, if new authorization request has say `public write email` scopes, then `scopes_match?` returns `true` because new scopes array contains previous one and in this case authorization form is skipped.


tute:
Thank you! Merged in https://github.com/doorkeeper-gem/doorkeeper/commit/03008f59878f03c4387f7c7f96df7fa9facb962b.

PR #592, opened 2015-03-05
zh-TW locales

hothero:


tute:
Thank you! Instead of putting all locales into main doorkeeper though, I'll create a `doorkeeper-i18n` repo à la https://github.com/tigrish/devise-i18n. If you'd like to start it I'll greatly appreciate it, otherwise I'll do in the upcoming weeks.
Thanks again!

tute:
Thank you @hothero! Created https://github.com/doorkeeper-gem/doorkeeper-i18n, and merged your commit there. You can install the gem and configure the locale in your app now. :)

PR #599, opened 2015-03-10
Included MassAssignmentSecurity in mixins if ::ProtectedAttributes is defined

knovoselic:
Fixed problem described in #541.
This is alternative approach to #555 .

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,11 @@ User-visible changes worth mentioning.
 
 ---
 
+## (unreleased)
+
+- [#541] Fixed `undefined method attr_accessible` problem on Rails 4
+    (happens only when ProtectedAttributes gem is used)
+
 ## 2.2.1
 
 - [#636] `custom_access_token_expires_in` bugfixes
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,7 +7,7 @@ User-visible changes worth mentioning.
 ## (unreleased)
 
 - [#541] Fixed `undefined method attr_accessible` problem on Rails 4
-    (happens only when ProtectedAttributes gem is used)
+    (happens only when ProtectedAttributes gem is used) in #599
 
 ## 2.2.1
 
--- a/lib/doorkeeper/models/access_grant_mixin.rb
+++ b/lib/doorkeeper/models/access_grant_mixin.rb
@@ -7,6 +7,7 @@ module Doorkeeper
     include Models::Revocable
     include Models::Accessible
     include Models::Scopes
+    include ActiveModel::MassAssignmentSecurity if defined?(::ProtectedAttributes)
 
     included do
       belongs_to :application, class_name: 'Doorkeeper::Application', inverse_of: :access_grants
--- a/lib/doorkeeper/models/access_grant_mixin.rb
+++ b/lib/doorkeeper/models/access_grant_mixin.rb
@@ -12,7 +12,7 @@ module Doorkeeper
     included do
       belongs_to :application, class_name: 'Doorkeeper::Application', inverse_of: :access_grants
 
-      if ::Rails.version.to_i < 4 || defined?(::ProtectedAttributes)
+      if respond_to?(:attr_accessible)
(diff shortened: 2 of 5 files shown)


knovoselic:
@tute What do you think about this one?

tute:
This looks ok to me. An alternative I like better is the one I quote from https://github.com/doorkeeper-gem/doorkeeper/pull/555#issuecomment-70371765:

> What I would love is not need parameters protection in the codebase, remove this lines, and anywhere we call `save` or the `update` flavors, inline each attribute to make sure this vulnerability doesn't exist.

Will think of this in two weeks, when I assign some workweek time for doorkeeper. Thanks for your input!

tute:
I think the condition we want here is that `ProtectedAttributes` is defined, and also the module is not included yet. Otherwise we'll include it twice in ActiveRecord.

tute:
This looks good to merge. Three steps to do before:

1. Can you please improve your commit message, with context on the three related issues/PRs, before?
2. Can you update the CHANGELOG?
3. Can you rebase on top of latest master?

Thank you!

knovoselic:
Hey, unfortunately I don't have enough time to work on it, should be able in two weeks or so. I'll let you know when it's done. Thanks.

tute:
ping @knovoselic. Thanks again for your good contribution!

tute:
@knovoselic after your change, we can assume in https://github.com/doorkeeper-gem/doorkeeper/blob/master/lib/doorkeeper/models/access_grant_mixin.rb#L14 that if it responds to `attr_accessible` it can safely be called, simplifying the condition. Would you update those, please?

tute:
ping @knovoselic. Thank you!

knovoselic:
@tute Sorry for prolonging this so much, I'll do it 100% next weekend.

tute:
> I'll do it 100% next weekend.

That's this weekend! :stuck_out_tongue:

knovoselic:
Yep, I'll work on it today :) Thanks for being patient.

knovoselic:
@tute Hey, I think I've done everything you've asked me to do. Any idea why this build has failed?
https://travis-ci.org/doorkeeper-gem/doorkeeper/builds/61950209
The next one has passed, and I haven't updated anything related to the code that has failed...

tute:
Second commit ("Simplified condition...") shouldn't be there, as it fails. See https://github.com/doorkeeper-gem/doorkeeper/pull/555#issuecomment-70370978. After you remove that, I'll merge. Thank you! :smiley:

tute:
Regarding the spec, it's a flappy test, I'm working on it in the "extract ORMs" PR. Thank you!

NEWS.md
```diff
@@ -4,6 +4,11 @@ User-visible changes worth mentioning.
 
 ---
 
+## (unreleased)
+
+- [#541] Fixed `undefined method attr_accessible` problem on Rails 4
```
tute: Let's add #599 too to the NEWS.

knovoselic:
@tute  I've added #599 to the news. I'm a little confused by this message:
> Second commit ("Simplified condition...") shouldn't be there, as it fails. See #555 (comment). After you remove that, I'll merge. Thank you! :smiley:

I do not see it failing, so I'm not sure if there's something that's not covered by the tests? If there is, what were you thinking about when you've suggested that those lines can be simplified?

tute:
> I do not see it failing, so I'm not sure if there's something that's not covered by the tests? If there is, what were you thinking about when you've suggested that those lines can be simplified?

You are totally right, we don't care any longer because we don't test against end-of-life'd Rails 4 (where that situation happened).
Thanks!

knovoselic:
Thanks :) When do you think it will be released?

tute:
Past week I released 2.2.1; I'd wait at least two weeks and two patches before releasing next version. Meanwhile you can point your Gemfile to this specific merge commit. Thanks!

knovoselic:
Will do, thanks!

PR #604, opened 2015-03-18
Add option to specific access token's complexity

jamesdphillips:
I may be daft, however, while I did notice that UniqueToken#generate's method signature includes options to change a token's complexity I couldn't find a nice way to configure it.

For your consideration, I've added a configuration option to change the token's complexity.


tute:
Hi @jamesdphillips, thanks for your work! Rather than a configuration, I'd like to take the "best" decision for users, if such a general enough definition exists. Is there a reason why the default behavior is not good enough? If so, how can we improve it, always and for everyone?

Thanks again!

jamesdphillips:
> Is there a reason why the default behavior is not good enough?

Sorry, I think I was in part confused and thought that SecureRandom's `#hex` method's size argument meant that it returned a random string of `n bits` not bytes. In that case I think the base configuration is quite sufficient. ;)

tute:
Thanks for your review! :)

PR #606, opened 2015-03-24
Add custom access token expiration support for Client Credentials flow

jzw:
8d17a7a introduced a configuration option for a custom application-specific
access token expiration, however this option did not work with the Client
Credentials flow.

Modify Doorkeeper::OAuth::ClientCredentials::Issuer class so that it utilizes
the Doorkeeper.configuration.custom_access_token_expires_in option when
present.


tute:
Thank you!

PR #609, opened 2015-03-25
Upgrade Rails to the 4.2.0 release from RC2

maclover7:
No need to add anything to the Changelog, right? Does not affect end users at all.


tute:
Thank you! No need indeed.

Issue #598, opened 2015-03-06
JWT Support


chriswarren:
Less an issue and more a question, I wanted to check on how you'd prefer JWT support be implemented, if at all.

I'm looking in to adding support for [JWT](http://jwt.io) support to Doorkeeper, and haven't found anyone discussing it. My questions are A) do you know of anyone already doing this and B) if not, would you want this to be part of Doorkeeper or an extension like the assertion granting?

I've started on a fork of doorkeeper to add it in, but can move it elsewhere if that's preferred.

tute:
Why should it be related with doorkeeper? Is it related with OAuth 2? Would JWT share relevant pieces of code with doorkeeper?

chriswarren:
Spec is at http://self-issued.info/docs/draft-ietf-oauth-jwt-bearer.html as well.

I'm looking at creating a flow like [Zendesks's SSO](https://support.zendesk.com/hc/en-us/articles/203663816-Setting-up-single-sign-on-with-JWT-JSON-Web-Token-).

tute:
I see. It should be an extension, like the assertion grants feature. And it will need a maintainer, which probably is you! :)

tute:
Closing for now, let me know if I can help you in any way. Thanks!

PR #610, opened 2015-03-25
JWT Support

chriswarren:
We talked about making this an extension in #598 but I wanted to submit it as a PR and get your thoughts before attempting to extract it, both to check that I'm not missing anything and in case seeing it written out changes your perspective on having it in the gem directly.

Adds support for JWT tokens, set up through the regular Doorkeeper config process. Supports custom payloads and secrets and multiple encryption types, all disabled by default.

Use case for JWT tokens is allowing applications to identify tokens provided by the Doorkeeper app and communicate amongst themselves without having to check in with the central application. If you know the secret used to encode the token you can decode it and access identifying information for the user.

### DIFF
--- a/doorkeeper.gemspec
+++ b/doorkeeper.gemspec
@@ -17,6 +17,7 @@ Gem::Specification.new do |s|
   s.require_paths = ["lib"]
 
   s.add_dependency "railties", ">= 3.2"
+  s.add_dependency "jwt", "~> 1.3.0"
 
   s.add_development_dependency "sqlite3", "~> 1.3.5"
   s.add_development_dependency "rspec-rails", "~> 2.99.0"
--- a/lib/doorkeeper.rb
+++ b/lib/doorkeeper.rb
@@ -52,6 +52,8 @@ require 'doorkeeper/orm/mongoid2'
 require 'doorkeeper/orm/mongoid3'
 require 'doorkeeper/orm/mongoid4'
 
+require 'jwt'
+
 module Doorkeeper
   def self.configured?
     @config.present?
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -108,6 +108,18 @@ and that your `initialize_models!` method doesn't raise any errors.\n
       def force_ssl_in_redirect_uri(boolean)
         @config.instance_variable_set("@force_ssl_in_redirect_uri", boolean)
       end
+
+      def use_jwt_token
+        @config.instance_variable_set('@jwt_token_enabled', true)
+      end
+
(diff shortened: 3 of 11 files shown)


tute:
Thanks for your work @chriswarren. At first glance code looked good, didn't find anything obvious to comment on. That is, without having enough context on JWT tokens though!

I can't reliably code-review and maintain this code because I don't use JWT tokens. Even if I code, it seems to me independent enough to deserve it's own project.

How can I make doorkeeper easier for you to extend it from another gem?

chriswarren:
My biggest concern/hurdle in trying to extract this so far has been figuring out the best approach to setting up the test infrastructure, given the necessary dependencies. The rest of the process pretty easy -  just aliasing, monkey patching, or adding the already written methods.

I'd gladly help maintain the JWT component within Doorkeeper if you want to leave it as part of the gem. Otherwise, any guidance or suggestions on testing this in the context of an extension would be appreciated. I'll keep at it and see how it goes.

tute:
I'm very much looking forward for getting that done cause that's what prevents me from having real extensions indeed. I didn't do it yet in a good part because I'm not sure how to solve it. Do you know of any projects that have done it? Do you imagine a solution that might work for doorkeeper?

Thanks for your input!

chriswarren:
I don't have any ideas for a more generic way of doing extensions yet, but I think we could solve this specific situation by adding a Doorkeeper config that'd allow the ability to declare how the token is generated.

```
Doorkeeper.configure do
  token_generator Doorkeeper::JWT
  # token_generator :default
end
```

This could  take a module that just has to respond to `generate_token`, so I'd add that plus the few other methods I added to `access_token_mixin` and when an Access Token is created it'd call that method instead.

Then I could put my configuration in to a `Doorkeeper::JWT.configure` block in either the Doorkeeper initializer or in an initializer of its own, keeping things separate.

This approach, I think, would let a doorkeeper-jwt gem live without having to include anything from Doorkeeper at all, and make testing pretty simple to do, since the only public interfaces to test are that configurations are set properly and that a JWT token is generated according to them. Doorkeeper would do everything else and just need to be changed to let the generator of the access token be configurable.

tute:
@chriswarren your description makes perfect sense. After defining the interface between doorkeeper and its extensions, testing on the extension gets simpler. As it should be. :)

I'm very happy about that. I won't be able to add the extension myself for a few weeks, but I'll gladly code review and merge it if you get to it before me.

Many thanks!

chriswarren:
Closing in favor of https://github.com/doorkeeper-gem/doorkeeper/pull/611 and https://github.com/chriswarren/doorkeeper-jwt.

PR #611, opened 2015-03-27
Allow custom access token generators to be used

chriswarren:
In response to #610, allows for custom access token generators to be used. Once this is in I'll move the functionality from the PR over to a separate gem.

### DIFF
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -204,7 +204,8 @@ and that your `initialize_models!` method doesn't raise any errors.\n
     option :realm,                          default: 'Doorkeeper'
     option :force_ssl_in_redirect_uri,      default: !Rails.env.development?
     option :grant_flows,                    default: %w(authorization_code client_credentials)
-    option :access_token_generator, default: "UniqueToken"
+    option :access_token_generator,
+      default: "Doorkeeper::OAuth::Helpers::UniqueToken"
 
     attr_reader :reuse_access_token
 
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -204,8 +204,7 @@ and that your `initialize_models!` method doesn't raise any errors.\n
     option :realm,                          default: 'Doorkeeper'
     option :force_ssl_in_redirect_uri,      default: !Rails.env.development?
     option :grant_flows,                    default: %w(authorization_code client_credentials)
-    option :access_token_generator,
-    default: "Doorkeeper::OAuth::Helpers::UniqueToken"
+    option :access_token_generator,         default: "Doorkeeper::OAuth::Helpers::UniqueToken"
 
     attr_reader :reuse_access_token
 
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -110,7 +110,9 @@ and that your `initialize_models!` method doesn't raise any errors.\n
       end
 
       def access_token_generator(access_token_generator)
-        @config.instance_variable_set('@access_token_generator', access_token_generator)
+        @config.instance_variable_set(
+          '@access_token_generator', access_token_generator
+        )
       end
     end
 
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -108,6 +108,10 @@ and that your `initialize_models!` method doesn't raise any errors.\n
       def force_ssl_in_redirect_uri(boolean)
         @config.instance_variable_set("@force_ssl_in_redirect_uri", boolean)
       end
+
+      def access_token_generator(access_token_generator)
+        @config.instance_variable_set('@access_token_generator', access_token_generator)
+      end
     end
 
     module Option
@@ -198,6 +202,7 @@ and that your `initialize_models!` method doesn't raise any errors.\n
     option :realm,                          default: 'Doorkeeper'
     option :force_ssl_in_redirect_uri,      default: !Rails.env.development?
     option :grant_flows,                    default: %w(authorization_code client_credentials)
+    option :access_token_generator, default: "UniqueToken"
 
     attr_reader :reuse_access_token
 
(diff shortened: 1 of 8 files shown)


chriswarren:
I'm working on my doorkeeper-jwt gem now and testing it to ensure that everything works end to end. Please wait to accept this until I've finished that up (next week most likely) so I can be sure that there's nothing else that needs to be done.

alfredw:
Hey Chris, if you need any help on this I would like to help as I'm starting a new project and would love to use this.

chriswarren:
@alfredw Thanks for the offer. Hopefully I'll have something out early next week; it'd be great to have someone else using it right away to help make sure I didn't miss anything. I'll push the unfinished gem up on Monday so you can take a peek.

@tute The problem I've run in to at this point is that I don't have access to `env` in my gem's config to get at Devise/Clearance/etc to get the current user's info, which would/should be used to populate the JWT payload. I'm going to see about passing that along to `.generate` to see if that'll get me what I need, but if you have any other suggestions I'd be open to them.

I also checked the `request` to see if I could get the `code` param from there and use that to look up the user's ID, but `code` was missing from the params. I haven't dug in too far yet, but it looks like it might be getting sanitized - do you know if that's the case offhand?

chriswarren:
@tute Any thoughts on what I need to do to get at the `env` var (or set it up similarly to Doorkeeper)? I'd like to be able to access the user/session similarly to how it's done in `resource_owner_authenticator`, but so far I haven't been able to work out where that's being set.

tute:
@chriswarren doorkeeper hooks as an engine (see `engine.rb`) into the ActionController stack, grabbing from there the `request` object built from rack. You can go lower level and hook into the middleware, but AFAIK that would leave you in between rack and doorkeeper. There though you would get anything warden, devise or clearance use. I'm not sure if you where exactly you need to be in the call chain: if before, after or both doorkeeper's code.

When we get a better common ground I may be able to suggest something more concrete.

lib/doorkeeper/models/access_token_mixin.rb
```diff
@@ -128,7 +128,16 @@ def generate_refresh_token
...
+      generator = Doorkeeper.configuration.access_token_generator.constantize
+
+      if generator.method(:generate).arity == 1
+        self.token = generator.generate(resource_owner_id)
+      else
+        self.token = generator.generate
+      end
+    rescue ArgumentError, TypeError, NoMethodError, NameError
+      raise Errors::UnableToGenerateToken,
+      "#{generator} does not respond to `.generate`."
```
chriswarren: not sure what it's looking for here. they are aligned.

chriswarren:
@tute just created https://github.com/chriswarren/doorkeeper-jwt, which depends on this change to work. Testing locally with an app that uses Doorkeeper as well, and things are looking good.

chriswarren:
Trying to figure out why there's a build fail - it's only failing on a single ruby/rails version, and in a file that I didn't touch. I looked for a way to rerun the tests but didn't see one, other than by making another commit.

tute:
@chriswarren please ignore the failure, we still have a flappy spec in mongo ORMs.

I'm happy it works! Soon I'll give some styling feedback so we merge away! Thank you very much for your work! :+1:

lib/doorkeeper/config.rb
```diff
@@ -198,6 +204,8 @@ def extended(base)
     option :realm,                          default: 'Doorkeeper'
     option :force_ssl_in_redirect_uri,      default: !Rails.env.development?
     option :grant_flows,                    default: %w(authorization_code client_credentials)
+    option :access_token_generator,
+    default: "Doorkeeper::OAuth::Helpers::UniqueToken"
```
tute: To keep it consistent with the other lines, even if it's a "long line", I'd align the `default` as the others are.

lib/doorkeeper/models/access_token_mixin.rb
```diff
@@ -128,7 +128,16 @@ def generate_refresh_token
...
 
     def generate_token
-      self.token = UniqueToken.generate
+      generator = Doorkeeper.configuration.access_token_generator.constantize
+
+      if generator.method(:generate).arity == 1
+        self.token = generator.generate(resource_owner_id)
+      else
+        self.token = generator.generate
+      end
```
tute: Rather than this conditional, what if we pass a hash (`resource_owner_id: resource_owner_id`) always, and make it an optional parameter in the default generator, keeping backwards compatibility?

lib/doorkeeper/models/access_token_mixin.rb
tute: I think it expects two spaces of indentation, to make it clear that it's part of the previous statement, and not a new one.

lib/doorkeeper/models/access_token_mixin.rb
```diff
@@ -128,7 +128,16 @@ def generate_refresh_token
...
+    rescue ArgumentError, TypeError, NoMethodError, NameError
```
tute: I'd trap `NoMethodError` as we are doing it, and `NameError` in another block with a proper error message (something like "generator not found"). I don't think we'll need to trap the other two?

lib/generators/doorkeeper/templates/initializer.rb
```diff
@@ -31,6 +31,13 @@
   #   oauth_client.application.additional_settings.implicit_oauth_expiration
   # end
 
+  # Use a custom method for generating the access token. Provide a class that
+  # responds to `.generate` and returns a token string. If the method requires
+  # identifying information about the token owner it can accept one parameter,
+  # and `resource_owner_id` will be passed.
+  # Rationale: https://github.com/doorkeeper-gem/doorkeeper/pull/610
```
tute: What if instead of linking to the code we link to a section in the README?

tute:
Left some comments. Please add a note in the CHANGELOG, and we'll merge. Thanks again!

chriswarren:
Responded to comments.

tute:
Thank you!

PR #614, opened 2015-04-01
Development-URIs

Uepsilon:
This adds a feature to define static Development-URIs. When requesting access from one of those (or subdomains), the request will be considered valid even when the redirect_uri does not match a Client-URI. 

This helps testing the OAuth-Integeration while developing apps without the need to define specific Development-URIs per Application. 

Since the default setting disables this feature, there is no need to update the configuration of existing apps.  

Matching is based on the host only. Ports, Protocol, Params, etc. will be ignored. So keep in mind, that you could create security issues when using it irresponsible. (e.g. don't add http://com as a Dev-URI :see_no_evil:)

### DIFF
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -194,6 +194,7 @@ and that your `initialize_models!` method doesn't raise any errors.\n
     option :authorization_code_expires_in,  default: 600
     option :orm,                            default: :active_record
     option :native_redirect_uri,            default: 'urn:ietf:wg:oauth:2.0:oob'
+    option :development_uris,               default: :disabled
     option :active_record_options,          default: {}
     option :realm,                          default: 'Doorkeeper'
     option :force_ssl_in_redirect_uri,      default: !Rails.env.development?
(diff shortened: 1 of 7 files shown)


tute:
Thanks for your work, @Uepsilon, but I don't think this is a good addition to doorkeeper. It adds complexity to a key step for the security of the protocol, and it's an option that doesn't conform with the spec, violating the "principle of least surprise". Moreover, it will force developers to choose de right parameters and test them in staging or production, rather than in their comfy development environments, making the process it tries to simplify even more cumbersome.

Would you still like doorkeeper to have that? Can you imagine how can you extend the project in another plugin or in your apps to have this option?

Thank you again for sharing your work.

tute:
Closing for now, happy to continue further discussion. Thank you for sharing your work.

Uepsilon:
Sorry for not responding. 

I thought this might be a good addition, as it is disabled by default and can be used for experienced users. 

We've switched to a different oauth gem though, so i cannot provide further support for that. 

If you'd like to add it or keep it, do it, maybe extract it or so. I could imagine this will be useful for some developers.

In our current solution, we've implemented that, but we can enable and disable it per client. This might be something to add as it will reduce the risk of leaving some security holes open. If you are done using the dev-mode, just turn it off and nothing has ever happend :)

PR #615, opened 2015-04-01
Pt-br locale

igorcalabria:
Also changed install generator to look for all locale files inside "config/locale" dir, I think it makes more sense to just scan for everything there.

### DIFF
--- a/config/locales/pt-BR.yml
+++ b/config/locales/pt-BR.yml
@@ -1,0 +1,151 @@
+pt-BE:
+  activerecord:
+    attributes:
+      doorkeeper/application:
+        name: 'Nome'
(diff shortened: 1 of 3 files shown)


tute:
Hey, thanks Igor! Can you send the locale to the i18n project? https://github.com/doorkeeper-gem/doorkeeper-i18n

If you'd like the installer to do something different, let's discuss it in a different PR please.

Thanks again!

tute:
ping @igorcalabria

tute:
Merged in https://github.com/doorkeeper-gem/doorkeeper-i18n/commit/f02ebde65b615807d58cf8b51bbad8ed05bca8c8. Thanks!

PR #624, opened 2015-04-10
Feature: Allow regex expressions of what constitutes a valid redirect uri

jackquack:
Different Applications are unable to use both urn:ietf:wg:oauth:2.0:oob:auto and urn:ietf:wg:oauth:2.0:oob. I thought the best way to solve this was to allow for the `native_redirect_uri` to be a regular expression, such as /urn:ietf:wg:oauth:2.0:oob:auto|urn:ietf:wg:oauth:2.0:oob/

Further, applications may have their own schema requirements that aren't either of the those.

### DIFF
--- a/app/validators/redirect_uri_validator.rb
+++ b/app/validators/redirect_uri_validator.rb
@@ -24,7 +24,7 @@ class RedirectUriValidator < ActiveModel::EachValidator
   private
 
   def native_redirect_uri?(uri)
-    self.class.native_redirect_uri.present? && uri.to_s == self.class.native_redirect_uri.to_s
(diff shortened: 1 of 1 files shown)


tute:
Paraphrasing https://github.com/doorkeeper-gem/doorkeeper/pull/437#issuecomment-49836106:

> This can potentially enable doorkeeper apps to act as "open redirectors", unless very carefuly checking the regexps. Prefer to leave this task in the hands of developers (monkey patch doorkeeper, knowing the associated risks) rather than making it easy for everyone.
> See http://tools.ietf.org/html/rfc6749#section-10.15.

Thank you for your work.

PR #628, opened 2015-04-12
Improve orm adapter

jasl:
- Let ORM adapter handle `setup_application_owner` to give a chance to custom it
- Let ORM adapter check environment requirements and decide whether raise an error or just warn

### DIFF
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -50,8 +50,7 @@ doorkeeper.
   end
 
   def self.setup_application_owner
-    require File.join(File.dirname(__FILE__), 'models', 'concerns', 'ownership')
-    Application.send :include, Models::Ownership
+    @orm_adapter.initialize_application_owner!
   end
 
   class Config
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -9,28 +9,16 @@ module Doorkeeper
     @config = Config::Builder.new(&block).build
     setup_orm_adapter
     setup_orm_models
-    check_for_missing_columns
     setup_application_owner if @config.enable_application_owner?
+    check_requirements
   end
 
   def self.configuration
     @config || (fail MissingConfiguration.new)
   end
 
-  def self.check_for_missing_columns
-    if Doorkeeper.configuration.orm == :active_record &&
-        ActiveRecord::Base.connected? &&
-        ActiveRecord::Base.connection.table_exists?(
-          Doorkeeper::Application.table_name
-        ) &&
-        !Doorkeeper::Application.new.attributes.include?("scopes")
-
-      puts <<-MSG.squish
-[doorkeeper] Missing column: `oauth_applications.scopes`.
-If you are using ActiveRecord run `rails generate doorkeeper:application_scopes
-&& rake db:migrate` to add it.
-      MSG
-    end
+  def self.check_requirements
+    @orm_adapter.check_requirements!(configuration)
   end
 
   def self.setup_orm_adapter
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -7,7 +7,8 @@ module Doorkeeper
 
   def self.configure(&block)
     @config = Config::Builder.new(&block).build
-    enable_orm
+    setup_orm_adapter
+    setup_orm_models
     check_for_missing_columns
     setup_application_owner if @config.enable_application_owner?
   end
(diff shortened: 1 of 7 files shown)


jasl:
`Your Gemfile lists the gem database_cleaner (~> 1.3.0) more than once.` seems not my fault

lib/doorkeeper/config.rb
```diff
@@ -7,54 +7,42 @@ def initialize
...
-      fail e, "ORM adapter not found (#{configuration.orm})", <<-error_msg
+      fail e, "ORM adapter not found (#{configuration.orm})", <<-ERROR_MSG
 [doorkeeper] ORM adapter not found (#{configuration.orm}), or there was an error
 trying to load it.
 
 You probably need to add the related gem for this adapter to work with
 doorkeeper.
-
-If you are working on the adapter itself, double check that the constant exists,
-and that your `initialize_models!` method doesn't raise any errors.\n
```
tute: Why do you remove this last paragraph?

lib/doorkeeper/orm/active_record.rb
```diff
@@ -12,6 +12,27 @@ def self.initialize_models!
           end
         end
       end
+
+      def self.initialize_application_owner!
```
tute: If this method is the same in every ORM, should we leave the definition only in ` lib/doorkeeper/config.rb`?

lib/doorkeeper/orm/active_record.rb
```diff
@@ -12,6 +12,27 @@ def self.initialize_models!
...
+        require 'doorkeeper/models/concerns/ownership'
+
+        Doorkeeper::Application.send :include, Doorkeeper::Models::Ownership
+      end
+
+      def self.check_requirements(_)
```
tute: If this method doesn't take any arguments, we can remove the `(_)`. If it does, let's name it for what it is (`(_variable_name)`, for example).

tute:
This looks good. Can you please address the comments and rebase on top of latest master?

lib/doorkeeper/config.rb
jasl: `initialize_models!` not call in this method, in `setup_orm_models` it will show the full error stack when `initialize_models!` raise error, so the tip is no use now.

lib/doorkeeper/orm/active_record.rb
jasl: It's just coincide
- I remember some people want to override it when they implement new ORM adapter
- this is an option, so keep separate between `initialize_models!`

jasl:
fixed and travis-ci is green

spec/lib/config_spec.rb
```diff
@@ -14,20 +14,12 @@
...
 
-  describe 'enable_orm' do
+  describe 'setup_orm_adapter' do
     it 'adds specific error message to NameError exception' do
       expect do
         Doorkeeper.configure { orm 'hibernate' }
       end.to raise_error(NameError, /ORM adapter not found \(hibernate\)/)
     end
-
-    it 'does not change other exceptions' do
```
jasl: no need test this case, cause `setup_orm_adapter` can only raise `NameError` by `constantize`

lib/doorkeeper/config.rb
```diff
@@ -7,54 +7,38 @@ def initialize
...
-        !Doorkeeper::Application.new.attributes.include?("scopes")
-
-      puts <<-MSG.squish
-[doorkeeper] Missing column: `oauth_applications.scopes`.
-If you are using ActiveRecord run `rails generate doorkeeper:application_scopes
-&& rake db:migrate` to add it.
-      MSG
-    end
+  def self.check_requirements
+    @orm_adapter.check_requirements!(@config)
```
tute: Is there a specific reason why we need to pass `@config`?

lib/doorkeeper/config.rb
```diff
@@ -7,54 +7,38 @@ def initialize
...
 
-  def self.enable_orm
-    class_name = "doorkeeper/orm/#{configuration.orm}".classify
-    class_name.constantize.initialize_models!
-  rescue NameError => e
-    if e.instance_of?(NameError)
-      fail e, "ORM adapter not found (#{configuration.orm})", <<-error_msg
+  def self.setup_orm_adapter
+    @orm_adapter = "doorkeeper/orm/#{configuration.orm}".classify.constantize
+  rescue => e
```
tute: Does this catch _any_ exception? Which are the ones we care about? Can the code raise other type of exceptions?

lib/doorkeeper/orm/active_record.rb
tute: > I remember some people want to override it when they implement new ORM adapter

Do you remember the specific cases where this change makes it better?


> this is an option, so keep separate between `initialize_models!`

It was already its own method in the config file. Why is it better to "unroll" it in each ORM implementation?

lib/doorkeeper/config.rb
jasl: there can only raise `NameError` by `constantize`, no matter `configuration.orm` is nil or just invalid.

lib/doorkeeper/config.rb
jasl: I'v no a strong reason, but option like `enable_application_owner`(there just one for now) may need a extra check, so I allow pass `@config` to easy to catch option

lib/doorkeeper/orm/active_record.rb
jasl: It refer to #281 .
`initialize_application_owner!` implemented in ORM Adapter is good to override, or can only extend by monkey patch

lib/doorkeeper/config.rb
tute: We should continue to specify `NameError`, then. See http://stackoverflow.com/a/10048406/356060.

lib/doorkeeper/config.rb
tute: Sounds good. We can use the `configuration` method, rather than accessing directly the instance variable.

lib/doorkeeper/orm/active_record.rb
tute: :+1:

spec/lib/config_spec.rb
tute: True. This test should continue to exist though. For example, if we raise another exception due to a typo while developing, the exception should not be trapped but raised.

jasl:
updated codes, rebased master, passed the Travis-CI

tute:
Beautiful! :clap: :clap: :clap:

PR #629, opened 2015-04-12
upgrade to rspec 3

jasl:
`rspec-rails` support `rails` 4.2 starts at version `3.1`, so I upgrade it and move all old `rspec` syntax to new. the changes seems too big.


jasl:
I quick reviewed all test cases, found lots of cases have no assertion, add `skip 'No assertion.'` to mask, maybe I'm wrong, I'm not skilled on testing.

Gemfile
```diff
@@ -1,14 +1,7 @@
...
 source 'https://rubygems.org'
 
 gem 'rails', "~> #{ENV['rails']}"
 
-if ENV['rails'][0] == '4'
-  gem 'database_cleaner'
-end
-if ENV['rails'] =~ /4.0|3.2/
-  gem 'rubysl-test-unit'
-end
```
tute: Can we remove this also from https://github.com/doorkeeper-gem/doorkeeper/blob/master/gemfiles/Gemfile.common.rb#L7-L12?

tute: nevermind, you do!

spec/controllers/tokens_controller_spec.rb
```diff
@@ -41,16 +43,15 @@
     end
   end
 
-  describe 'authorize response memoization' do
-    it "memoizes the result of the authorization" do
+  describe 'authorize response memorization' do
+    it 'memorizes the result of the authorization' do
+      skip 'No assertion.'
```
tute: These `skip`s are super useful, these specs are wrong and it's good to mark them as such. Thanks!! :+1:

spec/controllers/tokens_controller_spec.rb
```diff
@@ -41,16 +43,15 @@
...
+
       strategy = double(:strategy, authorize: true)
-      expect(strategy).to receive(:authorize).once
+      allow(strategy).to receive(:authorize).once
```
tute: `expect` is an assertion, and we should have both `allow` first, and `expect` after the execution of the code (to denote better that the `expect` is indeed an assertion.

spec/lib/models/revocable_spec.rb
```diff
@@ -11,6 +11,8 @@
 
   describe :revoke do
     it 'updates :revoked_at attribute with current time' do
+      skip 'No assertion.'
+
```
tute: Same with this spec: `expect` is indeed an assertion. We can rewrite as:

```ruby
clock = double(now: double)
allow(subject).to receive(:update_attribute).with(:revoked_at, clock.now)

subject.revoke(clock)

expect(subject).to receive(:update_attribute).with(:revoked_at, clock.now)
```

tute: This follows the setup / exercise / assert flow, I'd like most specs (if not all) to read like that.

spec/lib/oauth/client/credentials_spec.rb
```diff
@@ -17,16 +17,22 @@ class Doorkeeper::OAuth::Client
       end
 
       it 'accepts anything that responds to #call' do
+        skip 'No assertion.'
+
```
tute: Same about `expect` being an assertion (and needing a rewrite).

spec/lib/oauth/client/credentials_spec.rb
```diff
@@ -17,16 +17,22 @@ class Doorkeeper::OAuth::Client
...
         expect(method).to receive(:call).with(request)
         Credentials.from_request request, method
       end
 
       it 'delegates methods received as symbols to Credentials class' do
+        skip 'No assertion.'
```
tute: Same.

spec/lib/oauth/client/credentials_spec.rb
```diff
@@ -17,16 +17,22 @@ class Doorkeeper::OAuth::Client
...
+
         expect(Credentials).to receive(:from_params).with(request)
         Credentials.from_request request, :from_params
       end
 
       it 'stops at the first credentials found' do
+        skip 'No assertion.'
```
tute: Same.

spec/lib/oauth/client_credentials/issuer_spec.rb
```diff
@@ -28,6 +28,8 @@ class Doorkeeper::OAuth::ClientCredentialsRequest
       end
 
       it 'creates with correct token parameters' do
+        skip 'No assertion.'
```
tute: Sane,

spec/lib/oauth/client_credentials/issuer_spec.rb
```diff
@@ -72,6 +74,8 @@ class Doorkeeper::OAuth::ClientCredentialsRequest
         end
 
         it 'creates with correct token parameters' do
+          skip 'No assertion.'
```
tute: Same.

spec/lib/oauth/client_credentials_request_spec.rb
```diff
@@ -22,6 +22,8 @@ module Doorkeeper::OAuth
     end
 
     it 'issues an access token for the current client' do
+      skip 'No assertion.'
```
tute: Same.

spec/lib/oauth/token_spec.rb
```diff
@@ -18,16 +18,22 @@ module OAuth
         end
 
         it 'accepts anything that responds to #call' do
+          skip 'No assertion.'
```
tute: Same.

spec/lib/oauth/token_spec.rb
```diff
@@ -18,16 +18,22 @@ module OAuth
...
+
           expect(method).to receive(:call).with(request)
           Token.from_request request, method
         end
 
         it 'delegates methods received as symbols to Token class' do
+          skip 'No assertion.'
```
tute: Same. It seems the `skip`s were actually not needed, but I would like a rewrite of these specs to denote better the assertion after the execution. It is a common confusion.

spec/rails_helper.rb
```diff
@@ -0,0 +1,2 @@
+$LOAD_PATH.unshift File.expand_path(File.join(File.dirname(__FILE__), '../lib'))
+$LOAD_PATH.unshift File.expand_path(File.join(File.dirname(__FILE__), '../app'))
```
tute: Do we use this file?

spec/requests/applications/applications_request_spec.rb
```diff
@@ -37,7 +37,7 @@
 end
 
 feature 'Show application' do
-  let :app do
+  given :app do
```
tute: How is `given` different than `let`? Do we need this change?

spec/spec_helper.rb
```diff
@@ -1,2 +1,55 @@
...
+    DatabaseCleaner.start
+    Doorkeeper.configure { orm DOORKEEPER_ORM }
+  end
+
+  config.after do
+    DatabaseCleaner.clean
+  end
+
+  config.order = 'random'
+end
```
tute: Are you sure we want to load always everything? It's good to be able to run a subset of tests with less code while developing, so they run faster.

spec/requests/applications/applications_request_spec.rb
tute: I see request specs are features now. Why is that?

tute:
This is great work, thank you @jasl! Left many comments to address.

spec/requests/applications/applications_request_spec.rb
jasl: I'v read the Rspec's source code, `given` is just an alias for `let`, and it's match for `feature`

spec/rails_helper.rb
jasl: seems no need, I just follow the upgrade guide and not think about it.

spec/spec_helper.rb
jasl: your mean?
we could just let `before` hook run on initializing(using `before(:suite)`

jasl:
I comment some, but maybe I need a long time to understand these cases.

spec/spec_helper.rb
tute: I mean, why not keeping separate `spec/spec_helper.rb` and `spec/spec_helper_integration.rb` files?

spec/requests/applications/applications_request_spec.rb
tute: I see. Let's keep them as request specs, rather than features, please. That way we can upgrade and changeset will be smaller, if we need/want to change the type of spec we can do it in follow-up PRs.
Thank you!

spec/spec_helper.rb
tute: These files that now are combined might be related with the change on request specs. Let's try to roll that back.

jasl: oh, I found the official upgrade guide use the name `spec_helper.rb`, but its contents are the same as our `spec_helper_integration.rb`, so I google this name, found all result is all about doorkeeper...

tute: Some projects separate the helpers to load only the necessary code, according to the type of spec. That is non-standard.

jasl: well, I think follow standard is better.

tute: For this case, why?

jasl: It would easy to maintain I think, when I reading the upgrade guide, I was confusion that we don't have the `spec_helper.rb`. so I use CLI to created one, and realize they are do the same thing. so I guess it increase the cost to understanding for me.

jasl:
Sorry I can't work for doorkeeper this weekend(my company's site is under attacking), but I would reply your comments ASAP.

tute:
No worries @jasl, thanks for your work so far! I incorporated your commit in a new PR (https://github.com/doorkeeper-gem/doorkeeper/pull/635), and will apply my comments on top of it. Will close this Pr in favor of that one.

Best with your company's website!

PR #632, opened 2015-04-15
Properly fall back to default_scopes when no scope is specified.

knu:
The parameter `scopes` is always an array which evaluates as truthy, so the intended fallback has never worked.

This bug caused Doorkeeper to allow clients to access to protected resources under the default scopes regardless of the scopes their token has.

### DIFF
--- a/lib/doorkeeper/rails/helpers.rb
+++ b/lib/doorkeeper/rails/helpers.rb
@@ -4,7 +4,7 @@ module Doorkeeper
       extend ActiveSupport::Concern
 
       def doorkeeper_authorize!(*scopes)
-        @_doorkeeper_scopes = scopes || Doorkeeper.configuration.default_scopes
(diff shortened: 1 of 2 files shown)


knu:
I found a problem in the RSpec configuration where Doorkeeper's configuration is cleared before running each example, making what's in `spec/dummy/config/initializers/doorkeeper.rb` completely ignored.
This is the reason for the ugly `Doorkeeper.configuration.instance_eval` in the added spec.

knu:
Looks like `scope.presence` is used in `doorkeeper_for.rb` in the 1.4-stable branch, so I guess this is new in the 2.x series.

tute:
This is great, thank you very much. This was discussed (and not addressed) in https://github.com/doorkeeper-gem/doorkeeper/pull/567#issuecomment-75345436. I'm very happy to see it fixed.

PR #633, opened 2015-04-16
Extra Details to Custom Token Generator

BlakeDickie:
Added scopes, application and expires_in to the arguments passed to custom token generators.

Scopes and application are needed for use with client credentials grants as in such cases the resource_owner_id is nil and so the scope and application can be used to identify the authorization context of the request.

Expires in is passed so that JWT tokens created with Doorkeeper::JWT gem can set JWT "exp" attribute and allow other systems looking at the JWT token to check the expiry.

### DIFF
--- a/lib/doorkeeper/models/access_token_mixin.rb
+++ b/lib/doorkeeper/models/access_token_mixin.rb
@@ -129,7 +129,9 @@ module Doorkeeper
 
     def generate_token
       generator = Doorkeeper.configuration.access_token_generator.constantize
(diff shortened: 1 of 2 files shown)


tute:
Thanks! Can you please add a regression test? Also, can you squash your commits together, and rebase on top of latest master (instructions in https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history)?
Thanks again!

tute:
Thank you!

Issue #649, opened 2015-05-05
Doorkeeper overrides I18n.fallbacks


nvloff:
Doorkeeper 2.2.0 ( #627  ) broke the custom I18n.fallbacks we set in our app. We worked around that by having the fallbacks set in rails's `after_initialize`.

I don't think gems should change application's I18n fallbacks, it's just breaking things :)

jasl:
The current way is working well on our app, but could you describe more details? and the codes. actually I don't know how to test this.

tute:
We should append to the config, rather than set it.

jasl:
you're right, but I thought the `app.config.i18n.fallbacks = [:en]` is only affect doorkeeper originally, let's me check the doc.

tute:
Closing in favor of PR. Thank you both!

PR #650, opened 2015-05-05
only fallback i18n when it's not set

jasl:
PR for #649

The option fallback can be set as many types, such as `ActiveSupport::OrderedOption`, `Array`, `Hash`, `true`, `false` and `nil`.

I also found the `lib/doorkeeper/engine.rb` is run after `config/appliction.rb`

To conservative, let only set i18n fallback when the application not set it.

### DIFF
--- a/lib/doorkeeper/engine.rb
+++ b/lib/doorkeeper/engine.rb
@@ -5,7 +5,9 @@ module Doorkeeper
     end
 
     initializer "doorkeeper.locales" do |app|
-      app.config.i18n.fallbacks = [:en]
+      if app.config.i18n.fallbacks.blank?
+        app.config.i18n.fallbacks = [:en]
+      end
     end
 
     initializer "doorkeeper.routes" do


lib/doorkeeper/engine.rb
```diff
@@ -5,7 +5,9 @@ class Engine < Rails::Engine
     end
 
     initializer "doorkeeper.locales" do |app|
-      app.config.i18n.fallbacks = [:en]
+      if app.config.i18n.fallbacks.blank?
+        app.config.i18n.fallbacks = [:en]
+      end
     end
```
tute: I believe we want `app.config.i18n.fallbacks += [:en]`. Because doorkeeper can't default to languages it doesn't support we want to _add_ the English option to the ones the application might have defined. This implementation assumes `app.config.i18n.fallbacks` is an array (otherwise we need to initialize it before: `app.config.i18n.fallbacks ||= []`).

jasl: no. 
Because this file is load after app's `application.rb`, when the app set `config.i18n.fallbacks = false`, it will raise `undefined method`.
`config.i18n.fallbacks` has a default value, an empty set, so `app.config.i18n.fallbacks ||= []` has no effect.

tute: I couldn't find Rails documentation on the `fallbacks` option. When not set, in a Rails application I see it as an empty hash (it's in fact an `ActiveSupport::OrderedOptions`, a small subclass of `Hash`), which is close to what the i18n rubygem defines (https://github.com/svenfuchs/i18n/blob/cb679b8cdbab675703a3f88de4d48a48f7b50e06/lib/i18n/locale/fallbacks.rb). A code search in GitHub shows many projects using it as a boolean?

What's the correct way of adding a general default to English, when none is set? Even if it works with current implementation, I'd like to make sure we are using and setting the expected APIs (either by Rails or the i18n project).

jasl: the current way should work, but I found this file load after the app's, so it may override app's config. 
that's should be avoid.

tute:
> This file load after the app's, so it may override app's config(app may already set fallbacks, may not en).
> 
> that's should be avoid.

Can you explain further please? Also, do you think this is already a good step, or we should improve it still?

jasl:
That's mean, is we use `=` to set fallback, we may override app's configure, besides, `config.i18n.fallbacks` accepts many kinds of data-type, we can't hold all cases.

Rails doesn't set it by default, so I guess we only set fallbacks when its blank is enough, when its not blank, means that's an intentional conduct of app developer.

tute:
Thank you!

PR #654, opened 2015-05-12
check should not abort the actual migration

### LINKED ISSUE #653: check_requirements! aborting the scope migration
I have tested this with Rails 3.2. I upgraded Doorkeeper from `0.6.7` to `2.2.1` and I am not being able to run the migration that is suppose to create add the `scopes` column to `oauth_applications`. 

Since #628, `Doorkeeper::Orm::ActiveRecord.check_requirements!` was introduced and an error is raised when the ActiveRecord connection is available, the table exist and attribute `scopes` is found.

Should we consider changing the code to at least check if there are pending migrations? Any other ideas?

repinel:
Fix #653

### DIFF
--- a/lib/doorkeeper/orm/active_record.rb
+++ b/lib/doorkeeper/orm/active_record.rb
@@ -25,11 +25,11 @@ module Doorkeeper
              Doorkeeper::Application.table_name
            )
           unless Doorkeeper::Application.new.attributes.include?("scopes")
-            fail <<-MSG.squish
+            puts <<-MSG.squish
 [doorkeeper] Missing column: `oauth_applications.scopes`.
-Run `rails generate doorkeeper:application_scopes
-&& rake db:migrate` to add it.
+Create the following migration and run `rake db:migrate`.
             MSG
+            puts File.read(File.expand_path('../../../generators/doorkeeper/templates/add_scopes_to_oauth_applications.rb', __FILE__))
           end
         end
       end


lib/doorkeeper/orm/active_record.rb
```diff
@@ -25,11 +25,11 @@ def self.check_requirements!(_config)
              Doorkeeper::Application.table_name
            )
           unless Doorkeeper::Application.new.attributes.include?("scopes")
-            fail <<-MSG.squish
+            puts <<-MSG.squish
```
tute: I'd leave this as `fail`.

lib/doorkeeper/orm/active_record.rb
```diff
@@ -25,11 +25,11 @@ def self.check_requirements!(_config)
...
 [doorkeeper] Missing column: `oauth_applications.scopes`.
-Run `rails generate doorkeeper:application_scopes
-&& rake db:migrate` to add it.
+Create the following migration and run `rake db:migrate`.
             MSG
+            puts File.read(File.expand_path('../../../generators/doorkeeper/templates/add_scopes_to_oauth_applications.rb', __FILE__))
```
tute: Probably we can store the contents on a local variable, and interpolate it in the main failure string.

tute:
Let's add a regression test for this. It's looking good! You shall edit the NEWS.md file too, but I guess there's going to be some massaging of the code before we merge it. Thanks!

lib/doorkeeper/orm/active_record.rb
repinel: If you fail here, the migration that adds `scopes` to `oauth_applications` will just be aborted.

repinel: @tute any thoughts? Thanks!

tute: You are right, we should just `puts` the error message.

tute:
Merged into 2.2-stable in https://github.com/doorkeeper-gem/doorkeeper/commit/0b3284f39f776857cfb4ed6d0bf3200350d86d5e. Thank you!

repinel:
Cool :smiley: 

Shouldn't I add the test and update the `NEWS.md`?

tute:
I don't think we'll need a test for this. That branch works and will stay stable for posterity. :)

If you'd like to update NEWS describing changes from latest release until the tip of the branch that will be helpful. Thank you!

PR #657, opened 2015-05-12
Array#join for multiple URI errors

godsent:
If there are more than one errors `reduce(&:join)` will fail, because of trying to call `String#join`. When there is single error - reduce will return first element of the list without evaluating block (and without exception). I guess there is should be just Array#join.


tute:
@jasl this looks good to me. It's code you touched, so if it looks to you too, feel free to merge!

tute:
Actually, @godsent, if you can add a regression test, that will make it better.

godsent:
I added few specs.

PR #662, opened 2015-05-18
Change scopes for absent application/client to restrict to empty set

jasl:
refer PR #577 

The work is not finish, the problem is `valid_scopes` in `/lib/doorkeeper/oauth/helpers/scope_checker.rb` change old behaviour and break 3 cases.

some testing code
```ruby
application = Doorkeeper::Application.first
access_token = Doorkeeper::AccessToken.first
user = User.first

application.scopes # => ['public', 'write']
access_token.scopes # => ['public', 'write', 'yooo']

v = Validator.new('public write', access_token, application)
v.send :@valid_scopes # => ['public', 'write'] MAY NOT MAKE SENSE HERE
v.parsed_scopes # => ['public', 'write']

Doorkeeper::AccessToken.matching_token_for(application, user.id, 'public write yooo') == access_token # => true
```

### DIFF
--- a/lib/doorkeeper/models/access_token_mixin.rb
+++ b/lib/doorkeeper/models/access_token_mixin.rb
@@ -62,7 +62,7 @@ module Doorkeeper
         return false unless token.present?
         return true if token.scopes.blank? && param_scopes.blank?
 
-        Doorkeeper::OAuth::Helpers::ScopeChecker.valid?(
+        Doorkeeper::OAuth::Helpers::ScopeChecker.match?(
           param_scopes.to_s,
           token,
           application
--- a/lib/doorkeeper/models/access_token_mixin.rb
+++ b/lib/doorkeeper/models/access_token_mixin.rb
@@ -60,16 +60,13 @@ module Doorkeeper
 
       def scopes_match?(token, param_scopes, application)
         return false unless token.present?
+        return true if token.scopes.blank? && param_scopes.blank?
 
-        if token.scopes.blank? && param_scopes.blank?
-          true
-        else
-          Doorkeeper::OAuth::Helpers::ScopeChecker.valid?(
-            param_scopes.to_s,
-            token,
-            application
-          )
-        end
+        Doorkeeper::OAuth::Helpers::ScopeChecker.valid?(
+          param_scopes.to_s,
+          token,
+          application
+        )
       end
 
       def find_or_create_for(application, resource_owner_id, scopes, expires_in, use_refresh_token)
--- a/lib/doorkeeper/models/access_token_mixin.rb
+++ b/lib/doorkeeper/models/access_token_mixin.rb
@@ -53,18 +53,23 @@ module Doorkeeper
                               resource_owner_or_id
                             end
         token = last_authorized_token_for(application.try(:id), resource_owner_id)
-        if token && scopes_match?(token.scopes, scopes, application.try(:scopes))
+        if token && scopes_match?(token, scopes, application)
           token
         end
       end
 
-      def scopes_match?(token_scopes, param_scopes, app_scopes)
-        (!token_scopes.present? && !param_scopes.present?) ||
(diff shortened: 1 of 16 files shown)


jasl:
ping @tute

tute:
Are we doing correctly this check in...?

* `AccessTokenMixin#scopes_match?`
* `PasswordAccessTokenRequest#validate_scopes`
* `ClientCredentialsRequest#validate_scopes`

If we need to change code to handle those checks, we'll need some refactor so we can write this logic only once. This area of the code is very sensitive, and should be as simple as we can.

Also, what happens with applications with empty (default) scopes? Should they allow any scope?

tute:
Let's address most of Hound comments, and make CI green.

lib/doorkeeper/models/access_token_mixin.rb
```diff
@@ -53,18 +53,20 @@ def matching_token_for(application, resource_owner_or_id, scopes)
...
 
-      def scopes_match?(token_scopes, param_scopes, app_scopes)
-        (!token_scopes.present? && !param_scopes.present?) ||
-          Doorkeeper::OAuth::Helpers::ScopeChecker.match?(
-            token_scopes.to_s,
-            param_scopes,
-            app_scopes
-          )
+      def scopes_match?(token, param_scopes, application)
+        return false unless token.present?
```
tute: Let's use `if` whenever possible: `return false if token.empty?`.

lib/doorkeeper/models/access_token_mixin.rb
```diff
@@ -53,18 +53,20 @@ def matching_token_for(application, resource_owner_or_id, scopes)
...
+        return true if token.scopes.blank? && param_scopes.blank?
```
tute: Can the `ScopeChecker` handle this case, where both scopes are blank?

lib/doorkeeper/oauth/client_credentials/validation.rb
```diff
@@ -25,18 +25,12 @@ def validate_client
...
 
         def validate_scopes
-          return true unless @request.scopes.present?
-
-          application_scopes = if @client.present?
-                                 @client.application.scopes
-                               else
-                                 ''
-                               end
+          return true unless @request.original_scopes.present?
```
tute: Let's use `if` in favor of `unless`.

lib/doorkeeper/oauth/helpers/scope_checker.rb
```diff
@@ -18,26 +18,35 @@ def valid?
...
+            valid? && parsed_scopes.equal?(@valid_scopes)
           end
 
           private
 
-          def valid_scopes(server_scopes, application_scopes)
-            if application_scopes.present?
-              server_scopes & application_scopes
+          def valid_scopes(scopes_source, application)
+            # TODO: Refactor this
```
tute: Let's address now this `TODO`. A first idea that come to mind is `return` early in positive cases, and leave as implicit return (outside of every conditional, at the end of the function) the `scopes_source.scopes` part.

lib/doorkeeper/oauth/scopes.rb
```diff
@@ -60,6 +60,12 @@ def &(other)
         other_array = other.present? ? other.all : []
         self.class.from_array(all & other_array)
       end
+
+      def equal?(other)
+        self_scopes = self.map(&:to_s).sort
+        other_scopes = other.map(&:to_s).sort
+        (self_scopes.size == other_scopes.size) && ((self_scopes & other_scopes) == self_scopes)
```
tute: Let's put a semantic name to the `(self_scopes & other_scopes) == self_scopes`, and extract it as a method.

tute:
Don't forget, @jasl! :) No rush if you have no time, just making sure it's not forgotten.

tute:
Closing due to lack of activity.

PR #663, opened 2015-05-20
[ci skip] add news for #654

repinel:


tute:
Thanks! Can you please also add following ones?

```
Merge pull request #599 from knovoselic/attr_accessible_fix
Included MassAssignmentSecurity in mixins if ::ProtectedAttributes is defined
We need this for mongodb ORMs, that don't automatically include the
mixin.

Merge pull request #650 from jasl/i18n-fallback
only fallback i18n when it's not set
```

I'd also swap the 599 and 541 references below your changes. Thank you again!

repinel:
Sure. Should be good now.

tute:
Thank you!

Issue #664, opened 2015-05-21
Feature Request: Display Built-In Unauthorized 


edmondchui:
Per the wiki page [Customizing the response body when unauthorized](https://github.com/doorkeeper-gem/doorkeeper/wiki/Customizing-the-response-body-when-unauthorized), we can customize the respond body overwriting the method ```doorkeeper_unauthorized_render_options```.

However, it is currently impossible to display the three different reasons of unauthorized: ```revoked```, ```expired```, ```unknown``` which is accessible via the object ```OAuth::InvalidTokenResponse```.

It would be great if we can modify the definition of ```doorkeeper_unauthorized_render_options``` to passing the ```error``` object inside the method ```doorkeeper_render_error_with``` via the method ```doorkeeper_render_options``` to ```doorkeeper_unauthorized_render_options```:

```ruby
      def doorkeeper_render_error_with(error)
        options = doorkeeper_render_options(error) || {}
        if options.blank?
          head error.status
        else
          options[:status] = error.status
          options[:layout] = false if options[:layout].nil?
          render options
        end
      end
```

```ruby
      def doorkeeper_render_options(error)
        if doorkeeper_invalid_token_response?
          doorkeeper_unauthorized_render_options(error)
        else
          doorkeeper_forbidden_render_options(error)
        end
      end
```

tute:
I'm not against this. How exactly would you use it? Would it still comply with the spec https://tools.ietf.org/html/rfc6749#section-5.2?

edmondchui:
Yes, for the ```error_description``` field in that [section 5.2](https://tools.ietf.org/html/rfc6749#section-5.2) of the RFC.

tute:
I feel positive about it but I'm not entirely sure. Would like to see sample code, and see if it's backwards compatible. If it looks all good we can merge a nice PR!

edmondchui:
I actually copy and paste those modifications from my overwrite of the doorkeeper code. Sorry I don't have time to put up a crisp sample code for that.

If you have existing test suites we can simply run against them.

tute:
It's really good timing. Because we didn't place any arguments on the template methods this would be a backwards incompatible change, but we are about to release version 3, so let's go ahead with a PR!

tute:
Closing in favor of PR.

PR #665, opened 2015-05-22
Enable Display of Built-In Unauthorized

edmondchui:
For issue #664

### DIFF
--- a/lib/doorkeeper/rails/helpers.rb
+++ b/lib/doorkeeper/rails/helpers.rb
@@ -11,11 +11,11 @@ module Doorkeeper
         end
       end
 
-      def doorkeeper_unauthorized_render_options
+      def doorkeeper_unauthorized_render_options(error)
         nil
       end
 
-      def doorkeeper_forbidden_render_options
+      def doorkeeper_forbidden_render_options(error)
         nil
       end
 
@@ -32,7 +32,7 @@ module Doorkeeper
       end
 
       def doorkeeper_render_error_with(error)
-        options = doorkeeper_render_options || {}
+        options = doorkeeper_render_options(error) || {}
         if options.blank?
           head error.status
         else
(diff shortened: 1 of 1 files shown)


lib/doorkeeper/rails/helpers.rb
```diff
@@ -11,11 +11,11 @@ def doorkeeper_authorize!(*scopes)
         end
       end
 
-      def doorkeeper_unauthorized_render_options
+      def doorkeeper_unauthorized_render_options(error)
```
tute: I'd pick `_error` as a name.

tute: Also, I'd make it a keyword argument: `(_error:)`.

lib/doorkeeper/rails/helpers.rb
```diff
@@ -11,11 +11,11 @@ def doorkeeper_authorize!(*scopes)
         end
       end
 
-      def doorkeeper_unauthorized_render_options
+      def doorkeeper_unauthorized_render_options(_error:)
```
tute: I'm sorry, we don't need the underscore here, makes no sense to require such a key. Can you please remove it?

lib/doorkeeper/rails/helpers.rb
```diff
@@ -32,7 +32,7 @@ def doorkeeper_render_error
       end
 
       def doorkeeper_render_error_with(error)
-        options = doorkeeper_render_options || {}
+        options = doorkeeper_render_options(error) || {}
```
tute: This should read: `error: error`.

lib/doorkeeper/rails/helpers.rb
```diff
@@ -50,11 +50,11 @@ def doorkeeper_error
...
       end
 
-      def doorkeeper_render_options
+      def doorkeeper_render_options(error)
         if doorkeeper_invalid_token_response?
-          doorkeeper_unauthorized_render_options
+          doorkeeper_unauthorized_render_options(error)
         else
-          doorkeeper_forbidden_render_options
+          doorkeeper_forbidden_render_options(error)
```
tute: These three should read: `error: error`.

lib/doorkeeper/rails/helpers.rb
edmondchui: Oh, I thought the underscore was suggested by houndci and concurred by you in my last commit. Anyway, I'm changing it now.

lib/doorkeeper/rails/helpers.rb
edmondchui: This is the method call line, not the declaration line. Should we still use ```error: error``` as well?

tute:
Thank you!

Issue #668, opened 2015-05-28
Issue with use of DateTime in Recoverable concern


emeryamiller:
I'm getting a NoMethodError - undefined method `getlocal' when calling the /oauth/token path. 

I'm using Ruby 2.1.5 and Rails 4.2 with Doorkeeper 2.2.1.

It looks like the issue stems from ActiveRecord calling `getlocal` on the object it's passed when updating the database.  The issue is that the Recoverable concern passes a DateTime object, which doesn't have a `getlocal` method.  Only `Time` has that.

As an expedient fix for my code, I've updated the line:
```ruby
def revoke(clock = DateTime)
        update_attribute :revoked_at, clock.now
end
```
to 
```ruby
def revoke(clock = Time)
        update_attribute :revoked_at, clock.now
end
```
This fixes the issue, but this looks more like an ActiveRecord bug, since I don't see why it wouldn't be able to handle DateTime objects.

Here is part of the stacktrace:

> Started POST "/oauth/token" for ::1 at 2015-05-28 10:34:26 -0400
> ...
> NoMethodError - undefined method `getlocal' for Thu, 28 May 2015 10:34:26 -0400:DateTime:
>  activerecord (4.2.0) lib/active_record/type/date_time.rb:14:in `type_cast_for_database'
>  activerecord (4.2.0) lib/active_record/attribute.rb:46:in `value_for_database'
>  activerecord (4.2.0) lib/active_record/attribute_methods/dirty.rb:164:in `store_original_raw_attribute'
>  activerecord (4.2.0) lib/active_record/attribute_methods/dirty.rb:93:in `write_attribute'
>  activerecord (4.2.0) lib/active_record/attribute_methods.rb:50:in `__temp__275667f6b65646f51647'
>  activerecord (4.2.0) lib/active_record/persistence.rb:238:in `update_attribute'
>  doorkeeper (2.2.1) lib/doorkeeper/models/concerns/revocable.rb:5:in `revoke'

And the relevant code from active record:

> activerecord (4.2.0) lib/active_record/type/date_time.rb, line 14
``` ruby
    9
   10         def type_cast_for_database(value)
   11           zone_conversion_method = ActiveRecord::Base.default_timezone == :utc ? :getutc : :getlocal
   12
   13           if value.acts_like?(:time)
   14             value.send(zone_conversion_method)
   15           else
   16             super
   17           end
   18         end
   19
```

tute:
[...]

PR #671, opened 2015-06-01
Use Time instead of DateTime

emeryamiller:
- Active record works better with Time, as opposed to DateTime objects,
  see issue #668 for more details.

Note: I updated the code, but the test cases kept working despite the change.  I updated the revokable test cases anyway, to match the change above, and the test cases are still passing.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,12 @@ User-visible changes worth mentioning.
 
 ---
 
+## 3.0.0 (rc2)
+
+- [#671] Fixes NoMethodError - undefined method 'getlocal' when calling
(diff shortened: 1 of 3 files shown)


tute:
That is great, thank you! Would please update the NEWS file too? And, if you can add more context (your Note in the PR description, the linked gist in the issue, and the bug itself) to the commit message this will be world class. Thank you! :)

emeryamiller:
Ok, I've updated.  Just one question, I updated the News file by adding an 3.0.0 (rc2) version, I'm not sure if that's right, let me know if it's not.

tute:
Thanks so much!

PR #678, opened 2015-06-12
add support for using only application-level scopes

redterror:
This adds a config option to allow application scopes to fully override server-level scopes instead of the current behavior where the valid set is based on the intersection of the two sets.  It defaults to false.

If this is mostly in order I am happy to expand the generated config and add documentation.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -16,6 +16,9 @@ User-visible changes worth mentioning.
 
 ### Backward incompatible changes
 
+- [#678] Change application-specific scopes to take precedence over server-wide
+  scopes. This removes the previous behavior where the intersection between
+  application and server scopes was used.
 - [#648] Extracts mongodb ORMs to
   https://github.com/doorkeeper-gem/doorkeeper-mongodb. If you use ActiveRecord
   you don’t need to do any change, otherwise you will need to install the new
--- a/lib/doorkeeper/oauth/helpers/scope_checker.rb
+++ b/lib/doorkeeper/oauth/helpers/scope_checker.rb
@@ -25,7 +25,7 @@ module Doorkeeper
 
           def valid_scopes(server_scopes, application_scopes)
             if application_scopes.present?
-              server_scopes & application_scopes
+              application_scopes
             else
               server_scopes
             end
--- a/spec/lib/oauth/helpers/scope_checker_spec.rb
+++ b/spec/lib/oauth/helpers/scope_checker_spec.rb
@@ -41,12 +41,12 @@ module Doorkeeper::OAuth::Helpers
         Doorkeeper::OAuth::Scopes.from_string 'common svr'
       end
       let(:application_scopes) do
-        Doorkeeper::OAuth::Scopes.from_string 'common'
+        Doorkeeper::OAuth::Scopes.from_string 'app123'
       end
 
-      it 'is valid if scope is included in the server and the application' do
+      it 'is valid if scope is included in the application scope list' do
         expect(ScopeChecker.valid?(
-          'common',
+          'app123',
           server_scopes,
           application_scopes
         )).to be_truthy
--- a/spec/lib/oauth/token_request_spec.rb
+++ b/spec/lib/oauth/token_request_spec.rb
@@ -84,8 +84,11 @@ module Doorkeeper::OAuth
 
       it 'skips token creation if there is a matching one' do
         allow(Doorkeeper.configuration).to receive(:reuse_access_token).and_return(true)
+        allow(application.scopes).to receive(:has_scopes?).and_return(true)
+        allow(application.scopes).to receive(:all?).and_return(true)
         FactoryGirl.create(:access_token, application_id: pre_auth.client.id,
                            resource_owner_id: owner.id, scopes: 'public')
+
         expect do
           subject.authorize
         end.to_not change { Doorkeeper::AccessToken.count }


tute:
I'm hesitant about adding new configuration options into doorkeeper. I rather take some out at this point, actually. :) Why do you think we need this, @redterror?

redterror:
The statically configured server-level scopes override the dynamic application-level ones b/c of the array-intersection operator [here](https://github.com/doorkeeper-gem/doorkeeper/pull/678/files#diff-2552b78abd4a0e8e3be0084a1f0ec3a3R30).  That operator choice means that I can add application-level scopes all day but they won't be reflected until a change in my auth server code + deploy + restart are performed.

Application-level scopes are adjustable with a running app, but server-level scopes require a restart.  This change removes that requirement by changing the precedence.

redterror:
This got me thinking about what doorkeeper really is as it relates to "policy".  My understanding was that the goal of doorkeeper was to try to provide the *how*, with hooks for policy to be configured by the user along with a reasonable default.  But this is an example of a hard-coded policy without a hook for the user to change, hence this pull request.  Granted I have added a 2nd hardcoded policy and not a hook, but I didn't realize that until now!

Is there a better way of turning this into a hook the user could adjust?  Either way I am looking to add a locus of control for the end-user here.

Anyway, perhaps this is off-topic, but I wanted to share my perspective so you know where I'm coming from with this PR (and some others I had planned).

tute:
This is interesting discussion, thanks for your input @redterror.

I am not against the need of a code change, and redeploy of the app to change scopes in the app. I do prefer doorkeeper to make developers think thrice before introducing authorization behavior changes.

I actually feel uncomfortable about the fact that we moved authorization logic into the database. Think for example of a scenario where a malicious user gains access to the doorkeeper admin interface, say, by toggling it's `is_admin` flag due to a bug in the application code. Now that person can perform relevant changes in absolute silence. Now, if the applications data, and scopes, and everything were in static files in the application source code, the malicious user would need to gain access to the filesystem, which is a few steps further.

Convenience, security... I lean towards security for doorkeeper.

redterror:
I see this is less of a question about security vs convenience than it is about policy and who decides it.  If Doorkeeper has options or hooks that allow me to implement my own policy, then I can take your example under advisement and solve my problems as my organization needs to depending on how we see the risks.  As it is now there are no such hooks, so my choices are to maintain my own fork with my own custom policy or to monkey patch.

Are you interested in adding hooks or options for users to specify their own policy?  I get the sense not.

tute:
> If Doorkeeper has options or hooks that allow me to implement my own policy [...]

You have the option to put any scope in the initializer that you will need, and with current doorkeeper that effectively results in the behavior you expect. The question is not _if_ doorkeeper will enable you to do what your organization needs, but _how_.

---

When I think of this feature, your PR makes sense to me. It is confusing that doorkeeper allows two very distinct flows: application level scopes or global ones. And then programmers need to remember that those two are to be intersected. I'd rather have one, or the other, and that's your new flag from this PR.

What current implementation allows is users to easily define global scopes, and then configure application-level scopes if so they need. Now, what if we didn't define any "server" scope, and relied only in application level scopes? That would be your use case, would result in a simpler implementation and behavior too.

The impossible to answer question I now have is: how many doorkeeper users need application specific scopes? Because we can't answer it, I feel hesitant as of today to choose one over the other. But supporting everything brings complexity, and your new flag is a new option (state) that the code can be in, multiplied by all the others it already has.

So my question again: would you be able to work with doorkeeper as your organization needs, having added all possible scopes to the `server_scopes`, so when you edit them in the admin they are reflected instantly? If not, why?

redterror:
> So my question again: would you be able to work with doorkeeper as your organization needs, having added all possible scopes to the server_scopes, so when you edit them in the admin they are reflected instantly? If not, why?

It is possible, but it is inconvenient.  Now we have an extra step to remember when we add or change a scope, and this causes confusion for developers if the step is forgotten.  Not everyone's primary job or interest is interacting with the auth server.  When communicated to developers in my organization, this behavior is a received as if it is a bug.

If ultimately you want to retain the current policy without adding this option, I would then submit that the ability to add a scope that cannot be granted is a bug that causes confusion and should be addressed.  Validation on scope input or some kind of help text indicating what are the legal scopes would go towards addressing that.  Still, it seems the root of the issue is expecting to be able to control this behavior dynamically, like all the other pieces that are dynamically controlled.

tute:
> I would then submit that the ability to add a scope that cannot be granted is a bug that causes confusion and should be addressed. Validation on scope input or some kind of help text indicating what are the legal scopes would go towards addressing that. Still, it seems the root of the issue is expecting to be able to control this behavior dynamically, like all the other pieces that are dynamically controlled.

I agree. Adding the currently available scopes, or wording explaining this funkiness, is a workaround for now that I will automatically merge if a PR is submitted.

I'm open to rethinking how do we declare scopes and how do we compute them. The problem we need to solve is how to make it clear and usable for both users who use the global initializer scopes, and people who use the application level specific scopes.

Thank you very much for your time and input on this.

tute:
So we didn't release v3. After this discussion, I wonder if we need the `server_scopes & application_scopes` branch at all or not.

What if we had 100% of one or the other, only? Can it even be simpler?

redterror:
One suggestion - if it was instead `application_scopes || server_scopes`, that is a simple way to support both styles with one code path without any configuration options.  This should also make it easy to explain to end users - simply: "application scopes take precedence over scopes defined in the initializer".

This would represent a change from the current behavior, of course.  But perhaps the timing is right since you're getting ready for v3?

tute:
That makes perfect sense to me. I like that idea.

redterror:
Great!  I'll revise the PR.

redterror:
Revisions made - turns out we didn't even need the `||` here, the `present?` call covers it.  I'm not sure why it says there are conflicts though?  I just rebased this on master a moment ago.

NEWS.md
```diff
@@ -10,6 +10,9 @@ User-visible changes worth mentioning.
   the /oauth/token path. Switch from using a DateTime object to update
   AR to using a Time object. (Issue #668)
 - [#677] Support editing application-specific scopes via the standard forms.
+- [#678] Change application-specific scopes to take precedence over server-wide
+  scopes. This removes the previous behavior where the intersection between
+  application and server scopes was used.
```
tute: This is super important (and so much better)! Can you please add it at the top, in a subsection `### Backward incompatible changes`?

tute:
This is great, thank you! So much better. :+1: 

Left a tiny comment, ready for merge very soon!

NEWS.md
redterror: Changed!

tute:
Thank you!

Issue #667, opened 2015-05-27
Question about authorizing users to particular scopes


redterror:
I have a need/desire to control which users are authorized tokens in particular scopes.  Think an `admin` scope that I want to restrict to ... admins.  I hacked up something that works for my needs but is likely not general enough for inclusion directly:

```diff
diff --git a/app/controllers/doorkeeper/authorizations_controller.rb b/app/controllers/doorkeeper/authorizations_controller.rb
index 660ca52..44ce324 100644
--- a/app/controllers/doorkeeper/authorizations_controller.rb
+++ b/app/controllers/doorkeeper/authorizations_controller.rb
@@ -41,9 +41,9 @@ module Doorkeeper
     end

     def pre_auth
-      @pre_auth ||= OAuth::PreAuthorization.new(Doorkeeper.configuration,
-                                                server.client_via_uid,
-                                                params)
+      @pre_auth ||= Doorkeeper.configuration.preauth_class.new(Doorkeeper.configuration,
+                                                               server.client_via_uid,
+                                                               params.merge(current_resource_owner: current_resource_owner))
     end

     def authorization
diff --git a/lib/doorkeeper/config.rb b/lib/doorkeeper/config.rb
index 32af28e..22a092a 100644
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -98,6 +98,12 @@ doorkeeper.
           '@access_token_generator', access_token_generator
         )
       end
+
+      def preauth_class(klass)
+        @config.instance_variable_set(
+          '@preauth_class', klass
+        )
+      end
     end

     module Option
@@ -189,6 +195,7 @@ doorkeeper.
     option :force_ssl_in_redirect_uri,      default: !Rails.env.development?
     option :grant_flows,                    default: %w(authorization_code client_credentials)
     option :access_token_generator,         default: "Doorkeeper::OAuth::Helpers::UniqueToken"
+    option :preauth_class,                  default: "DoorKeeper::OAuth::PreAuthorization"

     attr_reader :reuse_access_token
```

A few questions:

1. Are you interested in supporting this sort of feature at all, or should I just maintain my own fork?
1. If yes, how close is this to something you'd accept?

One other approach I was thinking was adding a method similar to the `resource_owner_authenticator` method, but called, say, `scope_access_authenticator`.  It would *probably* get 2 params - the current_resource_owner and the requested scopes and act like a before_filter.

[...]

PR #681, opened 2015-06-26
Feature/alt preauth

redterror:
Just following up on [PR 667](https://github.com/doorkeeper-gem/doorkeeper/issues/667) for completeness sake and for other users who might run into this same need.

I know in that PR the plan was to spend time thinking about this feature, but I wanted to throw this out to jumpstart that process.


redterror:
Ah sorry, this was supposed to be branched off master, not off my other pending PR.  Will revise / resubmit later.

PR #683, opened 2015-07-01
Application UID/Secret generation should happen for blank fields

harrisj:
We are using Doorkeeper in a project, and there was a strange issue where the `oauth_applications` table's schema was somehow set to use the empty string as defaults for the `secret` and `uid` tables. (see [this pull request for MyUSA for details on the problem](https://github.com/18F/myusa/pull/684)). This would then make it impossible to create new applications because the blank string would not trigger the `generate_secret` and `generate_uid` before_create callbacks, but they would cause Rails to throw an exception for violating the `validates :presence` restriction on those fields.

While this ultimately turned out to be an issue with our DB schema and not with Doorkeeper, I feel like it would be good to modify this callback code for creates to also get triggered if the UID and Secret fields are blank strings, just because it turned out to be a difficult problem to find and resolve. So, I'm submitting a pull request back to you in case you agree. Thank you.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,7 @@ User-visible changes worth mentioning.
   the /oauth/token path. Switch from using a DateTime object to update
   AR to using a Time object. (Issue #668)
 - [#677] Support editing application-specific scopes via the standard forms.
+- [#682] Pass error hash to Grape `error!`
 
 ## 3.0.0 (rc1)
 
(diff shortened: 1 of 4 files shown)


tute:
I like this, thanks! I like removing `nil` from anywhere it could appear, and I like the idea of having DB schemas defaulting to empty strings.

Can you please rebase on top of latest master, merge the commits together, and add a note to the `NEWS.md` file? Thank I'll merge.

Thanks a lot!

harrisj:
Did a rebase. Again, I'm not really sure yet how one copy of the database wound up defaulting those to empty strings (which is why they stayed empty strings), but I figured it doesn't hurt to have the extra safeguard. Thank you.

tute:
Thank you! Merged in your changes.

PR #685, opened 2015-07-07
add support for specifying an alternative class for preauthorization

redterror:
I'm submitting this PR mostly to follow up on #667 with what I'm using in my project right now.  I was hoping we could resume discussion and find a way to add a supported hook of some kind for users to control preauthorization behavior.

What do you think?  The crux of my argument is that doorkeeper lacks a few key control loci (this PR addresses just one) that are required to implement various policies beyond the current one where all users can access all scopes.  Monkey patching is one way to go, but it seems cleaner to provide a way that users can rely on.  With the current implementation in this PR, I can build a class that takes the user, application, and scopes in the auth request and apply whatever logic is appropriate to my company to decide if it should be approved.

### DIFF
--- a/app/controllers/doorkeeper/authorizations_controller.rb
+++ b/app/controllers/doorkeeper/authorizations_controller.rb
@@ -41,9 +41,9 @@ module Doorkeeper
     end
 
     def pre_auth
-      @pre_auth ||= OAuth::PreAuthorization.new(Doorkeeper.configuration,
-                                                server.client_via_uid,
-                                                params)
+      @pre_auth ||= Doorkeeper.configuration.preauth_class.new(Doorkeeper.configuration,
+                                                               server.client_via_uid,
+                                                               params.merge(current_resource_owner: current_resource_owner))
     end
 
     def authorization
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -98,6 +98,12 @@ doorkeeper.
           '@access_token_generator', access_token_generator
         )
       end
+
+      def preauth_class(klass)
+        @config.instance_variable_set(
+          '@preauth_class', klass
+        )
+      end
     end
 
     module Option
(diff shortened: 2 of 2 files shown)


app/controllers/doorkeeper/authorizations_controller.rb
```diff
@@ -41,9 +41,9 @@ def redirect_or_render(auth)
     end
 
     def pre_auth
-      @pre_auth ||= OAuth::PreAuthorization.new(Doorkeeper.configuration,
-                                                server.client_via_uid,
-                                                params)
+      @pre_auth ||= Doorkeeper.configuration.preauth_class.new(Doorkeeper.configuration,
```
tute: `Doorkeeper.configuration.preauth_class.new(Doorkeeper.configuration`

This reads like hell. I understand why we want it, but I don't like it as is.

redterror: Breaking the line is no problem.  For context, there's similar code in [the use of the access_token_generator option](https://github.com/doorkeeper-gem/doorkeeper/blob/master/lib/doorkeeper/models/access_token_mixin.rb#L132).

tute: Yes I don't mean the white space, which can be better or worse but at last it's superficial. I mean that code is harder to read, and I don't know if it's because of how we are tackling the problem, or because of naming, or because we are getting too many configuration parameters. It looks like a smell of something not going well somewhere. Next week I will have more time to think about this and come up with more concrete answers.

tute:
@redterror I like the idea of making the preauth class configurable. But how can we sent in the params your custom class needs as an extension point, rather than explicitly stating the resource owner, as in current PR?

Do you want to rebase this branch on top of master? Thanks!

redterror:
@tute - my team that was working on using a central doorkeeper server has moved on.  In my case my philosophy was simply to send as much as I could, and since its a class, implementors could simply choose which arguments they were interested in.

tute:
Thanks @redterror.

If you don't need to move this PR forward we can close this PR, and reopen if your team or other parties need this feature in doorkeeper.

tute:
Closing for now, thanks for your ongoing help with paperclip @redterror. 👍

tute:
Closing for now, thanks for your ongoing help with doorkeeper @redterror. 👍

soonya:
I am not sure if this is the right forum for me to ask this question. We need a way for groups of users of our application (roles for eg) to map to certain scopes of allowed operation within the application. To give an anology: in the datacentre we can have the dcadmin(say root user), admin(for a certain class of devices), network operator(admin for a subset of class of devices) and network monitor(read only access to certain limited devices).

PR #686, opened 2015-07-08
Ensure applications scopes

andreacfm:
This is yet another way to add a constraint to the grant flow for limiting the scopes that can be authorized. The scenario is:

* doorkeeper has defined scopes a,b,c,d
* client A select as available scopes a,b,c
* the authorize process for A fails if scope d is required

This is a security concern especially when an implicit authorization flow is in place.
The behaviour is enabled adding a conf called **ensure_application_scopes**. Without this conf doorkeeper fallback to the normal behaviour.

Is this something can be discussed?

### DIFF
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -89,6 +89,10 @@ doorkeeper.
         @config.instance_variable_set("@reuse_access_token", true)
       end
 
+      def ensure_application_scopes
+        @config.instance_variable_set("@ensure_application_scopes", true)
+      end
+
       def force_ssl_in_redirect_uri(boolean)
         @config.instance_variable_set("@force_ssl_in_redirect_uri", boolean)
       end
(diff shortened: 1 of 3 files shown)


tute:
We just made a related merge, in which if there are application scopes those are the only ones taken into account, and if there are not, we only take into account the global doorkeeper scopes: https://github.com/doorkeeper-gem/doorkeeper/pull/678/files#diff-2552b78abd4a0e8e3be0084a1f0ec3a3L28.

Would your scenario work well with this new implementation?

andreacfm:
@tute Yes is very similar and should work. Any idea when this will be released and/or will be backported to 2.2. stable? Feel free to close this PR.

tute:
> Any idea when this will be released

Already released in v3.0.0.rc2: https://github.com/doorkeeper-gem/doorkeeper/blob/v3.0.0.rc2/NEWS.md

> When will be backported to 2.2. stable?

It can't because it's backwards incompatible.

> Feel free to close this PR.

Thanks for your input!

PR #688, opened 2015-07-09
Updates the size of the `token` column

bswinnerton:
This commit updates the size of the `token` column inside the default migration template from a `string` column type to `text`.

The rationale of this proposal is to accomdate users who may use an alternative token generator (accomplished in #611), which uses tokens larger than 255 characters.

### DIFF
--- a/lib/generators/doorkeeper/templates/migration.rb
+++ b/lib/generators/doorkeeper/templates/migration.rb
@@ -27,7 +27,15 @@ class CreateDoorkeeperTables < ActiveRecord::Migration
     create_table :oauth_access_tokens do |t|
       t.integer  :resource_owner_id
       t.integer  :application_id
+
+      # If you use a custom token generator you may need to change this column
+      # from string to text, so that it accepts tokens larger than 255
+      # characters. More info on custom token generators in:
+      # https://github.com/doorkeeper-gem/doorkeeper/tree/v3.0.0.rc1#custom-access-token-generator
+      #
+      # t.text     :token,             null: false
       t.string   :token,             null: false
+
       t.string   :refresh_token
       t.integer  :expires_in
       t.datetime :revoked_at


tute:
Thanks for your contribution @bswinnerton! This is not default behavior though, and should not be part of the default migration. I'm ok with adding the row commented out, explaining why it might be useful.

bswinnerton:
Fair enough, @tute! I've updated the migration with a comment as you suggested. If you'd like me to change the spacing, feel free to let me know which looks best.

lib/generators/doorkeeper/templates/migration.rb
```diff
@@ -27,7 +27,13 @@ def change
     create_table :oauth_access_tokens do |t|
       t.integer  :resource_owner_id
       t.integer  :application_id
+
+      # In the event you would like to use a custom generator as detailed in
+      # http://git.io/vqDqe, you may need to increase the size of this column
+      # to allow for tokens larger than 255 characters.
       t.string   :token,             null: false
+      #t.text     :token,             null: false
```
tute: Let's do something like:

```diff
diff --git a/lib/generators/doorkeeper/templates/migration.rb b/lib/generators/doorkeeper/templates/migration.rb
index af5aa7d..d5a51c4 100644
--- a/lib/generators/doorkeeper/templates/migration.rb
+++ b/lib/generators/doorkeeper/templates/migration.rb
@@ -27,7 +27,13 @@ class CreateDoorkeeperTables < ActiveRecord::Migration
     create_table :oauth_access_tokens do |t|
       t.integer  :resource_owner_id
       t.integer  :application_id
+
+      # If you use a custom token generator you may need to change this column
+      # from string to text, so that it accepts tokens larger than 255
+      # characters. More info on custom token generators in:
+      # https://github.com/doorkeeper-gem/doorkeeper/tree/v3.0.0.rc1#custom-access-token-generator
+      #
+      # t.text     :token,             null: false
       t.string   :token,             null: false
+
       t.string   :refresh_token
       t.integer  :expires_in
       t.datetime :revoked_at
```

Thanks again!

tute:
After the adjustment of the wording and rebasing commits together (https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history) I'll merge.

Thank you Brooks!

bswinnerton:
Updated @tute! :beers:

tute:
:clap: Thanks!

tute:
I'm curious: are you using this custom token feature? If so, with JWT tokens?

bswinnerton:
We are @tute! Or at least, we're in the process of implementing it into Doorkeeper. We found it very difficult when building services for the resource server to communicate with the authorization server on each request to determine if a given token is valid. JWT addresses these issues for us by using a shared secret between both trusted parties (the authorization and resource server).

```
  +--------+                                           +---------------+
  |        |--(A)------- Authorization Grant --------->|               |
  |        |                                           |               |
  |        |<-(B)----------- Access Token -------------|               |
  |        |               & Refresh Token             |               |
  |        |                                           |               |
  |        |                            +----------+   |               |
  |        |--(C)---- Access Token ---->|          |   |               |
  |        |                            |          |   |               |
  |        |<-(D)- Protected Resource --| Resource |   | Authorization |
  | Client |                            |  Server  |   |     Server    |
  |        |--(E)---- Access Token ---->|          |   |               |
  |        |                            |          |   |               |
  |        |<-(F)- Invalid Token Error -|          |   |               |
  |        |                            +----------+   |               |
  |        |                                           |               |
  |        |--(G)----------- Refresh Token ----------->|               |
  |        |                                           |               |
  |        |<-(H)----------- Access Token -------------|               |
  +--------+           & Optional Refresh Token        +---------------+
```

PR #689, opened 2015-07-10
Allow scopes to me resetted to empty in Mongoid

andreacfm:
Back porting for https://github.com/doorkeeper-gem/doorkeeper-mongodb/pull/5


tute:
Thank you!

PR #691, opened 2015-07-13

pmdeazeta:
Creating another PR based on this https://github.com/doorkeeper-gem/doorkeeper/pull/578. Copying description:

> This PR allows refresh tokens to not either revoke at some point in the future (if you want to allow some wiggle room for network failures, etc) set refresh_token_revoked_in to some number of seconds in the future.
>
> Also allows for refresh tokens to not be revoked until an access token created with that refresh token is successfully used once. Set refresh_token_revoked_on_use to true.
>
> Both are off by default.

### DIFF
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -189,6 +189,7 @@ doorkeeper.
     option :force_ssl_in_redirect_uri,      default: !Rails.env.development?
     option :grant_flows,                    default: %w(authorization_code client_credentials)
     option :access_token_generator,         default: "Doorkeeper::OAuth::Helpers::UniqueToken"
+    option :refresh_token_revoked_in,       default: 0
 
     attr_reader :reuse_access_token
 
--- a/lib/doorkeeper/models/concerns/revocable.rb
+++ b/lib/doorkeeper/models/concerns/revocable.rb
@@ -17,7 +17,9 @@ module Doorkeeper
         if old_refresh_token && !old_refresh_token.revoked?
           old_refresh_token.revoke_in(refresh_token_revoked_in)
         end
-        update_attribute :previous_refresh_token, "" if previous_refresh_token != ""
+        if previous_refresh_token.present?
+          update_attribute :previous_refresh_token, ""
+        end
       end
 
       def old_refresh_token
@@ -26,10 +28,6 @@ module Doorkeeper
 
       def refresh_token_revoked_in
         Doorkeeper.configuration.refresh_token_revoked_in
-      end
-
-      def configuration
-        Doorkeeper.configuration
       end
     end
   end
--- a/lib/doorkeeper/models/concerns/revocable.rb
+++ b/lib/doorkeeper/models/concerns/revocable.rb
@@ -8,6 +8,29 @@ module Doorkeeper
       def revoked?
         !!(revoked_at && revoked_at <= Time.now)
       end
+
+      def revoke_in(time)
+        update_attribute :revoked_at, Time.now + time
+      end
+
+      def revoke_previous_refresh_token!
+        if old_refresh_token && !old_refresh_token.revoked?
+          old_refresh_token.revoke_in(refresh_token_revoked_in)
+        end
+        update_attribute :previous_refresh_token, "" if previous_refresh_token != ""
+      end
+
+      def old_refresh_token
+        @old_refresh_token ||= AccessToken.by_refresh_token(previous_refresh_token)
+      end
+
+      def refresh_token_revoked_in
+        Doorkeeper.configuration.refresh_token_revoked_in
+      end
+
+      def configuration
+        Doorkeeper.configuration
+      end
     end
   end
 end
--- a/lib/doorkeeper/oauth/refresh_token_request.rb
+++ b/lib/doorkeeper/oauth/refresh_token_request.rb
@@ -35,8 +35,10 @@ module Doorkeeper
         refresh_token.transaction do
           refresh_token.lock!
           raise Errors::InvalidTokenReuse if refresh_token.revoked?
+          unless refresh_token.revoked_at
+            refresh_token.revoke_in(server.refresh_token_revoked_in)
+          end
 
-          refresh_token.revoke
           create_access_token
         end
       end
@@ -51,12 +53,19 @@ module Doorkeeper
           client
         )
 
-        @access_token = AccessToken.create!(
+        create_params = {
           application_id: refresh_token.application_id,
+          expires_in: expires_in,
           resource_owner_id: refresh_token.resource_owner_id,
           scopes: scopes.to_s,
-          expires_in: expires_in,
-          use_refresh_token: true)
+          use_refresh_token: true
+        }
+
+        unless server.refresh_token_revoked_in.zero?
+          create_params[:previous_refresh_token] = refresh_token.refresh_token
+        end
+
+        @access_token = AccessToken.create!(create_params)
       end
 
       def validate_token_presence
--- a/lib/doorkeeper/oauth/refresh_token_request.rb
+++ b/lib/doorkeeper/oauth/refresh_token_request.rb
@@ -35,7 +35,7 @@ module Doorkeeper
         refresh_token.transaction do
           refresh_token.lock!
           raise Errors::InvalidTokenReuse if refresh_token.revoked?
-          unless refresh_token.revoked_at
+          if !refresh_token.revoked_at
             refresh_token.revoke_in(server.refresh_token_revoked_in)
           end
 
@@ -56,14 +56,11 @@ module Doorkeeper
         create_params = {
           application_id: refresh_token.application_id,
           expires_in: expires_in,
+          previous_refresh_token: refresh_token.refresh_token,
           resource_owner_id: refresh_token.resource_owner_id,
           scopes: scopes.to_s,
           use_refresh_token: true
         }
-
-        unless server.refresh_token_revoked_in.zero?
-          create_params[:previous_refresh_token] = refresh_token.refresh_token
-        end
 
         @access_token = AccessToken.create!(create_params)
       end
--- a/lib/doorkeeper/oauth/refresh_token_request.rb
+++ b/lib/doorkeeper/oauth/refresh_token_request.rb
@@ -35,7 +35,7 @@ module Doorkeeper
         refresh_token.transaction do
           refresh_token.lock!
           raise Errors::InvalidTokenReuse if refresh_token.revoked?
-          if !refresh_token.revoked_at
+          if refresh_token.revoked_at.nil?
             refresh_token.revoke_in(server.refresh_token_revoked_in)
           end
 
@@ -56,7 +56,7 @@ module Doorkeeper
         create_params = {
           application_id: refresh_token.application_id,
           expires_in: expires_in,
-          previous_refresh_token: refresh_token.refresh_token,
+          previous_refresh_token: refresh_token.refresh_token.to_s,
           resource_owner_id: refresh_token.resource_owner_id,
           scopes: scopes.to_s,
           use_refresh_token: true
--- a/lib/doorkeeper/oauth/token.rb
+++ b/lib/doorkeeper/oauth/token.rb
@@ -55,7 +55,9 @@ module Doorkeeper
 
       def self.authenticate(request, *methods)
         if token = from_request(request, *methods)
-          AccessToken.by_token(token)
+          access_token = AccessToken.by_token(token)
+          access_token.revoke_previous_refresh_token! if access_token
+          access_token
         end
       end
     end
--- a/lib/generators/doorkeeper/previous_refresh_token_generator.rb
+++ b/lib/generators/doorkeeper/previous_refresh_token_generator.rb
@@ -1,0 +1,33 @@
+require 'rails/generators/active_record'
+
+class Doorkeeper::PreviousRefreshTokenGenerator < Rails::Generators::Base
+  include Rails::Generators::Migration
+  source_root File.expand_path('../templates', __FILE__)
+  desc 'Support revoke refresh token on access token use'
+
+  def self.next_migration_number(path)
+    ActiveRecord::Generators::Base.next_migration_number(path)
+  end
+
+  def previous_refresh_token
+    if oauth_applications_exists? && !previous_refresh_token_column_exists?
+      migration_template(
+        'add_previous_refresh_token_to_access_tokens.rb',
+        'db/migrate/add_previous_refresh_token_to_access_tokens.rb'
+      )
+    end
+  end
+
+  private
+
+  def previous_refresh_token_column_exists?
+    ActiveRecord::Base.connection.column_exists?(
+      :oauth_access_tokens,
+      :previous_refresh_token
+    )
+  end
+
+  def oauth_applications_exists?
+    ActiveRecord::Base.connection.table_exists? :oauth_applications
+  end
+end
--- a/lib/generators/doorkeeper/previous_refresh_token_generator.rb
+++ b/lib/generators/doorkeeper/previous_refresh_token_generator.rb
@@ -1,33 +1,31 @@
 require 'rails/generators/active_record'
 
-class Doorkeeper::PreviousRefreshTokenGenerator < Rails::Generators::Base
-  include Rails::Generators::Migration
-  source_root File.expand_path('../templates', __FILE__)
-  desc 'Support revoke refresh token on access token use'
+module Doorkeeper
+  class PreviousRefreshTokenGenerator < Rails::Generators::Base
+    include Rails::Generators::Migration
+    source_root File.expand_path('../templates', __FILE__)
+    desc 'Support revoke refresh token on access token use'
 
-  def self.next_migration_number(path)
-    ActiveRecord::Generators::Base.next_migration_number(path)
-  end
+    def self.next_migration_number(path)
+      ActiveRecord::Generators::Base.next_migration_number(path)
+    end
 
-  def previous_refresh_token
-    if oauth_applications_exists? && !previous_refresh_token_column_exists?
-      migration_template(
-        'add_previous_refresh_token_to_access_tokens.rb',
-        'db/migrate/add_previous_refresh_token_to_access_tokens.rb'
+    def previous_refresh_token
+      if !previous_refresh_token_column_exists?
+        migration_template(
+          'add_previous_refresh_token_to_access_tokens.rb',
+          'db/migrate/add_previous_refresh_token_to_access_tokens.rb'
+        )
+      end
+    end
+
+    private
+
+    def previous_refresh_token_column_exists?
+      ActiveRecord::Base.connection.column_exists?(
+        :oauth_access_tokens,
+        :previous_refresh_token
       )
     end
-  end
-
-  private
-
-  def previous_refresh_token_column_exists?
-    ActiveRecord::Base.connection.column_exists?(
-      :oauth_access_tokens,
-      :previous_refresh_token
-    )
-  end
-
-  def oauth_applications_exists?
-    ActiveRecord::Base.connection.table_exists? :oauth_applications
   end
 end
--- a/lib/generators/doorkeeper/templates/add_previous_refresh_token_to_access_tokens.rb
+++ b/lib/generators/doorkeeper/templates/add_previous_refresh_token_to_access_tokens.rb
@@ -1,0 +1,5 @@
+class AddPreviousRefreshTokenToAccessTokens < ActiveRecord::Migration
+  def change
+    add_column :oauth_access_tokens, :previous_refresh_token, :string, default: "", null: false
+  end
+end
--- a/lib/generators/doorkeeper/templates/migration.rb
+++ b/lib/generators/doorkeeper/templates/migration.rb
@@ -41,6 +41,7 @@ class CreateDoorkeeperTables < ActiveRecord::Migration
       t.datetime :revoked_at
       t.datetime :created_at,        null: false
       t.string   :scopes
+      t.string   :previous_refresh_token
     end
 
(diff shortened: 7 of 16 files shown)


lib/doorkeeper/config.rb
```diff
@@ -189,6 +189,8 @@ def extended(base)
     option :force_ssl_in_redirect_uri,      default: !Rails.env.development?
     option :grant_flows,                    default: %w(authorization_code client_credentials)
     option :access_token_generator,         default: "Doorkeeper::OAuth::Helpers::UniqueToken"
+    option :refresh_token_revoked_on_use,   default: false
+    option :refresh_token_revoked_in,       default: 0
```
tute: Having `refresh_token_revoked: 0` s equivalent to `refresh_token_revoked_on_use: false`. I would remove the `refresh_token_revoked_on_use` option, and infer it from the revoked in option.

lib/doorkeeper/models/concerns/revocable.rb
```diff
@@ -8,6 +8,29 @@ def revoke(clock = Time)
       def revoked?
         !!(revoked_at && revoked_at <= Time.now)
       end
+
+      def revoke_in(time)
+        update_attribute :revoked_at, (Time.now + time)
```
tute: I'd rewrite as `update revoked_at: Time.now + time`

tute: This reads weird though. The key is written in the past, but the value is in the future. This seems confusing.

lib/doorkeeper/oauth/refresh_token_request.rb
```diff
@@ -41,17 +43,37 @@ def default_scopes
       end
 
       def create_access_token
+        create_params = {
+          application_id:    refresh_token.application_id,
+          resource_owner_id: refresh_token.resource_owner_id,
+          scopes:            scopes.to_s,
+          expires_in:        server.access_token_expires_in,
+          use_refresh_token: true
```
tute: I would sort these alphabetically by key, and remove the unnecessary spacing.

lib/doorkeeper/oauth/refresh_token_request.rb
```diff
@@ -40,18 +42,36 @@ def default_scopes
...
+
+      def create_access_token
+        create_params = {
+          application_id:    refresh_token.application_id,
+          resource_owner_id: refresh_token.resource_owner_id,
+          scopes:            scopes.to_s,
+          expires_in:        server.access_token_expires_in,
+          use_refresh_token: true
+        }
+        create_params[:previous_refresh_token] = refresh_token.refresh_token if server.refresh_token_revoked_on_use
```
tute: Why make this an option? Should it always behave this way?

lib/doorkeeper/oauth/refresh_token_request.rb
```diff
@@ -41,17 +43,37 @@ def default_scopes
...
+
+        @access_token = AccessToken.create! create_params
+      end
+
+      def create_access_token
```
tute: Why do we have this method defined repeatedly?

lib/generators/doorkeeper/templates/add_previous_refresh_token_to_access_tokens.rb
```diff
@@ -0,0 +1,5 @@
+class AddPreviousRefreshTokenToAccessTokens < ActiveRecord::Migration
+  def change
+    add_column :oauth_access_tokens, :previous_refresh_token, :string
+  end
+end
```
tute: We need to define a way for existing applications to upgrade this. See the rake task for adding scopes to applications for reference.

tute:
Ping @pmdeazeta

tute:
Ping @pmdeazeta!

lib/generators/doorkeeper/templates/add_previous_refresh_token_to_access_tokens.rb
pmdeazeta: I've added the generator

lib/generators/doorkeeper/templates/add_previous_refresh_token_to_access_tokens.rb
```diff
@@ -0,0 +1,5 @@
+class AddPreviousRefreshTokenToAccessTokens < ActiveRecord::Migration
+  def change
+    add_column :oauth_access_tokens, :previous_refresh_token, :string
```
tute: Can you add `default: "", null: false`, to make sure we always get strings?

tute:
Thank you! This is looking great. We have some Rails 3.2 failures though: https://travis-ci.org/doorkeeper-gem/doorkeeper/jobs/79513732. Can you fix CI?

lib/doorkeeper/models/concerns/revocable.rb
```diff
@@ -8,6 +8,29 @@ def revoke(clock = Time)
       def revoked?
         !!(revoked_at && revoked_at <= Time.now)
       end
+
+      def revoke_in(time)
+        update_attribute :revoked_at, Time.now + time
+      end
```
pmdeazeta: @tute tried your suggestion here `update(revoked_at: Time.now + time)` but it was the once causing the failed tests on 3.2

lib/doorkeeper/models/concerns/revocable.rb
```diff
@@ -8,6 +8,29 @@ def revoke(clock = Time)
...
+
+      def revoke_previous_refresh_token!
+        old_refresh_token = AccessToken.by_refresh_token(previous_refresh_token)
+        if old_refresh_token && !old_refresh_token.revoked?
+          if configuration.refresh_token_revoked_in.zero?
```
tute: We don't need to put this conditional, because both branches will end up having the same behavior.

lib/doorkeeper/models/concerns/revocable.rb
```diff
@@ -8,6 +8,29 @@ def revoke(clock = Time)
...
+        end
+
+        if previous_refresh_token.present?
+          update_attribute(:previous_refresh_token, nil)
+        end
+      end
+
+      def configuration
+        Doorkeeper.configuration
+      end
```
tute: Refactor suggestion:

```ruby
    old_refresh_token.revoke_in configuration.refresh_token_revoked_in
  end

  update_attribute(:previous_refresh_token, nil)
end

def old_refresh_token
  @old_refresh_token ||= AccessToken.by_refresh_token(previous_refresh_token)
end

def configuration
  Doorkeeper.configuration
end
```

lib/doorkeeper/oauth/refresh_token_request.rb
```diff
@@ -51,12 +53,19 @@ def create_access_token
...
           resource_owner_id: refresh_token.resource_owner_id,
           scopes: scopes.to_s,
-          expires_in: expires_in,
-          use_refresh_token: true)
+          use_refresh_token: true
+        }
+
+        unless server.refresh_token_revoked_in.zero?
+          create_params[:previous_refresh_token] = refresh_token.refresh_token
+        end
```
tute: Why does this need to be surrounded by the conditional? Can't we/shouldn't be always store the previous refresh token?

lib/doorkeeper/oauth/token.rb
```diff
@@ -55,7 +55,11 @@ def self.from_request(request, *methods)
 
       def self.authenticate(request, *methods)
         if token = from_request(request, *methods)
-          AccessToken.by_token(token)
+          access_token = AccessToken.by_token(token)
+          if access_token && access_token.previous_refresh_token.present?
+            access_token.revoke_previous_refresh_token!
```
tute: Do we need this conditional and extra `revoke_previous_refresh_token` method? Can we directly always revoke it, whether it already is or not?

tute:
Looking better, thank you! Left some comments in a commit and the PR, I think we can still work through it to make this simpler. Thanks a lot for your work!

lib/doorkeeper/models/concerns/revocable.rb
```diff
@@ -8,6 +8,29 @@ def revoke(clock = Time)
       def revoked?
         !!(revoked_at && revoked_at <= Time.now)
       end
+
+      def revoke_in(time)
+        update_attribute :revoked_at, Time.now + time
```
lcpriest: Should we ever be using `Time.now` as opposed to `Time.now.utc`

lib/doorkeeper/oauth/token.rb
lcpriest: We don't need it if we don't use the confditional in https://github.com/doorkeeper-gem/doorkeeper/pull/691/files#diff-aa3b637335d18183d5e2717f3b25689aR64

lib/doorkeeper/models/concerns/revocable.rb
lcpriest: Done. Fixing tests now.

lib/doorkeeper/oauth/token.rb
lcpriest: Unless the where returns nil*

lcpriest:
How serious are we about hound? I can clean all of those up, anything we are relaxed on?

Fredar:
@tute @lcpriest Thanks for developing this, would help a lot to have this in doorkeeper.

lib/doorkeeper/oauth/refresh_token_request.rb
```diff
@@ -35,8 +35,10 @@ def before_successful_response
         refresh_token.transaction do
           refresh_token.lock!
           raise Errors::InvalidTokenReuse if refresh_token.revoked?
+          if !refresh_token.revoked_at
```
tute: > Favor unless over if for negative conditions.

Better yet: `if refresh_token.revoked_at.nil?`

tute:
> How serious are we about hound? I can clean all of those up, anything we are relaxed on?

We follow its barks. I removed some irrelevant comments in `spec/dummy/db/`; would be nice to configure it out of that path.

tute:
Just excluded that path from Hound (https://github.com/doorkeeper-gem/doorkeeper/commit/21731d02b766f6978bf3b2bf69137338b081fcaf). Left the comments that I'd care about though! Thank you.

lib/doorkeeper/models/concerns/revocable.rb
tute: We define the `configuration` method but don't use it, and we'd use it only once in this class. I'd remove the stray method.

lib/generators/doorkeeper/previous_refresh_token_generator.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  def previous_refresh_token_column_exists?
+    ActiveRecord::Base.connection.column_exists?(
+      :oauth_access_tokens,
+      :previous_refresh_token
+    )
+  end
+
+  def oauth_applications_exists?
+    ActiveRecord::Base.connection.table_exists? :oauth_applications
```
tute: Why do we need this check? It looks like we can remove this method and its call.

tute:
This is going to be a backwards incompatible change. Can you please add upgrade notes in NEWS, à la https://github.com/doorkeeper-gem/doorkeeper/blob/master/NEWS.md#backward-incompatible-changes? Thanks!

spec/factories.rb
```diff
@@ -11,6 +11,7 @@
     sequence(:resource_owner_id) { |n| n }
     application
     expires_in 2.hours
+    previous_refresh_token "token"
```
tute: I'd personally remove this line from the factory. First tokens don't have previous tokens, and it makes more sense to me as default state.

lib/doorkeeper/models/concerns/revocable.rb
```diff
@@ -8,6 +8,29 @@ def revoke(clock = Time)
...
+      end
+
+      def revoke_previous_refresh_token!
+        if old_refresh_token && !old_refresh_token.revoked?
+          old_refresh_token.revoke_in(refresh_token_revoked_in)
+        end
+        update_attribute :previous_refresh_token, "" if previous_refresh_token != ""
```
tute: Suggestion:

```ruby
if previous_refresh_token.present?
  update_attribute :previous_refresh_token, ""
end
```

lib/doorkeeper/oauth/refresh_token_request.rb
```diff
@@ -51,12 +53,16 @@ def create_access_token
           client
         )
 
-        @access_token = AccessToken.create!(
+        create_params = {
           application_id: refresh_token.application_id,
+          expires_in: expires_in,
+          previous_refresh_token: refresh_token.refresh_token,
```
tute: The `refresh_token` attribute might be nil, but `previous_refresh_token` doesn't allow for that. We need to add `.to_s`: `previous_refresh_token: refresh_token.refresh_token.to_s,`.

spec/lib/models/revocable_spec.rb
```diff
@@ -33,4 +33,26 @@
...
     end
   end
+
+  describe :revoke_previous_refresh_token! do
+    subject { FactoryGirl.build(:access_token, previous_refresh_token: 'old_refresh_token') }
+    previous_token = FactoryGirl.build(:access_token, token: "previous_token")
+    after { subject.revoke_previous_refresh_token! }
+
+    it 'revokes the previous token if present and sets the attribute :previous_refresh_token to ""' do
+      expect(Doorkeeper::AccessToken).to receive(:by_refresh_token).with(subject.previous_refresh_token).and_return(previous_token)
```
tute: Revocable is a module, and as such we can only test it through a concrete class like `AccessToken`. When we stub it like this, we are [stubbing the system under test](https://robots.thoughtbot.com/don-t-stub-the-system-under-test), potentially hiding bugs in the very same class we are testing.

We can also make it more readable as a "story" following the [four phase test](https://robots.thoughtbot.com/four-phase-test):   setup, exercise, verify, and teardown. That way it will read more like a story: "this is the state of the world, and then this happens, and then this is what we'd expect", without needing to jump from line to line parsing the code in our heads.

These two comments might be a departure from current doorkeeper style of specs, and thus might look a little different than other parts of this file and other files too. I'm fine with that, as I want to move towards more readable specs.

Example for the first spec (might unveil bugs in your current branch):

```ruby
describe :revoke_previous_refresh_token! do
  it "revokes the previous token if existing, and resets the `previous_refresh_token` attribute" do
    previous_token = FactoryGirl.create(:access_token, refresh_token: "refresh_token")
    current_token = FactoryGirl.create(
      :access_token,
      previous_refresh_token: previous_token.refresh_token
    )

    current_token.revoke_previous_refresh_token!

    expect(current_token.previous_refresh_token).to be_empty
    expect(previous_token.reload).to be_revoked
  end
end
```

tute:
Left a few comments. Looking great, thank you! If you'd like to pair on this, let me know!

spec/lib/oauth/refresh_token_request_spec.rb
```diff
@@ -67,6 +69,53 @@ module Doorkeeper::OAuth
...
+    context 'longer lived refresh tokens' do
+      let(:server) do
+        double :server,
+               access_token_expires_in: 2.minutes,
+               custom_access_token_expires_in: ->(_oauth_client) { 1234 },
+               refresh_token_revoked_in: 1.day
+      end
+
+      it 'revokes the previous token' do
+        expect { subject.authorize } .to change { refresh_token.revoked_at }.from(nil).to(be_within(1).of(DateTime.now + 1.day))
```
lcpriest: Do you mind if I leave these for #730

lib/doorkeeper/oauth/refresh_token_request.rb
lcpriest: Thoughts on a database constraint? `null: false, default: ""`

spec/lib/models/revocable_spec.rb
lcpriest:
```ruby
describe :revoke_previous_refresh_token! do
  let(:previous_token) { FactoryGirl.create(:access_token, refresh_token: "refresh_token") }
  let(:current_token)   {  FactoryGirl.create(:access_token, previous_refresh_token: previous_token.refresh_token)
  it "revokes the previous token if existing, and resets the `previous_refresh_token` attribute" do
    current_token.revoke_previous_refresh_token!

    expect(current_token.previous_refresh_token).to be_empty
    expect(previous_token.reload).to be_revoked
  end
end
```

lcpriest: Ping @tute.

lib/doorkeeper/oauth/refresh_token_request.rb
Fredar: First token starts with `refresh_token: nil`.
I think thats on purpose, no need to prevent it.

lcpriest:
I am unable to recreate these test failures on my local machine. Could you please have a look @tute?

tute:
> I am unable to recreate these test failures on my local machine. Could you please have a look @tute?

I'll be able to work on this in around three weeks. Thanks for your work!

spec/lib/oauth/refresh_token_request_spec.rb
tute: > Do you mind if I leave these for #730

If Hound is complaining about code that looks like code elsewhere in the project, I think we should delete/ignore its comments and address them in other PRs indeed, if we find them better.

tute:
> I am unable to recreate these test failures on my local machine. Could you please have a look @tute?

It's surprising, but I cannot reproduce in my computer either!

lib/doorkeeper/models/concerns/revocable.rb
tute: > Should we ever be using Time.now as opposed to Time.now.utc

I think it's best to keep consistent. Which makes me want to merge your other PR which makes everything UTC.

lib/doorkeeper/models/concerns/revocable.rb
```diff
@@ -8,6 +8,27 @@ def revoke(clock = Time)
...
+        update_attribute :revoked_at, Time.now + time
+      end
+
+      def revoke_previous_refresh_token!
+        if old_refresh_token && !old_refresh_token.revoked?
+          old_refresh_token.revoke_in(refresh_token_revoked_in)
+        end
+        if previous_refresh_token.present?
+          update_attribute :previous_refresh_token, ""
+        end
```
tute: I think we don't need this conditional, and we can always execute line 21. Would replace these three lines by `update_attribute :previous_refresh_token, ""`.

lib/doorkeeper/oauth/token.rb
```diff
@@ -55,7 +55,9 @@ def self.from_request(request, *methods)
 
       def self.authenticate(request, *methods)
         if token = from_request(request, *methods)
-          AccessToken.by_token(token)
+          access_token = AccessToken.by_token(token)
+          access_token.revoke_previous_refresh_token! if access_token
```
tute: Do we know here that this token has been requested via the previous refresh token?

lib/generators/doorkeeper/previous_refresh_token_generator.rb
```diff
@@ -0,0 +1,31 @@
...
+    include Rails::Generators::Migration
+    source_root File.expand_path('../templates', __FILE__)
+    desc 'Support revoke refresh token on access token use'
+
+    def self.next_migration_number(path)
+      ActiveRecord::Generators::Base.next_migration_number(path)
+    end
+
+    def previous_refresh_token
+      if !previous_refresh_token_column_exists?
```
tute: > Favor `unless` over `if` for negative conditions.

Ir we can say `if no_previous_refresh_token_column?`, which I think would read best.

lib/generators/doorkeeper/templates/migration.rb
```diff
@@ -41,6 +41,7 @@ def change
       t.datetime :revoked_at
       t.datetime :created_at,        null: false
       t.string   :scopes
+      t.string   :previous_refresh_token
```
tute: We need to add `default: "", null: false` here

spec/lib/models/revocable_spec.rb
tute: I'd rather have the `let`s inside of the `it` block (see https://robots.thoughtbot.com/lets-not), but I if you strongly prefer that alternative I'm fine with that.

Fredar:
@tute @lcpriest The specs that are failing on Travis are not present in this PR - they are however present in the master branch.
I guess rebasing this PR on current master would be a good idea (would remove dummy db barks as well).

tute:
Good catch @Fredar, thanks! Can you please rebase, @pmdeazeta?

Fredar:
I think you meant @lcpriest :wink:

lcpriest:
I believe that this branch is now up to date with master, still failing though.

Fredar:
@lcpriest It fails cause setup for those new tests needs to be updated.
We need to add options here https://github.com/pmdeazeta/doorkeeper/blob/feature/revoke-on-use/spec/controllers/protected_resources_controller_spec.rb#L197

`, previous_refresh_token: "", revoke_previous_refresh_token!: true`

Fredar:
@lcpriest Should I take over, or do you want to continue this PR?

lcpriest:
Happy for you to take over.

Fredar:
Very well, should have some time during the christmas break.
Thank you for your work :wink:

Fredar:
@pmdeazeta @lcpriest Can you grant me permissions for this repo please?
Otherwise I will have to create a new PR.

tute:
It's ok to open a new PR I think, and keep this for history.

Fredar:
Ok then, will open a new one and link.

lcpriest:
Sorry, internet is a bit flakey where I'm Christmassing.

tute:
Closing in favor of https://github.com/doorkeeper-gem/doorkeeper/pull/769. Thank you all!

PR #692, opened 2015-07-13
Ensure applications scopes

andreacfm:
This PR adds a config that forces the Scope Checker to be runned only vs application scopes. No server scopes will be used and authorization will fail if the required scopes does not match an application scope.
Sometimes falling back on server scopes can be too weak because if an application does not define any scope == any server scope can be granted.
The **ensure_application_scope** add a strong rule that limit the authorizable scopes ONLY to the application scopes.

### DIFF
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -89,6 +89,10 @@ doorkeeper.
         @config.instance_variable_set("@reuse_access_token", true)
       end
 
+      def ensure_application_scopes
+        @config.instance_variable_set("@ensure_application_scopes", true)
+      end
+
       def force_ssl_in_redirect_uri(boolean)
         @config.instance_variable_set("@force_ssl_in_redirect_uri", boolean)
       end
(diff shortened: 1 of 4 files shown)


tute:
Would this code have saved you from forgetting scopes from applications? How many applications do we have defined? Does the admin make it clear from the applications index which apps have scopes defined and which do not?

I'd want to increase the chance for users to know this, but I would not add a new option to the code.

andreacfm:
@tute Our use case is having a lot of applications. Btw the weak point that this PR tries to make stronger is that not defining any application scope is the same that allowing any scope to be requested.
Do you think that simply communicating to the app owner about this behaviour should be enough?

tute:
What about adding a validation to the Application model for example,
instead of adding a condition to how we compute scopes to check against? T
would probably be the same option, but it would enable a Rails validation.

andreacfm:
@tute This makes sense. Let me have some test. Thanks

tute:
ping @andreacfm. Thank you!

andreacfm:
@tute A validation solves the use case. Closing this.

tute:
> @tute A validation solves the use case.

Cool, thanks! We should add it to the `ApplicationMixin`.

PR #704, opened 2015-08-11
Make scopes model attribute accessible on application create/editing, to avoid mass assignment errors.

amcnamara:
Tests pass.

Fixes:
[2015-08-11 14:41:16] WARN  ActiveRecord::Base : WARNING: Can't mass-assign protected attributes for Doorkeeper::Application: scopes

### DIFF
--- a/lib/doorkeeper/models/application_mixin.rb
+++ b/lib/doorkeeper/models/application_mixin.rb
@@ -17,7 +17,7 @@ module Doorkeeper
       before_validation :generate_uid, :generate_secret, on: :create
 
       if respond_to?(:attr_accessible)
-        attr_accessible :name, :redirect_uri
+        attr_accessible :name, :redirect_uri, :scopes
       end
     end
 


tute:
When/where exactly do you see this?

> [2015-08-11 14:41:16] WARN ActiveRecord::Base : WARNING: Can't mass-assign protected attributes for Doorkeeper::Application: scopes

It's probably a security feature, not a bug. :)

amcnamara:
I get this whenever I try to create or edit an app with the default views, the controller is unable to persist the scopes but otherwise succeeds.  No doubt protected attributes are a security feature, if there's a way around this that I'm missing please let me know (I've only been using rails for a week or two now).

flynfish:
:+1:

amcnamara:
To dive slightly deeper, it seems that the applications controller already filters the input parameters so that only :name, :redirect_uri, and :scopes can be specified [1]; but the attributes are still mass-assigned on create [2] and update [3] which conflicts with the attr_accessible whitelist in this PR.

[1] https://github.com/doorkeeper-gem/doorkeeper/blob/master/app/controllers/doorkeeper/applications_controller.rb#L48
[2] https://github.com/doorkeeper-gem/doorkeeper/blob/master/app/controllers/doorkeeper/applications_controller.rb#L17
[3] https://github.com/doorkeeper-gem/doorkeeper/blob/master/app/controllers/doorkeeper/applications_controller.rb#L27

amcnamara:
Note: when I created a vanilla rails app and pulled in Doorkeeper I didn't have any issues persisting scopes; it's likely an issue with the configuration of my other rails project.  I'll see if I can track down the conflict; however I don't think this PR has any negative side effects for existing implementations (though I'm curious to hear your thoughts on it!).

tute:
You probably have a Rails app <= 3.2 or with protected_attributes gem. I'm pretty sure your fix will be merged as is, want to run it before. I hope before this weekend. Thank you!

tute:
Actually, merging now. Thank you!

PR #708, opened 2015-08-18
Guard against nil ‘created_at’ in Models::Expirable

mattvague:
I've noticed an issue in Doorkeeper where, if AccessToken#created_at is nil, the Models::Expirable#expires_in_seconds method will raise `NoMethodError (undefined method '+' for nil:NilClass)`.

I came across this issue when playing around with the doorkeeper API and submitting multiple Client Credentials requests in quick succession to `/oauth/token` so I guess there is some kind of race condition (didn't have time to fully explore it, but I think there should be a guard for this case no matter what).

### DIFF
--- a/lib/doorkeeper/models/concerns/expirable.rb
+++ b/lib/doorkeeper/models/concerns/expirable.rb
@@ -6,8 +6,8 @@ module Doorkeeper
       end
 
       def expires_in_seconds
-        return nil if expires_in.nil?
-        expires = (created_at + expires_in.seconds) - Time.now
+        return nil if created_at.nil? || expires_in.nil?
+        expires = expired_time - Time.now
         expires_sec = expires.seconds.round(0)
         expires_sec > 0 ? expires_sec : 0
       end
--- a/spec/lib/models/expirable_spec.rb
+++ b/spec/lib/models/expirable_spec.rb
@@ -47,5 +47,10 @@ describe 'Expirable' do
       expect(subject.expires_in_seconds).to be_nil
     end
 
+    it 'should return nil when created_at is nil' do
+      allow(subject).to receive(:expires_in).and_return(30.seconds)
+      allow(subject).to receive(:created_at).and_return(nil)
+      expect(subject.expires_in_seconds).to be_nil
+    end
   end
 end


tute:
And what does it mean that a token `expires_in_seconds` is `nil`? Also, how can it ever happen that a `created_at` attribute is `nil`? I'd rather fix that, and wouldn't add a new conditional.
Thanks for your work!

tute:
ping @mattvague

mattvague:
@tute Hey, sorry got caught up. As I say, I didn't have a lot of time to look into this but I can spend a little more time to try to find the root cause if you'd like. I'm actually fairly surprised no one else seems to have encountered this, I wonder if it's something to do with my set up.

Can you reproduce it?

mattvague:
@tute Yeah, my solution is definitely not ideal. The response in this case would look like this (note the 0):

```
{
  "access_token": "[TOKEN]",
  "token_type": "bearer",
  "scope": "public",
  "created_at": 0
}
```

mattvague:
Ah hah! So, what's happening is that since I am using `Doorkeeper::JWT` token generator from https://github.com/chriswarren/doorkeeper-jwt and since the requests are so close together, I get a token collision causing a validation error.

I wonder what the solution to that is on the doorkeeper side? Should I add a check somewhere for invalid AccessTokens?

tute:
That project we'll probably need transaction blocks? See https://github.com/doorkeeper-gem/doorkeeper/pull/712

mattvague:
@tute Right, but you don't think there is also an issue with doorkeeper here? Seems like we could make the problem a bit more explicit, it took me a fair bit of digging to figure out that I just had an invalid AccessToken

tute:
> Don't you think there is an issue with doorkeeper here as well? Seems like we could make the error a bit more explicit, it took me a fair bit of digging to figure out that the problem was an invalid AccessToken

I would argue the error is at even lower level than doorkeeper, it's the data itself. Rails makes `created_at` columns default to `NOT NULL`, does your schema.rb file show `null: false` in those attributes? If so, the error would have been caught by a database exception before persisting the object, making the source of the problem more obvious.

If it's not, I encourage you to add DB constraints to the data. Doorkeeper generates the tables with not null attributes in timestamps, so I don't think we can change anything here to fix your issue.

Will close this issue because of this, but I'm interested in learning more about your problem and solving it.

Thank you!

PR #709, opened 2015-08-20
Check count, not merely presence of `scopes`

GeekOnCoffee:
Without this, we're getting the `able_to` string even with no scopes defined

### DIFF
--- a/app/views/doorkeeper/authorizations/new.html.erb
+++ b/app/views/doorkeeper/authorizations/new.html.erb
@@ -7,7 +7,7 @@
     <%= raw t('.prompt', client_name: "<strong class=\"text-info\">#{ @pre_auth.client.name }</strong>") %>
   </p>
 
-  <% if @pre_auth.scopes %>
+  <% if @pre_auth.scopes.count > 0 %>
     <div id="oauth-permissions">
       <p><%= t('.able_to') %>:</p>
 


tute:
Looks good to me. Should we add an `else` branch explicitly stating "no scopes defined"?

GeekOnCoffee:
My gut reaction is that not showing anything is the appropriate behavior if there are no scopes defined... most users wouldn't understand what it means.

tute:
Yeah, I'd add it though. Your one liner is an improvement but it potentially hides information that we'll start to hide. While we are at it, let's be explicit of the no-scopes scenario adding the `else`. Thank you! :+1:

naw:
@tute: Could you clarify the purpose of the text you're requesting?

Is it so developers are warned that they haven't defined scopes?

or 

Is it so end-users (non-developers) who are authorizing applications see that the application is requesting permission for "no scopes defined"?

If the former, perhaps there is a better place to deliver that warning

If the latter, perhaps there is a better way to display this to the end-user?

tute:
> Is it so end-users (non-developers) who are authorizing applications see that the application is requesting permission for "no scopes defined"?

I was mistaken, I somehow thought this was in the admin for applications! Indeed it's not needed, thank you both.

PR #711, opened 2015-08-21
Fix rspec warnings, require env in spec_helper.rb

rngtng:
My first little takeaway:

Rspec threw warnings as raise_error didn’t provide an Error Class to assert on. Indeed in one case the wrong error was raised than expected. In addition, requiring the app env in spec_helper allows to run single test.

### DIFF
--- a/spec/models/doorkeeper/access_token_spec.rb
+++ b/spec/models/doorkeeper/access_token_spec.rb
@@ -131,9 +131,9 @@ module Doorkeeper
         token1 = FactoryGirl.create :access_token, use_refresh_token: true
         token2 = FactoryGirl.create :access_token, use_refresh_token: true
         expect do
-          token2.write_attribute :refresh_token, token1.refresh_token
+          token2.send(:write_attribute, :refresh_token, token1.refresh_token)
           token2.save(validate: false)
-        end.to raise_error
+        end.to raise_error(ActiveRecord::RecordNotUnique)
       end
     end
 
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,2 +1,5 @@
 $LOAD_PATH.unshift File.expand_path(File.join(File.dirname(__FILE__), '../lib'))
 $LOAD_PATH.unshift File.expand_path(File.join(File.dirname(__FILE__), '../app'))
+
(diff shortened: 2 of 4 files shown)


spec/spec_helper.rb
```diff
@@ -1,2 +1,5 @@
 $LOAD_PATH.unshift File.expand_path(File.join(File.dirname(__FILE__), '../lib'))
 $LOAD_PATH.unshift File.expand_path(File.join(File.dirname(__FILE__), '../app'))
+
+require 'capybara/rspec'
+require 'dummy/config/environment'
```
tute: The point of `spec_helper` is running unit tests that don't require Rails. For the ones that do, we require `spec_helper_integration` (https://github.com/doorkeeper-gem/doorkeeper/blob/master/spec/spec_helper_integration.rb#L11). Can we change the file that's loaded in teh ones that do require Rails, and leave this skinnier?

spec/models/doorkeeper/access_token_spec.rb
```diff
@@ -131,9 +131,9 @@ module NoGenerate
         token1 = FactoryGirl.create :access_token, use_refresh_token: true
         token2 = FactoryGirl.create :access_token, use_refresh_token: true
         expect do
-          token2.write_attribute :refresh_token, token1.refresh_token
+          token2.send(:write_attribute, :refresh_token, token1.refresh_token)
```
tute: Why do we have this change? Can this read `token2.refresh_token = token1.refresh_token`?

tute:
Left small comments, thanks for your work!

tute:
Merged your changes into master. Thank you! :)

spec/models/doorkeeper/access_token_spec.rb
rngtng: `write_attribute` is private method. And `raise_error` caught this exception instead of `ActiveRecord::RecordNotUnique`

spec/spec_helper.rb
rngtng: true, loading rails at this point doesn't make sense. I'll check which specs are not integration test but depend on rails (classes)

rngtng:
cool thanks!

PR #712, opened 2015-08-23
Wrap access token grant and refresh  in transactions.

phillbaker:
It's possible to exchange a single grant/refresh token for multiple access tokens. This is definitely _unexpected_ behavior, whether its a vulnerability is a bit murkier. In a highly concurrent environment, it may provide unexpected results to consumers of the API as well.

For the initial exchange of grant for access token, using a grant more than once
seems to violate Section 4.2.1 [1]:

```
4.1.2.  Authorization Response
code
The authorization code generated by the
authorization server. ...The client MUST NOT use the authorization code
more than once.
```

In the transaction, the `lock!` and error raise is the important part to prevent
multiple writes concurrently. In ActiveRecord, `lock!` reloads the record after
doing a SELECT ... FOR UPDATE so if two parallel requests come, the first one to
the DB locks and revokes the token, while the second waits for the lock to be
be released, then acquires the lock, reloading the (now revoked) record and raises
an error.

A side effect of this change is that granting and refreshing tokens is now
transactional - a token cannot be granted without also revoking the grant which
should improve consistency.

[1] https://tools.ietf.org/html/rfc6749#section-4.1.2

### DIFF
--- a/spec/requests/flows/authorization_code_spec.rb
+++ b/spec/requests/flows/authorization_code_spec.rb
@@ -128,3 +128,29 @@ feature 'Authorization Code Flow' do
     end
   end
 end
+
+describe 'Authorization Code Flow' do
+  before do
+    Doorkeeper.configure do
+      orm DOORKEEPER_ORM
+      use_refresh_token
+    end
+    client_exists
+  end
+
+  context 'issuing a refresh token' do
+    before do
+      authorization_code_exists application: @client
+    end
+
+    it 'second of simultaneous client requests get an error for revoked acccess token' do
+      authorization_code = Doorkeeper::AccessGrant.first.token
+      allow_any_instance_of(Doorkeeper::AccessGrant).to receive(:revoked?).and_return(false, true)
+
+      post token_endpoint_url(code: authorization_code, client: @client)
+
+      should_not_have_json 'access_token'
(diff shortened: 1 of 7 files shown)


tute:
Thank you @phillbaker! This looks great. Can you please add notes to `NEWS.md` (and add the missing new line on the end of  `spec/requests/flows/authorization_code_spec.rb`)?
Will merge this week.

spec/requests/flows/authorization_code_spec.rb
```diff
@@ -128,3 +128,29 @@
...
+      authorization_code = Doorkeeper::AccessGrant.first.token
+      allow_any_instance_of(Doorkeeper::AccessGrant).to receive(:revoked?).and_return(false, true)
+
+      post token_endpoint_url(code: authorization_code, client: @client)
+
+      should_not_have_json 'access_token'
+      should_have_json 'error', 'invalid_grant'
+    end
+  end
+end
```
phillbaker: Added, not sure why hound is still complaining, it's also very weird that it doesn't show up on github.com, but it's definitely in the file when I either go to edit it online and when I pull it down. :confused:

phillbaker:
Added to NEWS.md, thanks @tute

tute:
Thank you!

PR #722, opened 2015-09-19
Add support for overriding forbidden status code

jeffreyc:
Doorkeeper allows you to specify the message returned when a forbidden error is triggered (*e.g.*, a token does not have the correct scope). However, it does not allow you to specify the status code. A common practice is to return a 404 for 403s. This updates `doorkeeper_forbidden_render_options` to allow specifying `not_found_for_forbidden: true` as well as `text`/`json` and `layout` in order to return `:not_found` instead of `:forbidden`.

I didn't find any tests for `doorkeeper_forbidden_render_options`, so I added coverage similar to `doorkeeper_unauthorized_render_options`.

This includes a changelog message, though it might make more sense to update https://github.com/doorkeeper-gem/doorkeeper/wiki/Customizing-the-response-body-when-unauthorized to detail both the unauthorized and forbidden options 

I also fixed a misnamed test.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,10 @@ User-visible changes worth mentioning.
 
 ---
 
+- [#722] `doorkeeper_forbidden_render_options` now supports returning a 404 by
+  specifying `respond_not_found_when_forbidden: true` in the
+  `doorkeeper_forbidden_render_options` method.
+
 ## 3.0.1
 
 - [#712] Wrap exchange of grant token for access token and access token refresh
--- a/lib/doorkeeper/rails/helpers.rb
+++ b/lib/doorkeeper/rails/helpers.rb
@@ -31,10 +31,12 @@ module Doorkeeper
 
       def doorkeeper_render_error_with(error)
         options = doorkeeper_render_options(error) || {}
+        status = doorkeeper_status_for_error(
+          error, options.delete(:respond_not_found_when_forbidden))
         if options.blank?
-          head error.status
+          head status
         else
-          options[:status] = error.status
+          options[:status] = status
           options[:layout] = false if options[:layout].nil?
           render options
         end
@@ -53,6 +55,14 @@ module Doorkeeper
           doorkeeper_unauthorized_render_options(error: error)
         else
           doorkeeper_forbidden_render_options(error: error)
+        end
+      end
+
+      def doorkeeper_status_for_error(error, respond_not_found_when_forbidden)
+        if respond_not_found_when_forbidden && error.status == :forbidden
+          :not_found
+        else
+          error.status
         end
       end
 
--- a/spec/controllers/protected_resources_controller_spec.rb
+++ b/spec/controllers/protected_resources_controller_spec.rb
@@ -165,12 +165,106 @@ describe 'doorkeeper authorize filter' do
         end
       end
 
-      it 'it renders a custom JSON response', token: :invalid do
+      it 'it renders a custom text response', token: :invalid do
         get :index, access_token: token_string
         expect(response.status).to eq 401
         expect(response.content_type).to eq('text/html')
         expect(response.header['WWW-Authenticate']).to match(/^Bearer/)
         expect(response.body).to eq('Unauthorized')
+      end
+    end
+  end
+
+  context 'when custom forbidden render options are configured' do
+    before do
+      expect(Doorkeeper::AccessToken).to receive(:by_token).with(token_string).and_return(token)
+      expect(token).to receive(:acceptable?).with([:write]).and_return(false)
+    end
+
+    after do
+      module ControllerActions
+        def doorkeeper_forbidden_render_options(*)
+        end
+      end
+    end
+
+    controller do
+      before_filter -> { doorkeeper_authorize! :write }
+
+      include ControllerActions
+    end
+
+    let(:token) do
+      double(Doorkeeper::AccessToken,
+             accessible?: true, scopes: ['public'], revoked?: false, expired?: false)
+    end
+    let(:token_string) { '1A2DUWE' }
+
+    context 'with a JSON custom render' do
+      before do
+        module ControllerActions
+          def doorkeeper_forbidden_render_options(*)
+            { json: { error_message: 'Forbidden' } }
+          end
+        end
+      end
+
+      it 'renders a custom JSON response' do
+        get :index, access_token: token_string
+        expect(response.header).to_not include('WWW-Authenticate')
+        expect(response.content_type).to eq('application/json')
+        expect(response.status).to eq 403
+        parsed_body = JSON.parse(response.body)
+        expect(parsed_body).not_to be_nil
+        expect(parsed_body['error_message']).to match('Forbidden')
+      end
+    end
+
+    context 'with a status and JSON custom render' do
+      before do
+        module ControllerActions
+          def doorkeeper_forbidden_render_options(*)
+            { json: { error_message: 'Not Found' },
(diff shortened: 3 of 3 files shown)


tute:
Will check the HTTP spec this week: isn't this change against it? I don't
think unforbidden should ever return other status code.
Thanks for your contribution.

jeffreyc:
> Will check the HTTP spec this week: isn't this change against it? I don't think unforbidden should ever return other status code.

RFC 2616 section 10.4.4, 403 Forbidden, states:

> If the server does not wish to make this information available to the client, the status code 404 (Not Found) can be used instead.

If you would rather see something like:

```
status = if options.delete(:not_found_instead_of_forbidden)
           :not_found
         else
           error.status
         end
```

(*i.e.*, pass a bool to return a 404 instead of a 403, vs. explicitly specifying the status code), I'd be happy to make that change.

tute:
> If you would rather pass a bool to return a 404 instead of a 403, vs. explicitly specifying the status code, I'd be happy to make that change.

That sounds better to me, after your quote of the HTTP spec. Thanks for your work!

jeffreyc:
Updated per feedback. I didn't squash the commits, but would be happy to if you'd prefer (or make any other necessary changes). Happy to contribute!

NEWS.md
```diff
@@ -4,6 +4,9 @@ User-visible changes worth mentioning.
 
 ---
 
+- [#722] `doorkeeper_forbidden_render_options` now supports returning a 404 by
+  specifying `not_found_for_forbidden: true`.
```
tute: Let's rename `not_found_for_forbidden` to `respond_not_found_when_forbidden`?

lib/doorkeeper/rails/helpers.rb
```diff
@@ -31,10 +31,16 @@ def doorkeeper_render_error
...
+                 end
         if options.blank?
-          head error.status
+          head status
         else
-          options[:status] = error.status
+          options[:status] = status
           options[:layout] = false if options[:layout].nil?
           render options
         end
```
tute: We can extract a method:

```ruby
def http_status_code
  if error.status == :forbidden && options[:respond_not_found_when_forbidden]
    :not_found
  else
    error.status
  end
end
```

And use it instead of `error.status` in this block, instead of adding the conditional in it.

spec/controllers/protected_resources_controller_spec.rb
```diff
@@ -174,4 +174,106 @@ def doorkeeper_unauthorized_render_options(error: nil)
...
+      before do
+        module ControllerActions
+          def doorkeeper_forbidden_render_options(*)
+            { not_found_for_forbidden: true, text: 'Not Found' }
+          end
+        end
+      end
+
+      it 'renders a custom status code and text response' do
+        forbidden_render_test
```
tute: I wouldn't extract a `forbidden_render_test` method. It's body is key to what's happening in the test, it reads better if we inline it so we don't need to jump back and forth to grasp the whole story.

spec/controllers/protected_resources_controller_spec.rb
```diff
@@ -174,4 +174,106 @@ def doorkeeper_unauthorized_render_options(error: nil)
...
+      module ControllerActions
+        def doorkeeper_forbidden_render_options(*)
+        end
+      end
+    end
+
+    controller do
+      before_filter -> { doorkeeper_authorize! :write }
+
+      include ControllerActions
```
tute: Why do we need to define `ControllerActions`? Can we directly inline the method in this `controller do` block?

NEWS.md
tute: Also I'd add context:

> specifying `respond_not_found_when_forbidden: true` in the `doorkeeper_unauthorized_render_options` and `doorkeeper_forbidden_render_options` methods.

spec/controllers/protected_resources_controller_spec.rb
```diff
@@ -174,4 +174,106 @@ def doorkeeper_unauthorized_render_options(error: nil)
...
+        end
+      end
+
+      it 'renders a custom JSON response' do
+        forbidden_render_test
+        expect(response.content_type).to eq('application/json')
+        expect(response.status).to eq 404
+        parsed_body = JSON.parse(response.body)
+        expect(parsed_body).not_to be_nil
+        expect(parsed_body['error_message']).to match('Not Found')
```
tute: Do we need to assert in the `error_message` option in these specs? Isn't that already covered in other specs?

tute:
Left some PR feedback. Looking great, thanks so much for your work!

lib/doorkeeper/rails/helpers.rb
jeffreyc: I was attempting to avoid leaving the `:not_found_for_forbidden` key in the `options` hash, and that gets awkward when using a helper (it's possible, but then the helper has side effects, and that makes the code harder to read), which is why I wrote it that way. If you'd rather see it split out, and you're OK leaving the key in the `options` hash, I can make the change.

NEWS.md
jeffreyc: > Let's rename `not_found_for_forbidden` to `respond_not_found_when_forbidden`?

Done.

> Also I'd add context

Also done. NEWS.md is a little terse, so I wasn't sure how verbose to be.

spec/controllers/protected_resources_controller_spec.rb
jeffreyc: Fair. Both "with a JSON custom render" and "with a status and JSON custom render" have the same set of assertions. I'll remove the duplicate assertions from the second test.

spec/controllers/protected_resources_controller_spec.rb
jeffreyc: Removed.

spec/controllers/protected_resources_controller_spec.rb
jeffreyc: This is existing behavior I copied from elsewhere in the file. It helps DRY up the code. We could inline `ControllerActions`, but it would add LOC and repetition, plus it would make sense to also change the existing specs so this PR would include additional, unrelated changes.

Your call as to what you would prefer.

lib/doorkeeper/rails/helpers.rb
tute: Extract method + `delete` sounds great to me.

spec/controllers/protected_resources_controller_spec.rb
tute: > This is existing behavior I copied from elsewhere in the file.

Sounds ok to me, indeed a different PR might be best. I find existing specs in doorkeeper rather unclear, would like a revamp. Thanks!

tute:
Looking great to me! Feel free to extract method, rebase and squash, and I'll merge. Thanks!

lib/doorkeeper/rails/helpers.rb
jeffreyc: Refactored, with the `delete` staying in `doorkeeper_render_error_with`, so `doorkeeper_status_for_error` doesn't have hidden side effects. `doorkeeper_status_for_error` felt more consistent/descriptive, but if you prefer `http_status_code`, I'm happy to rename it.

jeffreyc:
> Feel free to extract method, rebase and squash, and I'll merge.

Extracted (see above), rebased, and squashed. If you'd like any other changes, just let me know. Happy to contribute!

tute:
Merged into master. Thanks so much! :clap:

PR #730, opened 2015-10-22
Prefer forcing timezones

lcpriest:
I believe this repositories hound settings encourage not using Time.now without specifying something like timezone or UTC. This should fix those hound errors.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,8 +3,11 @@
 User-visible changes worth mentioning.
 
 ---
-
+## 4.0.0.rc3
 - Toughen parameters filter with exact match
+
+### Backward incompatible changes
+- Force all timezones to use UTC to prevent comparison issues.
 
 ## 4.0.0.rc2
 
--- a/lib/doorkeeper/models/concerns/expirable.rb
+++ b/lib/doorkeeper/models/concerns/expirable.rb
@@ -2,12 +2,12 @@ module Doorkeeper
   module Models
     module Expirable
       def expired?
-        expires_in && Time.now > expired_time
+        expires_in && Time.now.utc > expired_time
       end
 
       def expires_in_seconds
         return nil if expires_in.nil?
-        expires = (created_at + expires_in.seconds) - Time.now
+        expires = (created_at + expires_in.seconds) - Time.now.utc
         expires_sec = expires.seconds.round(0)
         expires_sec > 0 ? expires_sec : 0
       end
--- a/lib/doorkeeper/version.rb
+++ b/lib/doorkeeper/version.rb
@@ -1,3 +1,3 @@
 module Doorkeeper
-  VERSION = "4.0.0.rc2"
+  VERSION = "4.0.0.rc3"
 end
--- a/lib/doorkeeper/models/concerns/revocable.rb
+++ b/lib/doorkeeper/models/concerns/revocable.rb
@@ -2,11 +2,11 @@ module Doorkeeper
   module Models
     module Revocable
       def revoke(clock = Time)
-        update_attribute :revoked_at, clock.now
+        update_attribute :revoked_at, clock.now.utc
       end
 
       def revoked?
-        !!(revoked_at && revoked_at <= Time.now)
+        !!(revoked_at && revoked_at <= Time.now.utc)
       end
     end
   end
--- a/spec/lib/models/revocable_spec.rb
+++ b/spec/lib/models/revocable_spec.rb
@@ -11,20 +11,21 @@ describe 'Revocable' do
 
   describe :revoke do
     it 'updates :revoked_at attribute with current time' do
-      clock = double now: double
-      expect(subject).to receive(:update_attribute).with(:revoked_at, clock.now)
+      utc = double utc: double
+      clock = double now: utc
+      expect(subject).to receive(:update_attribute).with(:revoked_at, clock.now.utc)
       subject.revoke(clock)
     end
   end
 
   describe :revoked? do
     it 'is revoked if :revoked_at has passed' do
-      allow(subject).to receive(:revoked_at).and_return(Time.now - 1000)
+      allow(subject).to receive(:revoked_at).and_return(Time.now.utc - 1000)
       expect(subject).to be_revoked
     end
 
     it 'is not revoked if :revoked_at has not passed' do
-      allow(subject).to receive(:revoked_at).and_return(Time.now + 1000)
+      allow(subject).to receive(:revoked_at).and_return(Time.now.utc + 1000)
       expect(subject).not_to be_revoked
     end
 
--- a/spec/models/doorkeeper/application_spec.rb
+++ b/spec/models/doorkeeper/application_spec.rb
@@ -129,7 +129,7 @@ module Doorkeeper
 
       it 'should destroy its access tokens' do
         FactoryGirl.create(:access_token, application: new_application)
-        FactoryGirl.create(:access_token, application: new_application, revoked_at: Time.now)
+        FactoryGirl.create(:access_token, application: new_application, revoked_at: Time.now.utc)
         expect do
           new_application.destroy
         end.to change { Doorkeeper::AccessToken.count }.by(-2)


tute:
@elle should doorkeeper force UTC times? It's used in Rails and Grape frameworks. Thank you!

elle:
@tute We should never use `Time.now` but rather `Time.current`
Doing so solves most issues with querying the database.

tute:
And how would doorkeeper dor non rails frameworks define current?

elle:
Good point. My apologies. So, yes, we would like to use `.utc`.

Would be good to freeze time in the specs and check these methods in various time zones.

lcpriest:
@elle is there a standard way to freeze time without timecop?

elle:
@lcpriest looking at the gemspec file, timecop is already a development dependency [https://github.com/doorkeeper-gem/doorkeeper/blob/master/doorkeeper.gemspec#L25]

Alternatively, we can use http://api.rubyonrails.org/classes/ActiveSupport/Testing/TimeHelpers.html

Looking at http://api.rubyonrails.org/classes/ActiveRecord/Timestamp.html, timestamps are saved in UTC by default. So it's a good thing that the methods compare both times in utc. Possible problem would be if someone ever set: `config.active_record.default_timezone = :local`

lcpriest:
Given that timecop isn't used anywhere, could we drop the dependency and use the TimeHelpers instead?

elle:
Would be good to give it a go

lcpriest:
I'm at work now, so will be unable to have a go at that - would you be okay to have a look?

tute:
Thank you Lachlan and Elle for your work on this. I'm totally up for dropping Timecop;  if it's not used anywhere It shouldn't break tests if you remove it as part of these changset.

tute:
@lcpriest: is this potentially backwards incompatible? In what cases?

lcpriest:
Any incompatibilities would be if your server was in non UTC time, this could expire a token too soon (or give it an extra day or so)

tute:
And so I understand the implications of this change: what are we missing while we don't have doorkeeper do UTC? What issue are we solving?

Thanks, again!

lcpriest:
It's really just a best practice:
http://rails-bestpractices.com/posts/2014/10/22/use-time-zone-now-instead-of-time-now/
http://danilenko.org/2012/7/6/rails_timezones/

tute:
> It's really just a best practice:

I understand. There are two reasons I feel the need to know:

* Best practices arise from real world problems that probably arise when those are not followed. I want to understand what those are
*

tute:
Whoops, sorry. :) Will check those links out to learn the why of this best practice. Elle also wrote about this:

* https://robots.thoughtbot.com/its-about-time-zones
* https://robots.thoughtbot.com/a-case-study-in-multiple-time-zones

Won't be able to do this homework today, hopefully next Friday! Thanks.

tute:
@lcpriest, the two links you show are Rails specific. Questions still stand:

* What bugs can arise because doorkeeper is not calling `.utc`?
* Would this changeset introduce potentially backwards incompatible changes? What would be the upgrade process?

If it was a new library I would definitly follow your suggestion and code, but there are thousands of existing installations, not only in Rails.

elle:
@tute this might be an issue with comparing two times that are not equal.

`Time.now` reads the local time on the machine, which can be anywhere, but let's assume for the example that it is in NY, so UTC -0400. Then it is compared to `revoked_at` which I am not sure how it is saved (i.e. in what time zone), and to `created_at` that is by default saved as `utc` (unless someone changed the defaults in their app).

Comparing two times in two different time zones is like comparing apples and oranges. To be useful, both times should be in the same time zone. And the least painful way to solve this is to have both compared times in `utc`.

I think we should use TimeCop and stub the times and time zones in the specs to make sure the time comparison works as intended. I would also probably call `.utc` on `created_at` and `revoked_at` to be safe.

tute:
Thank you, Elle! Copying bug report #738 that will be fixed when we merge this PR.

---

I found this issue when following a tutorial, as mentioned in my other issue.

In file `lib/doorkeeper/models/concerns/expirable.rb`, method `expired?` executes:

```ruby
expires_in && Time.now > expired_time
```

In my dev setup, I had to change the line to:

```ruby
expires_in && Time.now.utc > expired_time
```

so it would correctly validate. 

This was preventing me from getting to the URI validation and any other validation happening after this.

This was an issue because grant dates in the DB were in UTC and my Mac is running in timezone -06.00 (Mexico's CST). So when comparing dates, it wasn't comparing them correctly. I thought they would since they are supposed to be Date/Time objects and comparison would be done internally, accounting for time zones, but it seems it is not like that.

---

This answers my first question (what bugs are right now present), but I still didn't answer my second quesition:

Would this changeset introduce potentially backwards incompatible changes? If so, what would be the upgrade process?

lcpriest:
Any code outside of the gem written to compare times would not have taken this change into account, but I think if this is adequately changelogged, that isn't necessarily an issue for the gem.

The other is that we need to convert both sides of the equivalences to use utc, in cases of servers with non-utc timezones set.

lib/doorkeeper/models/concerns/expirable.rb
```diff
@@ -2,12 +2,12 @@ module Doorkeeper
...
     module Expirable
       def expired?
-        expires_in && Time.now > expired_time
+        expires_in && Time.now.utc > expired_time
       end
 
       def expires_in_seconds
         return nil if expires_in.nil?
-        expires = (created_at + expires_in.seconds) - Time.now
+        expires = (created_at + expires_in.seconds) - Time.now.utc
```
lcpriest: How does the .seconds method work outside of rails? It's added to the Numeric module:
http://api.rubyonrails.org/classes/Numeric.html

tute: > How does the .seconds method work outside of rails? It's added to the Numeric module:

Pretty sure it does not, I'm surprised Grape users didn't submit related issues.

tute:
> The other is that we need to convert both sides of the equivalences to use utc, in cases of servers with non-utc timezones set.

Right, and potentially convert doorkeeper's legacy data in different time zones.

lcpriest:
I've not got time right now to work on this, should I close this or is there someone else who can work on it?

tute:
If no one has found issues yet, I'm inclined to close it, as it might introduce new bugs for existing apps.

That said, I will have time starting in December to investigate these issues, and I'm ok having this open until then, as it is a potentially useful fix.

Your call. Thanks!

tute:
`master` is now targetting the next major release, with Rails 5 support, and dropping support for rails < 5. I think we should merge this in with the upcoming release. Is there any work left in this branch, or is it ready to be rebased and merged? Thank you!

lcpriest:
There are no timezone specific tests on this branch.

johnnyshields:
@tute your issue here sounds like a problem in your database serialization layer/ORM. You should be persisting dates in UTC in the DB, and setting TZ when you load them into Ruby. If you are storing dates in local TZ, then you should to ensure the TZ is set correctly when they are converted from DB to Ruby objects. (In other words, if you are serializing/deserializing from DB with correct TZ handling, then `.utc` is needed.)

tute:
@lcpriest would you update the NEWS file, and rebase this branch on top of latest master? I'd like to get this merged in.

lcpriest:
@tute Done :)

tute:
Nice! Can you add more context from this PR's discussion into your commit message, and explicitly state this as a potentially backwards incompatible change in the NEWS file? Then we merge. Thanks!

lcpriest:
rc3?

tute:
> rc3?

Yep!

NEWS.md
```diff
@@ -3,9 +3,12 @@
 User-visible changes worth mentioning.
 
 ---
-
+## 4.0.0.rc3
 - Toughen parameters filter with exact match
```
lcpriest: I'm not sure where this should have been put @tute

lib/doorkeeper/version.rb
```diff
@@ -1,3 +1,3 @@
 module Doorkeeper
-  VERSION = "4.0.0.rc2"
+  VERSION = "4.0.0.rc3"
```
lcpriest: If this shouldn't be another commit, I think github now allows PRs to be squashed on merge.
https://github.com/blog/2141-squash-your-commits

NEWS.md
tute: > I'm not sure where this should have been put @tute

I'll take care of that when I publish the release. Thanks!

tute:
Thanks so much! 🎉

lcpriest:
No, Thank you! Doorkeeper is such a great project and everyone I have spoken to so far has been really nice.

PR #731, opened 2015-10-22
Update specs to use expect

lcpriest:
When working on a PR for this repo, the test suite made a lot of notes about should matchers. This doesn't cover all of them, I noticed a lot in the flow directory, but that seemed to be a different testing framework.

### DIFF
--- a/spec/models/doorkeeper/access_grant_spec.rb
+++ b/spec/models/doorkeeper/access_grant_spec.rb
@@ -3,7 +3,7 @@ require 'spec_helper_integration'
 describe Doorkeeper::AccessGrant do
   subject { FactoryGirl.build(:access_grant) }
 
-  it { should be_valid }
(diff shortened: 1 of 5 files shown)


spec/models/doorkeeper/access_grant_spec.rb
```diff
@@ -14,23 +14,23 @@
...
       subject.save
       subject.token = nil
-      should_not be_valid
+      expect(subject).to_not be_valid
     end
 
     it 'is invalid without expires_in' do
       subject.expires_in = nil
-      should_not be_valid
+      expect(subject).to_not be_valid
```
tute: Small-y: let's switch the `to_not` to `not_to`s?

tute:
Thanks for your work!

tute:
Thank you! Merged into master. :smiley: :heart:

PR #733, opened 2015-10-27
Add registration for grant flows

reidab:
This is a proposal inspired by comments on https://github.com/doorkeeper-gem/doorkeeper-grants_assertion/issues/9, where the possibility of adding strategy registration to Doorkeeper is mentioned.

This adds a `Doorkeeper::GrantFlow` module, which handles registration, and a `Doorkeeper::GrantFlow::Flow` class to handle the actual matching of `grant_type` and `response_type` values.

A new flow is registered like so:

```ruby
Doorkeeper::GrantFlow.register :authorization_code,
  response_type_matches: 'code',
  response_type_strategy: Doorkeeper::Request::Code,
  grant_type_matches: 'authorization_code',
  grant_type_strategy: Doorkeeper::Request::AuthorizationCode
```

Instead of relying on constantizing the param values, requests are now routed to their corresponding strategy by matching against the `response_type_matches` and `grant_type_matches` attributes. This uses the `===` operator, so regular expressions may be used.

Is this an approach you'd be interested in? I'd be happy to iterate and improve on this idea.

### DIFF
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -229,7 +229,7 @@ doorkeeper.
     end
 
     def enabled_grant_flows
-      @enabled_grant_flows ||= grant_flows.map{|name| GrantFlow.get(name) }
+      @enabled_grant_flows ||= grant_flows.map{ |name| GrantFlow.get(name) }
     end
 
     def authorization_response_flows
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -228,33 +228,24 @@ doorkeeper.
       @realm ||= 'Doorkeeper'
     end
 
-    def authorization_response_types
-      @authorization_response_types ||= calculate_authorization_response_types
+    def enabled_grant_flows
+      @enabled_grant_flows ||= grant_flows.map{|name| GrantFlow.get(name) }
     end
 
-    def token_grant_types
-      @token_grant_types ||= calculate_token_grant_types
+    def authorization_response_flows
+      @authorization_response_types ||= enabled_grant_flows.select(&:handles_response_type?)
+    end
+
+    def token_grant_flows
+      @token_grant_types ||= calculate_token_grant_flows
     end
 
     private
 
-    # Determines what values are acceptable for 'response_type' param in
-    # authorization request endpoint, and return them as an array of strings.
-    #
-    def calculate_authorization_response_types
-      types = []
-      types << 'code'  if grant_flows.include? 'authorization_code'
-      types << 'token' if grant_flows.include? 'implicit'
-      types
-    end
-
-    # Determines what values are acceptable for 'grant_type' param token
-    # request endpoint, and return them in array.
-    #
-    def calculate_token_grant_types
-      types = grant_flows - ['implicit']
-      types << 'refresh_token' if refresh_token_enabled?
-      types
+    def calculate_token_grant_flows
+      flows = enabled_grant_flows.select(&:handles_grant_type?)
+      flows << GrantFlow.get('refresh_token') if refresh_token_enabled?
+      flows
     end
   end
 end
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -229,7 +229,7 @@ doorkeeper.
     end
 
     def enabled_grant_flows
-      @enabled_grant_flows ||= grant_flows.map{ |name| GrantFlow.get(name) }
+      @enabled_grant_flows ||= grant_flows.map { |name| GrantFlow.get(name) }
     end
 
     def authorization_response_flows
--- a/lib/doorkeeper/grant_flow.rb
+++ b/lib/doorkeeper/grant_flow.rb
@@ -1,0 +1,44 @@
+require 'doorkeeper/grant_flow/flow'
+
+module Doorkeeper
+  module GrantFlow
+    mattr_accessor(:registered_flows) { Hash.new }
+
+    module_function
+
+    def register(name_or_flow, options = {})
+      if name_or_flow.is_a? Doorkeeper::GrantFlow::Flow
+        registered_flows[name_or_flow.name.to_sym] = name_or_flow
+      else
+        registered_flows[name_or_flow.to_sym] = Flow.new(name_or_flow, options)
+      end
+    end
+
+    def get(name)
+      registered_flows.fetch(name.to_sym)
+    end
+  end
+end
+
+Doorkeeper::GrantFlow.register :implicit, 
+  response_type_matches: 'token',
+  response_type_strategy: Doorkeeper::Request::Token
+
+Doorkeeper::GrantFlow.register :authorization_code,
+  response_type_matches: 'code',
+  response_type_strategy: Doorkeeper::Request::Code,
+  grant_type_matches: 'authorization_code',
+  grant_type_strategy: Doorkeeper::Request::AuthorizationCode
+
+Doorkeeper::GrantFlow.register :client_credentials,
+  grant_type_matches: 'client_credentials',
+  grant_type_strategy: Doorkeeper::Request::ClientCredentials
+
+Doorkeeper::GrantFlow.register :password,
+  grant_type_matches: 'password',
+  grant_type_strategy: Doorkeeper::Request::Password
+
+Doorkeeper::GrantFlow.register :refresh_token,
+  grant_type_matches: 'refresh_token',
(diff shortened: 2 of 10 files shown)


lib/doorkeeper/grant_flow/flow.rb
```diff
@@ -0,0 +1,33 @@
...
+      def handles_response_type?
+        response_type_matches.present?
+      end
+
+      def matches_grant_type?(value)
+        grant_type_matches === value
+      end
+
+      def matches_response_type?(value)
+        response_type_matches === value
```
reidab: I normally would avoid `===`, but it seems like the right tool for this particular job.

lib/doorkeeper/config.rb
```diff
@@ -228,33 +228,24 @@ def realm
       @realm ||= 'Doorkeeper'
     end
 
-    def authorization_response_types
-      @authorization_response_types ||= calculate_authorization_response_types
+    def enabled_grant_flows
+      @enabled_grant_flows ||= grant_flows.map{ |name| GrantFlow.get(name) }
```
reidab: It would be nice to have grant_flows here be an array of Flow objects, but I don't want to break compatibility with existing configurations.

lib/doorkeeper/config.rb
```diff
@@ -228,33 +228,24 @@ def realm
...
-    def authorization_response_types
-      @authorization_response_types ||= calculate_authorization_response_types
+    def enabled_grant_flows
+      @enabled_grant_flows ||= grant_flows.map{|name| GrantFlow.get(name) }
     end
 
-    def token_grant_types
-      @token_grant_types ||= calculate_token_grant_types
+    def authorization_response_flows
+      @authorization_response_types ||= enabled_grant_flows.select(&:handles_response_type?)
```
reidab: I feel like this ends up being more readable than splitting the select across multiple lines.

lib/doorkeeper/grant_flow.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+
+    def get(name)
+      registered_flows.fetch(name.to_sym)
+    end
+  end
+end
+
+Doorkeeper::GrantFlow.register :implicit,
+  response_type_matches: 'token',
```
reidab: Aligning these here would start to push up against line-length limits. I may poke around with Hound in my own repo to see if there's a readable format that it likes…

NuckChorris:
I'm wondering if this would make more sense to handle this inside each strategy a la `respond_to?` instead of a hash of regexes.

Then the routing could consist of a simple array of strategies — much simpler to configure from a user's perspective (and probably from the perspective of this PR) if strategies can be turned on and off

Instead of
```ruby
Doorkeeper::GrantFlow.register(
  :assertion,
  grant_type_matches: /^urn:ietf:params:oauth:grant-type:assertion:(.*)/,
  grant_type_strategy: Doorkeeper::Request::Assertion
)
```
it could be 
```ruby
Doorkeeper.configuration.grant_flows << Doorkeeper::Request::Assertion
```

I guess there's probably some complication in the generation of helper DSL methods though

Anyways, as it is right now, this should be enough for me to get full RFC7521 compliance on doorkeeper-grants_assertion and take that to a 1.0

reidab:
I considered that option and I agree that it might be a better choice. I'd be happy to push things further in that direction. (In fact, this is what drove me towards the refactor in #734 that introduces a common base class for request strategies.)

reidab:
Looking at this again, modeling the grant flows separately from the request strategies seems necessary. The `authentication_code` flow requires separate strategies for response_type and grant_type requests.

reidab:
@tute Are there any changes I can make or different approaches I might try to make this more attractive?

tute:
> @tute Are there any changes I can make or different approaches I might try to make this more attractive?

I'll be able to think about this in around two weeks. Sorry I can't assign enough time before. Thank you for your work!

tute:
Added a question in: https://github.com/doorkeeper-gem/doorkeeper-grants_assertion/issues/9#issuecomment-166970162.

Your changeset might still be useful, not sure if for that one case though.

tute:
This PR seems to be needed for SAML too: https://github.com/doorkeeper-gem/doorkeeper/issues/764#issuecomment-167248444

lib/doorkeeper/config.rb
tute: > I feel like this ends up being more readable than splitting the select across multiple lines.

Sounds good to me. Can you please make the name of the variable and the method match?

lib/doorkeeper/grant_flow.rb
```diff
@@ -0,0 +1,54 @@
+require 'doorkeeper/grant_flow/flow'
+
+module Doorkeeper
+  module GrantFlow
+    mattr_accessor :registered_flows
+    self.registered_flows = {}
+
+    module_function
```
tute: Can we make this a singleton class, rather than a module?

lib/doorkeeper/grant_flow.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    def register(name_or_flow, options = {})
+      if name_or_flow.is_a? Doorkeeper::GrantFlow::Flow
+        registered_flows[name_or_flow.name.to_sym] = name_or_flow
```
tute: It looks like we never use this branch. Do we need this conditional?

reidab: Hmm… I'm trying to remember what led me to add it in the first place. I think the idea was to let plugins do more complex matching by subclassing `Doorkeeper::GrantFlow::Flow` and registering a pre-built flow from their subclass. I'm happy to drop it if we don't anticipate needing that flexibility.

tute: In that case, let's add sample code as documentation in the README or in a code comment. Thank you! :+1:

lib/doorkeeper/grant_flow/flow.rb
tute: > I normally would avoid `===`, but it seems like the right tool for this particular job.

Do we intend to do any kind of regular expression here, given we use the term `matches`? If yes, code should make that fact more explicit; if not, we should rename. Thanks!

lib/doorkeeper/grant_flow.rb
tute: > Can we make this a singleton class, rather than a module?

I see doorkeeper uses this pattern elsewhere, I don't like that. As this is new code, hopefully we can refactor it into classes.

lib/doorkeeper/grant_flow/flow.rb
reidab: The design is definitely to support regular expressions in the `grant_type_matches` and `response_type_matches` options. One example of the regexp use case is given in https://github.com/doorkeeper-gem/doorkeeper-grants_assertion/issues/9#issuecomment-151669702, where the grants_assertion extension may want to claim all grant types starting with `urn:ietf:params:oauth:grant-type:assertion:`.

hadees:
Is this close to being merged in?  I need this for SAML integration as well.

tute:
There are unaddressed comments. @reidab can you continue working on it? If not, @hadees would you continue work on this and open a new PR? Thank you all!

tute:
Closing due to lack of activity. Happy to work through it with whoever takes this work over!

NuckChorris:
I'll see if I can pick this up and continue it in a couple weeks. Current client isn't using Doorkeeper, but in May I should be back to it.

MarkMurphy:
Bump! we're another year in what's happening with this?

tute:
> Bump! we're another year in what's happening with this?

There is no person investing enough time to make this happen in a way that's good for most doorkeeper users. Are you available to help pushing it forward?

MarkMurphy:
> There is no person investing enough time to make this happen in a way that's good for most doorkeeper users. Are you available to help pushing it forward?

Possibly, what would you like me to do.

tute:
I don't work in this project now. https://github.com/maclover7 can help you if you'd like to contribute the changes you look after. Best!

moger777:
@maclover7 would you have any interest in someone reviving this pr (or potentially a new pr that solves this in a different way). I might try to get this working since it's a blocker for using assertion grants.

maclover7:
Hi @reidab! Unfortunately I can't promise I'll have time to review, but feel free to submit a PR here and tag me on it :)

nbulaj:
Hi @reidab . I think feature would be great to have in the Doorkeeper core. Would you like to propose it again?

Issue #661, opened 2015-05-13
Reuse access tokens for the client_credentials grant type


gilbertquri:
Doorkeeper does not seem to reuse access tokens for the client_credentials grant type in any case. Is this intentional?

tute:
This sounds like a bug, not intentional for sure.

vovimayhem:
I read somewhere in this Issue Tracker and in the RFC specs that client credential tokens should not be allowed to refresh... let's validate if that's true first.

tute:
> I think I read somewhere in this Issue Tracker and in the RFC specs that client credential tokens should not be allowed to refresh... I could be remembering it wrong, but let's validate if that's true first.

Please let us know where did you read this @vovimayhem, or the reasons behind that assertion that you remember. Thank you!

vovimayhem:
@tute searched, but can't find it. :( I'll give it an eye in the RFC specs

jimmykarily:
Guys any progress on this?

jimmykarily:
@vovimayhem you probably refer to [this](https://tools.ietf.org/html/rfc6749#section-4.4.3). I don't think the original problem is a missing refresh token though. My problem is that when I request an access token for a client, even if a valid and not expired one exists, a new one is generated and returned. This is not a problem if the client can store the token and reuse it until it expires, but in my case the token is used from a short lived client (a docker container) and every time the container is launched the token generating process starts again resulting in many valid tokens for the same client. I can't find anything in the RFC that justifies this behavior. Still, I might missed it.

jimmykarily:
I created a [Pull Request](https://github.com/jimmykarily/doorkeeper/pull/1) with what seems to fix the issue, at least on my project. I can change the base branch to the original gem's master if you prefer to keep the conversation on this repo (I just didn't want to create a PR until it's ready).

tute:
> I created a Pull Request with what seems to fix the issue, at least on my project. I can change the base branch to the original gem's master if you prefer to keep the conversation on this repo (I just didn't want to create a PR until it's ready).

@jimmykarily that looks good to me, feel free to complete the tests and open up a PR here. Thank you!

jimmykarily:
I will try to get it ready in the next couple of days. I remember having some trouble with the tests. I will ask for help if I get stuck.

tute:
Closing in favor of PR: https://github.com/doorkeeper-gem/doorkeeper/pull/736. Thank you all!

PR #736, opened 2015-10-29
Reuse the existing token in client_credentials flow

jimmykarily:
Fixes issue 661: https://github.com/doorkeeper-gem/doorkeeper/issues/661

### DIFF
--- a/spec/lib/oauth/client_credentials/creator_spec.rb
+++ b/spec/lib/oauth/client_credentials/creator_spec.rb
@@ -11,8 +11,32 @@ class Doorkeeper::OAuth::ClientCredentialsRequest
       end.to change { Doorkeeper::AccessToken.count }.by(1)
     end
 
+    context "when reuse_access_token is true" do
+      it "returns the existing valid token" do
+        allow(Doorkeeper.configuration).to receive(:reuse_access_token).and_return(true)
+        existing_token = subject.call(client, scopes)
+
+        result = subject.call(client, scopes)
+
+        expect(Doorkeeper::AccessToken.count).to eq(1)
+        expect(result).to eq(existing_token)
+      end
+    end
+
+    context "when reuse_access_token is false" do
+      it "returns a new token" do
+        allow(Doorkeeper.configuration).to receive(:reuse_access_token).and_return(false)
+        existing_token = subject.call(client, scopes)
+
+        result = subject.call(client, scopes)
+
+        expect(Doorkeeper::AccessToken.count).to eq(2)
+        expect(result).not_to eq(existing_token)
+      end
+    end
+
     it 'returns false if creation fails' do
-      expect(Doorkeeper::AccessToken).to receive(:create).and_return(false)
+      expect(Doorkeeper::AccessToken).to receive(:find_or_create_for).and_return(false)
       created = subject.call(client, scopes)
       expect(created).to be_falsey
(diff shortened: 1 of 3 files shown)


spec/lib/oauth/token_request_spec.rb
```diff
@@ -1,6 +1,12 @@
 require 'spec_helper_integration'
 
 module Doorkeeper::OAuth
+  describe ClientCredentialsRequest do
+    subject do
+      # TODO: Do we need something here too?
```
jimmykarily: I don't know if there are any testing conventions on the project regarding "side" checking. This file and ```spec/lib/oauth/client_credentials_request_spec.rb``` seem relevant although they would test the same thing as on the file above. Should I keep only the above test?

tute: I don't think we need to add this indeed, would you remove this new block please?

spec/lib/oauth/client_credentials/creator_spec.rb
```diff
@@ -11,8 +11,35 @@ class Doorkeeper::OAuth::ClientCredentialsRequest
...
+        allow(Doorkeeper.configuration).to receive(:reuse_access_token).and_return(false)
+      end
+
+      it 'returns the existing valid token when one exists' do
+        subject.call(client, scopes)
+        expect do
+          subject.call(client, scopes)
+        end.to change { Doorkeeper::AccessToken.count }.by(1)
+      end
+    end
```
tute: I suggest the following rewrite of this spec:

```ruby
context "when reuse_access_token is true" do
  it "returns the existing valid token" do
    allow(Doorkeeper.configuration).to receive(:reuse_access_token).and_return(true)
    existing_token = subject.call(client, scopes)

    result = subject.call(client, scopes)

    expect(Doorkeeper::AccessToken.count).to eq(1)
    expect(result).to eq(existing_token)
  end
end

context "when reuse_access_token is false" do
  it "returns a new token" do
    allow(Doorkeeper.configuration).to receive(:reuse_access_token).and_return(false)
    existing_token = subject.call(client, scopes)

    result = subject.call(client, scopes)

    expect(Doorkeeper::AccessToken.count).to eq(2)
    expect(result).not_to eq(existing_token)
  end
end
```

* It follows the [setup/execute/assertions structure](https://robots.thoughtbot.com/four-phase-test), making it easier to read
* It removes the unneeded `before` block
* It makes both options look more alike, to better illustrate the difference (with the config on or off, it's easier to visualize how behavior changes)
* It corrects the description for the second test
* It introduces an explaining variable for `existing_token` in the second spec
* It consistently uses double quotes

Could you update your copy? It might not be correct syntax (don't have my tests set up right now), and Hound might have one or two complaints, but I'd like to keep these advantages.

Thank you!

tute:
Left two comments, thank you! Can you please rebase on top of latest master when you address them? Thank you!

tute:
Also, please add a note to [NEWS.md file](https://github.com/doorkeeper-gem/doorkeeper/blob/master/NEWS.md). Thank you! :smiley: :clap:

jimmykarily:
Rebased, fixed comments and added entry in NEWS.md (I amended the commit since the changes were pretty clear). @tute thanks.

tute:
Thanks so much!

PR #740, opened 2015-11-04
Allow localhost in redirect URI option, and fixed attr_accessible issue with Rails 4.0.x

TMSCH:
Hey,

This pull request covers 2 things:

- The main one is the addition of a `allow_localhost_in_redirect_uri` option which allows to specify a localhost (e.g. `http://localhost:3000/callback`) even when `force_ssl_in_redirect_uri` is true. This has been a use case in our situation where an internal OAuth provider could be used for local dev (just as you can use Google OAuth for local dev too)
- The second, minor one, was an issue when upgrading from Doorkeeper 2.2.0 to 2.2.2 where `attr_accessible` was added to model mixins. In the `included` block, it tests for `respond_to?(:attr_accessible)` before using it but Rails 4.0.x still provides the method and raises a deprecation error... I've changed the condition to fix that.

Let me know if there's any issue with my updates!

Thanks
Thomas

### DIFF
--- a/app/validators/redirect_uri_validator.rb
+++ b/app/validators/redirect_uri_validator.rb
@@ -14,7 +14,7 @@ class RedirectUriValidator < ActiveModel::EachValidator
         return if native_redirect_uri?(uri)
         record.errors.add(attribute, :fragment_present) unless uri.fragment.nil?
         record.errors.add(attribute, :relative_uri) if uri.scheme.nil? || uri.host.nil?
-        record.errors.add(attribute, :secured_uri) if invalid_ssl_uri?(uri)
+        record.errors.add(attribute, :secured_uri) if invalid_ssl_uri?(uri) && !allowed_localhost_uri?(uri)
       end
     end
   rescue URI::InvalidURIError
@@ -30,5 +30,10 @@ class RedirectUriValidator < ActiveModel::EachValidator
   def invalid_ssl_uri?(uri)
     forces_ssl = Doorkeeper.configuration.force_ssl_in_redirect_uri
     forces_ssl && uri.try(:scheme) == 'http'
+  end
+
+  def allowed_localhost_uri?(uri)
+    allow_localhost = Doorkeeper.configuration.allow_localhost_in_redirect_uri
+    allow_localhost && uri.host == 'localhost'
   end
 end
--- a/app/validators/redirect_uri_validator.rb
+++ b/app/validators/redirect_uri_validator.rb
@@ -14,7 +14,9 @@ class RedirectUriValidator < ActiveModel::EachValidator
         return if native_redirect_uri?(uri)
         record.errors.add(attribute, :fragment_present) unless uri.fragment.nil?
         record.errors.add(attribute, :relative_uri) if uri.scheme.nil? || uri.host.nil?
-        record.errors.add(attribute, :secured_uri) if invalid_ssl_uri?(uri) && !allowed_localhost_uri?(uri)
+        if invalid_ssl_uri?(uri) && !allowed_localhost_uri?(uri)
+          record.errors.add(attribute, :secured_uri)
+        end
       end
     end
   rescue URI::InvalidURIError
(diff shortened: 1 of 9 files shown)


tute:
Hey @TMSCH! Thank you very much for your submission!

> Addition of a `allow_localhost_in_redirect_uri` option which allows to specify a localhost (e.g. http://localhost:3000/callback) even when `force_ssl_in_redirect_uri` is true. This has been a use case in our situation where an internal OAuth provider could be used for local dev (just as you can use Google OAuth for local dev too)

Default value for `force_ssl_in_redirect_uri` (if you don't set it to anything) is `!Rails.env.development?`. I assume that if you are in localhost you are in development environment, in which case you wouldn't need the new option that you add. Is that the case? If not, what is your situation, and why current doorkeeper design doesn't cut it for you?

> An issue when upgrading from Doorkeeper 2.2.0 to 2.2.2 where attr_accessible was added to model mixins. In the included block, it tests for respond_to?(:attr_accessible) before using it but Rails 4.0.x still provides the method and raises a deprecation error... I've changed the condition to fix that.

Rails 4.0 is not end-of-life'd, and is thus insecure. Doorkeeper doesn't support that version.

---

These notes make me think that your PR is not useful to add to doorkeeper as is. What do you think?

Thank you again for your time and work!

TMSCH:
Hi @tute !

Thanks for having a look at my PR!

I understand that you don't want to support Rails 4.0, though the changes required are small so I think it can always be worth it.

For the first point, I didn't explain clearly what I want to achieve. I'm talking about a use case where you want to use a **production** OAuth server (with Doorkeeper) to authenticate a **local** environment. We have a SOA architecture and we don't want people to have to setup every single service on their machine when they can use a remote instance. Our internal authentication service relies on Doorkeeper, and ideally, our developers should be able to authenticate through this service while working on their local environment (think about when you're using Google OAuth on a local machine, you can set it up so it redirects to localhost).

We _could_ deactivate `force_ssl_in_redirect_uri` but then any not secure address could be used in the production instance when we only want to authorize the use of `localhost`address.

Does that make more sense?

Thanks!

tute:
> I understand that you don't want to support Rails 4.0, though the changes required are small so I think it can always be worth it.

This small change has been [a source of pain](https://github.com/merit-gem/merit/search?q=attr_accessible&type=Issues&utf8=%E2%9C%93) in other OSS projects. I'm happy to not need to try to solve it anymore. As a library maintainer, I want the software we work on to run as reliably and securely as possible, and that includes filtering end-of-life'd (and thus insecure) dependencies.

> We could deactivate force_ssl_in_redirect_uri but then any not secure address could be used in the production instance when we only want to authorize the use of localhostaddress.

Disabling it is actually a good solution. This is only used as a validator for admins editing allowed applications (people with access to `/oauth/*` endpoints; this is you and your team mates. If you disable this option, it is your job (and not the software's) to enforce SSL whenever it makes sense, and not use it for, say, `localhost`. You are disabling a good check, because it makes sense in your scenario (but not most).

Does this make sense to you? Would it solve your problem?

Thanks!

TMSCH:
That's true indeed. Thanks for reviewing the PR! I'm not really fond of disabling the SSL requirement completely in our use case, but I understand the general use of Doorkeeper doesn't need this feature - and the Rails 4.0.x fix doesn't look very clean indeed.

tute:
Thanks! If we find more use cases for the `localhost` check, we can discuss again. Thanks for your work!

jredburn:
We're running into this need as well in order to support installed applications (a la https://developers.google.com/identity/protocols/OAuth2InstalledApp). Some developers we're working with have a strong preference for localhost vs the oob URN but https complicates things. We'd like to require https for any non-localhost redirect URIs but still allow a redirect URI like http://localhost:54321.

jcarbo:
@jredburn @TMSCH - A little late here but we have the same need and figured out a relatively painless workaround so thought I'd share.

The `force_ssl_in_redirect_uri ` is only enforced on save so you can just set it via the console and skip the validation. Either:
```ruby
# a) Using update_column skips validations
application.update_column(:redirect_uri, 'http://localhost:3000')

# b) Set the value and save without validation
application.redirect_uri = 'http://localhost:3000'
application.save(:validate => false)
```

Some drawbacks:
- Has to be done manually in the terminal.
- The validation is run on every save, so any future updates to the application (e.g. name change) would fail or need to be done without validations.

PR #741, opened 2015-11-05
Use application scopes as default if no global default scope

adstage-david:
This allows scopes defined on applications that are available outside of
the globally defined default/optional scope list (as allowed from #678)
to still show up in the default scopes for resource owners going through an
authorization flow. The application scope list is chosen as the scope list if:

1. There is no scope defined in the request and a default must be determined
2. There is no global Doorkeeper.default_scope option configured
3. The application is configured with a list of scopes.

I think would be considered backwards incompatible, so I could see the case for adding a config option (or a special value that can be passed to the existing default_scopes config option) to enable this feature. I could add that if that's preferable. 

My particular use case for wanting this change is I have some applications relying on default scopes right now and have added new scopes I want to only give to a subset of those applications. I upgraded to the latest Doorkeeper to support application specific scopes, but now I'm stuck because I will not be able to access these scopes without going back to edit these clients and pass them in (unless this PR is merged). I feel like it's logical if you have application specific scopes to pick those as default when no scope is defined, but if not that's okay.

### DIFF
--- a/lib/doorkeeper/oauth/pre_authorization.rb
+++ b/lib/doorkeeper/oauth/pre_authorization.rb
@@ -29,7 +29,7 @@ module Doorkeeper
       end
 
       def scope
-        @scope.presence || server.default_scopes.to_s
+        @scope.presence || default_scope
       end
 
       def error_response
@@ -37,6 +37,14 @@ module Doorkeeper
       end
 
       private
+
+      def default_scope
+        if server.default_scopes.empty? && client.application.scopes.present?
+          client.application.scopes.to_s
+        else
+          server.default_scopes.to_s
+        end
+      end
 
       def validate_response_type
         server.authorization_response_types.include? response_type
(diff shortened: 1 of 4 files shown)


ryansch:
:+1:

ryansch:
This PR doesn't work for client credentials flow.

tute:
> This PR doesn't work for client credentials flow.

How exactly does it not? And when we find, we should add the regression tests. Thanks!

adstage-david:
@ryansch - could you be more specific? Happy to dig in and fix, but details on how it's not working would be helpful. Is it just that if you sign up it doesn't take the default scopes like when you're doing a resource owner based flow, or does it cause a breaking change?

ryansch:
I've adopted an alternate approach which adds a `default_token_scopes` column as this PR didn't give me enough flexibility.  I'll be crafting a PR next week.

tute:
> I've adopted an alternate approach which adds a default_token_scopes column as this PR didn't give me enough flexibility. I'll be crafting a PR next week.

ping @ryansch. Thanks!

tute:
@adstage-david do you still need this PR in doorkeeper? Is there anyway we can solve your use case without adding this change into the library? Thanks!

tute:
Closing due to lack of activity. Please reopen if you'd like to continue working on this. Thank you!

PR #745, opened 2015-11-17
Add created_at to token generation options

ryansch:
I'm using doorkeeper-jwt to add a JWT as my access token.  I'd like to support stateless auth but I need to be able to use the `iat` (issued at) and `exp` (expires at) claims and I need them to match doorkeeper's internal state.  This PR ensures the `created_at` and `updated_at` are set on the access token before generating the token string itself.

Example use:
```ruby
Doorkeeper::JWT.configure do
  token_payload do |opts|
    {
      iat: opts[:created_at].to_i,
      exp: opts[:created_at].to_i + opts[:expires_in],
      sub: opts[:resource_owner_id]
    }
  end
end
```

### DIFF
--- a/lib/doorkeeper/models/access_token_mixin.rb
+++ b/lib/doorkeeper/models/access_token_mixin.rb
@@ -129,10 +129,16 @@ module Doorkeeper
     end
 
     def generate_token
+      self.created_at ||= Time.now.utc
+
       generator = Doorkeeper.configuration.access_token_generator.constantize
-      self.token = generator.generate(resource_owner_id: resource_owner_id,
-                                      scopes: scopes, application: application,
-                                      expires_in: expires_in)
+      self.token = generator.generate(
+        resource_owner_id: resource_owner_id,
+        scopes: scopes,
+        application: application,
+        expires_in: expires_in,
+        created_at: created_at
+      )
     rescue NoMethodError
       raise Errors::UnableToGenerateToken, "#{generator} does not respond to `.generate`."
     rescue NameError
--- a/spec/models/doorkeeper/access_token_spec.rb
+++ b/spec/models/doorkeeper/access_token_spec.rb
@@ -106,6 +106,23 @@ module Doorkeeper
         expect(token.token).to eq 'custom_generator_token_7200'
       end
 
+      it 'allows the custom generator to access the created time' do
+        module CustomGeneratorArgs
+          def self.generate(opts = {})
+            "custom_generator_token_#{opts[:created_at].to_i}"
+          end
+        end
+
+        Doorkeeper.configure do
+          orm DOORKEEPER_ORM
+          access_token_generator "Doorkeeper::CustomGeneratorArgs"
+        end
+
+        token = FactoryGirl.create :access_token
+        created_at = token.created_at
+        expect(token.token).to eq "custom_generator_token_#{created_at.to_i}"
+      end
+
       it 'raises an error if the custom object does not support generate' do
         module NoGenerate
         end
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,7 @@ User-visible changes worth mentioning.
 
 - [#823] Make configuration and specs ORM independent
 - [#777] Add support for public client in password grant flow
+- [#745] Add created_at timestamp to token generation options
 
 ---
 


lib/doorkeeper/models/access_token_mixin.rb
```diff
@@ -129,14 +129,37 @@ def generate_refresh_token
...
+          # Mongoid will not set the updated_at automatically if the created_at is set manually
+          self.updated_at = current_time
+        end
+      end
+    end
+
+    def current_time
+      if self.class.respond_to? :default_timezone
+        # Duplicate behavior from Rails 4
+        self.class.default_timezone == :utc ? Time.now.utc : Time.now
```
ryansch: Normally I'd agree with the bot, but this is straight out of Rails 4.2 ActiveRecord.

morganick:
:+1:

lib/doorkeeper/models/access_token_mixin.rb
```diff
@@ -129,14 +129,42 @@ def generate_refresh_token
...
+      )
     rescue NoMethodError
       raise Errors::UnableToGenerateToken, "#{generator} does not respond to `.generate`."
     rescue NameError
       raise Errors::TokenGeneratorNotFound, "#{generator} not found"
     end
+
+    def set_timestamps
+      if created_at.nil?
+        self.created_at = current_time
```
tute: Isn't this done automatically by the framework?

lib/doorkeeper/models/access_token_mixin.rb
```diff
@@ -129,14 +129,42 @@ def generate_refresh_token
...
+
+        if respond_to? :updated_at=
```
tute: Why do we need this `respond_to?` condition?

lib/doorkeeper/models/access_token_mixin.rb
```diff
@@ -129,14 +129,42 @@ def generate_refresh_token
...
+          # Mongoid will not set the updated_at automatically if the
+          # created_at is set manually
```
tute: Question repeats: why can't we leave them to be updated automatically?

lib/doorkeeper/models/access_token_mixin.rb
```diff
@@ -129,14 +129,42 @@ def generate_refresh_token
...
-      self.token = generator.generate(resource_owner_id: resource_owner_id,
-                                      scopes: scopes, application: application,
-                                      expires_in: expires_in)
+      self.token = generator.generate(
+        resource_owner_id: resource_owner_id,
+        scopes: scopes,
+        application: application,
+        expires_in: expires_in,
+        created_at: created_at
+      )
```
tute: :+1:

lib/doorkeeper/models/access_token_mixin.rb
```diff
@@ -129,14 +129,42 @@ def generate_refresh_token
...
+
+    def current_time
+      if self.class.respond_to? :default_timezone
+        # Duplicate behavior from Rails 4
+        self.class.default_timezone == :utc ? Time.now.utc : Time.now
+      else
+        # Mongoid is happy with this
+        Time.now.utc
+      end
+    end
```
tute: If we go with [UTC by default](https://github.com/doorkeeper-gem/doorkeeper/pull/730), we can remove all this method.

spec/models/doorkeeper/access_token_spec.rb
```diff
@@ -85,6 +85,23 @@ def self.generate(opts = {})
...
+
+        Doorkeeper.configure do
+          orm DOORKEEPER_ORM
+          access_token_generator "Doorkeeper::CustomGeneratorArgs"
+        end
+
+        token = FactoryGirl.create :access_token
+        created_at = token.created_at
+        expect(token.token).to eq "custom_generator_token_#{created_at.to_i}"
+      end
```
tute: :+1:

tute:
Left many little comments. In general, I'm good with adding the `created_at` option before calling the generator, but not with the other changes.

lib/doorkeeper/models/access_token_mixin.rb
ryansch: Normally yes but at the point when we call `generate_token` the timestamps have not yet been generated by rails and are `nil`.  For something like JWT where other systems are making decisions based on the `created_at` time burned into the token, it's important that doorkeeper's db agrees with the JWT token.

lib/doorkeeper/models/access_token_mixin.rb
ryansch: The Active Record model does not have an `updated_at` column but the Mongoid model does.

lib/doorkeeper/models/access_token_mixin.rb
ryansch: See above about `nil`.

lib/doorkeeper/models/access_token_mixin.rb
ryansch: That's fine with me.  As it says in that comment I'm just using the exact same code that Rails does when it autogenerates the timestamps.

ryansch:
@tute See responses one why we have to generate `created_at` to avoid `nil`.

lib/doorkeeper/models/access_token_mixin.rb
tute: > Normally yes but at the point when we call generate_token the timestamps have not yet been generated by rails and are nil. For something like JWT where other systems are making decisions based on the created_at time burned into the token, it's important that doorkeeper's db agrees with the JWT token.

:+1:

lib/doorkeeper/models/access_token_mixin.rb
tute: > The Active Record model does not have an updated_at column but the Mongoid model does.

Do you personally use mongo? I'd say that tokens are never update and should not have this column in any ORM (or have it as an extension); or that indeed we do need this data, in which case it should be incorporated into doorkeeper ("main").

tute: ping @ryansch

konsti:
+1 Any progress? Can I help?

lib/doorkeeper/models/access_token_mixin.rb
tute: To add context by mail, my ping was regarding the last pending comment, which discusses the necessity of `if respond_to? :updated_at=` in doorkeeeper. Thanks!

ryansch:
Oh wow.  Apparently all of these emails were buried.  I'll try to spend some time on this soon.  :christmas_tree: is around the corner so no guarantees this week.

lib/doorkeeper/models/access_token_mixin.rb
ryansch: The problem lies in the fact that doorkeeper has a pluggable ORM interface.  We can either play it safe and check for the existence of an `updated_at` field to ensure it gets set or we have to provide hooks for the ORM adapter to do the right thing here.

tute: > We can either play it safe and check for the existence of an updated_at field to ensure it gets set or we have to provide hooks for the ORM adapter to do the right thing here.

I don't think it makes sense for tokens to have an `updated_at` field, and I'd thus remove it from this PR. What hooks do you think external ORMs would need here?

tute:
@ryanjones interested in merging this PR in? Would you drop the code related with `updated_at`?

ryansch:
👋

tute:
@ryansch interested in merging this PR in? Would you drop the code related with updated_at?

ryansch:
The problem is here: https://github.com/doorkeeper-gem/doorkeeper-mongodb/blob/master/lib/doorkeeper/orm/mongoid5/access_token.rb#L9

That causes the token to have an `updated_at` field.  People are still using that gem and I'd prefer to keep the two lines of code to maintain the proper behavior.

The other option would be to open an issue against that gem to let them know that the parent doorkeeper gem doesn't expect or update an `updated_at` field on the token.

ryansch:
ping @tute ^

tute:
> The problem is here: https://github.com/doorkeeper-gem/doorkeeper-mongodb/blob/master/lib/doorkeeper/orm/mongoid5/access_token.rb#L9
> 
> That causes the token to have an updated_at field. People are still using that gem and I'd prefer to keep the two lines of code to maintain the proper behavior.
> 
> The other option would be to open an issue against that gem to let them know that the parent doorkeeper gem doesn't expect or update an updated_at field on the token.

That gem is already incompatible with current doorkeeper (https://github.com/doorkeeper-gem/doorkeeper-mongodb/issues/9). It can use the MongoId `able_to_set_updated_at` method to not try to update the non-existing attribute. Let's drop the code related with `updated_at`. Thank you!

ryansch:
You got it.

lib/doorkeeper/models/access_token_mixin.rb
```diff
@@ -129,14 +129,36 @@ def generate_refresh_token
...
     rescue NoMethodError
       raise Errors::UnableToGenerateToken, "#{generator} does not respond to `.generate`."
     rescue NameError
       raise Errors::TokenGeneratorNotFound, "#{generator} not found"
     end
+
+    def set_timestamps
+      if created_at.nil?
+        self.created_at = current_time
+      end
```
tute: This method can be rewritten as `self.created_at ||= current_time`. In that case, we don't need it as a method, and I'd be comfortable adding it verbatim to line 132 where we call `set_timestamps`. What do you think?

lib/doorkeeper/models/access_token_mixin.rb
tute: I see it comes from Rails, and I still feel rather uncomfortable because I don't understand it. In the case `self.class.default_timezone` is actually `utc`, wouldn't `Time.now.utc == Time.now`? And, what would happen if `:utc` were a string instead.

I don't want to change behavior to client Rails applications, but I feel so compelled to doing always `Time.now.utc` at the same time...

tute: Calling in @elle who has been thinking about timezone configuration in Rails apps for her input on these code changes ^.

elle: I would just use `Time.current` and if you need that in UTC then `Time.current.utc`

General rule of thumb to never use `Time.now` which is system time, not app time (i.e. the local machine's system time) and you are just setting yourself up to inconsistencies with time and zones.

And my guess is that using `Time.current` can probably eliminate a lot of that code. Maybe.

lib/doorkeeper/models/access_token_mixin.rb
elle: Also +1 on using UTC which is the default in Rails when saving timestamps in the database.

lib/doorkeeper/models/access_token_mixin.rb
tute: Thank you, Elle! Doorkeeper uses `Time.now.utc` instead of `Time.current` to play well with other frameworks.

I'd do `Time.now.utc` always for the `created_at` column then, @ryansch, and we can simplify some code in this file.

tute:
We have three items to go:

* Address the comment:
  > This method can be rewritten as `self.created_at ||= current_time`. In that case, we don't need it as a method, and I'd be comfortable adding it verbatim to line 132 where we call `set_timestamps`. What do you think?
* Add a comment to the [NEWS file](https://github.com/doorkeeper-gem/doorkeeper/blob/master/NEWS.md)
* Squash commits together

Then we merge. Thank you! :)

ryansch:
Acknowledged.

lib/doorkeeper/models/access_token_mixin.rb
ryansch: I agree that the use of `Time.now` in Rails is surprising.  I'll change it to `Time.now.utc`.

ryansch:
I think I've addressed all of your comments now @tute.

tute:
Thank you! 🎉 😃

PR #747, opened 2015-11-18
Specify default error translation key

ryansch:
I ran into this while working on my custom token generation.

- I accidentally triggered a `NoMethodError` which was rescued and wrapped here: https://github.com/doorkeeper-gem/doorkeeper/blob/master/lib/doorkeeper/models/access_token_mixin.rb#L137
- This was then fed into https://github.com/doorkeeper-gem/doorkeeper/blob/master/app/controllers/doorkeeper/tokens_controller.rb#L9 and I got an error message from I18n about a missing translation key.

This PR adds a default to prevent that from happening.


tute:
Thank you! :clap: :smiley:

Issue #746, opened 2015-11-17
Response with a custom error


ezilocchi:
Hi,

I'm working on a feature where I need to perform some custom logic as part of the user authentication process. As result of this extra logic I might need to stop the process and notify the user a situation (basically that the authentication has failed, but for a custom reason). In order to perform this I need to return an HTTP error and a message that need to be presented to the user.

The easiest way I found of saying to `Doorkeeper` that something went wrong during the authentication process, was raising a `Doorkeeper::Errors::DoorkeeperError` inside the `resource_owner_authenticator` block in the `config/initializers/doorkeeper.rb` file. The problem is that I can't specify the message I need to present to the user.

I looking into the code I found that doing this small change I was able to specify a message for my error

```
diff --git a/lib/doorkeeper/helpers/controller.rb b/lib/doorkeeper/helpers/controller.rb
index 71e33a2..ecdefd5 100644
--- a/lib/doorkeeper/helpers/controller.rb
+++ b/lib/doorkeeper/helpers/controller.rb
@@ -45,6 +45,8 @@ module Doorkeeper
                        :invalid_request
                      when Errors::InvalidGrantReuse
                        :invalid_grant
+                     else
+                       exception.message
                      end
```

`config/initializers/doorkeeper.rb`
```ruby
  resource_owner_from_credentials do |routes|
    raise Doorkeeper::Errors::DoorkeeperError.new('my_custom_message')
  end
```

`config/locales/doorkeeper.en.yml`
```yml
en.doorkeeper.errors.messages.my_custom_message: Some custom error message.
```

This is a hacky solution I know, but I would like to know if there is a better way or if it is worth to spend more time looking for a better one and then create a PR for it.

Thanks,
Emiliano.-

tute:
I'm not against this. I think it's worth it to discuss in a PR, and we might merge either this solution or a better one if we find one.
Thanks!

tute:
Thanks! Closing in favor of PR.

PR #749, opened 2015-11-19
Allow user to raise authorization error with a custom message

ezilocchi:
Adding the `otherwise` option on authorization error handler in order to allow users to add custom messages for custom authentication failures. The idea is that under `resource_owner_authenticator` block a user can `raise Doorkeeper::Errors::DoorkeeperError.new('custom_message')`.

This is related to: #746

Example:

`config/initializers/doorkeeper.rb`
```ruby
  resource_owner_from_credentials do |routes|
    # Some custom logic that needs to happen before raise the error
    raise Doorkeeper::Errors::DoorkeeperError.new('my_custom_message')
    # Some custom logic that should not be execute is the error happens
  end
```

`config/locales/doorkeeper.en.yml`
```yml
en.doorkeeper.errors.messages.my_custom_message: Dead customer some custom authentication logic failed.
```

### DIFF
--- a/lib/doorkeeper/helpers/controller.rb
+++ b/lib/doorkeeper/helpers/controller.rb
@@ -45,6 +45,8 @@ module Doorkeeper
                        :invalid_request
                      when Errors::InvalidGrantReuse
                        :invalid_grant
+                     when Errors::DoorkeeperError
+                       exception.message
                      end
 
         OAuth::ErrorResponse.new name: error_name, state: params[:state]
--- a/spec/controllers/tokens_controller_spec.rb
+++ b/spec/controllers/tokens_controller_spec.rb
@@ -31,6 +31,34 @@ describe Doorkeeper::TokensController do
     end
   end
 
+  describe 'when there is a failure due to a custom error' do
+    it 'returns the error response with a custom message' do
+      # NOTE: I18n will look for this key `en.doorkeeper.errors.messages.my_message` in a locale file
+      allow(I18n).to receive(:translate).
+        with("my_message", hash_including(scope: [:doorkeeper, :errors, :messages])).
+        and_return('This is my authorization custom message')
+
+      doorkeeper_error = Doorkeeper::Errors::DoorkeeperError.new('my_message')
+
+      strategy = double(:strategy)
+      allow(strategy).to receive(:authorize).and_raise(doorkeeper_error)
+
+      allow(controller).to receive(:server) { double(token_request: strategy) }
+
+      post :create
+
+      expect(response.status).to eq 401
+      expect(response.headers['WWW-Authenticate']).to match(/Bearer/)
+
+      expected_response_body = {
+        "error"             => "my_message",
+        "error_description" => "This is my authorization custom message"
+      }
+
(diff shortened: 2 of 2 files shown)


config/locales/en.yml
```diff
@@ -102,6 +102,8 @@ en:
           expired: "The access token expired"
           unknown: "The access token is invalid"
 
+        my_message: 'This is my authorization custom message'
```
tute: The value and the key are custom. This file shouldn't be changed then, right?

lib/doorkeeper/helpers/controller.rb
```diff
@@ -45,6 +45,8 @@ def get_error_response_from_exception(exception)
                        :invalid_request
                      when Errors::InvalidGrantReuse
                        :invalid_grant
+                     else
+                       exception.message
```
tute: Should we put this under a `when Errors` condition? Cause this will be tracking _any_ exception, and we don't want to [`catch Exception`](http://stackoverflow.com/a/10048406/356060). We will want to add test coverage for that, too.

ezilocchi: Agree. You mean something like this right?

```ruby
...
when Errors::DoorkeeperError
  exception.message
end
```
So if a different error is raised we bubble it up.

config/locales/en.yml
ezilocchi: Agree I've already thought about it. What about adding a new file in the same folder `en_custom_errors.yml` with:
```yml
my_message: 'This is my authorization custom message'
```

config/locales/en_custom_errors.yml
```diff
@@ -0,0 +1,5 @@
+en:
+  doorkeeper:
+    errors:
+      messages:
+        my_message: 'This is my authorization custom message'
```
tute: > What about adding a new file in the same folder `en_custom_errors.yml`

I'd say this is not needed. Can we mock the `I18n.t` call in tests, so we don't need to add this far-away file for that one particular spec?

tute:
Looking great, thank you!

config/locales/en_custom_errors.yml
ezilocchi: Makes sense. I will change this.

ezilocchi:
@tute All comments addressed.

tute:
Thank you!

ezilocchi:
@tute, just curious. Are you planning to include this PR in a future release? Or you found something that make you change your mind?

tute:
I rebased your commit, touched a bit the specs, and merged it into master, keeping you as author. The new behavior you added will be in the upcoming patch or minor release (not sure when will I release it yet).

Thank you!

tute:
https://github.com/doorkeeper-gem/doorkeeper/commit/fa4820ce2b6d33838c113fe3bb8b1584a45d0a1d is in master now.

ezilocchi:
Thank you!

icem:
Is this feature documented anywhere?

tute:
> Is this feature documented anywhere?

It is not. Anyone up for adding it to the README in a PR?

PR #756, opened 2015-11-25
Implement Device OAuth flow

kamal:
This implementation is based on Facebook and Google's Device OAuth flow

https://developers.facebook.com/docs/facebook-login/for-devices

https://developers.google.com/identity/protocols/OAuth2ForDevices?hl=en

Putting this out here for feedback. This is of course not part of the official spec but I couldn't find good resources on implementing doorkeeper extensions.


tute:
@kamal, thanks so much for sharing your work!

> This is of course not part of the official spec but I couldn't find good resources on implementing doorkeeper extensions.

There are two examples for extensions:

https://github.com/doorkeeper-gem/doorkeeper-mongodb incorporates doorkeeper "main" as a git submodule, and it's useful to run the original test suite across the different ORMs.

https://github.com/doorkeeper-gem/doorkeeper-grants_assertion is an extension, which is probably much more alike to the feature set you are planning on building here.

Is this useful?

Thank you, again! :heart: :smiley:

rectifyer:
This looks sweet! This will hep a ton with authentication on devices without text entry. Are there plans to accept this PR or will it be created as an outside extension?

tute:
> Are there plans to accept this PR or will it be created as an outside extension?

This should be maintained as an outside extension, following the two examples mentioned in https://github.com/doorkeeper-gem/doorkeeper/pull/756#issuecomment-159955599 (or better if we find any).

Do you want to create one, @kamal?

Thank you!

warmwaffles:
I may take a swing at pulling that out into an extension. It looks like it should be pretty straight forward to extract the changes into a gem.

warmwaffles:
For those interested, I did just rebase this off of the master branch upstream.

Update on the making this into its own extension:

Running in to some issues with the `lib/doorkeeper/config.rb` changes since those were on the doorkeeper config itself. I'm assuming I _could_ do the following, but it isn't pretty.

```rb
Doorkeeper::Config.option(:device_verification_url)
Doorkeeper::Config.option(:device_polling_interval, default: 5)
```

I have not tested this yet. Just looking for a little guidance on how this could be easily accomplished without over complicating the simple plug and play.

rectifyer:
@warmwaffles did you extract this into its own extension? I'm happy to test it out in my app if so.

warmwaffles:
I was not able to, but I may give it a shot again today. Though my question above still stands. I'm wondering if the config should just be apart of the extension itself instead of trying to be apart of the main config.

rectifyer:
I'm not exactly sure, but it appears as thought the config would be an extension to the main Doorkeeper config.

Looking the examples at 
https://github.com/doorkeeper-gem/doorkeeper-grants_assertion
https://github.com/doorkeeper-gem/doorkeeper-grants_assertion/blob/master/lib/doorkeeper/grants_assertion.rb

warmwaffles:
monkey patching ftw. Cool, I'll take a swing at this in a bit

warmwaffles:
If you feel like wanting to help our / critique / make it better head on over to https://github.com/envoy/doorkeeper-device_flow and I could add you as a collaborator @rectifyer

rectifyer:
@warmwaffles ok cool, yeah I'll help however I can.

rectifyer:
I tried to test this out today and so far I don't know enough to get it working. I was able to do a little more monkey patching to get the extension loaded, but after that I'm not sure where to go. Do you have any examples for the config or anything else I need to start testing this?

warmwaffles:
Yea I semi stopped porting it due to other outside factors. I'll take another crack at it soon though.

rectifyer:
@warmwaffles no worries at all. I'm actually making some progress on my end with some monkey patching. I got the new routes added in (you had a FIXME) and I'm going continue working on this tomorrow. I'll make a new branch and check in once I have something working.

tute:
Thanks for your work! Closing this PR, as we can continue discussing in https://github.com/envoy/doorkeeper-device_flow. Feel free to loop me in if we can make changes in doorkeeper to ease the work of extending it. 👏  ❤️

PR #762, opened 2015-12-09
Check should not abort the actual migration

arion:
You already did it in this pull request (https://github.com/doorkeeper-gem/doorkeeper/commit/0b3284f39f776857cfb4ed6d0bf3200350d86d5e). Please repeat for 3.0.1 version.


arion:
In my opinion: Versus this changes we could just add new condition `$0 =~ /rake$/` in checking. Then all rake jobs will be pass. What do you think?

tute:
> In my opinion: Versus this changes we could just add new condition $0 =~ /rake$/ in checking. Then all rake jobs will be pass. What do you think?

That could work, but this is good. Thank you!

PR #769, opened 2015-12-29

Fredar:
Creating another PR based on https://github.com/doorkeeper-gem/doorkeeper/pull/691. Copying description:

>This PR allows refresh tokens to not either revoke at some point in the future (if you want to allow some wiggle room for network failures, etc) set refresh_token_revoked_in to some number of seconds in the future.

>Also allows for refresh tokens to not be revoked until an access token created with that refresh token is successfully used once. Set refresh_token_revoked_on_use to true.

>Both are off by default.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,12 @@ User-visible changes worth mentioning.
 
 - Toughen parameters filter with exact match
 - Drop `attr_accessible` from models
+- [#769] Revoke refresh token on access token use. To make use of the new config
+  add `previous_refresh_token` column to `oauth_access_tokens`:
+
+  ```
+  rails generate doorkeeper:previous_refresh_token
+  ```
 
 ### Backward incompatible changes
 - Force all timezones to use UTC to prevent comparison issues.
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -236,6 +236,18 @@ doorkeeper.
       @token_grant_types ||= calculate_token_grant_types
     end
 
+    def refresh_token_revoked_on_use?
+      unless @refresh_token_revoked_on_use.nil?
+        return @refresh_token_revoked_on_use
+      end
+
+      @refresh_token_revoked_on_use =
+        ActiveRecord::Base.connection.column_exists?(
+          :oauth_access_tokens,
+          :previous_refresh_token
+        )
+    end
+
     private
 
     # Determines what values are acceptable for 'response_type' param in
--- a/lib/doorkeeper/oauth/refresh_token_request.rb
+++ b/lib/doorkeeper/oauth/refresh_token_request.rb
@@ -36,9 +36,13 @@ module Doorkeeper
           refresh_token.lock!
           raise Errors::InvalidTokenReuse if refresh_token.revoked?
 
-          refresh_token.revoke
+          refresh_token.revoke unless refresh_token_revoked_on_use?
           create_access_token
         end
+      end
+
+      def refresh_token_revoked_on_use?
+        server.refresh_token_revoked_on_use?
       end
 
       def default_scopes
@@ -46,17 +50,25 @@ module Doorkeeper
       end
 
       def create_access_token
-        expires_in = Authorization::Token.access_token_expires_in(
-          server,
-          client
-        )
+        @access_token = AccessToken.create!(access_token_attributes)
+      end
 
-        @access_token = AccessToken.create!(
+      def access_token_attributes
+        {
           application_id: refresh_token.application_id,
           resource_owner_id: refresh_token.resource_owner_id,
           scopes: scopes.to_s,
-          expires_in: expires_in,
-          use_refresh_token: true)
+          expires_in: access_token_expires_in,
+          use_refresh_token: true
+        }.tap do |attributes|
+          if refresh_token_revoked_on_use?
+            attributes[:previous_refresh_token] = refresh_token.refresh_token
+          end
+        end
+      end
+
+      def access_token_expires_in
+        Authorization::Token.access_token_expires_in(server, client)
       end
 
       def validate_token_presence
--- a/lib/doorkeeper/oauth/token.rb
+++ b/lib/doorkeeper/oauth/token.rb
@@ -55,7 +55,9 @@ module Doorkeeper
 
       def self.authenticate(request, *methods)
         if token = from_request(request, *methods)
-          AccessToken.by_token(token)
+          access_token = AccessToken.by_token(token)
+          access_token.revoke_previous_refresh_token! if access_token
+          access_token
         end
       end
     end
--- a/lib/generators/doorkeeper/previous_refresh_token_generator.rb
+++ b/lib/generators/doorkeeper/previous_refresh_token_generator.rb
@@ -1,0 +1,29 @@
+require 'rails/generators/active_record'
+
+class Doorkeeper::PreviousRefreshTokenGenerator < Rails::Generators::Base
+  include Rails::Generators::Migration
+  source_root File.expand_path('../templates', __FILE__)
+  desc 'Support revoke refresh token on access token use'
+
+  def self.next_migration_number(path)
+    ActiveRecord::Generators::Base.next_migration_number(path)
+  end
+
+  def previous_refresh_token
+    if no_previous_refresh_token_column?
+      migration_template(
+        'add_previous_refresh_token_to_access_tokens.rb',
+        'db/migrate/add_previous_refresh_token_to_access_tokens.rb'
+      )
+    end
+  end
+
+  private
+
+  def no_previous_refresh_token_column?
+    !ActiveRecord::Base.connection.column_exists?(
+      :oauth_access_tokens,
+      :previous_refresh_token
+    )
+  end
+end
--- a/lib/generators/doorkeeper/templates/migration.rb
+++ b/lib/generators/doorkeeper/templates/migration.rb
@@ -39,13 +39,21 @@ class CreateDoorkeeperTables < ActiveRecord::Migration
       # https://github.com/doorkeeper-gem/doorkeeper/tree/v3.0.0.rc1#custom-access-token-generator
       #
       # t.text     :token,             null: false
-      t.string   :token,             null: false
+      t.string   :token,                  null: false
 
       t.string   :refresh_token
       t.integer  :expires_in
       t.datetime :revoked_at
-      t.datetime :created_at,        null: false
+      t.datetime :created_at,             null: false
       t.string   :scopes
+
+      # If there is a previous_refresh_token column,
+      # refresh tokens will be revoked after a related access token is used.
+      # If there is no previous_refresh_token column,
+      # previous tokens are revoked as soon as a new access token is created.
+      # Comment out this line if you'd rather have refresh tokens
+      # instantly revoked.
+      t.string   :previous_refresh_token, null: false, default: ""
     end
 
     add_index :oauth_access_tokens, :token, unique: true
--- a/spec/lib/models/revocable_spec.rb
+++ b/spec/lib/models/revocable_spec.rb
@@ -34,4 +34,26 @@ describe 'Revocable' do
       expect(subject).not_to be_revoked
     end
   end
+
+  describe :revoke_previous_refresh_token! do
+    it "revokes the previous token if existing, and resets the
+      `previous_refresh_token` attribute" do
+      previous_token = FactoryGirl.create(
+        :access_token,
+        refresh_token: "refresh_token"
+      )
+      current_token = FactoryGirl.create(
+        :access_token,
+        previous_refresh_token: previous_token.refresh_token
+      )
+
+      expect_any_instance_of(
+        Doorkeeper::AccessToken
+      ).to receive(:revoke).and_call_original
+      current_token.revoke_previous_refresh_token!
+
+      expect(current_token.previous_refresh_token).to be_empty
+      expect(previous_token.reload).to be_revoked
+    end
+  end
 end
--- a/spec/lib/oauth/refresh_token_request_spec.rb
+++ b/spec/lib/oauth/refresh_token_request_spec.rb
@@ -5,7 +5,8 @@ module Doorkeeper::OAuth
     let(:server) do
       double :server,
              access_token_expires_in: 2.minutes,
-             custom_access_token_expires_in: -> (_oauth_client) { nil }
+             custom_access_token_expires_in: -> (_oauth_client) { nil },
+             refresh_token_revoked_on_use?: false
     end
     let(:refresh_token) do
       FactoryGirl.create(:access_token, use_refresh_token: true)
@@ -16,16 +17,15 @@ module Doorkeeper::OAuth
     subject { RefreshTokenRequest.new server, refresh_token, credentials }
 
     it 'issues a new token for the client' do
-      expect do
-        subject.authorize
-      end.to change { client.access_tokens.count }.by(1)
+      expect { subject.authorize }.to change { client.access_tokens.count }.by(1)
       expect(client.reload.access_tokens.last.expires_in).to eq(120)
     end
 
     it 'issues a new token for the client with custom expires_in' do
       server = double :server,
                       access_token_expires_in: 2.minutes,
-                      custom_access_token_expires_in: ->(_oauth_client) { 1234 }
+                      custom_access_token_expires_in: ->(_oauth_client) { 1234 },
+                      refresh_token_revoked_on_use?: false
 
       RefreshTokenRequest.new(server, refresh_token, credentials).authorize
 
@@ -65,6 +65,31 @@ module Doorkeeper::OAuth
       refresh_token.save
       subject.validate
       expect(subject).to be_valid
+    end
+
+    context 'refresh tokens expire on access token use' do
+      let(:server) do
+        double :server,
+               access_token_expires_in: 2.minutes,
+               custom_access_token_expires_in: ->(_oauth_client) { 1234 },
+               refresh_token_revoked_on_use?: true
+      end
+
+      it 'issues a new token for the client' do
+        expect { subject.authorize }.to change { client.access_tokens.count }.by(1)
+      end
+
+      it 'does not revoke the previous token' do
+        subject.authorize
+        expect(refresh_token).not_to be_revoked
+      end
+
+      it 'sets the previous refresh token in the new access token' do
+        subject.authorize
+        expect(
+          client.access_tokens.last.previous_refresh_token
+        ).to eq(refresh_token.refresh_token)
+      end
     end
 
     context 'clientless access tokens' do
--- a/lib/doorkeeper/models/concerns/revocable.rb
+++ b/lib/doorkeeper/models/concerns/revocable.rb
@@ -8,6 +8,23 @@ module Doorkeeper
       def revoked?
         !!(revoked_at && revoked_at <= Time.now.utc)
       end
+
+      def revoke_previous_refresh_token!
+        return unless refresh_token_revoked_on_use?
+        old_refresh_token.revoke if old_refresh_token
+        update_attribute :previous_refresh_token, ""
+      end
+
+      private
+
+      def old_refresh_token
+        @old_refresh_token ||=
+          AccessToken.by_refresh_token(previous_refresh_token)
+      end
+
+      def refresh_token_revoked_on_use?
+        Doorkeeper.configuration.refresh_token_revoked_on_use?
+      end
     end
   end
 end
--- a/lib/generators/doorkeeper/templates/add_previous_refresh_token_to_access_tokens.rb
+++ b/lib/generators/doorkeeper/templates/add_previous_refresh_token_to_access_tokens.rb
@@ -1,0 +1,11 @@
+class AddPreviousRefreshTokenToAccessTokens < ActiveRecord::Migration
+  def change
+    add_column(
+      :oauth_access_tokens,
+      :previous_refresh_token,
+      :string,
+      default: "",
+      null: false
+    )
+  end
+end
--- a/spec/controllers/protected_resources_controller_spec.rb
+++ b/spec/controllers/protected_resources_controller_spec.rb
@@ -28,7 +28,9 @@ describe 'doorkeeper authorize filter' do
 
     let(:token_string) { '1A2BC3' }
     let(:token) do
-      double(Doorkeeper::AccessToken, acceptable?: true)
+      double(Doorkeeper::AccessToken,
+             acceptable?: true, previous_refresh_token: "",
+             revoke_previous_refresh_token!: true)
     end
 
     it 'access_token param' do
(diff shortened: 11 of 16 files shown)


Fredar:
@tute I have updated code for all your comments from earlier PR, except this [one](https://github.com/doorkeeper-gem/doorkeeper/pull/691#discussion_r45563850). I dont think we need to check for that, should be safe to revoke either way.

What about `Time.now` and `Unnecessary spacing` barks in `lib/generators`?
Please have a general look, thanks!

tute:
> I have updated code for all your comments from earlier PR, except this one.
> I dont think we need to check for that, should be safe to revoke either way.
> What do you think?

You are probably right.

> What about `Time.now` and `Unnecessary spacing` barks in `lib/generators`, should we ignore them?

Yep!

> Please have a general look, thanks!

Doing it now. Thank you!

spec/dummy/db/migrate/20150909133655_add_previous_refresh_token_to_access_tokens.rb
```diff
@@ -0,0 +1,5 @@
+class AddPreviousRefreshTokenToAccessTokens < ActiveRecord::Migration
+  def change
+    add_column :oauth_access_tokens, :previous_refresh_token, :string, default: "", null: false
+  end
+end
```
tute: Do you want to update this migration with the one with the new coding style? Thanks!

lib/doorkeeper/oauth/refresh_token_request.rb
```diff
@@ -51,12 +53,16 @@ def create_access_token
...
+          expires_in: expires_in,
+          previous_refresh_token: refresh_token.refresh_token.to_s,
           resource_owner_id: refresh_token.resource_owner_id,
           scopes: scopes.to_s,
-          expires_in: expires_in,
-          use_refresh_token: true)
+          use_refresh_token: true
+        }
+
+        @access_token = AccessToken.create!(create_params)
```
tute: We can also write this as (if you prefer):

```ruby
@access_token = AccessToken.create!(
  application_id: refresh_token.application_id,
  expires_in: expires_in,
  previous_refresh_token: refresh_token.refresh_token.to_s,
  resource_owner_id: refresh_token.resource_owner_id,
  scopes: scopes.to_s,
  use_refresh_token: true
)
```

lib/doorkeeper/oauth/token.rb
```diff
@@ -55,7 +55,9 @@ def self.from_request(request, *methods)
 
       def self.authenticate(request, *methods)
         if token = from_request(request, *methods)
-          AccessToken.by_token(token)
+          access_token = AccessToken.by_token(token)
+          access_token.revoke_previous_refresh_token! if access_token
+          access_token
```
tute: What do you think of writing this as:

```ruby
AccessToken.by_token(token).tap do |access_token|
  access_token.revoke_previous_refresh_token!
end
```

lib/generators/doorkeeper/previous_refresh_token_generator.rb
```diff
@@ -0,0 +1,31 @@
...
+          'add_previous_refresh_token_to_access_tokens.rb',
+          'db/migrate/add_previous_refresh_token_to_access_tokens.rb'
+        )
+      end
+    end
+
+    private
+
+    def no_previous_refresh_token_column?
+      ActiveRecord::Base.connection.column_exists?(
```
tute: This needs a negation.

spec/lib/oauth/refresh_token_request_spec.rb
```diff
@@ -67,6 +69,53 @@ module Doorkeeper::OAuth
       expect(subject).to be_valid
     end
 
+    context 'refresh tokens expire on access token use' do
+      let(:server) do
+        double :server,
+               access_token_expires_in: 2.minutes,
+               custom_access_token_expires_in: ->(_oauth_client) { 1234 },
+               refresh_token_revoked_in: 1.day
+      end
```
tute: Why do we need this double?

spec/lib/oauth/refresh_token_request_spec.rb
```diff
@@ -67,6 +69,53 @@ module Doorkeeper::OAuth
...
+
+      it 'issues a new token for the client' do
+        expect do
+          subject.authorize
+        end.to change { client.access_tokens.count }.by(1)
+      end
+
+      it 'does not revoke the previous token' do
+        subject.authorize
+        refresh_token.revoked?.should eq(false)
```
tute: Suggestion: `expect(refresh_token).not_to be_revoked`

tute:
Left a few comments, this is looking great. Thank you!

spec/dummy/db/migrate/20150909133655_add_previous_refresh_token_to_access_tokens.rb
Fredar: Will do.

lib/doorkeeper/oauth/refresh_token_request.rb
Fredar: Ok

lib/generators/doorkeeper/previous_refresh_token_generator.rb
Fredar: Missed that, thanks!

spec/lib/oauth/refresh_token_request_spec.rb
Fredar: This file needs some refactor love, indeed.

spec/lib/oauth/refresh_token_request_spec.rb
Fredar: Used to trigger the new behaviour.
Without it the spec `it 'does not revoke the previous token' do` will fail.

lib/doorkeeper/oauth/token.rb
Fredar: It wont work, cause `access_token` can be `nil`.
We would have to write 
```
AccessToken.by_token(token).tap do |access_token|
  access_token.try(:revoke_previous_refresh_token!)
end
```

Fredar:
@tute Thanks for the comments.
I have updated or responded where needed.

Fredar:
@tute I think we should instantly revoke token here:
```ruby
    old_refresh_token.revoke_in(refresh_token_revoked_in)
  end
  update_attribute :previous_refresh_token, ""
end
```
I will change the code to:
```ruby
old_refresh_token.revoke
```
This method is only used during authentication.

Fredar:
@tute ping :wink:

Fredar:
Hmm for some reason hound is back with LineLength reporting on specs.

tute:
Thank you, @Fredar! I rebased and squashed your work into: https://github.com/doorkeeper-gem/doorkeeper/commit/a091053adc5d64b2ae2309e0ae5a6bb29cff7b77, please check the commit message.

A question arised again: why is it, again, that we need the `refresh_token_revoked_in` option? If the refresh token is only revoked upon successful use, why would we want to allow end users to configure _when_ that happens? I'm now leaning towards removing the option, and revoking on use.

Thanks!

Fredar:
@tute Thanks, commit message looks good.

You read my mind, as I was wondering the same thing(why we just dont revoke on use?) while refactoring this :smile: 

Anyway I have assumed that for some reason(security?), we should not keep `refresh_tokens` valid indefinitely. 
If we go down this path then every `previous_refresh_token` will be valid until `access_token` is used.
Having this as option allows users to give a bit of flexibility for network issues, but also to set a deadline in which the `previous_refresh_token` can still be used.

What do you think?

tute:
> Anyway I have assumed that for some reason(security?), we should not keep `refresh_tokens` valid indefinitely. If we go down this path then every `previous_refresh_token` will be valid until `access_token` is used. Having this as option allows users to give a bit of flexibility for network issues, but also to set a deadline in which the `previous_refresh_token` can still be used.

This PR doesn't change that, if anything it makes it worse. Refresh tokens are marked for revocation upon usage, before, instantly, and now using the new configuration option.

What I say is that if we undid this diff:

```diff
diff --git a/lib/doorkeeper/oauth/refresh_token_request.rb b/lib/doorkeeper/oauth/refresh_token_request.rb
index 8782571..59bad36 100644
--- a/lib/doorkeeper/oauth/refresh_token_request.rb
+++ b/lib/doorkeeper/oauth/refresh_token_request.rb
@@ -35,8 +35,10 @@ module Doorkeeper
         refresh_token.transaction do
           refresh_token.lock!
           raise Errors::InvalidTokenReuse if refresh_token.revoked?
+          if refresh_token.revoked_at.nil?
+            refresh_token.revoke_in(server.refresh_token_revoked_in)
+          end

-          refresh_token.revoke
```

We wouldn't need that configuration option, either.

Also, now that I inspect this, this is backwards incompatible and not configurable! See:

```diff
diff --git a/lib/doorkeeper/oauth/token.rb b/lib/doorkeeper/oauth/token.rb
index dcb17da..0c40510 100644
--- a/lib/doorkeeper/oauth/token.rb
+++ b/lib/doorkeeper/oauth/token.rb
@@ -55,7 +55,9 @@ module Doorkeeper

       def self.authenticate(request, *methods)
         if token = from_request(request, *methods)
-          AccessToken.by_token(token)
+          access_token = AccessToken.by_token(token)
+          access_token.revoke_previous_refresh_token! if access_token
```

That line should be if users want to use it that way. But for people using refresh tokens as long lived tokens to keep authentication open, doorkeeper won't work any more.

Fredar:
>This PR doesn't change that, if anything it makes it worse. Refresh tokens are marked for revocation upon usage, before, instantly, and now using the new configuration option.

I didnt mean to compare current state with PR state.
I was comparing the option to revoke on use with the option to revoke in the future + on use.

>What I say is that if we undid this diff:
We wouldn't need that configuration option, either.

Well true, the config is only needed for this. Do you propose to not revoke it here in any way?

>Also, now that I inspect this, this is backwards incompatible and not configurable! See:

I thought thats to be expected of this PR, after seeing your comment about adding it to the news https://github.com/doorkeeper-gem/doorkeeper/pull/691#issuecomment-152056581

To clarify:
I have had doubts, if we need the new config as well, but assumed its needed for better customization.
If you feel that removing the config is the right way to go then lets remove it.

Fredar:
@tute Regarding your comment:
>That line should be if users want to use it that way. But for people using refresh tokens as long lived tokens to keep authentication open, doorkeeper won't work any more.

Is this change causing those problems? https://github.com/doorkeeper-gem/doorkeeper/pull/769#issuecomment-170372426

If not, is it enough to just add a conditional like:
```
def self.authenticate(request, *methods)
  if token = from_request(request, *methods)
    access_token = AccessToken.by_token(token)
    access_token.revoke_previous_refresh_token! if access_token && refresh_token_revoked_on_use?
    access_token
  end
end

def refresh_token_revoked_on_use?
  server.refresh_token_revoked_in > 0
end
```

Fredar:
@tute Lets not leave it to be forgotten, please :wink:

tute:
@Fredar sorry about my delay, I've been on vacations. I keep this in mind, thanks for your time on this! :)

Fredar:
@tute No problem, hope you had a great time :smile:

rafaelsales:
@tute The final solution allows a refresh token to be reused forever unless many revoked? I thought the gem as intended to implement what's in the RFC, and according to it, reusing refresh tokens is OK.
If the gem were to only implement the most safe approaches of OAuth, it would not support password sign in, for example, cause these present the same security concerns that a lifetime refresh token might raise.
Google OAuth implementation and others that I've used allow lifetime refresh token, and that's mainly used for servers with a single account an API. The refresh token is stored in a secured constant and reuses for as long as the refresh token is not manually revoked or the OAuth App is not removed.

Fredar:
@tute Are you feeling to give this another chance? :wink: 
I am sure we can work something out.

tute:
Sorry about the delays! I do keep this in mind, but I struggle to make myself time. :-|

tute:
Thanks for your reminder :)

lib/doorkeeper/oauth/refresh_token_request.rb
```diff
@@ -35,8 +35,10 @@ def before_successful_response
         refresh_token.transaction do
           refresh_token.lock!
           raise Errors::InvalidTokenReuse if refresh_token.revoked?
+          if refresh_token.revoked_at.nil?
+            refresh_token.revoke_in(server.refresh_token_revoked_in)
+          end
 
-          refresh_token.revoke
```
tute: I would undo this changeset, dropping both the conditional and the new configuration option, and leaving it's previous behavior of always revoking.

lib/doorkeeper/oauth/refresh_token_request.rb
```diff
@@ -56,7 +58,9 @@ def create_access_token
           resource_owner_id: refresh_token.resource_owner_id,
           scopes: scopes.to_s,
           expires_in: expires_in,
-          use_refresh_token: true)
+          use_refresh_token: true,
+          previous_refresh_token: refresh_token.refresh_token.to_s
```
tute: Why do we need `.to_s`? Can `refresh_token.refresh_token` be nil?

spec/lib/models/revocable_spec.rb
```diff
@@ -33,4 +33,38 @@
...
+      )
+
+      expect_any_instance_of(Doorkeeper::AccessToken).to receive(:revoke).and_call_original
+      current_token.revoke_previous_refresh_token!
+
+      expect(current_token.previous_refresh_token).to be_empty
+      expect(previous_token.reload).to be_revoked
+    end
+
+    it "does not revoke the previous token if already revoked" do
```
tute: Is this important behavior to test, or an implementation detail?

spec/lib/oauth/refresh_token_request_spec.rb
```diff
@@ -5,7 +5,8 @@ module Doorkeeper::OAuth
     let(:server) do
       double :server,
              access_token_expires_in: 2.minutes,
-             custom_access_token_expires_in: -> (_oauth_client) { nil }
+             custom_access_token_expires_in: -> (_oauth_client) { nil },
+             refresh_token_revoked_in: 0
```
tute: Let's drop the `refresh_token_revoked_in` option.

tute:
Yay, I'm back! Left a few questions and comments in the diff. Can you rebase on top of latest master? Thank you!

tute:
About the backwards incompatibility, do you think we can test if the DB column is present, and exercise this new code only in that case? I don't want to make people need to modify their database if they don't intend to use this feature.

tute:
> If not, is it enough to just add a conditional like:
>
> ```ruby
> def self.authenticate(request, *methods)
>   if token = from_request(request, *methods)
>     access_token = AccessToken.by_token(token)
>     access_token.revoke_previous_refresh_token! if access_token && refresh_token_revoked_on_use?
>     access_token
>   end
> end
> 
> def refresh_token_revoked_on_use?
>   server.refresh_token_revoked_in > 0
> end
> ```

After dropping the `server.refresh_token_revoked_in` configuration option, the query method would test if the DB table has that attribute or not. Then this is perfect. 👍

tute:
> If the gem were to only implement the most safe paths of OAuth, it would not support password or client_credentials flows, for example, cause these present the same security concerns that a lifetime refresh token might raise.

@rafaelsales I intend to make doorkeeper as safe by default as possible, while following the spec. By default doorkeeper doesn't enable the password or implicit flows, but it does enable client credentials. Are you saying that it should not? If so, can you please explain your reasoning in more detail? Thank you!

spec/lib/models/revocable_spec.rb
Fredar: It was important with `revoked_in` option, as revoking would push the time forward.
Without the option it doesnt matter so much.

lib/doorkeeper/oauth/refresh_token_request.rb
Fredar: Hmm dont think so, will remove `to_s`

lib/doorkeeper/oauth/refresh_token_request.rb
Fredar: Actually we need the new condition here to not revoke refresh_token if the new config is used.

Fredar:
@tute Rebased and updated the PR with comment changes.
Please review the changes.

groe:
@Fredar Thanks for your efforts, great work!

lib/doorkeeper/config.rb
```diff
@@ -237,6 +236,18 @@ def token_grant_types
...
+    def refresh_token_revoked_on_use?
+      unless @refresh_token_revoked_on_use.nil?
+        return @refresh_token_revoked_on_use
+      end
+
+      @refresh_token_revoked_on_use =
+        ActiveRecord::Base.connection.column_exists?(
+          :oauth_access_tokens,
+          :previous_refresh_token
+        )
```
tute: We can rewrite this method like:

```ruby
@refresh_token_revoked_on_use ||=
  ActiveRecord::Base.connection.column_exists?(
    :oauth_access_tokens,
    :previous_refresh_token
)
```

tute: The condition is in the assignment (`||=`).

NEWS.md
```diff
@@ -3,6 +3,9 @@
 User-visible changes worth mentioning.
 
 ---
+
+- [#769] Revoke Refresh Token on access token use
```
tute: This will go out on rc3. Can you move it in to that subsection? Thanks!

lib/doorkeeper/config.rb
Fredar: @tute Unfortunately ||= will keep recomputing if the returned value is `false`
Thats why I have it explicitly for `nil?`

lib/doorkeeper/oauth/refresh_token_request.rb
```diff
@@ -56,7 +56,9 @@ def create_access_token
           resource_owner_id: refresh_token.resource_owner_id,
           scopes: scopes.to_s,
           expires_in: expires_in,
-          use_refresh_token: true)
+          use_refresh_token: true,
+          previous_refresh_token: refresh_token.refresh_token
```
tute: This will fail on installations that don't have that column in their database.

lib/doorkeeper/oauth/token.rb
```diff
@@ -55,7 +59,11 @@ def self.from_request(request, *methods)
 
       def self.authenticate(request, *methods)
         if token = from_request(request, *methods)
-          AccessToken.by_token(token)
+          access_token = AccessToken.by_token(token)
+          if access_token && refresh_token_revoked_on_use?
+            access_token.revoke_previous_refresh_token!
+          end
```
tute: What do you think of always calling `access_token.revoke_previous_refresh_token!` here, and adding the configuration conditional in the model?

tute:
Looking great @Fredar! Thanks for your work on this.

tute:
I would add new lines in the specs for lines to be shorter than 80 characters.

lib/doorkeeper/config.rb
tute: > @tute Unfortunately ||= will keep recomputing if the returned value is false
> Thats why I have it explicitly for `nil?`

Right, thanks.

lib/doorkeeper/oauth/refresh_token_request.rb
Fredar: Missed that, thanks.

lib/doorkeeper/oauth/token.rb
Fredar: No strong feelings either way.
Will move the conditional to `revocable` module though.

Fredar:
@tute Thanks for quick review!
Updated with your comments. I have only fixed line length for specs that are modified / added by this PR. Also didnt update in two cases where it was 81/80.

lib/generators/doorkeeper/templates/migration.rb
```diff
@@ -39,13 +39,14 @@ def change
...
-      t.string   :token,             null: false
+      t.string   :token,                  null: false
 
       t.string   :refresh_token
       t.integer  :expires_in
       t.datetime :revoked_at
-      t.datetime :created_at,        null: false
+      t.datetime :created_at,             null: false
       t.string   :scopes
+      t.string   :previous_refresh_token, null: false, default: ""
```
tute: Should we add a comment here, prompting developers to comment out this column if they wish to revoke tokens instantly, as it was before? Something like:

`# If you have a `previous_refresh_token` column, refresh tokens will be revoked only after a generated access token is used. If you don't, it will be revoked as soon as a new access token is issued.`

tute: Or is the `refresh_token_revoked_on_use` configuration still a thing? If so, we should add it with a comment to the initializer template.

tute:
Left a question around documentation for developers: how can they disable this option? Where should we document that?

Looking great, thanks for your work!

lib/generators/doorkeeper/templates/migration.rb
Fredar: @tute Comment sounds good.

> Or is the refresh_token_revoked_on_use configuration still a thing? If so, we should add it with a comment to the initializer template.

The `refresh_token_revoked_on_use?` is defined in `config.rb`, but should not be customized by user. The config is set only by presence of `previous_token` column.

Fredar: Is it ok like that? Or you would like to change something?

Fredar:
Thanks again for your time :wink: 
Added a comment to migration template and responded to your question [here](https://github.com/doorkeeper-gem/doorkeeper/pull/769#discussion_r60353888)

lib/generators/doorkeeper/templates/migration.rb
```diff
@@ -39,13 +39,19 @@ def change
...
       t.integer  :expires_in
       t.datetime :revoked_at
-      t.datetime :created_at,        null: false
+      t.datetime :created_at,             null: false
       t.string   :scopes
+
+      # If you have a `previous_refresh_token` column, refresh tokens
+      # will be revoked only after a generated access token is used.
+      # Remove `previous_refresh_token` column, if you would like
+      # to have them revoked as soon as a new access token is issued.
```
tute: This is great! Suggested rewording:

> If there is a `previous_refresh_token` column, refresh tokens will be revoked after a related access token is used. If there is no `previous_refresh_token` column, previous tokens are revoked as soon as a new access token is created. Comment out this line if you'd rather have refresh tokens instantly revoked.

lib/generators/doorkeeper/previous_refresh_token_generator.rb
```diff
@@ -0,0 +1,31 @@
...
+    private
+
+    def no_previous_refresh_token_column?
+      !ActiveRecord::Base.connection.column_exists?(
+        :oauth_access_tokens,
+        :previous_refresh_token
+      )
+    end
+  end
+end
```
tute: Can we add a note in NEWS.md detailing the command to run this generator?

tute:
Two small comments, ready to squash commits together after those! Thank you again. :)

lib/generators/doorkeeper/previous_refresh_token_generator.rb
Fredar: Will do.

Fredar:
@tute Updated migration comment, added news detail and squashed. Cheers!

lib/generators/doorkeeper/previous_refresh_token_generator.rb
```diff
@@ -0,0 +1,31 @@
+require 'rails/generators/active_record'
+
+module Doorkeeper
+  class PreviousRefreshTokenGenerator < Rails::Generators::Base
```
tute: Does `Rails` need a `::` prefix? I got:

```
%   rails generate doorkeeper:previous_refresh_token
[WARNING] Could not load generator "generators/doorkeeper/previous_refresh_token_generator". Error: uninitialized constant Doorkeeper::Rails::Generators.
/Users/tutec/Sites/opensource/doorkeeper-gem/doorkeeper/lib/generators/doorkeeper/previous_refresh_token_generator.rb:4:in `<module:Doorkeeper>'
```

tute:
The migration didn't run in my `spec/dummy` app.

Also, suggestion for the commit message:

> Revoke refresh token after access token use
>
> When `AccessToken#previous_refresh_token` attribute exists, it is populated with the refresh token that issues the access token. This refresh token will be revoked when the access token is authenticated for the first time.
>
> If the attribute doesn't exist in the database, doorkeeper will revoke the refresh token as soon as it issues a new access token, regardless of when is it used.

lib/generators/doorkeeper/previous_refresh_token_generator.rb
Fredar: Something must have changed since the time it was originally created.
Will move module directly into class definition.

Fredar:
Indeed, the migration in dummy app was scheduled before the table creation :wink: 
I have fixed the generator and recreated the migration.

tute:
😃 🎉 ❤️❤️❤️

groe:
Nice work – can't wait to use it in 4.0.0 stable! :+1:

kevinelliott:
I am attempting to use `refresh_token_revoked_on_use` in `config/initializers/doorkeeper.rb` but it is claiming that the setting does not exist. Thoughts here?

nbulaj:
@kevinelliott all you need is just to add 'previous_refresh_token' column to Access Tokens (see Doorkeeper rake tasks, there must be a generator for migration).

hickford:
@Fredar @tute Is there any documentation for the refresh_token_revoked_on_use option? I couldn't find it in https://doorkeeper.gitbook.io/guides/

Fredar:
@hickford Dont think so, but it should be pretty straightforward to use.

Commit msg explains roughly how it works
https://github.com/doorkeeper-gem/doorkeeper/pull/769/commits

There is also a migration generator. (if you installed doorkeeper before it was part of default migration set)

Issue #444, opened 2014-07-30
error code "invalid_resource_owner" is not in the standard


bagilevi:
When using Resource Owner Password Credentials Grant, if the password is invalid, Doorkeeper returns an error code [`invalid_resource_owner`](https://github.com/doorkeeper-gem/doorkeeper/blob/db6dd847827270bcd9ea5e2c8ffd06a25699b6e1/lib/doorkeeper/oauth/password_access_token_request.rb#L9), even though this is not defined in [Section 5.2](http://tools.ietf.org/html/rfc6749#section-5.2). It should be `invalid_grant` according to the standard.

I haven't found any mention of "invalid_resource_owner" on the web other than related to Doorkeeper.

Was there any reason to use this non-standard error code or is it defined somewhere as an extension?

PR #771, opened 2015-12-31
Fix validation error messages

rafaelsales:


tute:
Thank you!

hurty:
@rafaelsales @tute The invalid_resource_owner error is not in the spec. It should be invalid_grant when using Resource Owner Password Credentials Grant.

This has already been discussed and fixed in the past : #444

PR #778, opened 2016-01-16
Made code responses return the actual time that a token is still valid

exop-it:
I've spent the last hour debugging what turned out to be a problem with Doorkeeper.

If the `reuse_access_token` option is configured, then code responses will always return the token's initial `expires_in` value instead of the remaining time the token is still valid. This PR fixes that ;)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,9 @@ User-visible changes worth mentioning.
 
 ---
 
+- [#778] Use the remaining time that a token is still valid when building the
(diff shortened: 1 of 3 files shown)


tute:
Thank you! Can you please add notes to the NEWS.md file, and add a regression test? Thanks again!

irminsul:
Sure, here you go.

tute:
Thank you! Merged in.

PR #780, opened 2016-01-27
Make application relationship optional in access token model

### LINKED ISSUE #774: ActiveRecord::RecordInvalid - Validation failed: Application must exist with Rails 5 and EmberSimpleAuth
I'm using the github gem in Gemfile of Doorkeeper, Rails and Devise.

I'm using Ember Simple Auth to login Using Resource Owner Password Credentials flow. But when I login with correct username (email) and password I have this error:

```
Started POST "/oauth/token" for 127.0.0.1 at 2016-01-03 00:10:46 +0100
Processing by Doorkeeper::TokensController#create as HTML
  Parameters: {"grant_type"=>"password", "username"=>"email@email.com", "password"=>"[FILTERED]"}
  User Load (0.6ms)  SELECT  "users".* FROM "users" WHERE "users"."email" = $1 ORDER BY "users"."id" ASC LIMIT $2  [["email", "email@email.com"], ["LIMIT", 1]]
   (0.2ms)  BEGIN
  Doorkeeper::AccessToken Exists (0.7ms)  SELECT  1 AS one FROM "oauth_access_tokens" WHERE "oauth_access_tokens"."token" = $1 LIMIT $2  [["token", "e7c23c36acaa18b6076e69fefbf5092038b1b76cbde7502c537a98e822f9c107"], ["LIMIT", 1]]
   (0.2ms)  ROLLBACK
Completed 422 Unprocessable Entity in 127ms

ActiveRecord::RecordInvalid - Validation failed: Application must exist:
```

**Why I got this?**

Which version I have to use?

Everything is pretty default, i'm not using nothing than standard defaults.

ashishtajane:
This PR (https://github.com/rails/rails/pull/18937) makes all `belongs_to` relations in rails 5 as `required` by default. But application object need not be present for access token creation in Resource Owner Password Credentials grant flow. So this makes application relationship in access token model as optional.

This fixes #774

### DIFF
--- a/lib/doorkeeper/models/access_token_mixin.rb
+++ b/lib/doorkeeper/models/access_token_mixin.rb
@@ -12,7 +12,8 @@ module Doorkeeper
     included do
       belongs_to :application,
                  class_name: 'Doorkeeper::Application',
-                 inverse_of: :access_tokens
+                 inverse_of: :access_tokens,
+                 optional: true
 
       validates :token, presence: true, uniqueness: true
       validates :refresh_token, uniqueness: true, if: :use_refresh_token?


ashishtajane:
The tests for `rails 4.2` are failing because `optional` keyword is added in `rails 5`.
Is `doorkeeper#master` supporting `rails 4.2` also?

tute:
> Is doorkeeper#master supporting rails 4.2 also?

It is. We should do a conditional there. Yikes.

tute:
Thanks for your work on this! :)

relu:
Since `required` is deprecated in rails 5, I think a better approach that would suite both version groups would be to remove `optional: true` and just go with a validation: `validates :application_id, presence: false`. I tested this with rails master locally and it seems to work just fine. I know it doesn't look pretty but at least it works for both versions without any extra conditional garbage. Will follow-up with a PR.

tute:
Thank you, @ashishtajane and @relu! Merged in a variation of your solutions. :smiley: :clap: :heart:

PR #783, opened 2016-02-05
Explicitly specify full file path and fix grammar

pwim:
Minor change but makes copy & pasting file path easier.


tute:
Yep, thank you! :)

PR #787, opened 2016-02-13
compat with rails 5's default belongs_to presence validation

relu:
Rails 5 adds default presence validation on belongs_to associations:
rails/rails#18233

Application object need not be present for access token creation in Resource
Owner Password Credentials grant flow.

This is an alternative to #780 for backwards compatibility with rails 4

### DIFF
--- a/spec/models/doorkeeper/access_token_spec.rb
+++ b/spec/models/doorkeeper/access_token_spec.rb
@@ -143,6 +143,12 @@ module Doorkeeper
         subject.resource_owner_id = nil
         expect(subject).to be_valid
       end
+
+      it 'is valid without application_id' do
+        # For resource owner credentials flow
+        subject.application_id = nil
+        expect(subject).to be_valid
+      end
     end
 
     describe '#same_credential?' do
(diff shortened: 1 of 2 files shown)


tute:
Thank you, @relu! Merged in a variation of your solution. :smiley: :clap: :heart:

spec/models/doorkeeper/access_token_spec.rb
```diff
@@ -143,6 +143,12 @@ module NoGenerate
         subject.resource_owner_id = nil
         expect(subject).to be_valid
       end
+
+      it 'is valid without application_id' do
+        # For resource owner credentials flow
+        subject.application_id = nil
+        expect(subject).to be_valid
+      end
```
tute: I can't get this spec to fail in Rails 5.0.0.beta3. Why is that?

relu: I see. There is actually no conditional presence validation for this, so it doesn't make sense in it's current form anyway.

PR #793, opened 2016-02-25
Allow Doorkeeper::Models::Scopes#scopes to accept an array

supagroova:
Given that OAuth scopes can be many, it seems more intuitive to accept an array in addition to the current string with spaces.

Specs accompany the PR.

### DIFF
--- a/lib/doorkeeper/models/concerns/scopes.rb
+++ b/lib/doorkeeper/models/concerns/scopes.rb
@@ -4,7 +4,17 @@ module Doorkeeper
       def scopes
         OAuth::Scopes.from_string(self[:scopes])
       end
-
+      
+      # Accept a string or an array of scopes
+      def scopes=(*values)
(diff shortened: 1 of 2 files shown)


tute:
I find it weird that the setter accepts an array but the getter returns a string. Also that the setter allows two different types, I think.

If I started this class today I would probably pick arrays indeed, but I don't think it's worth to change the API by now.

Thank you very much for your suggestion and contribution!

supagroova:
@tute, thanks for the quick response! Just in case it wasn't clear, I intended this not to break the existing API, but just to add what would seem an intuitive use-case to it. If that makes any difference, then great, otherwise no problem.

Thanks for such a great gem!

chadwilken:
I would add that it would be nice to allow passing an array of scopes so that I could add checkboxes for scopes when registering an application so that it didn't allow a free for all.

PR #794, opened 2016-02-26
Mixin application relation spec

relu:


lib/doorkeeper/models/access_token_mixin.rb
```diff
@@ -15,13 +15,14 @@ module AccessTokenMixin
         inverse_of: :access_tokens
       }
       if defined?(ActiveRecord::Base) && ActiveRecord::VERSION::MAJOR >= 5
-        belongs_to_options.merge(optional: true)
+        belongs_to_options.merge!(optional: true)
```
relu: @tute seems hound prefers the above format, should I change it back (it was how my original PR had it)?

lib/doorkeeper/models/access_token_mixin.rb
```diff
@@ -15,13 +15,14 @@ module AccessTokenMixin
...
       end
 
       belongs_to :application, belongs_to_options
 
       validates :token, presence: true, uniqueness: true
       validates :refresh_token, uniqueness: true, if: :use_refresh_token?
+      validates :application_id, presence: true, allow_nil: true
```
relu: I consider this a bit of a stretch, but it was intended to provide verbosity. Feel free to ditch it if needed. Validation based on the `optional` key still stands anyway.

spec/models/doorkeeper/access_token_spec.rb
```diff
@@ -143,6 +143,12 @@ module NoGenerate
         subject.resource_owner_id = nil
         expect(subject).to be_valid
       end
+
+      it 'is valid without resource_owner_id' do
+        # For resource owner credentials flow
+        subject.application_id = nil
+        expect(subject).to be_valid
+      end
```
tute: This test doesn't fail in master for rails 5, so it's not catching the error.

lib/doorkeeper/models/access_token_mixin.rb
tute: > I consider this a bit of a stretch, but it was intended to provide verbosity. Feel free to ditch it if needed. Validation based on the optional key still stands anyway.

I would ditch it indeed. I think it's more confusing than useful ("check this, but allow it to not be checked too").

lib/doorkeeper/models/access_token_mixin.rb
tute: > @tute seems hound prefers the above format, should I change it back (it was how my original PR had it)?

I don't care enough, so if you feel strongly pick either. I personally liked `merge!`.

tute:
Thanks for your help! We need to make the spec fail in master, and then we are set!

spec/models/doorkeeper/access_token_spec.rb
relu: FYI, will need to test against pre-4.0.0rc1. I'm checking it at the moment and will follow-up.

relu: So it seems I found the issue. There's a default initializer that rails creates: https://github.com/rails/rails/commit/cdbf68599416e6b248d1d0e5bcbbbf73b90f1f42#diff-1846d4882fa63eb5b1271541995cbfdcR41 which controls this flag.

relu: Updated. Make sure when testing to include the initializer from `spec/dummy/config/initializers/active_record_belongs_to_required_by_default.rb`

spec/dummy/config/initializers/active_record_belongs_to_required_by_default.rb
```diff
@@ -0,0 +1,6 @@
+# Be sure to restart your server when you modify this file.
+
+# Require `belongs_to` associations by default. This is a new Rails 5.0
+# default, so it is introduced as a configuration option to ensure that apps
+# made on earlier versions of Rails are not affected when upgrading.
+Rails.application.config.active_record.belongs_to_required_by_default = true if Rails.version.to_i >= 5
```
tute: This got real! With this line in, master presents 544 examples, 150 failures, 2 pending. That's great! :+1:

tute:
Thank you so much! Squashed into two commits, rebased, and merged into master. :)

PR #627, opened 2015-04-12
fallback i18n

jasl:
I18n will broken when locale isn't `en`.

image

PR #802, opened 2016-03-11
Delete I18n logic from engine

johnnyshields:
The lines deleted in the PR are unnecessary, and they inadvertently change the I18n behavior of your entire app.

(In my case, it breaks the i18n-js gem's export, but others may be affected.)

It looks like this was added by @jasl here: https://github.com/doorkeeper-gem/doorkeeper/commit/78fa331710476532349b879fce0ccba536622952

My guess was he wanted the English error messages for Doorkeeper to appear even when using a different default locale. But the Doorkeeper gem should **NOT** reconfigure your Rails app globally in order to achieve this goal!

### DIFF
--- a/lib/doorkeeper/engine.rb
+++ b/lib/doorkeeper/engine.rb
@@ -4,12 +4,6 @@ module Doorkeeper
       app.config.filter_parameters += [:client_secret, :code, :token]
     end
 
-    initializer "doorkeeper.locales" do |app|
-      if app.config.i18n.fallbacks.blank?
-        app.config.i18n.fallbacks = [:en]
-      end
-    end
-
     initializer "doorkeeper.routes" do
       Doorkeeper::Rails::Routes.install!
     end


tute:
Thanks for your contribution!

I agree that doorkeeper shouldn't silently reconfigure your app, and that should be considered a bug. That said, this is a backwards incompatible change, which in turns will bring back the bug described in https://github.com/doorkeeper-gem/doorkeeper/pull/627.

How can we keep that bug fixed, while being considerate of the application's configuration? Is there a way to fallback only the translations with a `doorkeeper.` prefix?

johnnyshields:
#627 is **not** a bug. It is expected behavior of Rails if:
1) You do not have a translation for a given locale AND
2) You have not set locale fallbacks in your app.
This is how I18n works for every app and gem in existence, and it's not up to Doorkeeper to force an override on the user.

johnnyshields:
I think we can put a warning in the gem log when upgrading:

```
****
As of version x.y.z Doorkeeper no longer sets I18n.fallback_locales by default. If you are using I18n in your app, please either set fallbacks in your Rails configuration (see: http://guides.rubyonrails.org/i18n.html) or use one of our translations at https://github.com/doorkeeper-gem/doorkeeper-i18n
****
```

tute:
> #627 is not a bug. It is expected behavior of Rails if:
> 
>     You do not have a translation for a given locale AND
>     You have not set locale fallbacks in your app.

This make sense.

About backwards incompatibility, we didn't yet release v4, so we can be backwards incompatible. Merging, thank you!

tute:
Ouch, we should have added this context to the commit message. I'll add it to NEWS when we prepare a new release.

johnnyshields:
Cheers!

PR #803, opened 2016-03-11
Delete I18n logic from engine (2.2-stable branch)

johnnyshields:
Raising this to legacy branch as the presence of these lines break my app.

Same as https://github.com/doorkeeper-gem/doorkeeper/pull/802.

### DIFF
--- a/lib/doorkeeper/engine.rb
+++ b/lib/doorkeeper/engine.rb
@@ -4,12 +4,6 @@ module Doorkeeper
       app.config.filter_parameters += [:client_secret, :code, :token]
     end
 
-    initializer "doorkeeper.locales" do |app|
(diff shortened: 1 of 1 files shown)


tute:
The `2.2-stable` branch supports end of life'd dependencies, which are by definition insecure. I'd rather leave my libraries unmaintained for insecure versions, as an explicit statement the risks associated with not upgrading.

Also, this would be a breaking change. We discussed in the other PR why it's good to apply this breaking change to doorkeeper, and so we did in `master` but I wouldn't to do it in the `2.2-stable` branch.

Thank you for your contributions, but I will close this PR.

Is there any workaround for your application to work with these lines of code?

Issue #799, opened 2016-03-10
Fix Ruby warnings


tute:
`RUBYOPT=-w rake` runs all tests with ruby warnings enabled, which are printed out in [stderr](https://robots.thoughtbot.com/input-output-redirection-in-the-shell). They are very noisy, as it shows warnings for every dependency in use alongside the tests output, so they are most useful when filtering only the warnings coming from this gem. We do that redirecting tests `stdout` to `/dev/null` (ignoring it), and redirecting `stderr` to `stdout`, with the following command:

```bash
RUBYOPT=-w rake 2>&1 1>/dev/null | grep doorkeeper | sort --unique
```

Current output is:

```
/.../doorkeeper/lib/doorkeeper/config.rb:200: warning: instance variable @enable_application_owner not initialized
.*/./doorkeeper/lib/doorkeeper/config.rb:200: warning: instance variable @enable_application_owner not initialized
./../doorkeeper/lib/doorkeeper/config.rb:200: warning: instance variable @enable_application_owner not initialized
/.../doorkeeper/lib/doorkeeper/config.rb:136: warning: method redefined; discarding old access_token_generator
/.../doorkeeper/lib/doorkeeper/config.rb:136: warning: method redefined; discarding old force_ssl_in_redirect_uri
/.../doorkeeper/lib/doorkeeper/config.rb:136: warning: method redefined; discarding old realm
/.../doorkeeper/lib/doorkeeper/config.rb:148: warning: previous definition of realm was here
/.../doorkeeper/lib/doorkeeper/config.rb:196: warning: instance variable @refresh_token_enabled not initialized
/.../doorkeeper/lib/doorkeeper/config.rb:200: warning: instance variable @enable_application_owner not initialized
/.../doorkeeper/lib/doorkeeper/config.rb:204: warning: instance variable @confirm_application_owner not initialized
/.../doorkeeper/lib/doorkeeper/config.rb:227: warning: method redefined; discarding old realm
/.../doorkeeper/lib/doorkeeper/config.rb:84: warning: previous definition of realm was here
/.../doorkeeper/lib/doorkeeper/config.rb:92: warning: previous definition of force_ssl_in_redirect_uri was here
/.../doorkeeper/lib/doorkeeper/config.rb:96: warning: previous definition of access_token_generator was here
/.../doorkeeper/lib/doorkeeper/models/access_token_mixin.rb:108: warning: instance variable @use_refresh_token not initialized
[...]

PR #805, opened 2016-03-13
In Doorkeeper Config Option, Remove method if it already exists.

RowiDont:


tute:
Good for: https://github.com/doorkeeper-gem/doorkeeper/issues/799

tute:
Thank you! Added more context into the commit messages, and merged into master. :+1: :smiley:

PR #807, opened 2016-03-16
Use Ruby 1.9 hash syntax in initializer configuration file.

vrybas:
I change this line manually every time when install Doorkeeper on new
project.

Particularly relevant to recent drop of Ruby 2.0 support in v4.0.0.rc1

### DIFF
--- a/spec/dummy/config/initializers/doorkeeper.rb
+++ b/spec/dummy/config/initializers/doorkeeper.rb
@@ -30,10 +30,10 @@ Doorkeeper.configure do
   use_refresh_token
 
   # Provide support for an owner to be assigned to each registered application (disabled by default)
-  # Optional parameter :confirmation => true (default false) if you want to enforce ownership of
+  # Optional parameter confirmation: true (default false) if you want to enforce ownership of
   # a registered application
   # Note: you must also run the rails g doorkeeper:application_owner generator to provide the necessary support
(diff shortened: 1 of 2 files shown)


tute:
Thanks for your update. This is Ruby 2 hash syntax though, can you please update your commit message? Rails 4 support comment is not related with this by now, so I'd remove it as well. Thanks!

spec/dummy/config/initializers/doorkeeper.rb
```diff
@@ -30,10 +30,10 @@
   use_refresh_token
 
   # Provide support for an owner to be assigned to each registered application (disabled by default)
-  # Optional parameter :confirmation => true (default false) if you want to enforce ownership of
+  # Optional parameter confirmation: true (default false) if you want to enforce ownership of
   # a registered application
   # Note: you must also run the rails g doorkeeper:application_owner generator to provide the necessary support
-  # enable_application_owner :confirmation => false
+  # enable_application_owner :confirmation: false
```
christopherhein: I think you need to remove the leading `:` as well

tute:
A note: you don't need to close the PR. You can continue updating it, and even force push when you want to replace the commits, while keeping the discussion. Moving on to https://github.com/doorkeeper-gem/doorkeeper/pull/808 for now. Thanks!

PR #809, opened 2016-03-17
Read initializer config file from template, instead of duplicating it

vrybas:
To avoid maintaining both files.


tute:
Thanks for your idea! I'm fine with that file being copied verbatim into the dummy app. Makes it easier to debug and test certain issues. Thanks again!

vrybas:
Sure )

vrybas:
@tute maybe add the test, which asserts that those files content is the 
same? It already bit me on the last PR :)

tute:
I think it's ok that they may go out of sync. In your previous PR, if you hadn't updated the dummy app's initializer, it would have been fine as well.

The Rails initializers for that app are out of sync with either Rails 4.2 or 5, always, and I don't see that as a problem.

Do you think we should keep them in sync for other reasons?

vrybas:
I would be nice to run tests with the 'production' initializer. To catch
syntax errors at least.  But it doesn't really matter, as all the
configuration lines are commented out anyway.

Curious is there a good way to test default config file syntax
correctness.

PR #813, opened 2016-04-09
adding scopes to application authorization path

gregology:
Otherwise authorization fails because of missing scopes.
Thanks for this awesome gem team!


tute:
Thanks!

Issue #221, opened 2013-03-29
Changing model table/collection name (location)


ignatiusreza:
What is the suggested way to change the location of where the model is stored in the database. Particularly since mongoid3 support having multiple database session, I want to doorkeeper to store the data not in the default session.

The reason for this is the "application" model will be used across several different applications, each having their own different default session.

PR #819, opened 2016-04-20
Configurable table names

nbulaj:
This PR allows to customize `Application`, `AccessToken` and `AccessGrant` table names. It may be necessary for at least two reasons:

* Developer don't want to use default table names like "oauth_xxx" for any reason. Instead of using something like 
`Doorkeeper::Application.class_eval { self.table_name = :apps }` 
he can simply change corresponding Doorkeeper configuration option now.
* ORM extensions become independent of strong Doorkeeper table names.

All the current names is saved by default (`oauth_applications`, `oauth_access_tokens`, `oauth_access_grants`). Tests are present.

The problem with strong table names appeared when I was working on [doorkeeper-sequel](https://github.com/nbulaj/doorkeeper-sequel) adapter, in part of migrating associations (many-to-many :through, ).

### DIFF
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -197,7 +197,7 @@ doorkeeper.
     option :access_token_generator,         default: "Doorkeeper::OAuth::Helpers::UniqueToken"
 
     %w(applications access_grants access_tokens).each do |model_name|
-      option :"#{model_name}_table_name",   default: "oauth_#{model_name}"
+      option :"#{model_name}_table_name",   default: :"oauth_#{model_name}"
     end
 
     attr_reader :reuse_access_token
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -10,6 +10,7 @@ module Doorkeeper
     setup_orm_adapter
     setup_orm_models
     setup_application_owner if @config.enable_application_owner?
+    setup_table_names
     check_requirements
   end
 
@@ -39,6 +40,10 @@ doorkeeper.
 
   def self.setup_application_owner
     @orm_adapter.initialize_application_owner!
+  end
+
+  def self.setup_table_names
+    @orm_adapter.initialize_table_names!
   end
 
   class Config
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -191,6 +191,10 @@ doorkeeper.
     option :grant_flows,                    default: %w(authorization_code client_credentials)
     option :access_token_generator,         default: "Doorkeeper::OAuth::Helpers::UniqueToken"
 
+    %w(applications access_grants access_tokens).each do |model_name|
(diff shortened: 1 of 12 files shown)


tute:
Hey! Thanks for your submission!

> Developer don't want to use default table names like "oauth_xxx" for any reason.

What are those reasons? Is this a common enough problem to deserve configuration options in the main repository?

> Instead of using something like `Doorkeeper::Application.class_eval { self.table_name = :apps }` he can simply change corresponding Doorkeeper configuration option now.

While we don't merge this PR, does the workaround `Doorkeeper::Application.class_eval { self.table_name = :apps }` work for you?

> The problem with strong table names appeared when I was working on doorkeeper-sequel adapter, in part of migrating associations (many-to-many :through, ).

Can you please in further detail the issue you encountered?

tute:
Thank you again for your work and submission.

nbulaj:
> What are those reasons?

Database naming requirements, for example.  Though of course, we can always use the standard Doorkeeper names. People is [asking about](https://github.com/doorkeeper-gem/doorkeeper/issues/221) how to [change model](http://stackoverflow.com/questions/27264407/how-to-override-model-from-the-doorkeeper-gem)/table for the gem.

> does the workaround Doorkeeper::Application.class_eval { self.table_name = :apps } work for you?

Yes, it works. But it's a kind of monkey patching and might be placed into the _config/initializers_. A more flexible way through the Doorkeeper config preferred I think.

> Can you please explain in further detail the issue you encountered?

Here you are right: if the developer does not want to change default name then everything is OK. In another case `class_eval` will not work.

Thanks for the answer!

tute:
Doorkeeper models are properly namespaced in `Doorkeeper::`, and tables can be customized with prefixes and suffixes. Right now, the default prefix is `oauth_`, ensuring that it won't conflict with existing applications.

I don't mind Doorkeeper models being akin to reserved keywords and constants, and not configurable. Doorkeeper is an engine extending Ruby applications, and I'd expect certain restrictions to exist, when they are not likely to produce conflicts with user defined names.

Regarding table names, I only thought about their names being configurable (and adding the prefix) to avoid name conflicts. I am surprised that some users would use doorkeeper codebase on top of preexisting database schemas. Is that your use case? Or do you only have a restriction on the table name, regardless of it's columns?

It sounds to me like this is a very specific use case, and I wouldn't add a configuration and code in the main project to accommodate that. What do you think?

nbulaj:
I thought about it and decided that you're probably right. I will close the pull request at the moment, if someone has the same need - he will write about it.

tute:
Thanks for your time and work!

jtoy:
just to add my 2 cents, I wanted to customize the names because I hate have odd tables names in my schema.

nbulaj:
@jtoy you can use my proposed solution with:

```ruby
Doorkeeper::Application.class_eval { self.table_name = :apps }
```

PR #825, opened 2016-04-28
Style adjustments and legacy code refactoring

nbulaj:
Hi. During the work on the Sequel ORM adapter for Doorkeeper I found a lot of issues and code smells in the code base. So I decided to fix them.

This PR contains modification of a large part of the code base in accordance with the _Ruby Style Guide_: [single quotes](https://github.com/nbulaj/doorkeeper/commit/d91cc74801bc2ff06b7d9ad3b72bbda5a7381e85#diff-8b7db4d5cc4b8f6dc8feb7030baa2478L1), [guard clauses](https://github.com/nbulaj/doorkeeper/commit/d91cc74801bc2ff06b7d9ad3b72bbda5a7381e85#diff-d0fc797f13ea4de1171e4f62a72a28e0L38), [if statements](https://github.com/nbulaj/doorkeeper/commit/d91cc74801bc2ff06b7d9ad3b72bbda5a7381e85#diff-d5ebad835c0898d6e7daa5c127785334R22), [lambdas](https://github.com/nbulaj/doorkeeper/commit/d91cc74801bc2ff06b7d9ad3b72bbda5a7381e85#diff-2dac7b5ba275c2921a0ad8385afea15bL184), [mutable objects](https://github.com/nbulaj/doorkeeper/commit/d91cc74801bc2ff06b7d9ad3b72bbda5a7381e85#diff-19f61d7b331feabc5b91a716edc3c531R9), [raising of exceptions](https://github.com/nbulaj/doorkeeper/commit/d91cc74801bc2ff06b7d9ad3b72bbda5a7381e85#diff-2dac7b5ba275c2921a0ad8385afea15bR17), [unnecessary commas](https://github.com/nbulaj/doorkeeper/commit/d91cc74801bc2ff06b7d9ad3b72bbda5a7381e85#diff-3534f9b842cba947eb81b27bd9327fc9L17) ans so on.

Also this PR contains refactoring of the old code ([methods of ORM models and mixins](https://github.com/nbulaj/doorkeeper/commit/d91cc74801bc2ff06b7d9ad3b72bbda5a7381e85#diff-c42935829979fd1bcb5ed418c92f6386L36), [other stuffs](https://github.com/nbulaj/doorkeeper/commit/d91cc74801bc2ff06b7d9ad3b72bbda5a7381e85#diff-d0fc797f13ea4de1171e4f62a72a28e0L34)).

All specs are green.


lib/doorkeeper/config.rb
```diff
@@ -14,7 +14,7 @@ def self.configure(&block)
   end
 
   def self.configuration
-    @config || (fail MissingConfiguration.new)
+    @config || (fail MissingConfiguration)
```
tute: Is this a change in behavior?

tute:
I like this, thank you! Let's use double quotation marks as doorkeeper's default, since it will be more consistent for when we need interpolation, and we don't lose anything by using those. You can configure Hound to require this in subsequent PRs.

Thanks again!

lib/doorkeeper/oauth/refresh_token_request.rb
```diff
@@ -23,10 +23,8 @@ def initialize(server, refresh_token, credentials, parameters = {})
         @original_scopes = parameters[:scope] || parameters[:scopes]
         @refresh_token_parameter = parameters[:refresh_token]
 
-        if credentials
-          @client = Application.by_uid_and_secret credentials.uid,
-                                                  credentials.secret
-        end
+        @client = Application.by_uid_and_secret(credentials.uid,
+                                                credentials.secret) if credentials
```
tute: This is a surprising `if` back and down there. I would leave the multi-line `if`.

lib/doorkeeper/models/application_mixin.rb
```diff
@@ -19,31 +19,27 @@ module ApplicationMixin
...
-        self.uid = UniqueToken.generate
-      end
+      self.uid = UniqueToken.generate if uid.blank?
     end
 
     def generate_secret
-      if secret.blank?
-        self.secret = UniqueToken.generate
-      end
+      self.secret = UniqueToken.generate if secret.blank?
```
tute: Same thing, let's favor multi-line `if`s. They read less like english, but code behavior is more explicit I think. Again, we can configure Hound to enforce this rule.

lib/doorkeeper/config.rb
nbulaj: No, because message is initialized in the constructor (see [config.rb#L3](https://github.com/doorkeeper-gem/doorkeeper/blob/master/lib/doorkeeper/config.rb#L3)). You can check it in _irb_/_rails console_:

```ruby
2.3.0 :004 > raise Doorkeeper::MissingConfiguration
# Doorkeeper::MissingConfiguration: Configuration for doorkeeper missing. Do you have doorkeeper initializer?
```

tute:
For reference, the guide I would use when Hound doesn't specify its own, is https://github.com/thoughtbot/guides/blob/master/style/ruby/README.md#ruby.

tute:
Here is sample configuration, if you'd like to tackle these as part of the same PR: https://github.com/thoughtbot/guides/blob/master/style/ruby/.rubocop.yml

lib/doorkeeper/models/application_mixin.rb
nbulaj: @tute I usually guided those "[Ruby Style Guide](https://github.com/bbatsov/ruby-style-guide)".

I think that code is readable in both cases, but with one line it is more elegant. _thoughtbot/guides_ recommends to use multiline `if` to avoid ternary operators and nothing more.

tute: > I think that code is readable in both cases, but with one line it is more elegant. thoughtbot/guides recommends to use multiline if to avoid ternary operators and nothing more.

The very first line in https://github.com/thoughtbot/guides/blob/master/style/ruby/README.md advices against "conditional modifiers". I can find the related discussion if you are interested in it.

nbulaj:
As for the double quotation marks, Doorkeeper now "represents" a a porridge of quotes. 
I prefer to use quotation marks for its intended purpose: double quotation for interpolation, single quotation for simple strings. I think any Ruby developer needs to distinguish them

lib/doorkeeper/models/application_mixin.rb
nbulaj: Ok, I will return multiline `if`.

>  I can find the related discussion if you are interested in it.

If not difficult, I will be grateful.

tute:
> I prefer to use quotation marks for its intended purpose: double quotation for interpolation, single quotation for simple strings. I think any Ruby developer should be able to distinguish them

I'm interested in this discussion. Why do you think it's important to distinguish them? Historically, it was for performance reasons, but it's been long with both styles performing practically the same.

lib/doorkeeper/models/application_mixin.rb
tute: "if surprise" blog post: https://robots.thoughtbot.com/read-between-the-lines#readability

nbulaj:
> I'm interested in this discussion. Why do you think it's important to distinguish them? Historically, it was for performance reasons, but it's been long with both styles performing practically the same.

Single quotes means a "simple" string and it's slightly more explicit (visually) - there's no need to explore the string for any escaping sequences or expressions. Just like symbols - they means what they means. No problem to change the string to double quotation if it must (!!) be processed by special. Community _Ruby Style Guide_ offers both the first and second options, but I prefer to split a usual immutable string ('Doorkeeper 4.0.0rc3') and interpolated one ("Doorkeeper #{gem.version}").

tute:
Single quoted string is no immutable, it's a string where `#{}` has no special meaning. I don't need to know which strings in my codebases have that behavior. When I need to interpolate it either doesn't work if I forget, or I remember to switch the surrounding quotes. I'd rather not have to think about that, and know that all strings in my app can take `#{}` inside (a more common need than the need to show the actual `#{}` string).

See http://anti-pattern.com/always-use-double-quoted-strings-in-ruby.

nbulaj:
I did not mean in the literal sense (about "immutable").
Ok, so what we decide to do with the strings? They are different everywhere in the gem.

tute:
> Ok, so what we decide to do with the strings? They are different everywhere in the gem.

I typically only change code in the file that I'm touching and ignore others. I'm not against switching all strings to be double quoted (and adding the rule in Hound) if you prefer.

nbulaj:
Ok, take into account for the future. I believe that it is necessary to supplement the recommendations in the _Contribute.md_ file to avoid any questions in the future.

I think there is no need to rollback the strings from that commit if  PR will be merged to the master. Just return multiline `if`-s if there is no more corrections. What do you think?

tute:
> I believe that it is necessary to supplement the recommendations in the Contribute.md file to avoid any questions in the future.

👍 

> I think there is no need to rollback the strings from that commit if PR will be merged to the master.

I think moving double-quoted strings to single quotes is a step backwards, and I'd undo those. If we leave existing single-quoted strings untouched I'm fine with that.

tute:
> I believe that it is necessary to supplement the recommendations in the Contribute.md file to avoid any questions in the future.

The guidelines already mention the style guide, where the rules, rationale behind them, and "don't rewrite existing code to follow this guide" rule lie. Should we add something else that I'm missing?

nbulaj:
@tute I returned quotes and `if` statements as it was before my PR.

> Should we add something else that I'm missing?

There are two documents describes how to contribute to the Doorkeeper gem, that may be a bit confused:

1. https://github.com/doorkeeper-gem/doorkeeper/blob/master/CONTRIBUTING.md
2. https://github.com/doorkeeper-gem/doorkeeper/wiki/Contributing

 The main README links to the last one (wiki/Contributing - I have read just this) where are no info about style guides. I believe that it would be better to unify the documentation in one place.

lib/doorkeeper/oauth/client_credentials_request.rb
```diff
@@ -12,7 +12,7 @@ class ClientCredentialsRequest
       attr_reader :response
       attr_writer :issuer
 
-      alias :error_response :response
+      alias error_response response
```
tute: What do you think about using `alias_method` here instead? See http://stackoverflow.com/a/4763178/356060

lib/doorkeeper/oauth/token.rb
```diff
@@ -26,7 +26,7 @@ def from_basic_authorization(request)
 
         def token_from_basic_header(header, pattern)
           encoded_header = token_from_header(header, pattern)
-          token, _ = decode_basic_credentials(encoded_header)
+          token, = decode_basic_credentials(encoded_header)
```
tute: This is such weird syntax. How can we rewrite this in a clearer way than either the older or the newer option?

lib/doorkeeper/orm/active_record/application.rb
```diff
@@ -11,14 +11,9 @@ class Application < ActiveRecord::Base
...
-      self.column_names.map { |c| "#{table_name}.#{c}" }
-    end
-
     def self.authorized_for(resource_owner)
-      joins(:authorized_applications).
-        where(AccessToken.table_name => { resource_owner_id: resource_owner.id, revoked_at: nil }).
-        group(column_names_with_table.join(','))
+      where(id: Doorkeeper::AccessToken.
+        where(resource_owner_id: resource_owner.id, revoked_at: nil))
     end
```
tute: Do we not need the `group` call here? And why did we drop the `joins`?

tute:
Will now merge the two guidelines together, favouring the one versioned with the code over the wiki that sometimes gets neglected.

I'm happy about these changes, thanks for your work.

lib/doorkeeper/orm/active_record/application.rb
nbulaj: There is no need in the JOIN and GROUP BY.  That method returns applications, that has access tokes for the requested resource_owner.

JOINS was joining on the access tokens table and GROUP was acting like DISTINCT. There is no need in such a complex and suspicious logic (especially with the `column_names` method).  You can see at the `spec/models/doorkeeper/applications_spec.rb` : 139

lib/doorkeeper/oauth/client_credentials_request.rb
nbulaj: Yes, this is better solution.

lib/doorkeeper/oauth/token.rb
nbulaj: I can rewrite `decode_basic_credentials` to return only what is necessary. What do you think?

lib/doorkeeper/orm/active_record/application.rb
tute: Sounds good. I would extract to another method or local variable the argument:

```ruby
Doorkeeper::AccessToken.where(resource_owner_id: resource_owner.id, revoked_at: nil)
```

Question: how does the scope know to grab the application id's from the `AccessToken.where` call?

lib/doorkeeper/oauth/token.rb
tute: > I can rewrite `decode_basic_credentials` to return only what is necessary. What do you think?

Sounds great, let's do that.

lib/doorkeeper/orm/active_record/application.rb
nbulaj: I do not quite understand the question.

Rails is smart and merged argument query (` Doorkeeper::AccessToken.where(resource_owner_id: resource_owner.id, revoked_at: nil)`) to the first query `(where(id: ...))` and make a one SQL request with subquery:

```sql
SELECT * FROM oauth_applications
WHERE oauth_applications.id IN (
  SELECT id FROM oauth_access_tokens
  WHERE resource_owner_id: 1 AND revoked_at IS NOT NULL
)
```

I don't think there is a need in extracting of the `Doorkeeper::AccessToken.where`, but if you want - I can do it.

tute: But we don't want to match `oauth_applications.id` with `oauth_access_tokens.id`, we want to match them with `oauth_access_tokens.application_id` instead. I'm surprised the tests didn't fail.

> I don't think there is a need in extracting of the `Doorkeeper::AccessToken.where`, but if you want - I can do it.

I'm trying to find a blog post that explains how writing one statement per line makes code easier to reason about. Will post when I find it.

nbulaj: Oh, sorry, you are absolutely right! I will check and fix it .

lib/doorkeeper/models/access_grant_mixin.rb
```diff
@@ -28,7 +28,7 @@ module AccessGrantMixin
 
     module ClassMethods
       def by_token(token)
-        where(token: token.to_s).limit(1).to_a.first
```
tute: This were introduced in https://github.com/doorkeeper-gem/doorkeeper/pull/484, as performance improvements by @mtparet. Do you know if `find_by` performs an `ORDER` operation?

lib/doorkeeper/oauth/authorization/uri_builder.rb
```diff
@@ -20,7 +20,7 @@ def uri_with_fragment(url, parameters = {})
         end
 
         def build_query(parameters = {})
-          parameters = parameters.reject { |k, v| v.blank? }
+          parameters = parameters.reject { |_k, v| v.blank? }
```
tute: Does `_` work instead of `_k`?

lib/doorkeeper/orm/active_record/access_token.rb
```diff
@@ -10,6 +10,10 @@ def self.delete_all_for(application_id, resource_owner)
     end
     private_class_method :delete_all_for
 
+    def self.for(resource_owner)
```
tute: What's the opposite of `revoked`? Should this be called, for example, `active_for`, to make it explicit that we filter out revoked tokens?

lib/doorkeeper/models/access_grant_mixin.rb
nbulaj: No order, just where + take (http://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-find_by)

tute: > No order, just where + take (http://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-find_by)

💯

lib/doorkeeper/oauth/authorization/uri_builder.rb
nbulaj: Yeap, you are right. Аs well as for https://github.com/doorkeeper-gem/doorkeeper/blob/master/lib/doorkeeper/config.rb#L182

tute: > Yeap, you are right. Аs well as for https://github.com/doorkeeper-gem/doorkeeper/blob/master/lib/doorkeeper/config.rb#L182

In that case I like the documentation `_routes` provide. Here, I thought `_k` for key wasn't necessary (or I'd expand it to `_key, value`.

lib/doorkeeper/orm/active_record/access_token.rb
nbulaj: Yes, so would be more transparent

lib/doorkeeper/oauth/authorization/uri_builder.rb
nbulaj: Ok, and what about https://github.com/doorkeeper-gem/doorkeeper/blob/master/lib/doorkeeper/config.rb#L149 ? No args is used in the `define_method` block.

tute:
🎉 

Thank you for your work.

nbulaj:
@tute thank you for the interesting discussion and fruitful collaboration! :+1:

PR #829, opened 2016-05-06
Fix README instructions for resource owner foreign key

### LINKED ISSUE #781: Configuring activerecord
In the README it says:

> if you plan on using making User the resource owner, change the two lines in the migration file:
> 
> `t.integer    :resource_owner_id, null: false`
> to:
> `t.references :user, foreign_key: true, null: false`

However this doesn't seem to work, as the models appear to be hardcoded to search using the `resource_owner_id`. If there another configuration step to make this work, or is there an error in the documentation?

jclem:
This fixes the instructions in the README for adding a foreign key for resource owner models. Note that this closes #781.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,17 +76,11 @@ to generate the migration tables:
     rails generate doorkeeper:migration
 
 You may want to add foreign keys to your migration. For example, if you plan on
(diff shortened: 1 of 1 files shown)


tute:
Thanks so much! :smiley:

PR #835, opened 2016-05-16
Fix model_helper.rb to support different ORMs

### LINKED ISSUE #833: Requests flows: authorization_code_spec fails for Sequel ORM
I'm currently working on Doorkeeper Sequel adapter and found an interesting bug in specs.

Take a look at the next spec: **Requests flows** - **authorization_code_spec.rb** -  **scenario** "_resource owner authorizes using test url_" (source code in master - [spec/requests/flows/authorization_code_spec.rb#L30](https://github.com/doorkeeper-gem/doorkeeper/blob/master/spec/requests/flows/authorization_code_spec.rb#L30) ).

As you can see, after visiting `authorization_endpoint_url(client: @client)` and clicking `Authorize` button there should be a `AccessGrant` associated to the application (or `@client` as it passed with the params). We can see it there [spec/support/helpers/model_helper.rb#L16](https://github.com/doorkeeper-gem/doorkeeper/blob/master/spec/support/helpers/model_helper.rb#L16).

```ruby
  ...
  grant = Doorkeeper::AccessGrant.first
  expect(grant.application).to eq(client)
```

All is OK for the `ActiveRecord` ORM (test is green), but there is a problem for the `Sequel` gem.

>   3) Authorization Code Flow resource owner authorizes using test url
>      Failure/Error: expect(grant.application).to eq(client)
>      
>        expected: #<Doorkeeper::Application @values={:id=>1, :name=>"Application 128", :uid=>"7a0e555965bafb541a02ecb12dcac7aac2ad2735d83c47911d6077bde419ca6b", :secret=>"932aa79adef2ac7d7e1157b404400e42efd63e995aa0cd2a03973e5e1210d098", :scopes=>"", :redirect_uri=>"urn:ietf:wg:oauth:2.0:oob", :created_at=>2016-05-13 17:36:13 +0300, :updated_at=>2016-05-13 17:36:13 +0300, :owner_id=>nil, :owner_type=>nil}>
>             got: #<Doorkeeper::Application @values={:id=>1, :name=>"Application 128", :uid=>"7a0e555965bafb541a02ecb12dcac7aac2ad2735d83c47911d6077bde419ca6b", :secret=>"932aa79adef2ac7d7e1157b404400e42efd63e995aa0cd2a03973e5e1210d098", :scopes=>"", :redirect_uri=>"urn:ietf:wg:oauth:2.0:oob", :created_at=>2016-05-13 17:36:13 +0300, :updated_at=>2016-05-13 17:36:13 +0300, :owner_id=>nil, :owner_type=>nil}>
>      
>        (compared using ==)
>      
>        Diff:
>        
>      # ./spec/support/helpers/model_helper.rb:16:in `access_grant_should_exist_for'
>      # ./spec/requests/flows/authorization_code_spec.rb:30:in `block (2 levels) in <top (required)>'

As you can see, `Doorkeeper::Application` is the same, but RSpec matcher reported that it is not a truth (they are different). Let's take a look to the RSpec `BuiltIn::Eq` matcher description (or failure message above  -  `(compared using ==)`):

```ruby
# Passes if <tt>actual == expected</tt>.
```

So the objects is compared with the `==` operator. And if we open `ActiveRecord` sources, then we will see this ( [activerecord/lib/active_record/core.rb#L423](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/core.rb#L423) ):

```ruby
    def ==(comparison_object)
      super ||
        comparison_object.instance_of?(self.class) &&
        !id.nil? &&
        comparison_object.id == id
    end
    alias :eql? :==
```

So, if `object == object` returns `false`, then `ActiveRecord` checks the class and IDs of the models. And if they are the same (both is a instance of 'SomeClass' and IDs 1 == 1 for example) - then model objects are considered equal. Example:

```ruby
    u1 = User.first
    u2 = User.first
    u1.email = "something" # we change u1 email field
    u1 == u2  # and despite that the objects are identical
    #=> true 
```

But `Sequel` gem override `==` method in another way - it checks **the values** of the models, **not the IDs**. We must use `===` to check only the IDs of the models (see [Sequel documentation - "Instance Methods with Roughly the Same Behavior"](https://github.com/jeremyevans/sequel/blob/master/doc/active_record.rdoc#instance-methods-with-roughly-the-same-behavior)).

Now, as you can see at the 25 line of the _authorization_code_spec.rb_ code (https://github.com/doorkeeper-gem/doorkeeper/blob/master/spec/requests/flows/authorization_code_spec.rb#L25), the `@client` is changing before we click "Authorize" button:

```ruby
@client.redirect_uri = Doorkeeper.configuration.native_redirect_uri
@client.save!
```

Since we changed the values of the `@client` model, RSpec test now is broken, because it is trying to check the models with `==` comparison, that checks the values (not the IDs like `AR`) of the Sequel models.

So, I think that part of the spec must be rewritten to match all the ORMs (maybe `MongoMapper` or `Mongoid` do the same behaviour with the `==` method). We must check the ID of the records, not the model objects. So `access_grant_should_exist_for` method should be something like this:

```ruby
 def access_grant_should_exist_for(client, resource_owner)
    grant = Doorkeeper::AccessGrant.first
    expect(grant.application).to have_attributes(id: client.id).and(be_instance_of(Doorkeeper::Application))
    grant.resource_owner_id  == resource_owner.id
  end
```

What do you think?

nbulaj:
Fixes #833 - expectations of 'access_grant_should_exist_for' and 'access_token_should_exist_for' methods checks for the same object ID, class and Resource Owner ID.

### DIFF
--- a/spec/support/helpers/model_helper.rb
+++ b/spec/support/helpers/model_helper.rb
@@ -13,14 +13,20 @@ module ModelHelper
 
   def access_grant_should_exist_for(client, resource_owner)
     grant = Doorkeeper::AccessGrant.first
-    expect(grant.application).to eq(client)
-    grant.resource_owner_id  == resource_owner.id
+
+    expect(grant.application).to have_attributes(id: client.id).
+      and(be_instance_of(Doorkeeper::Application))
+
+    expect(grant.resource_owner_id).to eq(resource_owner.id)
   end
 
   def access_token_should_exist_for(client, resource_owner)
-    grant = Doorkeeper::AccessToken.first
-    expect(grant.application).to eq(client)
-    grant.resource_owner_id  == resource_owner.id
+    token = Doorkeeper::AccessToken.first
+
+    expect(token.application).to have_attributes(id: client.id).
+      and(be_instance_of(Doorkeeper::Application))
+
+    expect(token.resource_owner_id).to eq(resource_owner.id)
   end
 
   def access_grant_should_not_exist


tute:
Thank you!

PR #841, opened 2016-05-27
Allow to issue new token without client credentials

xendoc:
[This spec](https://github.com/doorkeeper-gem/doorkeeper/blob/master/spec/requests/flows/password_spec.rb#L49) is successful, but `ActiveRecord::StatementInvalid` is caused by the constraint of database schema.

I want you to check PR https://github.com/doorkeeper-gem/doorkeeper/pull/842 before to merge.


tute:
Good catch, thank you!

PR #842, opened 2016-05-27
Disallow to issue a new token without a client

xendoc:
A new token is issued without a client in case of ResourceOwnerPasswordCredentials flow.
However it doesn't work in version `4.0.0.rc4` by the constraint of database schema.

If you change as PR https://github.com/doorkeeper-gem/doorkeeper/pull/841, is necessary to change the database schema. Also a new token isn't issued without a client in case of RefreshToken flow.
I don't think it's a good way.

It should change `client_id` to required parameter. It is well enough to support public client.

I want you to merge one of https://github.com/doorkeeper-gem/doorkeeper/pull/841 or this PR.

### DIFF
--- a/lib/doorkeeper/oauth/password_access_token_request.rb
+++ b/lib/doorkeeper/oauth/password_access_token_request.rb
@@ -36,7 +36,7 @@ module Doorkeeper
       end
 
       def validate_client
-        !parameters[:client_id] || !!client
+        !!client
       end
     end
   end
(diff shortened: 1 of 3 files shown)


tute:
https://github.com/doorkeeper-gem/doorkeeper/pull/841 seems to be the correct changeset: tokens should be allowed to not have associated clients. Thanks!

Issue #465, opened 2014-08-14
Adding methods to `Doorkeeper::ApplicationController` (path helper related)


aymericbouzy:
oauth_authorized_application_path(application) fails systematically for me, with : 

No route matches {:action=>"destroy", :controller=>"doorkeeper/authorized_applications", :format=>nil, :id=>nil, :locale=>#<Doorkeeper::Application _id: 53e8f2696c0e32545e000001, created_at: 2014-08-11 16:42:17 UTC, updated_at: 2014-08-11 16:42:17 UTC, name: "Foobar", uid: "d938832444fe519704ec9c6536d8cc40017443f2db66a0bcc1c250bfaf6a4552", secret: "1b1772c69fcd5c55fcb75efbe531992891fdddcd88fe4e158880087d2197671e", redirect_uri: "urn:ietf:wg:oauth:2.0:oob">} missing required keys: [:id]

Am I doing something wrong ?

moskeyombus:
@tute Is the ability to configure which controller the Doorkeeper base controller inherits from something you would still a feature you would be willing to merge?  I have a use case where this would come in handy, would be happy to open a PR

tute:
What is your use case, @moskeyombus? Would reopening `Doorkeeper::ApplicationController` work for you?

moskeyombus:
Basically the application I'm working has a lot of before and after filters in `ApplicationController` that should be utilized by the Doorkeeper controllers.  It's not realistic for me to pull those filters out into helpers (although I realize this is probably best practice).  Ideally, I would be able to configure `Doorkeeper::ApplicationsController` to inherit from my `ApplicationController` instead of `ActionController::Base`.  That would also give me the ability to define my own `BaseController` that `Doorkeeper::ApplicationsController` could inherit from, which may prove useful down the road.

tute:
I am curious about your example: what do those before filters do that need to be run both for user facing, HTML endpoints and for token endpoints?

What I mean by reopening, is adding a `app/controllers/doorkeeper/application_controller.rb` file in your application with following contents:

```ruby
module Doorkeeper
  class ApplicationController < ActionController::Base
    before_filter :do_1
    after_filter :do_2
  end
end
```
This will add behavior at run time to doorkeeper's controller

moskeyombus:
[...]

PR #845, opened 2016-05-31
Allow customising the Doorkeeper::ApplicationController base controller

jvenezia:
Related to https://github.com/doorkeeper-gem/doorkeeper/issues/465.

Usefull to keep Doorkeeper controllers in the same context than the rest of your app.

### DIFF
--- a/README.md
+++ b/README.md
@@ -282,6 +282,16 @@ end
 JWT token support is available with
 [Doorkeeper-JWT](https://github.com/chriswarren/doorkeeper-jwt).
 
+### Custom Base Controller
+
+By default Doorkeeper's main controller `Doorkeeper::ApplicationController` inherits from `ActionController::Base`.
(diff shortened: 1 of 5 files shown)


jvenezia:
Actually we could do a devise-like solution. Devise has no `ApplicationController` of his own and inherits all controllers from the users `ApplicationController`, and you can choose another one with the `parent_controller` option. Then it injects helpers in the actual parent controller.

I think it's a nicer solution than what I actually done here.

What do you think ?

jvenezia:
Hi, any thought on this?

tute:
Thank you! 👏

apanzerj:
Thank you as well! We were just hacking around with this. Would love a release!

tute:
You got it: https://rubygems.org/gems/doorkeeper/versions/4.1.0

apanzerj:
Thanks!

PR #847, opened 2016-06-03
Specify Null option to timestamps

xendoc:
In Rails 5, default option of timestamps are changed to `null: false`.
If you do not specify the option deprecation messages are called.


tute:
I would make them default to `false`. What would it mean to have a timestamp set as `NULL`? Thanks!

xendoc:
I agree with your thoughts, but I thought compatibility of database schema is important.

tute:
Existing applications don't need to change this, and if they do, they can do it without the need of us adding a generator for applying this. This will affect only newly generated doorkeeper tables. Code should work with this setting.

xendoc:
Oh, I get it.

It has been changed default to false.

tute:
Thank you!

PR #848, opened 2016-06-04
Fix reuse access token

Exelord:
Hi,
I found an edge case where `#reuse_access_token` method was not working properly.
Here is a quick fix. I also added a test for that. I hope that code and test is self-explanatory.

If you have any questions, go ahead!


tute:
The failure without your fix looks good to me. You create an application that to be accessed requires scopes `public` and `write`, but try to access it with a token that can only access `public`. I think this is not a bug.

Exelord:
Hmm.. Not exactly, I think.
If you are creating an app e.g. in facebook you are passing there which scopes do you want to use in your application. And e.g. you are choosing `public_profile` and `email`.

But If you are requesting a token for user you can specify which scopes do you need for this token. Eg. only public_profile for commenting below an articles.

There should be `or` operator not `and`. Scopes for application are defined to prevent access for restricted scopes not for require ones.

If Im wrong, please correct me! I will be very appreciate. :) I dont know everything...

Exelord:
ping?

nbulaj:
Hi @Exelord . [This change looks strange to me](https://github.com/doorkeeper-gem/doorkeeper/pull/848/files#diff-2552b78abd4a0e8e3be0084a1f0ec3a3R22). 

You said:

> There should be or operator not and

But in the commit OR check was replaced with AND..

Exelord:
Hmm, It's been a while ago, I don't know why... Maybe it's a mistake... but! There is a test for that. If it works, then it means it's correct.

nbulaj:
Hi @Exelord . Seems like it is a stale PR. Could you please review it and approve that problem still exists? There were some changes in `matching_token_for` behavior. I'll take a look at it a little bit later.

Exelord:
I applied the changes. Lets see what tests would say

Exelord:
BTW. `has_scopes` require from the array to include all of them. So valid_scopes have to include all parsed_scopes. Thats the bug. They dont have to. parsed_scopes should be included in valid_scopes but they dont have to be all of them

bpieck:
I would disagree - a token should only be reused, if the scopes are identically. You never know for sure, what the token is requested for. And if it is forwarded to some other microservice that really should only have access to some information - doorkeeper should not just decide to respond with a token, that has some more access rights then there were requested ...

Exelord:
> doorkeeper should not just decide to respond with a token, that has some more access rights then there were requested

exactly! If my app requested a token with just a `public` scope, I should receive such a  token even my app is authorized to all the scopes like `public` and `private`.

bpieck:
Ah, sorry, I misunderstood you firstly. But I think, your fix broke much stuff. If I see correctly you did not fix it at the position, where you only check for matching_token_for. ScopeChecker.valid? is even used to check, if a given scope is valid at all. So with your "fix" it's not possible anymore, to ask for a token with a subset of allowed scopes. You always need to request complete set of scopes, the application is allowed to.

from /lib/doorkeeper/oauth/preauthorization I see:
```
      def validate_scopes
        return true if scope.blank?

        Helpers::ScopeChecker.valid?(
          scope,
          server.scopes,
          client.application.scopes
        )
      end
```
So with your change, any preauthorization will return an invalid for a request with just a subset of all allowed scoped, right?

bpieck:
So these two tests are green on master:
```
      it 'matches token with one of application`s scopes' do
        scopes = Doorkeeper::OAuth::Scopes.from_string('public')
        application = FactoryBot.create :application, scopes: "public write"
        token = FactoryBot.create :access_token, default_attributes.merge(
          application: application,
          scopes: 'public'
        )
        last_token = AccessToken.matching_token_for(application, resource_owner_id, scopes)
        expect(last_token).to eq(token)
      end

      it 'matches token with one of application`s scopes' do
        scopes = Doorkeeper::OAuth::Scopes.from_string('public')
        application = FactoryBot.create :application, scopes: "public write"
        token = FactoryBot.create :access_token, default_attributes.merge(
          application: application,
          scopes: 'public write'
        )
        last_token = AccessToken.matching_token_for(application, resource_owner_id, scopes)
        expect(last_token).to be_nil
      end
```
So in my understanding your issue is not there anymore.

Exelord:
Sure! because this PR has been submitted 2 years ago 😂It's natural it's failing now.
I just removed the "fix" and left only specs. If that will pass, its ok now :)

Exelord:
anyway, this test doesn't make any sense
```
      it 'matches no token with more then requested scopes' do
        scopes = Doorkeeper::OAuth::Scopes.from_string('public')
        application = FactoryBot.create :application, scopes: "public write"
        token = FactoryBot.create :access_token, default_attributes.merge(
          application: application,
          scopes: 'public write'
        )
        last_token = AccessToken.matching_token_for(application, resource_owner_id, scopes)
        expect(last_token).to be_nil
      end
```
it should be
```
      it 'does not match a token with more than requested scopes' do
        scopes = Doorkeeper::OAuth::Scopes.from_string('public write')
        application = FactoryBot.create :application, scopes: "public"
        token = FactoryBot.create :access_token, default_attributes.merge(
          application: application,
          scopes: 'public write'
        )
        last_token = AccessToken.matching_token_for(application, resource_owner_id, scopes)
        expect(last_token).to be_nil
      end
```

Exelord:
OK all's good now. You can merge the test or close it :)

bpieck:
No it shouldn't:

my test tested, that you do not get a token with "public write" access, when you ask for "public"
your test teste, that you do not get a token with "public" access, when you ask for "public write" ;)

Exelord:
You ask for `public write`
```
      token = FactoryBot.create :access_token, default_attributes.merge(
          application: application,
          scopes: 'public write'
        )
```
and expect to not have the token with just `public` scope

The assertion doesn't make sense for me.

It's just obvious that if your app has access to `public write` and you create a token which requests `public write` you should check if the token has `public write`. Not if there is no token with just `public`

bpieck:
Although it does not really matter, that you misread the code - here the correction to your interpretation:

I created a token, with public write (via FactoryBot). And with
`last_token = AccessToken.matching_token_for(application, resource_owner_id, scopes)`
where scopes is "public", the request is for token with scope public. While a Token with "public write" already existing.

But does not really matter. We agree, that this change is not needed anymore and it can be closed :)

Exelord:
Sure, but the spec would be useful to make this future proof.

bpieck:
If you want them to accept this MR - you need at least to squash the commits to one ^^

PR #853, opened 2016-06-23
i18n home nav link

GATAKAWAKACHICO:
Hi, 
Thank you very much providing such nice plugin.
How about using i18n header nav of the home text as well as that of applications?


tute:
Yes! Thank you.

Issue #55, opened 2012-02-23
Include 'spec' directory into s.files


miyagawa:
I wonder if it is intentional that the 'spec' directory is excluded from the s.files in the gemspec. Probably these spec files were meant to be used only for testing on your development end, but I've found that some of the helpers and factories in the doorkeeper git repository are useful on my own app testing. 

For example, i can use the `Factory(:application)` to create a new application, then use `client_is_authorized` to quickly create a new access token, then use `with_access_token_header` to send a new authorized request.

If they are installed in the `gem install`, we can load them using our own spec_helper by looking at `Gem.loaded_specs` and load from `spec/support/helpers` etc. What do you think?

jormon:
https://github.com/doorkeeper-gem/doorkeeper/commit/09cb679186a3197fcc29c7e315e3bcd13ae02de1 breaks this (in 2.1.4+)

jormon:
fixed in v4.2.0 which includes work done in #871

PR #871, opened 2016-08-05
Rename user factory to doorkeeper_testing_user

jormon:
Developers of Rails applications love to include the factories from the doorkeeper gem into their applications because the factories make it easy to do feature testing of the API, stubbing out the relevant doorkeeper models.

However, there is a name conflict on the user factory (which is a global namespace), so that developers cannot include `<path_to_doorkeeper>/spec/factories.rb` if their own factories.rb includes a :user factory (which most do).

This change simply renames the internal factory (which is only used in 1 test anyways) to a scoped name that is much less likely to conflict.  It does not change which class the factory is modeled off of (still the user.rb in the dummy spec app).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,9 @@ User-visible changes worth mentioning.
 
 ## master
 
+- [#871] Allow downstream users to better utilize doorkeeper spec factories by
+  eliminating name conflict on `:user` factory.
+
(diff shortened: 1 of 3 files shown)


jormon:
related to #55

tute:
Merged into master. Thank you! :)

Issue #859, opened 2016-07-03
Improve API documentation


Morpheu5:
I'm finding it extremely difficult to learn how to use Doorkeeper, especially since the most common use cases seem to be not mine. In particular, API documentation would be critical. There are some examples and tutorials, but they aren't exhaustive. In fact, I find myself going through Doorkeeper's code many times over to check whether certain methods exist, what their names are, what parameters they might accept, and so on.

I know this sounds like something I should be taking up myself, but as I said, I'm trying to navigate it, with limited success…

tute:
@Morpheu5, thanks for your comment. I agree that documentation is very lacking for this project. I won't have the time to work on this, but will leave this issue open, and retitle it to better describe the problem, welcoming contributors who might help.
Thanks!

treble37:
@Morpheu5 - do you have a specific issue/question you would like documented or covered better? I just went thru the process of using Doorkeeper to setup an OAuth provider with an existing rails application.  If you post a specific issue/question, then perhaps I (or others) can take a crack at it and improve the docs for everyone's benefit.

tute:
@Morpheu5 what would be actionable steps to move this issue forward? Is there another rubygem for which we can follow the example and improve our API documentation? Is there another software package (not necessarily a gem) you have in mind?
Thanks!

Morpheu5:
Hey - sorry for the delay.

@treble37 I couldn't find clear docs on setting up authentication using the password grant, for example, so I had to read the OAuth2 specification and proceed by trial and error with doorkeeper. I think I got it to work, but I'm not sure I got it to work correctly :) There are a lot of such holes in the docs. To answer @tute's question too, API level documentation is nearly non-existent — I mean documenting which methods do what, what parameters do they expect, and so on. It's really guesswork at this point, working backwards from the examples and the RFC. Then, a higher level set of docs on how to implement and use the different grants, and so on, that would be useful too. I would start with completing the API's docs first, that will eliminate a lot of guesswork.

treble37:
For the API level docs, is this [sample (from grape)](https://github.com/ruby-grape/grape/blob/master/lib/grape/dsl/parameters.rb) what you would consider good documentation for the API @Morpheu5?

Morpheu5:
[...]

PR #881, opened 2016-09-16
Add documentation to config

treble37:
- Document MissingConfiguration Error
- Document Config::Builder class
- Addresses #859

### DIFF
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -53,6 +53,9 @@ doorkeeper.
       # application (disabled by default)
       # Optional parameter confirmation: true (default false) if you want
       # to enforce ownership of a registered application
+      #
+      # @param [Hash] opts the options to confirm if an application owner is present
+      # @option opts[Boolean] :confirmation (false) Set confirm_application_owner variable
       def enable_application_owner(opts = {})
         @config.instance_variable_set('@enable_application_owner', true)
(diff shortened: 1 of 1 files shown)


lib/doorkeeper/config.rb
```diff
@@ -56,44 +67,87 @@ def confirm_application_owner
...
       def optional_scopes(*scopes)
         @config.instance_variable_set('@optional_scopes', OAuth::Scopes.from_array(scopes))
       end
 
+      # Change the way client credentials are retrieved from the request object.
+      # By default it retrieves first from the `HTTP_AUTHORIZATION` header, then
+      # falls back to the `:client_id` and `:client_secret` params from the
+      # `params` object.
+      #
+      # @param [Array] methods Define client credentials
```
Morpheu5: Yes, this is excellent! It could be more precise, like, this is an array of symbols? What are alle the valid symbols that I can pass in? And so on… but this is the kind of polishing that can be done further on. Better to start with covering as much code as possible, and then nailing details down. Good work, thanks a lot! :)

treble37: Ok, I like the idea of covering code as much as possible @Morpheu5, so let's try and do that.

tute:
Thank you!! Let's keep this going. 😃 🚢 ❤️

PR #885, opened 2016-09-23
fix issue with per_form csrf token

lionelperrin:
### Issue

The "destroy" button for application fails with `ActionController::InvalidAuthenticityToken`
### Explanation

Rails may check authenticity token with a 'per form' policy. When doing so, the authenticity token is generated with the form_tag and depends on the url and the method.
The form to delete application is not build with a simple call form_tag but also uses an <input> field for the method. This prevents rails from generating a suitable authenticity token.
### Fix

replace 

```
<%= form_tag target %>
<input type="hidden" name="_method" value="delete">
...
<% end %>
```

with 

```
<%= form_tag target, method: :delete %>
...
<% end %>
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,6 @@ if ENV['rails'].start_with?('5')
 end
 
 gem "activerecord-jdbcsqlite3-adapter", platform: :jruby
-gem "sqlite3", platform: [:ruby, :mswin, :mingw]
-
+gem "sqlite3", platform: [:ruby, :mswin, :mingw, :x64_mingw]
+gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw]
 gemspec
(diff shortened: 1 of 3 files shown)


tute:
Yess, much better. What about generating the form directly with a button to delete?

```ruby
button_to t('doorkeeper.authorized_applications.buttons.revoke'),
  oauth_authorized_application_path(application),
  data: { confirm: t('doorkeeper.authorized_applications.confirmations.revoke') },
  method: :delete,
  class: submit_btn_css
```

Would you do that? Does it work well?

Thanks for your help!

nbulaj:
@tute it will work, but only one moment: `doorkeeper.applications.buttons.destroy` instead of `doorkeeper.authorized_applications.buttons.revoke`.

And is it an expected behavior that we need a Resource Owner for doing that action (`resource_owner_authenticator` block is invoked for that request)?

tute:
Thank you! :)

PR #889, opened 2016-09-29
Replace `destroy` with `delete_all` on app dependent tokens and grants

### LINKED ISSUE #888: Destroying an OAuth Application can lead to a large number of SQL queries
Rails 4.2.7.1
Doorkeeper 4.2.0

When an OAuth Application is destroyed e.g. `Doorkeeper::Application.first.destroy` all associated Access Tokens are also destroyed.

However, this is done using `dependent: :destroy` as seen here in the application mixin: https://github.com/doorkeeper-gem/doorkeeper/blob/master/lib/doorkeeper/models/application_mixin.rb#L11

The problem is that ActiveRecord loads all the dependent tokens and destroys them one-by-one. For a long-running application, this can run into the thousands.

My personal experience with this was that a web request to destroy an application was bumping into our 10 second timeout thanks to all the DB calls.

I'm aware that there has been discussion about deleting tokens previously, and that my issue is more of a QoL problem than a hard 'bug', but it was certainly unexpected.

My gut feeling is that it's safe to simply do a `dependent: delete_all` on the tokens instead, which would make a single SQL query rather than one per token.

davefp:
Fixes #888 by specifying that access tokens and grants for an application are deleted using a single SQL query each rather than loading each one and running callbacks.

Re: Tests: The tests still pass but I'm not sure if it's reasonable to test that the number of queries executed during deletion is reduced. I'm happy to take advice on how best to handle this.

### DIFF
--- a/lib/doorkeeper/models/application_mixin.rb
+++ b/lib/doorkeeper/models/application_mixin.rb
@@ -7,8 +7,8 @@ module Doorkeeper
     include ActiveModel::MassAssignmentSecurity if defined?(::ProtectedAttributes)
 
     included do
-      has_many :access_grants, dependent: :destroy, class_name: 'Doorkeeper::AccessGrant'
-      has_many :access_tokens, dependent: :destroy, class_name: 'Doorkeeper::AccessToken'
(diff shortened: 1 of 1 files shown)


tute:
Thank you! :D

PR #892, opened 2016-10-05
Add a required minimum ruby version to gem spec.

knappe:


tute:
We support 2.1 too: https://github.com/doorkeeper-gem/doorkeeper/blob/master/.travis.yml#L6

Thanks!

knappe:
@tute Updated, thanks for the heads up.

tute:
Thank you!

Issue #879, opened 2016-09-12
OAuth2 functionality without management views for API only apps


RaoulDijksman:
I would love to see a way to use doorkeeper without the management views. I'm expecting I won't be the only one now that Rails 5 is released with API only apps. Would this be a route you are willing to go down?

This might be able to go hand in hand with issue #710

tute:
I'd like to see this, yes. In such a configuration, how would you edit the applications? Thanks!

RaoulDijksman:
It would be nice if it was possible to tell rails what parts of the engine you would like to mount.
Maybe a api only mode like: `use_doorkeeper :api_only` that wouldn't mount the routes nor add assets to the pipeline that belong to views.

Another option would be to split up the project in to 2 gems. The oauth part and a management part.

diddeb:
I tend to agree on this subject. Perhaps splitting the project is too radical, but some way of disabling the admin UI would be appreciated.

tute:
> Perhaps splitting the project is too radical, but some way of disabling the admin UI would be appreciated.

Wholeheartedly agree. The engine shouldn't always inject the routes, according to the `api_mode` Rails option.

tute:
@RaoulDijksman @diddeb can you please give #893 a try? If it works well, I'll release it soon. Thank you!

belgoros:
So if I'm not mistaken, I have to change in my routes as follows:
```
#routes.rb
use_doorkeeper :api_only
```
Right ? What kind of controller should I set as `base_controller` in `doorkeeper.rb`:
```
base_controller 'ApplicationController'
```
takin in account that in Rails API `ApplicationController` is inherited from `ActionController::API`?

prcongithub:
Is this resolved? I am facing a similar issue.

nbulaj:
@prcongithub nope, it is not implemented yet. We have two PRs with implementation for Rails API, but they need some exploration and work.

nbulaj:
@prcongithub @RaoulDijksman merged in ba374aaec7c031a2ae9695e6a5e097551a8d92c0 , can you check it with `master` branch?

prcongithub:
Cool. Will give it a try soon.

PR #893, opened 2016-10-07
Add api_only option

treble37:
- Use doorkeeper without management views for API only Rails apps
- Addresses #879

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,7 +4,7 @@ User-visible changes worth mentioning.
 
 ## master
 
-##4.2.1
+## 4.3.0
 
 - [#879] Add api_only option to use doorkeeper without management views
   for API only Rails applications
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,11 @@ User-visible changes worth mentioning.
 
 ## master
 
+##4.2.1
+
+- [#879] Add api_only option to use doorkeeper without management views
+  for API only Rails applications
+
 ## 4.2.0
 
 - Security fix: Address CVE-2016-6582, implement token revocation according to
--- a/README.md
+++ b/README.md
@@ -35,6 +35,8 @@ https://github.com/doorkeeper-gem/doorkeeper/releases
 - [Protecting resources with OAuth (a.k.a your API endpoint)](#protecting-resources-with-oauth-aka-your-api-endpoint)
   - [Protect your API with OAuth when using Grape](#protect-your-api-with-oauth-when-using-grape)
   - [Route Constraints and other integrations](#route-constraints-and-other-integrations)
+  - [API Only option for use with Rails
+    API](#api-only-option-for-use-with-rails-api)
   - [Access Token Scopes](#access-token-scopes)
   - [Custom Access Token Generator](#custom-access-token-generator)
   - [Authenticated resource owner](#authenticated-resource-owner)
@@ -227,6 +229,20 @@ end
 For more information about integration and other integrations, check out [the
 related wiki
 page](https://github.com/doorkeeper-gem/doorkeeper/wiki/ActionController::Metal-with-doorkeeper).
+
+### API only option for use with Rails API
+
+By default, Doorkeeper comes with a user interface showing the
+[list of registered client applications](#applications-list) as well as authorized applications
+list for an authorized resource owner. If you have a Rails API application and don't need this feature,
+you can disable it with the `:api_only` option in the `use_doorkeeper`
+block in your `routes.rb` file as follows:
+
+```ruby
+Rails.application.routes.draw do
+  use_doorkeeper api_only: true
+end
+```
 
 ### Access Token Scopes
 
--- a/lib/doorkeeper/rails/routes.rb
+++ b/lib/doorkeeper/rails/routes.rb
@@ -5,7 +5,6 @@ module Doorkeeper
   module Rails
     class Routes
       module Helper
-        # TODO: options hash is not being used
         def use_doorkeeper(options = {}, &block)
           Doorkeeper::Rails::Routes.new(self, &block).generate_routes!(options)
         end
@@ -24,6 +23,7 @@ module Doorkeeper
 
       def generate_routes!(options)
         @mapping = Mapper.new.map(&@block)
+        skip_controllers_if_api(options)
         routes.scope options[:scope] || 'oauth', as: 'oauth' do
           map_route(:authorizations, :authorization_routes)
           map_route(:tokens, :token_routes)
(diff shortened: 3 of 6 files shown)


tute:
Exciting changes. Thank you! A few comments:

- This is a new feature as much as a bug fix, next version should be 4.3.0.
- Please add a space in the NEWS title: `## 4.3.0`
- Please add documentation in the README on how to use this option
- Can you please address Hound's comments on new code?

Thank you so much!

treble37:
Thank you for the feedback, I updated with the changes requested!

lcpriest:
Anything I can do to help this PR along?

tute:
Hi! Thanks for your comment. I'd like someone with this issue in their app to verify this is the right approach (asked also in https://github.com/doorkeeper-gem/doorkeeper/issues/879#issuecomment-252651921). Can you do that? Thanks!

nbulaj:
@tute looks good. If question is only in routes - then it is a good and simple solution.

whatthewhat:
> I'd like someone with this issue in their app to verify this is the right approach (asked also in #879 (comment)). Can you do that? Thanks!

@tute Hi! We have been using doorkeeper with these options (which are identical to this PR from what I can tell) for a while now:

```ruby
use_doorkeeper do
  skip_controllers :applications, :authorized_applications
end
```

I think a more interesting problem for the `api_only` mode is actually the `authorizations` controller<sup>*</sup>, bit this is a good first step! 👍 


<sup>*</sup> at the moment we have a separate base controller that can render html to handle that, but it could be interesting to experiment with proxying these screens to a JS app..

cickes:
How do you authorize an app under this feature?  Assuming API Only access, using a POST request to /oauth/authorize yields a CSRF error.  
````ruby
Can't verify CSRF token authenticity.
Completed 422 Unprocessable Entity in 1ms (ActiveRecord: 0.0ms)

ActionController::InvalidAuthenticityToken (ActionController::InvalidAuthenticityToken)
````  

Normally this can be overriden via `protect_from_forgery with: :null_session`.  However, I am not having success.  Thoughts?

**Comment by nbulaj** [comment]
Authorization flow is something that is hard to implement for the API project. What to do with redirects? So there are some questions about using that grant flow in the API mode...

**Comment by cickes** [comment]
Need to create 2 services then?  
1) Public service (obviously protected) authorization service (via authorization flow)
2) API Only service for token creation / `doorkeeper_authorize!`

**Comment by leviathan** [comment]
I'm using `ruby=2.2.5` and `Rails=5.0.1`. 
My rails app is running in API only mode (`Rails::Application.config.api_only = true`).

I'm running this branch in my project:
```
gem 'doorkeeper', :git => 'https://github.com/treble37/doorkeeper', :branch => 'api-only-879'
```

And routing is configured as suggested:
```
Rails.application.routes.draw do
  use_doorkeeper api_only: true
  ...
```

The Rails docs suggest that the application controller should inherit from `ActionController::API`.
But when using: `class ApplicationController < ActionController::API`
then any request to the rails server fails with:

```
Uncaught exception: undefined method `layout' for Doorkeeper::ApplicationsController:Class
	/.rvm/gems/ruby-2.2.5/bundler/gems/doorkeeper-1cee208d2e6b/app/controllers/doorkeeper/applications_controller.rb:3:in `<class:ApplicationsController>'
	/.rvm/gems/ruby-2.2.5/bundler/gems/doorkeeper-1cee208d2e6b/app/controllers/doorkeeper/applications_controller.rb:2:in `<module:Doorkeeper>'
	/.rvm/gems/ruby-2.2.5/bundler/gems/doorkeeper-1cee208d2e6b/app/controllers/doorkeeper/applications_controller.rb:1:in `<top (required)>
```

**When** using the `ActionController::Base` controller, then it works.
But using `ActionController::Base` totally defeats the purpose of having a Rails api_only app!!
```
class ApplicationController < ActionController::Base # works, but totally defeats api_only
```

**Suggestion:** Layout should also be excluded when doorkeeper is running in api mode.

**Comment by maclover7** [comment]
@treble37 would you be able to address the issues that were pointed out above? thanks!

**Comment by treble37** [comment]
Hi @maclover7, just getting back to this after a while, let me read up on the comments and report back!

**Comment by anik3tra0** [comment]
any update?

**Comment by maclover7** [comment]
@anik3tra0 Currently waiting for @treble37 to update the PR, and for more people to test that this works with their app

**Comment by frodsan** [comment]
I had to add this to make it work with Rails-api:

```
# frozen_string_literal: true

class DoorkeeperController < ApplicationController
  def self.protect_from_forgery(*); end

  def self.layout(*); end
end
```

**Comment by ssilvius** [comment]
What is left to do here? @treble37, I'll take this PR off your hands if you could write up a quick todo list.

**Comment by treble37** [comment]
Thank you @ssilvius. I haven't had a chance to look at this and probably won't so thank you for being willing to help!   here's a quick todo list:

1) Rebase the branch against latest master
2) Address the above  comments https://github.com/doorkeeper-gem/doorkeeper/pull/893#issuecomment-274570199 and https://github.com/doorkeeper-gem/doorkeeper/pull/893#issuecomment-260742204
3) Test against the versions of Rails that need to be supported

Thanks again!

**Comment by andresperezl** [comment]
I'm proposing a simple change that addresses both problems in comments https://github.com/doorkeeper-gem/doorkeeper/pull/893#issuecomment-274570199 and https://github.com/doorkeeper-gem/doorkeeper/pull/893#issuecomment-260742204

Simply add to [`application_controller.rb`](https://github.com/doorkeeper-gem/doorkeeper/blob/master/app/controllers/doorkeeper/application_controller.rb):

```ruby
protect_from_forgery with: :exception unless ApplicationController <= ActionController::API
```

This then will involve a two steps process for the user, set the routes with the `user_doorkeeper api_only: true` option and changing the config of  `config.base_controller = 'ActionController::API`

But there is one problem that I'm seeing, and if that if we want to follow the suggestion on  to use `ActionController::API` we need to bump the minimum rails version to 5.0, which was the first release including this class, after merging into rails the  [`rails-api` gem](https://github.com/rails-api/rails-api).

The other option would be using the rails-api gem directly but when running rake, it outputs this errors 
```
Failure/Error: wrap_parameters format: [:json]

NoMethodError:
  undefined method `wrap_parameters' for ActionController::API:Class
```

When using Rails 5, rake runs cleanly.

Fork with proposed change and rebased: https://github.com/andresperezl/doorkeeper/commit/d9094a8b7383dca6a99fa601e6ac297f96cf5894

JoeWoodward:
Another option would be to use `respond_to?(: protect_from_forgery)`

nbulaj:
Hi @treble37. I prefer the solution proposed in #992. It also handles `protect_from_forgery`, Doorkeeper API option, routes and etc, but has more: helpers restriction, redirect cases, JSON responses for API requests, more specs. How to you think, can we continue our work in that direction?

nbulaj:
Closing this PR in order to concentrate work on more solid #992

PR #897, opened 2016-10-25
Implement nonce tracking (WIP, open questions)

toupeira:
This is a proof-of-concept to store `nonce` parameters passed in the pre-authorization stage, to be able to return them later in the token response etc.

My motivation to add this feature is for `doorkeeper-openid_connect`, which requires nonce support for clients that use it, and returns the nonce value as part of the JWT-encoded ID token ([spec reference](http://openid.net/specs/openid-connect-core-1_0.html#NonceNotes))

I'm not entirely sure if it has any use outside of OpenID Connect, the OAuth 2.0 RFCs ([6749](https://tools.ietf.org/html/rfc6749), [6819](https://tools.ietf.org/html/rfc6819)) mention nonces only in passing without further references. If they only make sense for OpenID Connect I'll probably look for other ways to make Doorkeeper more extensible, so I can implement these changes in the OpenID Connect gem instead. The main reason I tried doing it in Doorkeeper first is because I can just add another column to the `access_grants` table to store the nonces, which would be trickier in an extension gem.

So my questions to you:
- Do you have any more information on nonce usage in OAuth2, especially how they should be returned?
- Do you think the current approach makes sense, or would you prefer to keep this functionality out of the main Doorkeeper code?

Some relevant PRs I found:
- [a previous experiment to add nonces](https://github.com/doorkeeper-gem/doorkeeper/pull/476)
- [making pre-authorization fields more dynamic](https://github.com/doorkeeper-gem/doorkeeper/pull/420)

Pending tasks:
- [ ] add specs
- [ ] add migration to add `nonce` column
- [ ] check all flows (I only tested the code flow for now)

### DIFF
--- a/app/views/doorkeeper/authorizations/new.html.erb
+++ b/app/views/doorkeeper/authorizations/new.html.erb
@@ -26,6 +26,7 @@
       <%= hidden_field_tag :state, @pre_auth.state %>
       <%= hidden_field_tag :response_type, @pre_auth.response_type %>
       <%= hidden_field_tag :scope, @pre_auth.scope %>
+      <%= hidden_field_tag :nonce, @pre_auth.nonce %>
       <%= submit_tag t('doorkeeper.authorizations.buttons.authorize'), class: "btn btn-success btn-lg btn-block" %>
     <% end %>
     <%= form_tag oauth_authorization_path, method: :delete do %>
@@ -34,6 +35,7 @@
       <%= hidden_field_tag :state, @pre_auth.state %>
       <%= hidden_field_tag :response_type, @pre_auth.response_type %>
       <%= hidden_field_tag :scope, @pre_auth.scope %>
+      <%= hidden_field_tag :nonce, @pre_auth.nonce %>
       <%= submit_tag t('doorkeeper.authorizations.buttons.deny'), class: "btn btn-danger btn-lg btn-block" %>
     <% end %>
   </div>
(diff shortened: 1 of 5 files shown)


toupeira:
Here's how the integration in doorkeeper-openid_connect would look: https://github.com/doorkeeper-gem/doorkeeper-openid_connect/compare/feat/nonce-tracking?expand=1

tute:
> Do you have any more information on nonce usage in OAuth2, especially how they should be returned?

I don't.

> Do you think the current approach makes sense, or would you prefer to keep this functionality out of the main Doorkeeper code?

I'd think about https://github.com/doorkeeper-gem/doorkeeper/pull/420 again, which might make it easier for you to extend doorkeeper. Regarding the migration, `doorkeeper-openid_connect` could extend doorkeeper's schema with a new table with foreign key.

What are your thoughts?

toupeira:
Ok thanks, I agree that if there's no known use for nonces outside of OpenID Connect, that it's better to keep these modifications out of Doorkeeper. I'll see what changes I need to make to extend this in a somewhat clean fashion (#420 alone won't be enough), and will open another PR for that.

Regarding [your comment in #420](https://github.com/doorkeeper-gem/doorkeeper/pull/420#issuecomment-48112752):

> With this change you'd need to override/monkey patch `Doorkeeper::OAuth::PreAuthorization#hidden_fields`. It would be easier to override the view, using the standard Rails practice of recreating the view tree in the `app` folder.

At the moment it looks to me like overriding a method and making the template more generic will be cleaner than duplicating the whole template, especially if the end-user of the `doorkeeper-openid_connect` gem wants to customize it.

PR #898, opened 2016-10-31
Add some documentation around Helpers::Controller

treble37:
- Tell user where these methods would be called
- Addresses #859


tute:
Thanks! This looks like class-level documentation. Should we put it in the very first line? Thanks!

treble37:
Great point, added to top of file, thanks!

tute:
Thank you! Squashed and merged it in, keeping your authorship. 😃  🎉

PR #904, opened 2016-11-03
Pessimistic Gemfile rspec-rails dependency

nbulaj:
No need to lock rspec-rails on some specific version (like "3.5.2") and update
it with every new release. Rspec follows Semantic Versioning so we can make
a pessimistic dependency for it (>= 3.5, < 4).


tute:
Thanks a lot!

Issue #906, opened 2016-11-07
Problem with scope checking


toupeira:
I ran into a problem with the scope checking, if I have a token which has only some of the available scopes granted, it doesn't get matched (in `AuthorizationsController#matching_token?`) when performing a request with the already granted scopes in the `scope` parameter.

Steps to reproduce:
- Create an application with scope `a`
- Perform an authorization for the scope `a` to issue a token
- Add the scope `b` to the application
- Perform another authorization for the scope `a`
  - What I expect to happen: the authorization passes without user interaction
  - What actually happens: the user needs to confirm the authorization again

While investigating the issue I stumbled over several confusing things / possible problems:
- `AccessTokenMixin.scopes_match?` passes through the `scope` parameter to `ScopeChecker` as second argument, but `ScopeChecker` expects the `server_scopes` there
- `ScopeChecker#match?` looks like it has a wrong argument order (`@valid_scopes.has_scopes?(parsed_scopes)` would make more sense to me, i.e. each requested scope needs to be present in the server/application scopes)
- `access_token_spec` has several examples for `matching_token_for` where the description suggests that the search should be successful, but the expectation is looking for a `nil` token (this might just be ambiguous wording though)

Not sure yet what exactly needs fixing so I thought I'd ask for your input first :)

It seems to me like we might need to perform two checks:
- each requested scope needs to be present in the token
- each scope from the token needs to be present in the server/application configuration

PR #907, opened 2016-11-08
Fix lookup for matching tokens when application has additional scopes

toupeira:
Follow-up to #906 

I traced the changes back to https://github.com/doorkeeper-gem/doorkeeper/commit/03008f5 and https://github.com/doorkeeper-gem/doorkeeper/pull/588, which made me realize that the existing token needs to match the requested scopes *exactly*, otherwise the user will end up with a token that has more permissions than requested. So `ScopeChecker#match?` seems to be correct, and the swapped order is already checked in `#valid?` as well. Sorry for the confusion, but I guess that can't be avoided when dealing with OAuth ;-)

Looking at `PreAuthorization#authorizable?` the scopes from the params are already verified to be present in the server/application scopes, so I think `AccessTokenMixin.matching_token_for` only needs to check if there is a token that matches the requested scopes. These changes try to accomplish that (as well as fixing the ambiguous spec descriptions), though I guess we're still abusing `ScopeChecker` a bit :)

One thing I'm still unsure about: currently only the last authorized token is checked, but what if there is a previously authorized, still-valid token matching the requested scopes? Shouldn't that also cause the authorization to be skipped? If yes I'll change the code to return the newest one matching the scopes.

### DIFF
--- a/spec/requests/flows/skip_authorization_spec.rb
+++ b/spec/requests/flows/skip_authorization_spec.rb
@@ -15,13 +15,24 @@ feature 'Skip authorization form' do
     end
 
     scenario 'skips the authorization and return a new grant code' do
-      client_is_authorized(@client, @resource_owner, scopes: 'public')
-      visit authorization_endpoint_url(client: @client)
+      client_is_authorized(@client, @resource_owner, scopes: "public")
+      visit authorization_endpoint_url(client: @client, scope: "public")
 
-      i_should_not_see 'Authorize'
+      i_should_not_see "Authorize"
       client_should_be_authorized @client
       i_should_be_on_client_callback @client
-      url_should_have_param 'code', Doorkeeper::AccessGrant.first.token
+      url_should_have_param "code", Doorkeeper::AccessGrant.first.token
+    end
+
+    scenario "skips the authorization if other scopes are not requested" do
+      client_exists scopes: "public read write"
+      client_is_authorized(@client, @resource_owner, scopes: "public")
+      visit authorization_endpoint_url(client: @client, scope: "public")
+
+      i_should_not_see "Authorize"
+      client_should_be_authorized @client
+      i_should_be_on_client_callback @client
+      url_should_have_param "code", Doorkeeper::AccessGrant.first.token
     end
 
     scenario 'does not skip authorization when scopes differ (new request has fewer scopes)' do
@@ -41,12 +52,6 @@ feature 'Skip authorization form' do
       visit authorization_endpoint_url(client: @client, scope: 'public')
       click_on 'Authorize'
       access_grant_should_have_scopes :public
-    end
-
-    scenario 'doesn not skip authorization when scopes are greater' do
-      client_is_authorized(@client, @resource_owner, scopes: 'public')
-      visit authorization_endpoint_url(client: @client, scope: 'public write')
-      i_should_see 'Authorize'
     end
 
     scenario 'creates grant with new scope when scopes are greater' do
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,7 @@ User-visible changes worth mentioning.
 - [#1032] Refactor BaseRequest callbacks into configurable lambdas
 - [#1040] Clear mixins from ActiveRecord DSL and save only overridable API. It
   allows to use this mixins in Doorkeeper ORM extensions with minimum code boilerplate.
+- [#907] Fix lookup for matching tokens in certain edge-cases
 
 ## 4.3.0
 
--- a/lib/doorkeeper/models/access_token_mixin.rb
+++ b/lib/doorkeeper/models/access_token_mixin.rb
@@ -69,30 +69,33 @@ module Doorkeeper
                             else
                               resource_owner_or_id
                             end
-        token = last_authorized_token_for(application.try(:id), resource_owner_id)
-        if token && scopes_match?(token.scopes, scopes, application.try(:scopes))
-          token
+        tokens = authorized_tokens_for(application.try(:id), resource_owner_id)
+        tokens.detect do |token|
+          scopes_match?(token.scopes, scopes, application.try(:scopes))
         end
       end
 
-      # Checks whether the token scopes match the scopes from the parameters or
-      # Application scopes (if present).
+      # Checks whether the token scopes match the scopes from the parameters
       #
       # @param token_scopes [#to_s]
       #   set of scopes (any object that responds to `#to_s`)
-      # @param param_scopes [String]
+      # @param param_scopes [Doorkeeper::OAuth::Scopes]
       #   scopes from params
-      # @param app_scopes [String]
+      # @param app_scopes [Doorkeeper::OAuth::Scopes]
       #   Application scopes
       #
-      # @return [Boolean] true if all scopes are blank or matches
+      # @return [Boolean] true if the param scopes match the token scopes,
+      #   and all the param scopes are defined in the application (or in the
+      #   server configuration if the application doesn't define any scopes),
       #   and false in other cases
       #
       def scopes_match?(token_scopes, param_scopes, app_scopes)
-        (!token_scopes.present? && !param_scopes.present?) ||
-          Doorkeeper::OAuth::Helpers::ScopeChecker.match?(
-            token_scopes.to_s,
-            param_scopes,
+        return true if token_scopes.empty? && param_scopes.empty?
+
+        (token_scopes.sort == param_scopes.sort) &&
+          Doorkeeper::OAuth::Helpers::ScopeChecker.valid?(
+            param_scopes.to_s,
+            Doorkeeper.configuration.scopes,
             app_scopes
           )
       end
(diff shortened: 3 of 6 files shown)


spec/requests/flows/skip_authorization_spec.rb
```diff
@@ -16,7 +16,18 @@
...
+      i_should_not_see 'Authorize'
+      client_should_be_authorized @client
+      i_should_be_on_client_callback @client
+      url_should_have_param 'code', Doorkeeper::AccessGrant.first.token
+    end
+
+    scenario 'skips the authorization if the application has additional scopes which are not requested' do
+      client_exists scopes: 'public read write'
+      client_is_authorized(@client, @resource_owner, scopes: 'public')
+      visit authorization_endpoint_url(client: @client, scope: 'public')
```
toupeira: @tut should I ignore these or fix them?

toupeira:
@tute I was wondering if you already had time to take a look at this PR?

bufferoverflow:
@toupeira I would say just use double-quoted strings as mentioned by the bot.

toupeira:
@tute I wish you a happy new year :)

Do you have time for a review? Let me know if you have any questions about the bug or the fix.

tute:
> @tute I wish you a happy new year :)

Thank you very much for this and your patience. :) 

> Do you have time for a review? Let me know if you have any questions about the bug or the fix.

I don't have much time for a proper review. I just announced we need new maintainers here: https://twitter.com/tutec/status/819174354609143813. :-/

toupeira:
@tute ah too bad, but I can relate :) Thanks for all your work!

toupeira:
I reported this to the [Gitlab issue tracker](https://gitlab.com/gitlab-org/gitlab-ce/issues/26575), in case they're interested.

timreyn:
Hi all,

This issue came up in an app I'm working on; wondering if there's an update to this PR, or to doorkeeper's maintenance situation?

Many thanks, particularly to @toupeira for figuring out the details so that we didn't have to!

toupeira:
@maclover7 I noticed you did some maintenance on the doorkeeper gem recently, do you have time to take a look at this PR? Let me know if you have any questions, though I'll have to wrap my head around it again as well ;-)

nbulaj:
Hi @toupeira. What did OAuth2 specs says about this? I think we need to implement it according to standards, maybe you have a quick-link? Sounds good, but we need to check it I think.

f3ndot:
@nbulaj this looks like a major security issue that probably warrants a CVE (I'll request one once fix is out). In my opinion we should prioritize getting this merged and into a v4.3.1 release.

From what I can find, I see nothing in the specs defining how scopes should be handled explicitly. However, I think it's implied that an access token found should not have scopes be broader than what the end user has granted.

The closest paragraph on this matter I could find is [Section 1.5](https://tools.ietf.org/html/rfc6749#section-1.5) (emphasis added):

>   Refresh tokens are credentials used to obtain access tokens.  Refresh
   tokens are issued to the client by the authorization server and are
   used to obtain a new access token when the current access token
   becomes invalid or expires, or to **obtain additional access tokens
   with identical or narrower scope** (access tokens may have a shorter
   lifetime and fewer permissions than authorized by the resource
   owner).  Issuing a refresh token is optional at the discretion of the
   authorization server.  If the authorization server issues a refresh
   token, it is included when issuing an access token (i.e., step (D) in
   Figure 1).

* * *

And @toupeira RE: your final paragraph:

> One thing I'm still unsure about: currently only the last authorized token is checked, but what if there is a previously authorized, still-valid token matching the requested scopes? Shouldn't that also cause the authorization to be skipped? If yes I'll change the code to return the newest one matching the scopes.

I agree with this strategy. If we're searching for a matching token by their scopes exactly, we should continue to look for all still valid matches-- not just the latest one.

f3ndot:
I apologize. I think I spoke too soon. My concern was around was maybe a related issue that no longer applies. I couldn't perform the security exploit I thought could occur:

https://github.com/f3ndot/doorkeeper/commit/2a2f9279fab2748742b92a8dec84d327004e6508

@toupeira can you think of a scenario where the code could be exploited to erroneously get a token that should not happen?

nbulaj:
Hi @toupeira . Can't continue to work on Doorkeeper till we finish with this issue :)

Can you provide a failing spec where somebody can get the token with over-permissions?

nbulaj:
Hi @toupeira . Does your issue correlate with this one #957 ?

I see some similiar changes there: https://github.com/creatubbles/doorkeeper/commit/24320897e3c33adc2218bb4c789d72ab98979d39 (not shown in PR above)

toupeira:
@nbulaj sorry for the delay, finally got around to looking into this :)

I rebased on master and added the app scopes back into `scopes_match?`, the main difference is that we're calling `Doorkeeper::OAuth::Helpers::ScopeChecker.valid?` now instead of `.match?`, so now we're only verifying that each param scope is present in the application/server scopes, and not anymore that each application/server scope is present in the param scopes (which as far as I can tell never made any sense).

The param scopes are matched to the token scopes with a simple sort and comparison, from the specs in `spec/requests/flows/skip_authorization_spec.rb` it seems that this is indeed the desired behaviour (only return tokens that match *all* scopes).

I'm not really sure what to do when the client doesn't send a `scope` parameter at all, Doorkeeper grants all default scopes in this case. Maybe we should look for tokens with all the default scopes? This might have unintended consequences if the default scopes are changed after a token is issued, but I guess it's really an edge case.

> Can you provide a failing spec where somebody can get the token with over-permissions?

I'm not sure that's actually possible, the issue I had was that the authorization screen was shown again even if the client already had a valid token, so mostly just a minor inconvenience.

> Does your issue correlate with this one #957 ?
> I see some similiar changes there: creatubbles/doorkeeper@2432089 (not shown in PR above)
> And this one also #941

#957 seems to be related to the empty `scopes` param case. I added the specs from https://github.com/creatubbles/doorkeeper/commit/24320897e3c33adc2218bb4c789d72ab98979d39#diff-a5750277cf6e867d3b1c2649bf3c896dR373 and they succeed, but I'm still not sure if this is really the correct behaviour. I only included the first spec ("does not match token if empty scope requested and token/app scopes present") in this PR since the second one is already covered by other specs.

#941 looks like my original issue.

nbulaj:
Hi @toupeira . Something go wrong with the specs on Travis. As I can see it is connected to logic changes for `last_authorized_token_for` method. First of all, it can be painless for some Doorkeeper ORM extensions to use this update as it can break them (extensions implements this method in their own way). And, the more important, it now returns not one token, but the collection.

`last_authorized_token_for` is used in Access Token mixin in case when `reuse_access_token` configuration option is ON (to return token for update) and for the Refresh Token strategy (it finds last matching token and refreshes it). I don't see any problems if we return just latest matching token and work with it. All others just will expire after some configured time and even if somebody will use it - no error can occur because they all are valid.

toupeira:
@nbulaj I think we do need to consider older tokens, see previous discussion with @f3ndot:

> And @toupeira RE: your final paragraph:
>
> >  One thing I'm still unsure about: currently only the last authorized token is checked, but what if there is a previously authorized, still-valid token matching the requested scopes? Shouldn't that also cause the authorization to be skipped? If yes I'll change the code to return the newest one matching the scopes.
>
> I agree with this strategy. If we're searching for a matching token by their scopes exactly, we should continue to look for all still valid matches-- not just the latest one.

But I added back the `last_authorized_token_for` method now for backwards-compatibility, it delegates to `authorized_tokens_for` and returns the first match (sorted by `:created_at :desc`).

Some other failures were because of requests with empty scopes, these are matched again now if the token doesn't have any scopes either (in the case of #957 the token had scopes as well).

The remaining specs were failing because the used scopes weren't setup in the server configuration, they should run now as well.

nbulaj:
Hey @toupeira , that's much better I think :) I will take a close look tomorrow because today I have no time. Could you please add a changelog entry to NEWS.md ? Thanks!

toupeira:
@nbulaj sure, NEWS.md is updated.

nbulaj:
Ok, so @toupeira what do you suggest about processing ALL the matching tokens? How do you think we need to handle this situation?

toupeira:
@nbulaj not sure I understand the question, can you rephrase? :)

Previously the matching was incomplete, since the scopes were only compared on the last authorized token. So if the user created another token with other scopes, suddenly the previous token became invalid even though it was still authorized. Now all authorized tokens are considered to find one with matching scopes.

nbulaj:
Ah, OK, so now I get it @toupeira  :) I misunderstood you and it was exactly what you answered about. Just forget my previous question.

I think it is ready to be merged. The only question remained:

> I'm not really sure what to do when the client doesn't send a scope parameter at all, Doorkeeper grants all default scopes in this case. Maybe we should look for tokens with all the default scopes? This might have unintended consequences if the default scopes are changed after a token is issued, but I guess it's really an edge case.

Everybody need to understand that something may go wrong if he will change default scopes. The same as access token / grant TTL for example (old one will work with old value, not the new one). So it looks like the most suitable solution, but maybe requires some notification in Doorkeeper initializer comments in order to warn developers about the cases.

nbulaj:
Hi @toupeira . Big thanks for your work and patience :)

If you agree with my thoughts about wht to do if `client doesn't send a scope parameter at all` (see message above), then you can send a new PR 👍 Thanks once again!

toupeira:
@nbulaj thanks to you as well :)

I added a [note to the wiki](https://github.com/doorkeeper-gem/doorkeeper/wiki/Using-Scopes/_compare/e729bd70f4806dad6a20effbb520aefdbe7aa369) since I'm not too sure about the wording and current users won't see a newly added message to the initializer template anyway.

PR #909, opened 2016-11-30
Exposing InvalidTokenResponse reason.

rodrigopinto:
This PR is for making it possible to read the reason of an `InvalidTokenResponse` error. 

**Motivation:**

When trying to render a custom unauthorized response using [customize the response body when unauthorized][1] I needed to fetch the type of error so that the consumer could handle the response differently based on the type.

For example, if the error is `expired` they could try to refresh the token, but if it's `revoked`, the process would be different.

When I was implementing the custom method, I needed to do some metaprogramming to fetch the error type, it would be nice to avoid doing this.

What I did:

```ruby
def doorkeeper_unauthorized_render_options(error: nil)
  {
    json: {
      error: {
        code: error.instance_variable_get(:@reason), # <<<<<<< here
        title: error.description
      }
    }
  }
end
```

**Solution**

So my idea with this pull request is to expose the `reason` variable within the `InvalidTokenResponse`class so that I can access the reason. This then allows me to do the following instead:

```ruby
def doorkeeper_unauthorized_render_options(error: nil)
  {
    json: {
      error: {
        code: error.reason, # <<<<<<< here
        title: error.description
      }
    }
  }
end
``` 

ps: I'll squash the commits, after the review. :)

What do you think?

[1]: https://github.com/doorkeeper-gem/doorkeeper/wiki/Customizing-the-response-body-when-unauthorized

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,9 @@
 
 User-visible changes worth mentioning.
 
+- Add `InvalidTokenResponse#reason` reader method to allow read the kind of
+  invalid token error.
+
 ## master
 
 ## 4.2.0
--- a/lib/doorkeeper/oauth/invalid_token_response.rb
+++ b/lib/doorkeeper/oauth/invalid_token_response.rb
@@ -1,6 +1,8 @@
 module Doorkeeper
   module OAuth
     class InvalidTokenResponse < ErrorResponse
+      attr_reader :reason
+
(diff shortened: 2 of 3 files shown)


rodrigopinto:
@maclover7 Sorry to bother, but could you take a look at this proposal. Thanks.

maclover7:

PR #911, opened 2016-12-06
Remove unnecessarily extended ActiveSupport::Concern

amatsuda:
These modules do not use `included` nor `ClassMethods`, so they don't have to extend ActiveSupport::Concern.


tute:
Thank you! :)

PR #915, opened 2017-01-05
Update doorkeeper to be used with grape >= 0.10

tim-vandecasteele:
Grape has since long discouraged the use of options[:route_options] for custom additions. Since version 0.10 it cannot be used anymore like setup in the helpers in doorkeeper (grape is currently at 0.19.1).

This PR fixes the behaviour to properly work with grape, and describes a bit better in the example how you could actually use it and set it up (if you want I can elaborate more in the readme).

I think with this example and the change, most people should have enough with doorkeeper as is, and not need wine_bouncer at all.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ You can pass any option `before_action` accepts, such as `if`, `only`,
 
 ### Protect your API with OAuth when using Grape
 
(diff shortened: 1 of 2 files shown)


tute:
Thank you! :)

tim-vandecasteele:
Thanks for merging :)

PR #917, opened 2017-01-12
Change find_by to where/first to be mongoid friendly

alechoey:
Using find_by introduces breaking behavior for applications using Mongoid5 with doorkeeper-mongodb without the `raise_if_not_found` configuration flag set.

Setting the `raise_if_not_found` configuration flag produces significant behavior changes in applications using Mongoid throughout the application including calls to `find` and `reload`

### DIFF
--- a/lib/doorkeeper/models/access_grant_mixin.rb
+++ b/lib/doorkeeper/models/access_grant_mixin.rb
@@ -36,7 +36,7 @@ module Doorkeeper
       #   if there is no record with such token
       #
       def by_token(token)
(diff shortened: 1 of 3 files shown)


nbulaj:
Hi @alechoey . `find_by` caches query results (so more performant). Moreover, it is required to refactor doorkeeper-mongoid gem, not the doorkeeper itself.

nbulaj:
Will be fixed soon with new release of `doorkeeper-mongodb` gem. Anyway thanks for the PR! 👍

PR #920, opened 2017-01-15
Specify correct wikki pages from the way client credentials is config ured on doorkeeper initializer.

bekicot:
Doorkeeper currently have 35 pages of wikis, directing to the right wikis will save time for user. I found it difficult to find corresponding wiki, hopefully this change will help people having the same problem as i do.

I didn't add to news as it is minor changes. Will add if it required.


nbulaj:
Why not? I think it can be helpful. Can you please add an entry to [NEWS.md](NEWS.md) and squash all in one commit? Thanks!

nbulaj:
Thanks @bekicot , I decided that we can skip an entry in NEWS.md for this PR.

PR #933, opened 2017-01-21
Handle random newlines in the grant_type param

andrewclink:
When x-www-form-encoded data is posted to /oauth/token, the possibility of an invisible newline at the end can suck away more development time than one might want to spend on a Saturday morning.

Tests still pass. 

I don't know if this is the right place to handle it, but in lieu of more descriptive error messages (at least in development), this obviates the problem.

### DIFF
--- a/app/controllers/doorkeeper/tokens_controller.rb
+++ b/app/controllers/doorkeeper/tokens_controller.rb
@@ -71,7 +71,7 @@ module Doorkeeper
     end
 
     def strategy
-      @strategy ||= server.token_request (params[:grant_type] && params[:grant_type].chomp)
+      @strategy ||= server.token_request params[:grant_type].try(:chomp)
     end
 
     def authorize_response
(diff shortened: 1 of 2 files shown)


maclover7:
@andrewclink Thanks for the PR! Can you handle it [here](https://github.com/doorkeeper-gem/doorkeeper/blob/master/app/controllers/doorkeeper/tokens_controller.rb#L74), where we grab the param from the hash? Thanks!

maclover7:
@andrewclink Can you squash down your commits, and add a test for the new behavior? Looks to ready to go after that.

nbulaj:
@andrewclink what about squashing commits? And I think you need to add an entry to the `NEWS.md` (take a look at some previous items in the file for example).

andrewclink:
Sorry for the delay; I'll see if I can get some time this afternoon to take care of it.

nbulaj:
Hi @andrewclink. Any changes would come?

nbulaj:
Hi @andrewclink. Any changes here?

nbulaj:
What if newline will be not only in `grant_type` parameter (`code`, for example)?

nbulaj:
Closing due to lack of activity. Feel free to send this PR once again with fixed comments

PR #948, opened 2017-03-28
Fix comparing Scopes to non-enumerable instances

wktk:
For example, `scopes == false` raises a `NoMethodError` that says `#map` is not defined for `false`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,7 @@ User-visible changes worth mentioning.
 
 ## master
 
+- [#948] Make Scopes.<=> work with any "other" value.
 - [#970] Escape certain attributes in authorization forms.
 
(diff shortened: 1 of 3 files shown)


wktk:
This fixes #951

nbulaj:
@wktk looks good to me. I think it will be good if you will add an entry to the NEWS.md file and squash all the commits to a single one.

@maclover7 what do you think about dat?

wktk:
Thank you for reviewing.

> add an entry to the NEWS.md file and squash all the commits to a single one

done.

wktk:
@nbulaj @maclover7 Hello, is there any chance it gets merged?

nbulaj:
Thanks @wktk for your work and patience!

PR #950, opened 2017-04-18
Explicitly add controller actions

will-in-wi:
In https://github.com/doorkeeper-gem/doorkeeper/wiki/Customizing-routes, it is encouraged in some circumstances to inherit from `Doorkeeper::ApplicationsController`. In some cases, it works well to add some code to all the actions (in my case, authorization).

If I do this, it looks something like:

```ruby
def index
  authorize! :index, :oauth_application
  super
end
```

However, this doesn't work for the edit or show actions, as they rely on implicit logic from Rails. This patch allows super to be called for them, which makes it more robust in case you ever add logic to those methods.

Thanks!

### DIFF
--- a/app/controllers/doorkeeper/applications_controller.rb
+++ b/app/controllers/doorkeeper/applications_controller.rb
@@ -9,6 +9,8 @@ module Doorkeeper
       @applications = Application.all
     end
 
+    def show; end
+
     def new
       @application = Application.new
     end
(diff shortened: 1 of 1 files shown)


nbulaj:
@maclover7 what do you think?

maclover7:
@will-in-wi Can you squash to one commit?

will-in-wi:
Squashed!

maclover7:

PR #954, opened 2017-04-23
Change to allow optional scopes in the authorization code request

semnil:
As with Password Access Token Request, this is a change to make the optional scope available for authorization code request.

I cannot know the reason why this is limited to the default scopes.
If I should not make this change, please let me know why.


semnil:
I'm sorry. It seems it was a misunderstanding.

PR #956, opened 2017-04-26
Remove the dead code for legacy Rails

tricknotes:
The Rails 3.x support is dropped since 02949f96.


nbulaj:
@maclover7 I think it can be quickly merged too (yep, support of 3.x Rails was dropped).

maclover7:

Issue #957, opened 2017-04-28
AccessToken.matching_token_for returns wrong values


swilgosz:
I noticed, that in some cases the `Doorkeeper::AccessToken.matching_token_for` returns a token even if it should return nil

*Steps to reproduce.*

1. Get user: `owner = ResourceOwnerClass.find(:resource_owner_id)`
2. Create `app` with scope: `app = Doorkeeper::Application.create(name: "test-app", redirect_uri: "urn:ietf:wg:oauth:2.0:oob", scopes: "sample:scope")`
3. Create a user access token the same scope:  `token = Doorkeeper::AccessToken.create(application_id: app.id, scopes: "sample:scope", resource_owner_id: owner.id)`
4. `Doorkeeper::AccessToken.matching_token_for(app, u.id, "")` returns token , but should return `nil`

swilgosz:
I'm still working on the fix and adding some more tests coverage

nbulaj:
Hi @swilgosz , take a look at the last comments in #907 - you can refactor this PR to fix blank scopes.

PR #964, opened 2017-05-08
AccessToken.matching_token_for returns wrong results

swilgosz:
**Description**

If there is access_token and application already saved in database with matching scopes,
access_token_for returns first token, completely ignoring ‘scopes’ argument.

This happens, because in ’Doorkeeper::OAuth::ScopeChecker’ the ‘match?’ method
compares token scopes against application scopes if application scopes are present.

**Expected behaviour:**

If there is no token with scopes requested in ‘mathing_token_for’, ‘nil should be returned,
so new token could be created.

This affects custom authentication methods based on scope checking
and it is potential security problem.

**Resolves** #957

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,6 +14,7 @@ User-visible changes worth mentioning.
 - Small refactorings within the codebase
 - [#921] Switch to Appraisal, and test against Rails master
 - [#892] Add minimum Ruby version requirement
+- [#957] AccessToken.matching_token_for return wrong values for empty string
 
 ## 4.2.0
 
--- a/lib/doorkeeper/models/access_token_mixin.rb
+++ b/lib/doorkeeper/models/access_token_mixin.rb
@@ -95,7 +95,7 @@ module Doorkeeper
                               resource_owner_or_id
                             end
         token = last_authorized_token_for(application.try(:id), resource_owner_id)
-        if token && scopes_match?(token.scopes, scopes, application.try(:scopes))
+        if token && scopes_match_exactly?(token.scopes, scopes, application.try(:scopes))
           token
         end
       end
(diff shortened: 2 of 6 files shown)


piotrj:
@swilgosz I don't know the code that well and would have to do a more thorough investigation, but isn't the current behavior correct? 

Here is my reasoning (version 1):
1. User tries to authorize an app with empty set of scopes
2. He already has an access token with scope "read"
3. This means that he does not have to authorize this app cause he already authorized the app for a set of scopes that is larger than what he asked for now.

My reasoning (version 2):
1. User tries to authorize an app with empty set of scopes.
2. He already has an access token with scope "read".
3. Since he didn't specify any scopes (the ones specified are empty) then we fallback to app's scopes.  Let's say app's scope is "read"
4. This means that the token is valid, the user already authorized the app and we can successfully return.

Nevertheless, I think there is a bug there. But a different one that you are describing. The problem is that if application has some default scopes, then we never take the scopes passed in params into considerations. I guess there was some misunderstanding between 417b9d57 and then some commit somewhere else. The `ScopeChecker.match?` method's second argument is called `server_scopes` but we pass the scopes we get in params there. Seems wrong. I need to dig deeper there. 

So to sum up -  I don't think we want exact match of scopes. Cause if you are asking for 'read' and you have already access token that has scopes 'read write' then this AccessToken matches the request scopes (and has more). 

Thoughts?

toupeira:
@swilgosz I think my changes in https://github.com/doorkeeper-gem/doorkeeper/pull/907 might have resolved this issue as well, please check the new behaviour.

nbulaj:
Hi @swilgosz. Can you check the `master` branch please?

PR #965, opened 2017-05-08
Instead of returning a meaningless `nil`, use true [breaking]

krainboltgreene:


maclover7:
Hi @krainboltgreene! I think `doorkeeper_authorize!` is meant to be used as a `before_action` in a Rails controller. Can you provide more information about your use case for this method, and why you need this return type?

krainboltgreene:
As we know a before_action is just a method called before another method. In many cases it can be advantageous to combine multiple methods. For example, let's say you want to support OAuthentication *and* another form of authentication. Returning nil is a falsy value, despite being an approval from doorkeeper.

The other thing I noticed is that this function also has the chance of returning `" "` for a failed authentication, which is a truthy value.

piotrj:
@krainboltgreene as @maclover7 said, this method is supposed to be run as a before filter of Rails controller. It works under the assumption that if before filter renders something the action is not run. If it doesn't render anything (and doesn't return `false`) then the action will run (This is not true in Rails 5 anymore - before filters are supposed to `throw :abort` if they want to halt the chain. 

Probably there should be another method that answers a question "was the doorkeeper authorization successful". `valid_doorkeeper_token?` is almost what you need (but you would need to set `@_doorkeeper_scopes` manually before running it)

krainboltgreene:
Okay, that's great, except what if you have a before action that decides to do two different types of authentication?

piotrj:
@krainboltgreene sorry for such a delay in response.

I was thinking about this a little bit. And it seems that your case is valid. And indeed we could return true here, it doesn't really matter. But if you use something like this in your callback

```
before_action :authorize

def authorize
  doorkeeper_authorize! || other_authorize
end
```

Then if doorkeeper_authorize! returns true then it's fine. But if the authorization will not be successful then this method renders error. Which means that whatever you do in `other_authorize` your action won't run.

Does it make sense.

I guess maybe we could have some different method to check whether doorkeeper authorization is successful? Maybe you could just use `valid_doorkeeper_token?` method? Let me know what you think.

krainboltgreene:
Yeah, this isn't like the highest super blocking issue. It just seemed like "If return values matter in ruby, seems like the return value should kinda matter for this."

I'm currently using `valid_doorkeeper_token?` as well, though.

piotrj:
@krainboltgreene cool. Should we close this PR then?

nbulaj:
Hi @krainboltgreene, what the status of this PR?

maclover7:
Going to close this out -- my suggestion is to use use `valid_doorkeeper_token?`, which is what this callback is using anyways.

PR #970, opened 2017-05-25
Fix XSS issues in default views

### LINKED ISSUE #969: XSS in default views
Default authorization and application views use the "raw" function which inevitably lead to xss issue :

https://github.com/doorkeeper-gem/doorkeeper/blob/14c9ed154a4363593b18216f302d7fc2f8b62120/app/views/doorkeeper/authorizations/new.html.erb#L7
https://github.com/doorkeeper-gem/doorkeeper/blob/14c9ed154a4363593b18216f302d7fc2f8b62120/app/views/doorkeeper/applications/_form.html.erb#L24

simkim:
Default views allow client to xss resource owner to make him grant access, or client to xss admin.

Fix #969

### DIFF
--- a/app/views/doorkeeper/applications/_form.html.erb
+++ b/app/views/doorkeeper/applications/_form.html.erb
@@ -21,7 +21,7 @@
       </span>
       <% if Doorkeeper.configuration.native_redirect_uri %>
           <span class="help-block">
-            <%= raw t('doorkeeper.applications.help.native_redirect_uri', native_redirect_uri: "<code>#{ Doorkeeper.configuration.native_redirect_uri }</code>") %>
+            <%= raw t('doorkeeper.applications.help.native_redirect_uri', native_redirect_uri: content_tag(:code) { Doorkeeper.configuration.native_redirect_uri }) %>
           </span>
       <% end %>
     </div>
--- a/app/views/doorkeeper/authorizations/new.html.erb
+++ b/app/views/doorkeeper/authorizations/new.html.erb
@@ -4,7 +4,7 @@
 
 <main role="main">
   <p class="h4">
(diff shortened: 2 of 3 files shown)


maclover7:
Merging as a bugfix, will be released as part of normal process. Since we are putting restrictions on admins, they could technically just overwrite the forms, and do whatever they want. Such is a risk when using someone else's server for authorization 😬 

If you have any possible security issues to report in the future, please contact the maintainers of this project before posting about them publicly.

simkim:
Sure, how can I improve the CONTRIBUTING file, are there designated maintainers for those kind of issue ?

f3ndot:
Requested CVE for this because impact can be quite severe given it can execute on the authorization grant view.

f3ndot:
Assigned CVE-2018-1000088

f3ndot:
Writeup published: https://blog.justinbull.ca/cve-2018-1000088-stored-xss-in-doorkeeper/

tobypinder:
@f3ndot This writeup is comprehensive but the "Fix" section fails to identify the risk to client applications that have [customised views](https://github.com/doorkeeper-gem/doorkeeper/wiki/Customizing-views) via `rails generate doorkeeper:views` - if I understand correctly any future fixes to default views won't automatically be propagated back to these users as Doorkeeper will just use _their_ templates, which will be based on the vulnerable old defaults. 

I would imagine that this customisation would be performed by a significant proportion of Doorkeeper's userbase given that most would want to customise their user experiences, so I imagine this is not some niche edge-case.

f3ndot:
@tobypinder Ah! Thanks for pointing that out. The writeup has been updated to call this particular scenario out.

PR #976, opened 2017-06-29
Fix to invalidate the second redirect URI when the first URI is the native URI

eitoball:
This PR should fix to invalidates the second (or further) redirect URI when multiple redirect URIs are entered and the first URI is native URI like `urn:ietf:wg:oauth:2.0:oob`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,7 @@ User-visible changes worth mentioning.
 
 ## master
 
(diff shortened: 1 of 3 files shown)


nbulaj:
Hi @eitoball. Yep, it's a nice point! Could you please resolve the conflicts, add an entry to NEWS.md file and squash commits to a single one? Thank you!

eitoball:
All Done. Thank you for checking in.

PR #977, opened 2017-06-29
FIX migration 5.1

DmitryBash:


DmitryBash:
` No such file or directory @ dir_initialize - db/migrate`
Will `Dir.pwd`  => `Dir.entries("#{Dir.pwd}/db/migrate")` correct this issue?

nbulaj:
What about to replace `rails_v = "#{Rails::VERSION::MAJOR}.#{Rails::VERSION::MINOR}".to_f ; if rails_v <= 5.0` to `ActiveRecord::VERSION::MAJOR >= 5` ? I think it would be more compact and elegant without any problems and type conversions.

What about db/migrate path - take a look to the documentation http://api.rubyonrails.org/classes/Rails/Generators/Migration.html#method-i-migration_template:

> The migration version, migration file name, migration class name are available as instance variables in the template to be rendered.

So in the template you can render current Rails version if `respond_to?` version.

DmitryBash:
@nbulaj Hey, what about something like this?

maclover7:
👍 Thanks for this PR! Merged with tests via #981.

PR #982, opened 2017-07-29
chore: Do not allow failure on rails 5.1

sanemat:
:)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,5 +22,3 @@ matrix:
       rvm: 2.1
     - gemfile: gemfiles/rails_5_1.gemfile
       rvm: 2.1
-  allowed_failures:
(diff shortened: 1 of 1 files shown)


nbulaj:
Hi @sanemat . There is a problem: `gemfiles/rails_5_1.gemfile` refers to `gem "rails", :github => "rails/rails"`, so it means that `rails_5_1.gemfile` is looking into the master branch of Rails, where 5.2 version will be very soon. I think @maclover7 that we need to rename `rails_5_1.gemfile` to `rails-edge.gemfile` and add a new gemfile for Rails 5.1 with strong version definition.

nbulaj:
Thanks @sanemat . Already fixed (tested against 5.1 and Rails master).

PR #985, opened 2017-08-24
fix migration file for rails 5

jonhue:
The migration file generated with `rails generate doorkeeper:migration` did not have the appropriate `[5.x]` when I was installing it.

**Rails version:** 5.1.3

So here is a fix.

### DIFF
--- a/lib/generators/doorkeeper/application_owner_generator.rb
+++ b/lib/generators/doorkeeper/application_owner_generator.rb
@@ -18,7 +18,7 @@ class Doorkeeper::ApplicationOwnerGenerator < Rails::Generators::Base
   end
   
   def migration_version
-    if Rails.version >= 5
+    if Rails.version >= "5.0.0"
       "[#{Rails::VERSION::MAJOR}.#{Rails::VERSION::MINOR}]"
     end
   end


nbulaj:
Take a look at the #981 - already in master.

jonhue:
@nbulaj 

It has not been done for the other migration like adding applications & stuff.

jcamejo:
Hi
i'm having some troubles with this fix, the migration is being generated without  the migration version, even with the master fix. If i download the gem into my local environment it creates the migration with the version but not with the remote one. Is there anything that i'm doing wrong?

i've updated the gem to the last version.

Thanks!

jonhue:
@jcamejo 

This branch works fine. Master does not. 
For now, I suggest you just add it manually.

nbulaj:
Hi @maclover7 . I think we can merge it with the master branch.

@jonhue one thing that I would change: Rails version check. Why not to compare integers, not strings? I mean why not `Rails::VERSION::MAJOR >= 5` instead of `Rails.version >= "5.0.0"` ? It works, yes, but looks a little bit hacky

jonhue:
@nbulaj Here you go ...

I want to point out though, that the other generators in this repository are sing a string comparison:
https://github.com/doorkeeper-gem/doorkeeper/blob/master/lib/generators/doorkeeper/migration_generator.rb#L21

nbulaj:
@jonhue take a look at the TravisCI log - [there are some errors](https://travis-ci.org/doorkeeper-gem/doorkeeper/jobs/289052202#L693) :)

jonhue:
@nbulaj Seems like poor little Travis didn't like that change too much :)

nbulaj:
Hi @jonhue . Some comments:

**1.** Could you please check Rails version as:

`if Rails::VERSION::MAJOR >= 5`

 As I said before, I don't like the way of string comparison. Let's check integers :)

**2.** Could you add some specs for this feature (check Rails version in the specs, and if it more than 5, then check if migration template has a version specification) ? You can take a look at some [already existing examples](https://github.com/doorkeeper-gem/doorkeeper/tree/master/spec/generators)

After that please rebase your commits and squash them into one, and add an entry to [NEWS.md](NEWS.md). Thanks for your work and patience!

nbulaj:
Thanks @jonhue for your work! I already fixed this issue with some corrections and specs to prepare a new release of the gem.

PR #992, opened 2017-09-27
Rails API support

Borzik:
This was influenced by #893, but I took a bit different approach and extended functionality to other areas. It is still WIP, but I just wanted to share what is done already, and I'll be glad to get some advices.

The main premise for this PR is that usually Rails APIs and frontend apps are running on different domains, so we cannot easily identity users by cookies or something like that (e.g. in our project we use [devise_token_auth](https://github.com/lynndylanhurley/devise_token_auth) and authenticate users by tokens).

Switching on `api_mode` configuration will:
- disable request forgery protection, since we assume user is not authenticated by cookies - if he is, there's no much sense in turning `api_mode` on
- remove helpers since we don't rely on HTML views
- render JSON with `status: redirect` and `redirect_uri` instead of directly redirecting user
- disable applications and authorized_applications controller (maybe I will update it to support JSON as well and enable them back - not sure if it's needed though)

As a side effects, this PR:
- makes `Doorkeeper::Oauth::PreAuthorization` serializable by as_json method which will provide all information needed to show preauthorization page
- *(I need more information on this one from someone more familiar with project)* sets redirect_uri on `Doorkeeper::Oauth::ErrorResponse` - for some reason, it was not set before. If we render ErrorResponse with api_mode, it will render the same `status: redirect` and `redirect_uri` with error message

Anyone who will use `api_mode`, will also need some oauth page (which can be easily implemented as a small SPA and that will be running on the same domain as main frontend application). They can make main frontend app do this job, but I assume it is not worth loading the whole app just to show a single dialogue message with Allow/Deny buttons. It will have to be responsible for:
- logging users in
- making pre-auth requests which will return client name, and other information
- sending POST `oauth/authorize`
- redirecting user whenever they get `status: redirect` to `redirect_uri` provided in response

Please let me know your thoughts about it.

### DIFF
--- a/app/controllers/doorkeeper/application_controller.rb
+++ b/app/controllers/doorkeeper/application_controller.rb
@@ -4,8 +4,9 @@ module Doorkeeper
 
     include Helpers::Controller
 
-    protect_from_forgery with: :exception
-
-    helper 'doorkeeper/dashboard'
+    unless Doorkeeper.configuration.api_mode
+      protect_from_forgery with: :exception
+      helper 'doorkeeper/dashboard'
+    end
   end
 end
--- a/app/controllers/doorkeeper/authorizations_controller.rb
+++ b/app/controllers/doorkeeper/authorizations_controller.rb
@@ -4,14 +4,9 @@ module Doorkeeper
 
     def new
       if pre_auth.authorizable?
-        if skip_authorization? || matching_token?
-          auth = authorization.authorize
-          redirect_to auth.redirect_uri
-        else
-          render :new
-        end
+        render_success
       else
-        render :error
+        render_error
       end
     end
 
@@ -26,6 +21,26 @@ module Doorkeeper
 
     private
 
+    def render_success
+      if skip_authorization? || matching_token?
+        auth = authorization.authorize
+        redirect_or_render auth
+      elsif Doorkeeper.configuration.api_mode
+        render json: @pre_auth
+      else
+        render :new
+      end
+    end
+
+    def render_error
+      if Doorkeeper.configuration.api_mode
+        render json: @pre_auth.error_response.body[:error_description],
+               status: :bad_request
+      else
+        render :error
+      end
+    end
+
     def matching_token?
       AccessToken.matching_token_for pre_auth.client,
                                      current_resource_owner.id,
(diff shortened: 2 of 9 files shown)


JoeWoodward:
No movement on this PR?

nbulaj:
Hi @Borzik . I like this approach more than original PR. Does it still WIP , maybe you need some help? I will look at your questions (aka side-effects) ASAP.

Borzik:
Hi! When I submitted this PR I was working on a feature which was postponed, so I'm not working actively on this at the moment. Since you think this PR is worth attention, I'll look into it again next week, will rebase it on current master, and will try to get it out of WIP status (as I remember, it worked well for me on my app, but it still needs refactoring).

nbulaj:
Great! Let me know it you will need some help.

JoeWoodward:
Just found another issue with this configuration. I have an app that is api only but also one that is api and web. The api only works fine, however, the hybrid I want to remove protect_from forgery for authorization & token & revoke endpoints but would need to still have the routes defined for applications

Borzik:
@JoeWoodward 
I think such hybrid app support is out of this PR's scope. This update is intended to support Rails applications generated with `--api` argument only, and to do it in the simplest possible way (by toggling one configuration line).

@nbulaj 
I've done some cleanup and fixed most Hound comments.
I don't really like what's going on in `app/controllers/doorkeeper/authorizations_controller.rb`, specifically this [structure](https://github.com/doorkeeper-gem/doorkeeper/pull/992/files#diff-1282d211312f992c2a614d1f4a6302feR9):
```ruby
if success
  if api_mode?
  else
  end
else
  if api_mode?
  else
  end
end
```

And `redirectable?` doesn't seem to convey correct meaning [here](https://github.com/doorkeeper-gem/doorkeeper/pull/992/files#diff-1282d211312f992c2a614d1f4a6302feR43):
```ruby
if redirectable? and api_mode?
  render ...
elsif redirectable?
  redirect
else
  render
end
```

If you have any ideas how this can be improved, please let me know.

nbulaj:
Hi @Borzik. Can you please resolve the conflicts?

What about some code issues, you can extract successful and error handling to separate private methods, like

```ruby
def new
  if pre_auth.authorizable?
    render_successful
   else
     render_error
  end
end

private 

# ...

def render_successful
  if skip_authorization? || matching_token?
    auth = authorization.authorize
    redirect_or_render auth
  elsif Doorkeeper.configuration.api_mode
    render json: @pre_auth
  else
    render :new
  end
end

def render_error
  if Doorkeeper.configuration.api_mode
    render json: @pre_auth.error_response.body[:error_description],
           status: :bad_request
  else
    render :error
  end
end
```

nbulaj:
Hi @Borzik . Could you please rebase with the master to resolve the conflicts?

pacop:
Interested in seeing this in the new version. Hope that @Borzik can rebase these files

Borzik:
@nbulaj @pacop Done

nbulaj:
Thanks @Borzik ! It would be really great if somebody could test it in his app, maybe @pacop ? Try to use `master` branch to be sure we didn't break something

PR #1003, opened 2017-11-30
Make native redirect code a query param

dv:
Using native redirect, instead of redirecting to

```oauth/authorize/abc123abc123```

it redirects to

```oauth/authorize/native?code=abc123abc123```

This allows for automated flows to detect that an Authorization code was granted in much the same way as a normal redirect. This is used by e.g. Mac Paw (and possibly Postman) to detect whether a code was given or not. I'm not sure if this was intentional or not, but adding the code as an actual param would make sense to mimic the non-oob flow as much as possible, gaining support by any software that works with non-oob flow.

Before:

After:

### Related
- https://stackoverflow.com/questions/46834280/oauth-2-0-authentication-flow-with-paw-and-funky-redirect-uri/47563884#47563884

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,7 @@ User-visible changes worth mentioning.
 
 ## master
 
+- [#1003] Use URL query param to pass through native redirect auth code so automated apps can find it.
 - [#868] `Scopes#&` and `Scopes#+` now take an array or any other enumerable
   object.
 - [#1019] Remove translation not in use: `invalid_resource_owner`.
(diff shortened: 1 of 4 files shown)


nbulaj:
Hi @dv. Can you please add an entry on the top of the [NEWS.md](https://github.com/doorkeeper-gem/doorkeeper/blob/master/NEWS.md) and rebase with squashing? Thanks!

nbulaj:
Hi @dv. Looks good to me. One moment: can you improve specs to check if valid code present in the `code` parameter?

Also, could you please add an entry on the top of the [NEWS.md](https://github.com/doorkeeper-gem/doorkeeper/blob/master/NEWS.md) and rebase with squashing? Thanks!

dv:
Hey @nbulaj, thanks! Did as requested, is this ok?

nbulaj:
Great, thanks @dv!

PR #1004, opened 2017-12-03
Clarify how to add local redirect URIs

jfly:
One of our users was confused by the wording of this message and suggested this alternative wording, which I agree is superior.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,7 @@ User-visible changes worth mentioning.
 - [#948] Make Scopes.<=> work with any "other" value.
 - [#970] Escape certain attributes in authorization forms.
 - [#974] Redirect URI is checked without query params within AuthorizationCodeRequest.
+- [#1004] More explicit help text for `native_redirect_uri`.
 - [#1023] Update Ruby versions and test against 2.5.0 on Travis CI.
 - [#1024] Migrate from FactoryGirl to FactoryBot.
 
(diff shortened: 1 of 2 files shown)


nbulaj:
Hi @jfly . What confusing you find in phrase "local tests"? I think it is clearly explain the purpose of native redirect uris. What do you rhink?

jfly:
CC-ing @luis-j-ianez, who originally told me he found the message confusing. He can probably give more insight into this than I can.

If I were going to guess, I think the word "tests" is confusing because it makes a programmer think of test scripts, not development on localhost.

Luis-J-Ianez:
Well, it's difficult to reproduce what my mind thought first when I read "Use blablabla for local tests". As now I understand the process, I can't reproduce the mistake, so to speak. But I remember that at first I thought the "blablabla" was a kind of alias for "localhost". In my view the new wording is more helpful.

nbulaj:
So @jfly @Luis-J-Ianez maybe we can extend this message by adding something like "for development purposes" in order to improve understanding of the help message? What do you think?

Also can you please add a changelog entry to NEWS.md with descriptive comment of the changes and rebase to quash commits? Thanks for yo patience and work!

jfly:
Done and done!

nbulaj:
Great, thanks @jfly!

jonatanklosko:
Hey, this message seems no longer accurate. As far as I can see, this behaviour relayed on a bug fixed in #976, namely: if you added `urn:ietf:wg:oauth:2.0:oob` in the first line of `redirect_uri`, followed by other URLs (e.g. localhost ones) they were not validated and thus worked like a charm. #976 makes all URLs validated, so adding `urn:ietf:wg:oauth:2.0:oob` has no effect on the other URLs.

PR #1010, opened 2018-01-03
Configuration to enforce application scopes

### LINKED ISSUE #1009: Enforce scopes when creating/updating applications
Followup to #896 

Doorkeeper should enforce the scopes when creating/editing registered applications via `/oauth/applications/new` and `/oauth/applications/123/edit`.

Currently it is possible to enter nonexistent scopes (like "blahblah") when creating/editing an application, and those scopes will be used in the authorization step. When authorizing to the application, the `/oauth/authorize` page shows the untranslated scope names, and authorization succeeds. No error messages are displayed, other than the tooltip saying the scope is not translated.

In other words, there is not much purpose to the `optional_scopes` configuration, since those can be easily ignored/overridden when registering a new application. It seems to me that the default behavior should be to enforce each application's scopes are in either `default_scopes` or `optional_scopes`, during the application's create/update step.

talklittle:
Fixes #1009

### DIFF
--- a/lib/doorkeeper/orm/active_record/application.rb
+++ b/lib/doorkeeper/orm/active_record/application.rb
@@ -13,6 +13,8 @@ module Doorkeeper
     validates :redirect_uri, redirect_uri: true
     validates :confidential, inclusion: { in: [true, false] }
 
+    validate :validate_configured_scopes, if: :enforce_scopes?
+
     before_validation :generate_uid, :generate_secret, on: :create
 
     has_many :authorized_tokens, -> { where(revoked_at: nil) }, class_name: 'AccessToken'
@@ -40,6 +42,17 @@ module Doorkeeper
 
     def generate_secret
       self.secret = UniqueToken.generate if secret.blank?
+    end
+
+    def validate_configured_scopes
+      if !scopes.blank? &&
+         !ScopeChecker.valid?(scopes.to_s, Doorkeeper.configuration.scopes)
+        errors.add(:scopes, :invalid_scope)
+      end
+    end
+
+    def enforce_scopes?
+      Doorkeeper.configuration.enforce_configured_scopes?
     end
   end
 end
--- a/spec/requests/applications/applications_request_spec.rb
+++ b/spec/requests/applications/applications_request_spec.rb
@@ -20,6 +20,73 @@ feature 'Adding applications' do
       click_button 'Submit'
       i_should_see 'Whoops! Check your form for possible errors'
     end
+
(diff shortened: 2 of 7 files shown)


lib/doorkeeper/orm/active_record/application.rb
```diff
@@ -12,6 +12,14 @@ class Application < ActiveRecord::Base
     validates :uid, uniqueness: true
     validates :redirect_uri, redirect_uri: true
 
+    validates_each :scopes, if: :enforce_scopes? do |record, attr, value|
```
nbulaj: Let's move it it from the `validates_each` to `validate` with custom method or just `validate` with a block (because `validates_each` is responsible for processing bunch of attributes, but we process only one - `scopes`).

lib/doorkeeper/orm/active_record/application.rb
```diff
@@ -12,6 +12,14 @@ class Application < ActiveRecord::Base
...
+      valstr = value.to_s
```
nbulaj: No need to do `to_s` because ActiveModel / ActiveRecord already done type-casting for us.

lib/doorkeeper/orm/active_record/application.rb
```diff
@@ -12,6 +12,14 @@ class Application < ActiveRecord::Base
...
+      unless valstr.blank? ||
```
nbulaj: Lets rewrite it to `if` - `unless` with more than one condition is hard to read and maintain.

spec/requests/applications/applications_request_spec.rb
```diff
@@ -20,6 +20,44 @@
...
+      fill_in "doorkeeper_application[redirect_uri]",
+              with: "https://example.com"
+      fill_in "doorkeeper_application[scopes]", with: "blahblah"
+
+      click_button "Submit"
+      i_should_see "Whoops! Check your form for possible errors"
+    end
+
+    scenario "adding app validating scope, blank scope is accepted" do
+      config_is_set("enforce_configured_scopes", true)
```
nbulaj: Let's add a test with multiple scopes to be sure everything is OK

nbulaj:
Hi @talklittle . Could you please take an attention to my comments below? Also could you please add an entry to NEWS.md file and squash all the commits to a single one? Thanks!

talklittle:
@nbulaj Thanks for review. Applied your feedback.

nbulaj:
Thank you @talklittle :+1:

PR #1016, opened 2018-01-14
create configuration option to allow http callbacks to localhost uris

GrantJamesPowell:
Hey! 

We love Doorkeeper at the company I work at now.

We have an interesting use case that other people might have.

We would like to use the `force_ssl`configuration option, but we'd also like to allow http access to localhost callback URIs.

This PR creates the option `allow_non_secure_localhost_redirects` which allows doorkeeper users to allow http localhost uris while still forcing all other uris to be ssl.

### DIFF
--- a/app/validators/redirect_uri_validator.rb
+++ b/app/validators/redirect_uri_validator.rb
@@ -29,6 +29,12 @@ class RedirectUriValidator < ActiveModel::EachValidator
 
   def invalid_ssl_uri?(uri)
     forces_ssl = Doorkeeper.configuration.force_ssl_in_redirect_uri
-    forces_ssl && uri.try(:scheme) == 'http'
+    allow_non_ssl_localhost = Doorkeeper.configuration.allow_non_secure_localhost_redirects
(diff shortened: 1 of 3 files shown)


nbulaj:
Hi @GrantJamesPowell . Why not just :

```ruby
#config/initializers/doorkeeper.rb
Doorkeeper.configure do
  # ...

  force_ssl if Rails.env.production?
end
```

?

GrantJamesPowell:
@nbulaj . Hey! Yeah we want to be able to force SSL in production for all urls except localhost ones. I think its pretty common as a need too, I know facebook oauth allows non ssl redirects only for localhost.

nbulaj:
Hi @GrantJamesPowell. If you will use master branch of the Doorkeeper, then you can set `#force_ssl_in_redirect_uri` config option as a callable object (like block, proc, or even class with method `#call`):

```ruby
force_ssl_in_redirect_uri { |uri| uri.host != 'localhost' }
```

Then you will be able to handle localhost URIs (or non SSL) in any Rails env

PR #1018, opened 2018-01-20
Removing unnecessary <th>

soundasleep:
There are only three columns in the table - we don't need four headers


nbulaj:
Thanks @soundasleep

PR #1019, opened 2018-01-22
Remove translation not in use: invalid_resource_owner 

baxang:
The key was removed from use in #843 (discussed in #771 and formerly #444) and I think it is better removed to prevent confusion.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,7 @@ User-visible changes worth mentioning.
 
 ## master
 
+- [#1019] Remove translation not in use: `invalid_resource_owner`.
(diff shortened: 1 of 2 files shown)


nbulaj:
Hi @baxang . Can you please add a note to NEWS.md (see CONTRIBUTING section in order how to format it) and squash commits to one? Thanks

baxang:
Hi @nbulaj, I've updated NEWS.md and squashed commits too.

nbulaj:
Got it. Thanks @baxang!

baxang:
I thank you @nbulaj. Looking forward to a new release!

PR #1022, opened 2018-01-27
Fix testing Rails 5.1 version

tricknotes:
The version specification '~> 5.1' means '5.x'.
When Rails 5.2 is released, rails version will be updated to 5.2.0.


nbulaj:
Whoops, 0 lost :) Thanks!

PR #1024, opened 2018-01-27
Migrate from FactoryGirl to FactoryBot

tricknotes:
FactoryGirl has been renamed as FactoryBot.
Ref: https://robots.thoughtbot.com/factory_bot

This commit fixes the following warning message:
```
DEPRECATION WARNING: The factory_girl gem is deprecated. Please upgrade to factory_bot.
See https://github.com/thoughtbot/factory_bot/blob/v4.9.0/UPGRADE_FROM_FACTORY_GIRL.md for further instructions.
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,7 @@ User-visible changes worth mentioning.
 - [#970] Escape certain attributes in authorization forms.
 - [#974] Redirect URI is checked without query params within AuthorizationCodeRequest.
 - [#1023] Update Ruby versions and test against 2.5.0 on Travis CI.
+- [#1024] Migrate from FactoryGirl to FactoryBot.
 
 ## 4.2.5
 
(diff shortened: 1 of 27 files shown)


nbulaj:
Hi @tricknotes . Can we set more specific version of FactoryBot in dependencies like it is done for other deps ? Something can change in new major release, we don't want to brake our specs :) Also don't sure if it doesn't break doorkeeper ORM extensions (such as mongoid or sequel), because some of them uses original test cases with their own specs (that can use FactoryGirl). Need some exploration here, I will check it

tricknotes:
Thanks for your review, @nbulaj !
I agree with version specification.
So I updated my commit as https://github.com/doorkeeper-gem/doorkeeper/pull/1024/commits/b9290a92b894e8047aeb7a74dcba2dba5b4254a5.

(note: The current latest version of factory_bot is 4.8.2, not 4.9.x. https://rubygems.org/gems/factory_bot )

nbulaj:
Perfect, thanks you @tricknotes !

Issue #1007, opened 2017-12-19
Improve documentation around foreign keys


fschwahn:
In the readme there's the suggestion to add a foreign key on all tables with `resource_owner_id`-columns, however there's no note that this will make it impossible to destroy that record without adding an association with a dependent option (because a `ActiveRecord::InvalidForeignKey`-exception will be raised). This is easy to forget if it is mentioned nowhere. I think the README should be expanded to include that information, something along the lines:

```
Remember to add associations to your model so the related records are deleted:

has_many :oauth_access_grants, class_name: "Doorkeeper::AccessGrant", foreign_key: :resource_owner_id, dependent: :destroy
has_many :oauth_access_tokens, class_name: "Doorkeeper::AccessToken", foreign_key: :resource_owner_id, dependent: :destroy
```

nbulaj:
Hi @fschwahn. Would you like to submit a PR ?

nbulaj:
Impoved via #1025

PR #1025, opened 2018-01-29
Improve documentation for adding foreign keys

fschwahn:
Add explanation about necessity to define relations when adding foreign keys as mentioned in #1007

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,17 @@ for each table that includes a `resource_owner_id` column:
 add_foreign_key :table_name, :users, column: :resource_owner_id
 ```
 
+Remember to add associations to your model so the related records are deleted.
+If you don't do this a `ActiveRecord::InvalidForeignKey`-error will be raised
+when you try to delete a model with related access grants or access tokens.
+
+```ruby
+class User < ApplicationRecord
+  has_many :oauth_access_grants, class_name: "Doorkeeper::AccessGrant", foreign_key: :resource_owner_id, dependent: :destroy
+  has_many :oauth_access_tokens, class_name: "Doorkeeper::AccessToken", foreign_key: :resource_owner_id, dependent: :destroy
+end
+```
+
 Then run migrations:
 
 ```sh
--- a/README.md
+++ b/README.md
@@ -87,13 +87,13 @@ add_foreign_key :table_name, :users, column: :resource_owner_id
 ```
 
 Remember to add associations to your model so the related records are deleted.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -86,6 +86,17 @@ for each table that includes a `resource_owner_id` column:
 add_foreign_key :table_name, :users, column: :resource_owner_id
 ```
 
+Remember to add associations to your model so the related records are deleted.
+If you don't do this a `ActiveRecord::InvalidForeignKey`-error will be raised
```
nbulaj: Let's write "_a**n** `ActiveRecord::InvalidForeignKey`_", ok?

README.md
```diff
@@ -86,6 +86,17 @@ for each table that includes a `resource_owner_id` column:
...
+when you try to delete a model with related access grants or access tokens.
```
nbulaj: `delete` method will ignore callbacks for associations, so let's write "_try to **`destroy`** the model_" instead to avoid misunderstanding.

README.md
```diff
@@ -86,6 +86,17 @@ for each table that includes a `resource_owner_id` column:
...
+
+```ruby
+class User < ApplicationRecord
+  has_many :oauth_access_grants, class_name: "Doorkeeper::AccessGrant", foreign_key: :resource_owner_id, dependent: :destroy
```
nbulaj: Does it really required to name associations like `oauth_access_grants` ? `Doorkeeper::AccessGrant` already has table name prefixes, so Rails must recognize the name of the table for associations.

Maybe it will be better to name them like:

```ruby
has_many :access_grants, ...
has_many :access_tokens, ...
```
Also I suggest to change `dependent: :destroy` option to `dependent: :delete_all` - it will delete all the associated records in one SQL query instead of N+1. But we can save :destroy variant in comment, like " _# or :destroy if you need callbacks_"

What do you think?

nbulaj:
Hi @fschwahn . Can you please take a look at my comments above? Also please add a small changelog entry to NEWS.md and rebase commits to squash them into one. Thanks for your work!

nbulaj:
Hi @fschwahn . Any updates here?

fschwahn:
Had a busy week, thanks for reminding me. I incorporated all your suggestions, all of them make sense. I did all the editing using Github, I don't have the repository checked out, so I can't easily squash. But you can Squash & Merge using Github UI.

nbulaj:
Thank you @fschwahn, good work! 👍

PR #1028, opened 2018-02-07
Make it possible to have composit strategy names.

nashby:
This fix is needed for https://github.com/doorkeeper-gem/doorkeeper-openid_connect since
it's possible to have "response_type=id_token token" that we need to translate to
`IdTokenToken` strategy class.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,7 +7,7 @@ User-visible changes worth mentioning.
 - [#913] Deferred ORM (ActiveRecord) models loading
 - [#943] Fix Access Token token generation when certain errors occur in custom token generators
 - [#1026] Implement RFC7662 - OAuth 2.0 Token Introspection
-- [#985] Generate valid migration files for Rails >= 5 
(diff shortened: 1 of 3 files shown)


nbulaj:
Hi @nashby . Is it OK to have such `response_type` ? Does it corresponds the standarts?

nashby:
@nbulaj hey! Yep. You can find more info here http://openid.net/specs/openid-connect-core-1_0.html#id_token-tokenExample

nbulaj:
@nashby thats great and make sense, thank you. Maybe we can extract `strategy` param normalization to the separated private method? We can extend this part in the future due to some similar problems or in order to have more complex way of strategy class initialization. 

What do you think @nashby  ?

Also could you please add an entry to NEWS.md file with short description for this PR? Thank you!

nashby:
@nbulaj makes sense! Done.

nbulaj:
Thank you @nashby !

PR #1029, opened 2018-02-08
Sort applications by created_at

travisofthenorth:
### Summary

The impetus for this change is that the index page for applications doesn't apply any default ordering which can be a little disorienting.

This PR introduces a base AR model with an `ordered_by` method and removes the `order_method` + `public_send` implementation this is currently being used. It also adds a deprecation warning for this implementation in other ORMs. The idea is that a single `ordered_by` method that all ORMs implement will be a lot cleaner and more maintainable going forward.

Finally, this PR sorts applications by `created_at` in the index view.

### DIFF
--- a/lib/doorkeeper/models/access_token_mixin.rb
+++ b/lib/doorkeeper/models/access_token_mixin.rb
@@ -168,7 +168,7 @@ module Doorkeeper
       #   nil if nothing was found
       #
       def last_authorized_token_for(application_id, resource_owner_id)
-        public_send(order_method, created_at_desc).
+        ordered_by(:created_at, :desc).
           find_by(application_id: application_id,
                   resource_owner_id: resource_owner_id,
                   revoked_at: nil)
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@ User-visible changes worth mentioning.
 
 ## master
 
+- [#1029] Deprecate `order_method` and introduce `ordered_by`. Sort applications
+  by `created_at` in index action.
 - [#1033] Allow Doorkeeper configuration option #force_ssl_in_redirect_uri to be a callable object.
 - Fix Grape integration & add specs for it
 - [#913] Deferred ORM (ActiveRecord) models loading
--- a/app/controllers/doorkeeper/applications_controller.rb
+++ b/app/controllers/doorkeeper/applications_controller.rb
@@ -10,7 +10,10 @@ module Doorkeeper
       @applications = if applications.respond_to?(:ordered_by)
                         applications.ordered_by(:created_at)
                       else
-                        ActiveSupport::Deprecation.warn "#{Doorkeeper.configuration.orm} must implement #ordered_by method \that will be used by default in Doorkeeper 5."
+                        message = "#{Doorkeeper.configuration.orm} must \
+                          implement #ordered_by method that will be used by \
+                          default in Doorkeeper 5."
+                        ActiveSupport::Deprecation.warn(message)
                         applications
                       end
     end
--- a/app/controllers/doorkeeper/applications_controller.rb
+++ b/app/controllers/doorkeeper/applications_controller.rb
@@ -6,7 +6,13 @@ module Doorkeeper
     before_action :set_application, only: [:show, :edit, :update, :destroy]
 
     def index
-      @applications = Application.all
+      applications = Application.all
+      @applications = if applications.respond_to?(:ordered_by)
+                        applications.ordered_by(:created_at)
+                      else
+                        ActiveSupport::Deprecation.warn "#{Doorkeeper.configuration.orm} must implement #ordered_by method \that will be used by default in Doorkeeper 5."
+                        applications
+                      end
     end
 
     def show; end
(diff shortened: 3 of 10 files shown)


nbulaj:
Hi @travisofthenorth . Point about ordering is good for me. And doing it by `created_at` - this is OK.

But I'm worried about doorkeeper ORMs compatibility - MongoMapper and Mongoid have some specific API for this action, so users of this gems can be affected and Application page would be broken. I need to think how to bypass it...

What about gemspec changes - I prefer not to tie to some specific gems. Especially when we don't know what specific developer prefers to use (irb, pry, etc / rspec, minitest, etc). What about `assigns` - we can just check the names of IDs of the application in the table with some CSS/XPath selectors to be sure if they are placed in the right order. So I would recommend not to add this dependencies to the gem.

travisofthenorth:
@nbulaj feedback incorporated

travisofthenorth:
@nbulaj re: ORM compatibility, would that be as simple as updating these mixins or is there more to it? https://github.com/doorkeeper-gem/doorkeeper-mongodb/blob/master/lib/doorkeeper-mongodb/mixins/mongoid/application_mixin.rb#L30

travisofthenorth:
Should I just do this for now?
```ruby
applications = Application.all
@applications = applications.respond_to?(:order) ? applications.order(:created_at) : applications
```

I'm happy to do this, do the class method, try to get a PR into the adapter gem, a combination of all of them, etc. Just let me know.

travisofthenorth:
I added a class method and I'm testing for it being implemented in the controller.

nbulaj:
I like it. Let's make it more robust:

```ruby
def self.ordered_by(attribute, direction = :desc)
  order(attribute => direction)
end
```

It will allow to order records by any attribute both in DESC or ASC directions (in future if it will be required). And then use it in controller like this:

```ruby
# ...
if applications.respond_to?(:ordered_by)
  applications.ordered_by(:created_at)
else
  ActiveSupport::Deprecation.warn "#{Doorkeeper.configuration.orm} must implement #ordered_by method that ill be used by default in Doorkeeper 5."
  applications
end
```

We need to deprecate this situation in order to to allow every doorkeeper extension to implement this API for ordering.

travisofthenorth:
@nbulaj nice! I thought of doing that but I didn't want to totally diverge from that `order_method` approach you linked. I think this approach is a lot nicer.

nbulaj:
@travisofthenorth we can do the same as it is done for AccessToken mixin (`order_method, created_at_desc`), but I found it too limited and want to use above solution for all the Doorkeeper models in order to add flexibility and portability.

travisofthenorth:
@nbulaj updated. I added a base class for AR with `ordered_by` and I actually removed the AR `order_method` implementation. I put a deprecation warning in one of the mixins so the message would only print once.

nbulaj:
That is exactly what I was thinking about! The only point is that Doorkeeper orm extensions don't use AccessToken mixin (because it AR-tied), so it wouldn't be loaded ever (only for some old mongodb extension). So developers wouldn't see deprecation message and wouldn't update to latest extensuons versions.

Could you please also add a changelog entry to NEWS.md?

travisofthenorth:
@nbulaj cool. I added the deprecation warning to the applications controller as well. Is there a better place for it? I was going to put it into `setup_orm_adapter` in `config.rb` but I can't actually check the models since they're lazily-loaded (at least in AR's case).

Also added the new entry in news.md

lib/doorkeeper/models/access_token_mixin.rb
```diff
@@ -168,10 +174,15 @@ def find_or_create_for(application, resource_owner_id, scopes, expires_in, use_r
...
       #
       def last_authorized_token_for(application_id, resource_owner_id)
-        public_send(order_method, created_at_desc).
-          find_by(application_id: application_id,
-                  resource_owner_id: resource_owner_id,
-                  revoked_at: nil)
+        ordered = if respond_to?(:ordered_by)
+                    ordered_by(:created_at, :desc)
+                  else
+                    public_send(order_method, created_at_desc)
```
nbulaj: We don't need this check. First of all, because this a AR mixin and method already defined. Also because `order_method` was removed. Just use new API.

lib/doorkeeper/models/access_token_mixin.rb
```diff
@@ -10,6 +10,12 @@ module AccessTokenMixin
     include ActiveModel::MassAssignmentSecurity if defined?(::ProtectedAttributes)
 
     included do
+      unless respond_to?(:ordered_by)
```
nbulaj: Also can be removed, because `BaseRecord` implements this method, so it is a dead code. Let's ApplicationController do his work with deprecation :)

lib/doorkeeper/models/access_token_mixin.rb
travisofthenorth: Ah yeah. I was thinking that the mixin was used by more than just AR.

travisofthenorth: Updated

nbulaj: Yep, for old versions of mongodb extension. But it is a wrong way due to some difference in MongoMapper and Mongoid API. New version of the extension uses it's own, the same as Sequel.

nbulaj:
Thank you @travisofthenorth for your work and patience! 👍

PR #1030, opened 2018-02-10
Mavenlink with upstream 2018 02

diclophis:
### Summary

Hello,

This PR makes these changes:

1. Restrict `redirect_uri` to only http/https
2. Add `mongoid` adapter support for string literals


nbulaj:
Hi @diclophis 

> Restrict redirect_uri to only http/https

is the default behavior of the Doorkeeper gem unless it is configured in another way. [You implemented already existing validation](https://github.com/doorkeeper-gem/doorkeeper/blob/c846335ad75f2f9c7108e577cb84eaf8ab66b86f/app/validators/redirect_uri_validator.rb#L17)

> Add mongoid adapter support for string literals

Mongoid and MongoMapper I18n already presented in the [doorkeeper-mongodb](https://github.com/doorkeeper-gem/doorkeeper-mongodb) extension, that is published on Rubygems or can be used from the git.

> (_from the PR changes_): has_query_parameter: 'cannot contain a query parameter.'

No problem with the query params. It was fixed in 82e7f70519ceab688000953a5d7c908abf59f68c  and [corresponds with the standard](https://tools.ietf.org/html/rfc6749#section-3.1.2).

Issue #999, opened 2017-11-09
Secret token for public client using authorization code grant


lefty313:
as stated in https://tools.ietf.org/html/draft-ietf-oauth-native-apps-12#section-8.5

```
Secrets that are statically included as part of an app distributed to
   multiple users should not be treated as confidential secrets, as one
   user may inspect their copy and learn the shared secret.  For this
   reason, and those stated in Section 5.3.1 of [RFC6819], it is NOT
   RECOMMENDED for authorization servers to require client
   authentication of public native apps clients using a shared secret,
   as this serves little value beyond client identification which is
   already provided by the "client_id" request parameter.
```

is there a plan to allow `authorization code grant` without `client_secret` for `public clients` ?

Right now generating `access_token` without `secret_code` returns `401`

```ruby
params = {
  client_id:    "d47f9dd61dbc4a8393188414d2dffe141f571ad6fc41b4824aff1ebdae7e8140",
  code:         "1347d04acbe35d67e07580a02edd3e36cc9274336374b1667a8aa297a2de3c14",
  grant_type:   "authorization_code",
  redirect_uri: "urn:ietf:wg:oauth:2.0:oob",
}

RestClient.post 'http://localhost:3000/oauth/token', params
#=> RestClient::Unauthorized: 401 Unauthorized

params[:client_secret] = "50c1e44cf4be4de9ad960a82079b7254acc1bf8f2a0f476045f044127518dcdd"

RestClient.post("http://localhost:3000/oauth/token", params).body
#=>
{
 "access_token"=>"4079f12c7de20c4e3c9f4a7cc5b6552f15a82136244cea6c5fcbc3973184b7b7",
 "token_type"=>"bearer",
 "expires_in"=>7200,
 "refresh_token"=>
 "8b4fd7092cf1a4979f1085c307e1ce07ae187a350c0a683889481383bf075cf9",
 "created_at"=>1510242157
}
```

lefty313:
I could work on it in my free time but first I need to know maintainer opinion ;)

kapilmittal:
You should use Implicit grant flow for public clients. The implicit grant type is used for mobile apps and web applications, where the client secret confidentiality is not guaranteed.

Authorization code flow is meant for non public clients where client_secret confidentiality can be maintained.

lefty313:
[...]

PR #1031, opened 2018-02-10
Allow public clients to authenticate without client_secret

f3ndot:
### Summary

This PR aims to solve #999 and comply with [Section 8.5 of  draft-ietf-oauth-native-apps-12](https://tools.ietf.org/html/draft-ietf-oauth-native-apps-12#section-8.5):

```
   Secrets that are statically included as part of an app distributed to
   multiple users should not be treated as confidential secrets, as one
   user may inspect their copy and learn the shared secret.  For this
   reason, and those stated in Section 5.3.1 of [RFC6819], it is NOT
   RECOMMENDED for authorization servers to require client
   authentication of public native apps clients using a shared secret,
   as this serves little value beyond client identification which is
   already provided by the "client_id" request parameter.
```

### Public vs Private Clients

The original [OAuth 2 RFC](https://tools.ietf.org/html/rfc6749) makes a definition between public & private clients ([Section 2.1](https://tools.ietf.org/html/rfc6749#section-2.1)). To this end, along with modeling industry best practice, `Doorkeeper::Application` now has a `#confidential?` attribute. The developer must decide when creating an application, whether it is to be a confidential or non-confidential client.

This allows Doorkeeper to behave differently when receiving an authentication request, depending on the identified client and its confidentiality.

This decision maintains security by forcing the presence of a `client_secret` when the developer intends their app only to be used in secure environments.

For backwards compatibility, an `UPGRADING.md` document should be written asking developers to generate a migration that adds the `confidential` column. To be safe by default, the default value/backfill should be a confidential app.

### What's Next

I omitted some important work to keep this PR simple:

1. There's no tracking in the `Doorkeeper::AccessToken` for whether or not it was created under authenticated means (confidential app) or just identified (public app). A future PR should do this so it is clear to the developer that that Access Token's application relationship is guaranteed or just claimed.
2. We don't forbid certain grant types depending on the confidentiality of an app. For example, a public app cannot use Authorization Code (unless PKCE). Research on the RFC specs is required, to determine if/how we may prevent a public client from authenticating using a grant type intended for private apps and vice versa.

### DIFF
--- a/lib/doorkeeper/request/password.rb
+++ b/lib/doorkeeper/request/password.rb
@@ -3,7 +3,7 @@ require 'doorkeeper/request/strategy'
 module Doorkeeper
   module Request
     class Password < Strategy
-      delegate :credentials, :resource_owner, :parameters, to: :server
+      delegate :credentials, :resource_owner, :parameters, :client, to: :server
 
       def request
         @request ||= OAuth::PasswordAccessTokenRequest.new(
@@ -12,16 +12,6 @@ module Doorkeeper
           resource_owner,
           parameters
         )
-      end
-
-      private
-
-      def client
-        if credentials
-          server.client
-        elsif parameters[:client_id]
-          server.client_via_uid
-        end
       end
     end
   end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -2,6 +2,10 @@ AllCops:
   Exclude:
     - "spec/dummy/db/*"
 
+Metrics/BlockLength:
+  Exclude:
+    - spec/**/*
+
 LineLength:
   Exclude:
     - spec/**/*
--- a/app/controllers/doorkeeper/applications_controller.rb
+++ b/app/controllers/doorkeeper/applications_controller.rb
@@ -57,7 +57,8 @@ module Doorkeeper
     end
 
     def application_params
-      params.require(:doorkeeper_application).permit(:name, :redirect_uri, :scopes)
+      params.require(:doorkeeper_application).
+        permit(:name, :redirect_uri, :scopes, :confidential)
     end
   end
 end
--- a/app/views/doorkeeper/applications/_form.html.erb
+++ b/app/views/doorkeeper/applications/_form.html.erb
@@ -27,6 +27,17 @@
     </div>
   <% end %>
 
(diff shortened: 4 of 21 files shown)


f3ndot:
Screenshots of the views:

screen shot 2018-02-10 at 2 05 00 pm

screen shot 2018-02-10 at 2 02 20 pm

screen shot 2018-02-10 at 2 02 05 pm

f3ndot:
Ignoring the coverage decrease because the existing lines that became uncovered are guarded for Rails 5 only and the test envs that never reach it are 4.x

nbulaj:
@f3ndot I will take a look at it a little bit later - currently I do not have much time

f3ndot:
@nbulaj Take all the time you need, no rush.

nbulaj:
Hi @f3ndot . Could you please sync this branch with the master and squash the commits to a single one? I will take a look at this this week.

f3ndot:
👋 @nbulaj I'll take care of this Friday or on the weekend

baxang:
Thanks for your work @f3ndot. I'm looking forward to this becoming available soon. I tried testing this on my application but had an issue with database migration (the new column was not recognised for some reason). I will give it another try.

f3ndot:
@nbulaj all set!

lib/doorkeeper/request/password.rb
```diff
@@ -13,16 +13,6 @@ def request
...
         )
       end
-
-      private
-
-      def client
-        if credentials
-          server.client
-        elsif parameters[:client_id]
-          server.client_via_uid
```
nbulaj: Is it that we removed client_via_uid and introduced only client?

f3ndot: I think so, because this method was originally doing was allowing `client` to be set if:

1. Valid credentials aka good `client_id` and `client_secret`
2. Only good `client_id` set 

Because `credentials` now returns true if good `client_id` and `client_secret` *or* only good `client_id` set this private method is redundant. We can trust the `server.client` to be set when either (1) or (2) is satified.

nbulaj:
Hi @f3ndot . I've added upgrade notes to the [Migration from old versions doc](https://github.com/doorkeeper-gem/doorkeeper/wiki/Migration-from-old-versions) and introduced generator for generating migration for projects with old Doorkeeper version.

PR #1032, opened 2018-02-11
Refactor BaseRequest callbacks into configurable lambdas

f3ndot:
Refactors #1001 into configurable lambdas per @nbulaj's recommendation

Happy to change the name of the config options and description, if you have a better idea.


nbulaj:
Thank ypou @f3ndot for your work!

PR #1036, opened 2018-02-14
add failing spec for javascript urls failing validation

diclophis:
### Summary

Need help figuring out how to fix this particular scenario where we need to invalid redirect uri when it has `javascript:` schema

Thanks

### DIFF
--- a/spec/validators/redirect_uri_validator_spec.rb
+++ b/spec/validators/redirect_uri_validator_spec.rb
@@ -60,6 +60,14 @@ describe RedirectUriValidator do
       expect(subject).to be_valid
     end
 
+    it 'invalidates the uri when the uri does uses javascript:' do
+      subject.redirect_uri = 'javascript://%0aalert%28document.cookie%29//'
+
+      expect(subject).not_to be_valid
+      error = subject.errors[:redirect_uri].first
+      expect(error).to eq('must be an HTTPS/SSL URI.')
+    end
+
     it 'accepts a non secured protocol when disabled' do
       subject.redirect_uri = 'http://example.com/callback'
       allow(Doorkeeper.configuration).to receive(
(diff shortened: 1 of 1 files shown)


diclophis:
@nbulaj I could use some help figuring out how to configure doorkeeper / patch it, such that we could provide support for invalidating uris that are not HTTP/HTTPS

See also this PR: https://github.com/doorkeeper-gem/doorkeeper/pull/1030 Which was our internal orgs refactor... we just need to figure out how to get the ability to allow _only_ `https` URIs...

nbulaj:
Hi @diclophis. Nice point about JS URIs. It requires to add some constant with list of denied schemas like `javascript` in `RedirectUriValidator` and check the URI against such kind of redirect URIs. Don't sure we need to check the URI during authorization flow because such URIs will be denied by the provider, so I think we can skip it. I can implement this functionality, but you feel free to implement it yourself like:

```ruby
class RedirectUriValidator < ActiveModel::EachValidator
  INVALID_SCHEMAS = %w[javascript].freeze

  def validate_each(record, attribute, value)
     # ...
        return false if native_redirect_uri?(uri)
        record.errors.add(attribute, :invalid_uri) if invalid_schema?(uri)
        # ...
      end
    end
   # ...
  end

  private
  # ...

  def invalid_schema?(uri)
    uri.scheme.in?(INVALID_SCHEMAS)
  end
end
```

What about #1030 - I've already commented it.

> how to get the ability to allow only https URIs...

Just set responsible Doorkeeper configuration option for it:

```ruby
Doorkeeper.configure do
  force_ssl_in_redirect_uri true # for any Rails env
end
```

nbulaj:
Ah, I've got it.

> how to get the ability to allow only https URIs...

You mean how to allow only URIs with HTTPS schema , yep?

Currently it is impossible.

nbulaj:
@diclophis I've added an option to doorkeeper config that will allow to forbid some specific redirect URIs in  3756ec2.

It can be used as:

```ruby
Doorkeeper.configure do
  # ...
  forbid_redirect_uri { |uri| uri.scheme == 'javascript' }
end
```

I decided not to block `javascript` scheme in redirect URI because I couldn't find any point about this in OAuth 2 RFC or any other place. So lets developers decide if they want to forbid such URIs.

PR #1037, opened 2018-02-17
PKCE

bpieck:
I tried to do, as Nikita asked to do, but refactored a little. Hope it is now as expected - else: Please tell me, what to change.

### Summary

- merged work of Ross Sharrott into one commit
- fixed issue about code_challenge in redirect_uri
- ensured, doorkeeper is still working, although PCKE migrations are not generated and migrated
- addition to NEWS.md
- addition to README.md

### DIFF
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -103,6 +103,18 @@ doorkeeper.
         @config.instance_variable_set(:@access_token_methods, methods)
       end
 
+      # There is a big Android oAuth library, that does not send a client secret in PKCE authorization_code flow:
+      # https://github.com/openid/AppAuth-Android/blob/7089089ab6a5950da773d68168e2683da523bfd5/library/java/net/openid/
+      # appauth/AuthorizationManagementActivity.java
+      # It is quite easily understandable, why they do not support secrets. Although its defined in RCF7636 where secret
+      # is required, as it is in the standard authorization_code flow it's required. RCF7636 explains itself, that PKCE
+      # is invented, because you never can trust a static client_secret in a mobile app. Still it should be default,
+      # that secret is expected. But whoever wants to support a library as above, can use enable_pkce_without_secret
+      # in doorkeeper initializer to do so.
+      def enable_pkce_without_secret
+        @config.instance_variable_set(:@enable_pkce_without_secret, true)
+      end
+
       # Issue access tokens with refresh token (disabled by default)
       def use_refresh_token
         @config.instance_variable_set(:@refresh_token_enabled, true)
@@ -255,6 +267,10 @@ doorkeeper.
     def refresh_token_enabled?
       @refresh_token_enabled ||= false
       !!@refresh_token_enabled
+    end
+
+    def pkce_without_secret_enabled?
+      @enable_pkce_without_secret ||= false
     end
 
     def enable_application_owner?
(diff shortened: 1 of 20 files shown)


bpieck:
When I checked rubocop - there were 520 offenses. Since those are not all from this PR - I did not dare to start to fix that. I think that would not be reviewable...

nbulaj:
Hi @bpieck . Big thanks for your work! I can review it in detail only tomorrow. Could you please add some spec for new PKCE generator like [it is done here](https://github.com/doorkeeper-gem/doorkeeper/blob/master/spec/generators/application_owner_generator_spec.rb)? Also it would be great to cover `enable_pkce_without_secret` configuration option with tests.

bpieck:
Sorry for answering late. Having a little temperature - so lying most time in bed.
Thanks for pointing direction. I will do so later this day.

bpieck:
Thanks for pointing those out. Found two issues when implemented tests.

lib/doorkeeper/config.rb
```diff
@@ -103,6 +103,18 @@ def access_token_methods(*methods)
...
 
+      # There is a big Android oAuth library, that does not send a client secret in PKCE authorization_code flow:
+      # https://github.com/openid/AppAuth-Android/blob/7089089ab6a5950da773d68168e2683da523bfd5/library/java/net/openid/
+      # appauth/AuthorizationManagementActivity.java
+      # It is quite easily understandable, why they do not support secrets. Although its defined in RCF7636 where secret
+      # is required, as it is in the standard authorization_code flow it's required. RCF7636 explains itself, that PKCE
+      # is invented, because you never can trust a static client_secret in a mobile app. Still it should be default,
+      # that secret is expected. But whoever wants to support a library as above, can use enable_pkce_without_secret
+      # in doorkeeper initializer to do so.
+      def enable_pkce_without_secret()
```
nbulaj: Can you please remove the brackets? `()`

lib/doorkeeper/oauth/authorization_code_request.rb
```diff
@@ -6,18 +6,32 @@ class AuthorizationCodeRequest < BaseRequest
...
       private
 
+      def no_secret_allowed_for_pkce?
+        Doorkeeper.configuration.pkce_without_secret_enabled? &&
+          grant.respond_to?(:code_challenge) &&
+          grant.code_challenge.present? && @client.blank?
+      end
+
+      def client_by_uid(parameters)
+        Doorkeeper::Application.find_by(uid: parameters[:client_id])
```
nbulaj: Can you please use `by_uid` method from `Doorkeeper::Application` class ? https://github.com/doorkeeper-gem/doorkeeper/blob/master/lib/doorkeeper/models/application_mixin.rb#L29

nbulaj:
Hi @bpieck . Can you please take a look to my comments above? And rebase the branch to master to resolve the conflicts.

One thing I'm interested in: does PKCE flow enabled by default? And do we have an ability to disable it (like we do it in Doorkeeper configuration with `grant_flows ['client_credentials']` )?

bpieck:
Currently it's disabled, if you do not run this rails g doorkeeper:pkce

bpieck:
So I worked on your comments and rebased for newer master. And I added some tests for not executed migrations. You get now response invalid_request, if code_verifier is present in the request, but not supported by server.
I think to use an additional dummy would destroy maintainability and will need to have two different test setups. So instead I added the possibility to run rspec with environment variable WITHOUT_PKCE to simulate the dummy without generated migrations for pkce. I hope this is ok. So you can run all the other tests with and without pkce columns - and only pkce test specs differ depending on this environment variable set.

bpieck:
hm, so I have to rebase again.

nbulaj:
Sorry about that, to many commits :)

bpieck:
rebased - if anything left to be changed - just tell me.

PR #974, opened 2017-06-09
Fix uri comparing

dmitrytrager:
when doing checks from authorization code request

Related to: https://github.com/doorkeeper-gem/doorkeeper/issues/949

PR #1045, opened 2018-03-05
Validate redirect_uri as the native URI when making authorization code requests

kymmt90:
### Summary

[`AuthorizationCodeRequest#validate_redirect_uri`](https://github.com/doorkeeper-gem/doorkeeper/blob/8c337517cfde1237d5a1521b88e9bca364b210e1/lib/doorkeeper/oauth/authorization_code_request.rb#L48-L53) has started to use [`Helpers::URIChecker.valid_for_authorization?`](https://github.com/doorkeeper-gem/doorkeeper/blob/8c337517cfde1237d5a1521b88e9bca364b210e1/lib/doorkeeper/oauth/helpers/uri_checker.rb#L19-L21) since the change in #974. Although this change makes sense, this method does not care the native URI (default is `urn:ietf:wg:oauth:2.0:oob`). So, it seems that `POST /oauth/token` always returns a 401 error when using the native URI as a callback in authorization code flow since v4.3.0.

I think [`Helpers::URIChecker.native_uri?`](https://github.com/doorkeeper-gem/doorkeeper/blob/8c337517cfde1237d5a1521b88e9bca364b210e1/lib/doorkeeper/oauth/helpers/uri_checker.rb#L27-L29) is also needed in `AuthorizationCodeRequest#validate_redirect_uri` to validate as the native URI.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,7 @@ User-visible changes worth mentioning.
 ## master
 
 - [#907] Fix lookup for matching tokens in certain edge-cases
+- [#1045] Validate redirect_uri as the native URI when making authorization code requests
 
 ## 4.3.1
 
--- a/lib/doorkeeper/oauth/authorization_code_request.rb
+++ b/lib/doorkeeper/oauth/authorization_code_request.rb
@@ -46,10 +46,11 @@ module Doorkeeper
       end
 
       def validate_redirect_uri
-        Helpers::URIChecker.valid_for_authorization?(
-          redirect_uri,
-          grant.redirect_uri
-        )
+        Helpers::URIChecker.native_uri?(redirect_uri) ||
+          Helpers::URIChecker.valid_for_authorization?(
+            redirect_uri,
+            grant.redirect_uri
+          )
       end
     end
   end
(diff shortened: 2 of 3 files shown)


nbulaj:
Hi @kymmt90 . It make sense! Could you please add an entry to the changelog (NEWS.md) and squash commits to a single one? Thanks!

kymmt90:
@nbulaj Added the changelog and squash commits. Thank you!

hallucinations:
@kymmt90, @nbulaj It makes more sense to return `true` from [`Doorkeeper::OAuth::Helpers::URIChecker::valid?`](https://github.com/doorkeeper-gem/doorkeeper/blob/master/lib/doorkeeper/oauth/helpers/uri_checker.rb#L5), doesn't? 

This way we would make sure that the `params[:redirect_uri]` is the same as the redirect_uri for the grant since we are [matching them](https://github.com/doorkeeper-gem/doorkeeper/blob/master/lib/doorkeeper/oauth/helpers/uri_checker.rb#L27).

Here's the spec change:

```
Doorkeeper::OAuth::AuthorizationCodeRequest
  when redirect_uri is native
    invalidates when redirect_uri of the grant is not native
    validates when redirect_uri of the grant is also native
```

If you agree, please let me know – so that I can submit a patch.

nbulaj:
By changing `Doorkeeper::OAuth::Helpers::URIChecker::valid?`we will introduce changes for all the grant types that uses this class for checking URI. If you are sure that we don't break anything (especially for flows like client credentials or password) - you can try to submit a patch, I will check it.

hallucinations:
@nbulaj, Thanks! And sure, I will submit a patch.

Issue #996, opened 2017-10-26
Expiration Time Base On Grant Type


GeminPatel:
**Feature:** To provide different expiration to tokens based on different grant type.

**Why?:** We may want different expiration time for client_credentials, password, and implicit. Password tokens may have 1 hr of expiry while Client Credential token may have 1 day of expiry.

Do you guys think this is possible/feasible feature to implement?

_PS: We may also keep these expiry times per client in a json column in the DB._

kapilmittal:
+1 for this feature.
I also need to provide short lived user tokens and relatively long lived client tokens in my oauth app. 

Would you guys be interested in a PR for the same?

nbulaj:
Hi @kapilmittal . Feel free to send a PR with this feature. I think it must be a configurable lambda that must be used optionally (if set) for all strategies

nbulaj:
Merged in  55439ee

PR #1049, opened 2018-03-12
Adds custom expiration time based on grant type

rishabhsairawat:
### Summary
This PR aims to solve #996 **Expiration Time Base On Grant Type**

### DIFF
--- a/lib/doorkeeper/oauth/authorization/token.rb
+++ b/lib/doorkeeper/oauth/authorization/token.rb
@@ -5,8 +5,8 @@ module Doorkeeper
         attr_accessor :pre_auth, :resource_owner, :token
 
         class << self
-          def access_token_expires_in(server, pre_auth_or_oauth_client)
-            if (expiration = custom_expiration(server, pre_auth_or_oauth_client))
+          def access_token_expires_in(server, pre_auth_or_oauth_client, grant_type)
+            if (expiration = custom_expiration(server, pre_auth_or_oauth_client, grant_type))
               expiration
             else
               server.access_token_expires_in
@@ -15,14 +15,14 @@ module Doorkeeper
 
           private
 
-          def custom_expiration(server, pre_auth_or_oauth_client)
+          def custom_expiration(server, pre_auth_or_oauth_client, grant_type)
             oauth_client = if pre_auth_or_oauth_client.respond_to?(:client)
                              pre_auth_or_oauth_client.client
                            else
                              pre_auth_or_oauth_client
                            end
 
-            server.custom_access_token_expires_in.call(oauth_client)
+            server.custom_access_token_expires_in.call(oauth_client, grant_type)
           end
         end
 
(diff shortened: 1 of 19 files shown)


nbulaj:
Hi @rishabhsairawat . Thank you for th PR!

I think that we can extract grant type names (like `'client_credentials'` and other) in some constant like:

```ruby
module Doorkeeper 
  module OAuth
    FLOWS = [
      CLIENT_CREDENTIALS = 'client_credentials'.freeze,
      REFRESH_TOKEN = 'refresh_token'.freeze,
      # ...
    ].freeze
  
    # ...
  end
end
```

lib/doorkeeper/oauth/base_request.rb
```diff
@@ -2,6 +2,8 @@ module Doorkeeper
   module OAuth
     class BaseRequest
       include Validations
+      
+      attr_accessor :grant_type
```
nbulaj: Do we really need accessor ? I think `attr_reader` would be enough

lib/doorkeeper/oauth/authorization/token.rb
```diff
@@ -36,7 +36,7 @@ def issue_token
             pre_auth.client,
             resource_owner.id,
             pre_auth.scopes,
-            self.class.access_token_expires_in(configuration, pre_auth),
+            self.class.access_token_expires_in(configuration, pre_auth, 'implicit'),
```
nbulaj: Use constants for such things as I described in the topic comments.

spec/controllers/authorizations_controller_spec.rb
```diff
@@ -30,10 +30,13 @@ def translated_error_message(key)
   let(:client)        { FactoryBot.create :application }
   let(:user)          { User.create!(name: 'Joe', password: 'sekret') }
   let(:access_token)  { FactoryBot.build :access_token, resource_owner_id: user.id, application_id: client.id }
-
+  
   before do
     allow(Doorkeeper.configuration).to receive(:grant_flows).and_return(["implicit"])
     allow(controller).to receive(:current_resource_owner).and_return(user)
+    allow(Doorkeeper.configuration).to receive(:custom_access_token_expires_in).and_return(proc {|app, grant|
+        grant == 'implicit' ?  1234 : nil
```
nbulaj: Use flows constants in specs too.

lib/doorkeeper/oauth/authorization/token.rb
rishabhsairawat: Hi @nbulaj Where to define these constants for grant flows as there is no class for grants in Doorkeeper::OAuth module. Should I define them  in BaseRequest class

nbulaj: Let's place them into `Doorkeeper::Request` module as `GRANT_TYPES`

nbulaj: No, let's introduce  `oauth.rb` with explicit OAuth module:

```ruby
module Doorkeeper
  module OAuth
    GRANT_TYPES = [
     # ...
    ]. freeze
  end
end
```

rishabhsairawat: I was thinking the same as It would be better to place them in separate file.

nbulaj:
Thank you @rishabhsairawat 👍

PR #1052, opened 2018-03-14
Fix NameError with confidential apps generator

talklittle:
### Summary

Fixes typo which caused migration to fail with new Confidential applications migration generator

### Other Information

Error was `NameError: uninitialized constant AddConfidentialToApplications`


nbulaj:
Thanks!

PR #1053, opened 2018-03-15
Support redirect_uri's with registered query param(s)

### LINKED ISSUE #1050: why allow create oauth application with redirect_uri that has query params but cannot request
### Steps to reproduce
1. Create an application with redirect_uri that has query params like `https://github.com?allow=yes`. It works and no validate errors.
2. When test oauth request, which is http://localhost/oauth/authorize?client_id=xxxxxxx&redirect_uri=https%3A%2F%2Fgithub.com%3Fallow%3Dyes&response_type=code,  it render the invalid_redirect_uri page.

### Expected behavior
I hope the behavior of application's creation with query params and request oauth is the same, both allow or neither allow.

And I saw the source code.

1. when create application, the validation of query params is absent.   https://github.com/doorkeeper-gem/doorkeeper/blob/644d2ceb41411b9b320a6c8ee769762b37deba01/app/validators/redirect_uri_validator.rb#L8

2. when request the oauth, it checks the query params.   
https://github.com/doorkeeper-gem/doorkeeper/blob/master/lib/doorkeeper/oauth/helpers/uri_checker.rb#L15

Thank you!

f3ndot:
This amends the URIChecker to allow authorization for requests that contain query params. The query params all must be defined in the client app's #redirect_uri and all defined params must be present. Their order does not matter.

If there is no query param registered in the app's #redirect_uri then the previous behaviour occurs: stripping the query param from the request and comparing the URLs without it.

Fixes #1050

### DIFF
--- a/lib/doorkeeper/oauth/helpers/uri_checker.rb
+++ b/lib/doorkeeper/oauth/helpers/uri_checker.rb
@@ -12,6 +12,13 @@ module Doorkeeper
         def self.matches?(url, client_url)
           url = as_uri(url)
           client_url = as_uri(client_url)
+
+          if client_url.query.present?
+            return false unless query_matches?(url.query, client_url.query)
+            # Clear out queries so rest of URI can be tested. This allows query
+            # params to be in the request but order not mattering.
+            client_url.query = nil
+          end
           url.query = nil
           url == client_url
         end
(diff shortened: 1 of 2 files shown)


f3ndot:
Someone hilariously I *just* ran into this issue while trying to integrate Amazon Alexa into my Doorkeeper app: they require a `?vendorID=XXXX` in the redirect URI

nbulaj:
Whooops, once again no changelog entry :) Can you add it please?

Issue #1002, opened 2017-11-27
Bug: Able to get default scopes in the token even if they are not present in the application scopes using password grant type


kapilmittal:
When grant type is 'password' and application scopes are present:
     1.  when scope parameter is not present in the /oauth/token api then this api is giving token with default scopes (for exp. 'public' scope) even if 'public' scope is not present in the application scopes. 
     2. when scope parameter is present in the api and its value is 'public' then it gives following error:
         {
            "error": "invalid_scope",
            "error_description": "The requested scope is invalid, unknown, or malformed."
         }

Above 2nd point behaviour is the correct behaviour and api should behave like this for point 1 (when scope parameter is not present ) also.

Note: When grant type is client_credentials then api is working fine (It gives error when scope parameter is not present and 'public' scope is not present in the application scopes).  This issue exists for 'password' grant type.

f3ndot:
Hi @kapilmittal, I was unable to reproduce your issue with specs (see #1039). Can you put up a failing test or double check your findings?

Thanks

nbulaj:
I've double checked the specs and @f3ndot specs and could not find an error. Maybe you have a typo in your scopes or Doorkeeper initializer?

nbulaj:
Hi @kapilmittal . Any news here?

kapilmittal:
I am very sure that this issue was happening at that time. I will check this once again on Monday and let you know. Will check the specs also.

kapilmittal:
Also I remember that I looked into the doorkeeper code also and the code looked in sync with the mentioned bug because in the code, if scope parameter was not provided in the API then it was skipping the scope validation. 
Anyway, will confirm this on Monday.

kapilmittal:
Hi @f3ndot / @nbulaj 

I have verified once again and this issue still exists. 
I have checked the specs also. In the specs, 'scope' parameter is provided in the api but as mentioned in the bug, this issue exists when we don't provide the 'scope' parameter in the api (then api returns token with default scope even if that default scope is not present in the application scopes).

kapilmittal:
Hi. @f3ndot / @nbulaj . Did you have a chance to look into this again?

nbulaj:
Hi @kapilmittal . Can you please prepare the failing spec that I can use for tests and work on this issue?

kapilmittal:
Sure. will add failing spec.

nbulaj:
[...]

PR #1056, opened 2018-03-18
Adds failing test for password grant when application scopes are present and no scope is passed

rishabhsairawat:
### Summary

This is with reference to issue #1002 **Bug: Able to get default scopes in the token even if they are not present in the application scopes using password grant type**

### Other Information
 I am adding failing test for the bug.


nbulaj:
Thanks @rishabhsairawat  , we need to work on that! Feel free to send a PR if you have q quick fix

PR #1066, opened 2018-03-28
Use factory bot to save application rather than using a double

pakwfoley:
### Summary

Everywhere else in the doorkeeper specs FactoryBot is used to create applications. This change follows that consistency.  

### Other Information

This change came about while building a custom orm that proxys to our main application's database for application data. There is no nice way of handling application doubles in our mock, so this small change would help us greatly.


nbulaj:
Nice point, thanks!

Issue #536, opened 2014-12-22
How to keep database clean


chumakoff:
Should I do something with access_token table that grows infinitely ? Is there a way to automatically remove all expired or revoked tokens ?

jfcalvo:
I think that Doorkeeper should add a rake task for this.

francesle:
Similarly for `oauth_access_grants`, can we safely delete revoked or expired access grants?

jfcalvo:
I think that this issue should be reopened. Doorkeeper should implement some way to maintain the database clean (rake tasks or some other options) it's not very nice to know that you database will increase ad infinitum.

tute:
I won't work on this (I've never seen it being a problem, and I believe YAGNI), but I'm not opposed to a simple rake task that clears the unneeded objects from the DB. I would merge it in.

rossinkiwi:
I got around the problem of an endlessly growing *oauth_access_grants* table by over-riding a method in *lib/doorkeeper/oauth/authorization_code_request.rb*.  Here's what I did:
1. Create a file *config/initializers/doorkeeper_delete_grants.rb*
2. Insert the following:
```ruby
Doorkeeper::OAuth::AuthorizationCodeRequest.module_eval do
# Override this method so that oauth_access_grants table does not grow continuously.
# Original behavior was to retain all the revoked oauth_access_grants, they now 
# get deleted after revoking.

    private

    def before_successful_response
      grant.revoke
      gapp = grant.application
      groid = grant.resource_owner_id
      gscopes = grant.scopes
      Doorkeeper::AccessGrant.find(grant.id).destroy
      #find_or_create_access_token(grant.application,
      #                            grant.resource_owner_id,
      #                            grant.scopes,
      #                            server)
      find_or_create_access_token(gapp,
                                  groid,
                                  gscopes,
                                  server)
    end
  end
```

Charizard:
:+1: for a rake task.

Ashkin:
+1 for a rake task.

Specific use-case: I regularly copy my production database over to staging for testing with real-world data. As my staging database has a limit of 10 million rows I need a way to reduce row count, and the biggest culprit of excess data is Doorkeeper.

If there was an official rake task, I wouldn't need to write my own and hope I didn't overlook something.

As a secondary benefit, it would take a good deal less time to perform my daily/weekly backups and the staging imports without this excess data, as the oauth tables account for 25%+ of the total import time.

tute:
[...]

PR #1068, opened 2018-04-03
Add a rake task to cleanup stale database records

NobodysNightmare:
### Summary

Based on discussions in https://github.com/doorkeeper-gem/doorkeeper/issues/536,
this merge request wraps "safe to execute" maintenance tasks
in a rake task.

For now this is basically the one-liner provided by @tute, but I assume that there is more
data that we can safely remove during such a cleanup.

My main motivation is that migrations on the `oauth_access_tokens` table tend to take quite a long
time, when that table has grown rather large.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ https://github.com/doorkeeper-gem/doorkeeper/releases
   - [Routes](#routes)
   - [Authenticating](#authenticating)
   - [Internationalization (I18n)](#internationalization-i18n)
+  - [Rake Tasks](#rake-tasks)
 - [Protecting resources with OAuth (a.k.a your API endpoint)](#protecting-resources-with-oauth-aka-your-api-endpoint)
   - [Ruby on Rails controllers](#ruby-on-rails-controllers)
   - [Grape endpoints](#grape-endpoints)
@@ -106,7 +107,7 @@ for each table that includes a `resource_owner_id` column:
 add_foreign_key :table_name, :users, column: :resource_owner_id
 ```
 
-If you want to enable [PKCE flow] for mobile apps, you need to generate another 
+If you want to enable [PKCE flow] for mobile apps, you need to generate another
 migration:
 
 [PKCE flow]: https://tools.ietf.org/html/rfc7636
@@ -115,7 +116,7 @@ migration:
     rails generate doorkeeper:pkce
 ```
 
-If you want to allow PKCE flow without secrets, you can configure doorkeeper to 
+If you want to allow PKCE flow without secrets, you can configure doorkeeper to
 allow this:
 
 ```ruby
@@ -242,6 +243,34 @@ You may want to check other ways of authentication
 
 Doorkeeper support multiple languages. See language files in
 [the I18n repository](https://github.com/doorkeeper-gem/doorkeeper-i18n).
+
+### Rake Tasks
+
+If you are using `rake`, you can load rake tasks provided by this gem, by adding
+the following line to your `Rakefile`:
+
+```ruby
+Doorkeeper::Rake.load_tasks
+```
+
+#### Cleaning up
+
+By default Doorkeeper is retaining expired and revoked access tokens and grants.
+This allows to keep an audit log of those records, but it also leads to the
+corresponding tables to grow large over the lifetime of your application.
+
+If you are concerned about those tables growing too large,
+you can regularly run the following rake task to remove stale entries
+from the database:
+
+```rake
+rake doorkeeper:db:cleanup
+```
+
+Note that this will remove tokens that are expired according to the configured TTL
+in `Doorkeeper.configuration.access_token_expires_in`. The specific `expires_in`
+value of each access token **is not considered**. The same is true for access
+grants.
 
 ## Protecting resources with OAuth (a.k.a your API endpoint)
 
@@ -471,7 +500,7 @@ Doorkeeper 4.3.0 it uses [ActiveSupport lazy loading hooks](http://api.rubyonrai
 to load models. There are [known issue](https://github.com/doorkeeper-gem/doorkeeper/issues/1043)
 with the `factory_bot_rails` gem (it executes factories building before `ActiveRecord::Base`
 is initialized using hooks in gem railtie, so you can catch a `uninitialized constant` error).
-It is recommended to use pure `factory_bot` gem to solve this problem. 
+It is recommended to use pure `factory_bot` gem to solve this problem.
 
 ## Upgrading
 
--- a/lib/doorkeeper/orm/active_record/stale_records_cleaner.rb
+++ b/lib/doorkeeper/orm/active_record/stale_records_cleaner.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+module Doorkeeper
+  module Orm
+    module ActiveRecord
+      class StaleRecordsCleaner
+        def initialize(model)
+          @model = model
+        end
+
+        def clean_revoked
+          table = @model.arel_table
+          @model.where.not(revoked_at: nil)
+                .where(table[:revoked_at].lt(Time.current))
+                .delete_all
+        end
+
+        def clean_expired(ttl)
+          table = @model.arel_table
+          @model.where(table[:created_at].lt(Time.current - ttl))
+                .delete_all
+        end
+      end
+    end
+  end
+end
--- a/lib/doorkeeper/rake/db.rake
+++ b/lib/doorkeeper/rake/db.rake
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+namespace :doorkeeper do
+  namespace :db do
+    desc 'Removes stale data from doorkeeper related database tables'
+    task cleanup: [
+      'doorkeeper:db:cleanup:revoked_tokens',
+      'doorkeeper:db:cleanup:expired_tokens',
+      'doorkeeper:db:cleanup:revoked_grants',
+      'doorkeeper:db:cleanup:expired_grants'
+    ]
+
+    namespace :cleanup do
+      desc 'Removes stale access tokens'
+      task revoked_tokens: 'doorkeeper:setup' do
+        cleaner = Doorkeeper::Orm::ActiveRecord::StaleRecordsCleaner.new(Doorkeeper::AccessToken)
+        cleaner.clean_revoked
+      end
+
+      desc 'Removes expired (TTL passed) access tokens'
+      task expired_tokens: 'doorkeeper:setup' do
+        cleaner = Doorkeeper::Orm::ActiveRecord::StaleRecordsCleaner.new(Doorkeeper::AccessToken)
+        cleaner.clean_expired(Doorkeeper.configuration.access_token_expires_in)
+      end
+
+      desc 'Removes stale access grants'
+      task revoked_grants: 'doorkeeper:setup' do
+        cleaner = Doorkeeper::Orm::ActiveRecord::StaleRecordsCleaner.new(Doorkeeper::AccessGrant)
+        cleaner.clean_revoked
+      end
+
+      desc 'Removes expired (TTL passed) access grants'
+      task expired_grants: 'doorkeeper:setup' do
+        cleaner = Doorkeeper::Orm::ActiveRecord::StaleRecordsCleaner.new(Doorkeeper::AccessGrant)
+        cleaner.clean_expired(Doorkeeper.configuration.authorization_code_expires_in)
+      end
+    end
+  end
+end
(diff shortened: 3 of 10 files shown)


lib/doorkeeper/rake/setup.rake
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+namespace :doorkeeper do
+  task setup: :environment do
```
NobodysNightmare: This is inspired by `resque`, where you have a hook that is executed before each rake task of that gem.

I decided to depend on `:environment` by default, but to be honest I never used Doorkeeper outside a Rails-context. If it is neccessary, we could also make the dependency on `environment` optional, as they've done in `resque`.

README.md
```diff
@@ -243,6 +244,27 @@ You may want to check other ways of authentication
 Doorkeeper support multiple languages. See language files in
 [the I18n repository](https://github.com/doorkeeper-gem/doorkeeper-i18n).
 
+### Rake Tasks
+
+If you are using `rake`, you can load rake tasks provided by this gem, by adding
+the following line to your `Rakefile`:
+
+```ruby
+Doorkeeper::Rake.load_tasks
```
NobodysNightmare: Alternatively we could load all the rake tasks when requiring `doorkeeper`, but I feel that this would force people to use rake, when it is not a strict dependency of this gem.

README.md
```diff
@@ -243,6 +244,27 @@ You may want to check other ways of authentication
...
+
+If you are concerned with some of the tables added by Doorkeeper getting too large
+over time, you can regularly run the following rake task to remove stale entries
+from the database:
+
+```rake
+rake doorkeeper:db:cleanup
+```
+
+Note that by default those records are retained to allow stricter auditing.
```
NobodysNightmare: Actually I assume that this is the reason. Feel free to fill me in on a better explanation :)

NobodysNightmare:
@tute wrote:
> Doorkeeper keeps revoked tokens, you can delete them with a query like:
> ```
> DELETE FROM oauth_access_tokens WHERE revoked_at IS NOT NULL AND revoked_at < NOW();
> ```

On our database I can see that this does not hit the majority of tokens. We have lots of tokens, that simply expired (`created_at = 2016-...`, `expires_in = 7200`).

Is it safe to delete those after a short hold period? Should I keep them? If yes, why?

lib/doorkeeper/rake/setup.rake
nbulaj: Looks pretty good for me :+1:

README.md
nbulaj: Yep, everything is OK

nbulaj:
Hi @NobodysNightmare . What about to remove all the tokens / grants not for current time, but without token TTL?

```ruby
.where(tokens[:revoked_at].lt(Time.current - Doorkeeper.configuration.access_token_expires_in))
```

It will prevent from removing still-valid tokens. What do you think?

NobodysNightmare:
Hey @nbulaj, I assume you mean tokens that were **created** more than the TTL amount ago? Your code example refers to the revocation date. I.e.:

```ruby
.where(tokens[:created_at].lt(Time.current - Doorkeeper.configuration.access_token_expires_in))
```

Your approach seems nice to me and should be easy to implement. I am a little afraid of the edge case, where the TTL of tokens is being changed. For that purpose the table contains a column for the ttl of the token. So I could build a SQL query that calculates whether `created_at + ttl < Time.current`. However, calculations with dates are specific to your SQL backend (MySQL is different from Postgres and so on).

Do you think it is sufficient to rely on `Doorkeeper.configuration.access_token_expires_in`? I can implement that.

nbulaj:
Yep, sorry, I just copied your example :)

It's ok to remove revoked tokens, but also we can remove not revoked, but already not accessible tokens (Time.current - TTL).

> However, calculations with dates are specific to your SQL backend (MySQL is different from Postgres and so on).

But we use ORM's that must implement it for us. isn' it? :)

NobodysNightmare:
I just added a commit that also cleans up expired tokens and grants using the simple approach that you proposed.

I split the rake tasks further, so that anyone would be able to just run the cleanups they need.

> The only edge case is custom expiration time :\ I can't say right now if it can be used for rake tasks. What do you think?

I chose the static approach now (reading TTL from configuration) and also added a disclaimer for that to the README.

Alternative approaches:
* Let the SQL backend decide which records are matching and delete them
    * this would be fast (just one `DELETE FROM` query)
    * I have no clue how to do this portable, since the `where` clause had to contain some DBMS specific date calculation (e.g. for MySQL: `where('DATE_ADD(created_at, INTERVAL expires_in SECOND) < ?', Time.current)`)
        * this is where my ORM might help me, but I don't know of the relevant ActiveRecord feature
* implement it using expiration calculation in Ruby (`Doorkeeper::AccessToken.select(&:expired?).each(&:destroy)`)
    * slow N+1 query
    * more accurate than current implementation

lib/doorkeeper/rake/db.rake
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+namespace :doorkeeper do
+  namespace :db do
+    desc 'Removes stale data from doorkeeper related database tables'
+    task cleanup: [
+                    'doorkeeper:db:cleanup:revoked_tokens',
```
nbulaj: Let's indent it (all the array items) with just 2 spaces and final `]` under the `t` of `task`

lib/doorkeeper/rake/db.rake
```diff
@@ -0,0 +1,51 @@
...
+                    'doorkeeper:db:cleanup:expired_tokens',
```
nbulaj: Let's indent it (all the array items) with just 2 spaces and final `]` under the `t` of `task`

nbulaj:
Also could you please squash all your commits to a single one after you finish? Thanks!

lib/doorkeeper/orm/active_record/stale_record_cleaner.rb
```diff
@@ -0,0 +1,26 @@
...
+          table = @model.arel_table
+          @model.where.not(revoked_at: nil)
+                .where(table[:revoked_at].lt(Time.current))
+                .delete_all
+        end
+
+        def clean_expired(ttl)
+          table = @model.arel_table
+          @model.where(table[:created_at].lt(Time.current - ttl))
+                .delete_all
```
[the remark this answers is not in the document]
NobodysNightmare: Is @houndci-bot right on this one?

The rubocop default is having leading dots and I don't see doorkeeper overriding that...

lib/doorkeeper/rake/db.rake
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+namespace :doorkeeper do
```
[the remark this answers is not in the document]
NobodysNightmare: I think that rake namespaces should not be affected by this, while I agree that `task` should not be too long.

We are using a rubocop config like this at our projects:

```yml
# We need to configure exemptions for blocks that we generally accept to be
# long, since they are less comparable to methods and more comparable to
# modules/classes.
Metrics/BlockLength:
  ExcludedMethods:
    - context
    - describe
    - namespace
    - factory
    - define
    - shared_examples_for
    - shared_examples
    - xcontext
    - xdescribe
```

What do you think?

NobodysNightmare:
Addressed your points and squashed to a single commit.

I extracted the beefy part of the rake task into a service class. This also helped in reducing the duplication that was present between the tasks.

I am a little confused by the hound results, see my responses above :)

spec/lib/orm/active_record/stale_record_cleaner_spec.rb
```diff
@@ -0,0 +1,77 @@
+require 'spec_helper'
```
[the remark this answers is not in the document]
NobodysNightmare: fixed

lib/doorkeeper/rake/db.rake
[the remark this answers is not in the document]
nbulaj: I just disable this check for all RSpec suite because it is too noisy :) Just ignore this one

lib/doorkeeper/orm/active_record/stale_record_cleaner.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module Doorkeeper
+  module Orm
+    module ActiveRecord
+      class StaleRecordCleaner
```
nbulaj: Just a small hint: maybe better to name it '`StaleRecordsCleaner`' (as we remove more than one record)? :)

lib/doorkeeper/orm/active_record/stale_record_cleaner.rb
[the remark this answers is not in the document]
nbulaj: Just ignore it. Not a big deal.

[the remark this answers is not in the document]
nbulaj: Just ignore it. Not a big deal.

lib/doorkeeper/orm/active_record/stale_record_cleaner.rb
NobodysNightmare: True :)

fixed

NobodysNightmare:
@nbulaj I was thinking about this MR again and I am not sure anymore, that deleting expired access tokens is a good idea.

As far as I can tell, these tokens are still used (after their expiry) to obtain new access tokens using their `refresh_token`.

By cleaning expired tokens, we basically also invalidate the corresponding `refresh_tokens`... After which amount of time should this be considered a safe operation? As far as I can tell, the refresh tokens are so far never expiring...

nbulaj:
Hey, @NobodysNightmare . We completely forgot about refresh tokens, you are right! But it is a configurable option (`use_refresh_token`), so by default there are no refresh tokens on OAuth2 provider with Doorkeeper. Therefore this task can be helpful in this case.

What I can suggest is to add configuration check to rake tasks that will see if this option is activated. If yes, it must: **1)** stop working with the warning message, **2)** show warning message and ask if user is sure he want's to remove the tokens and if yes - continue the operation.

What do you think?

NobodysNightmare:
Warning and having an option to continue anyways sounds good... What should we do about the general clean task then? (the one that executes all four cleanups)

Failing it completely just when one independent step does not work sounds not so nice...

I see three options:
* remove the common task and just expose the separate tasks
* skip expired tokens in the common task when it is not safe
    * log that it was skipped
* ~~fail common task, but ensure expired tokens are handled last~~

edit: Thinking about it, the last option is probably lazy and not useful for anyone...

NobodysNightmare:
Of course one additional idea:

Even if `use_refresh_token` is true, not all access tokens will have a refresh token associated. When obtaining a token via client credentials flow, there is never a refresh token associated.

So alternatively, we could only delete expired refresh tokens that also have **no** refresh token.

For our database this would still help a lot, since we have many more tokens issued via client credentials flow...

It would also be a solution that has the rake task always behave the same. We are always removing all the expired tokens that have surely no purpose anymore...

nbulaj:
To skip expired tokens and to log that is a good idea I think. And every developer can then run another task that will destroy expired tokens.

However, this one sounds better to me:

> we could only delete expired refresh tokens that also have **no** refresh token.

Other cases depends on the purposes of the project, so I think every developer can write his own tasks.

NobodysNightmare:
> we could only delete expired refresh tokens that also have **no** refresh token.

Implemented that proposal here: https://github.com/doorkeeper-gem/doorkeeper/pull/1080

lib/doorkeeper/orm/active_record/stale_records_cleaner.rb
```diff
@@ -0,0 +1,26 @@
...
+        end
+
+        def clean_revoked
+          table = @model.arel_table
+          @model.where.not(revoked_at: nil)
+                .where(table[:revoked_at].lt(Time.current))
+                .delete_all
+        end
+
+        def clean_expired(ttl)
```
lefty313: @nbulaj first of all sorry for commenting already merged pull request. 

IMO people will use this rake task when database is already bloated but in such case it can easily kill production database because doorkeeper does not generate index for `AccessToken#created_at` field. Maybe we should consider adding this index by default or add warning in `Readme.md` ?

nbulaj: @NobodysNightmare it definitely makes sense. Maybe `find_each` could help us. Would you like to produce a PR?

nbulaj: Fixed in https://github.com/doorkeeper-gem/doorkeeper/pull/1234

NobodysNightmare: Sorry for not responding. I had this somewhere in my inbox with "I want to answer there" :)

Cool that you already went ahead and fixed it :+1:

PR #1072, opened 2018-04-04
AuthorizationsController: Memoize strategy.authorize result

mdeutsch:
This allows easy access to the authorize response in a subclass of Doorkeeper::AuthorizationsController. It's equivalent to #568, which made the same change to TokensController.

PR #1064, opened 2018-03-26
Added :after_successful_authorization callback

nattfodd:
### Summary

This PR adds an ability to run some code after successful authorization. TLDR: https://github.com/doorkeeper-gem/doorkeeper/issues/1062

Example:

```ruby
Doorkeeper.configure do
  after_successful_authorization do |controller|
    controller.session[:logout_urls] << 
      Doorkeeper::Application
      .find_by(controller.request.params.slice(:redirect_uri))
      .logout_uri
  end
end
```

PR #1075, opened 2018-04-06
Call before/after_successful_authorization callbacks on create as well as new

mdeutsch:
### Summary

The before/after_successful_authorization callbacks added in #1064 are only called on the `new` action. I think they should also be called on `create` to cover situations where the user must explicitly authorize the client.

### Other Information

I also found that the callbacks can be run when authorization is not successful. I added some more tests and modified the existing code a bit.

This PR has some conflicts with my other PR, #1072. I'll resolve them but wanted to put this up for review.

cc @nattfodd

### DIFF
--- a/app/controllers/doorkeeper/authorizations_controller.rb
+++ b/app/controllers/doorkeeper/authorizations_controller.rb
@@ -22,13 +22,9 @@ module Doorkeeper
     private
 
     def render_success
-      before_successful_authorization
       if skip_authorization? || matching_token?
-        auth = authorize_response
-        after_successful_authorization
-        redirect_or_render auth
+        redirect_or_render authorize_response
       elsif Doorkeeper.configuration.api_only
-        after_successful_authorization
         render json: pre_auth
       else
         render :new
@@ -84,7 +80,13 @@ module Doorkeeper
     end
 
     def authorize_response
-      @authorize_response ||= strategy.authorize
+      @authorize_response ||= begin
+        authorizable = pre_auth.authorizable?
+        before_successful_authorization if authorizable
+        auth = strategy.authorize
+        after_successful_authorization if authorizable
+        auth
+      end
     end
 
     def after_successful_authorization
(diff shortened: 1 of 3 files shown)


nbulaj:
So the callback will be invoked twice? Looks strange from my point of view...

Moreover, maybe it will be better to move the hooks under the Authorization class instead of controller? It will DRY the code at least

nattfodd:
> So the callback will be invoked twice? Looks strange from my point of view...

+1. I would rather rename callback in `create`(or both?)  into something different, so it would be 2 different callbacks.

> Moreover, maybe it will be better to move the hooks under the Authorization class instead of controller? It will DRY the code at least

As I mentioned in previous PR, we need controller's scope in that callback (i.e. for interacting with cookies/sessions).

nbulaj:
We can use existing callbacks, but add an extra argument - action ('new' or 'create'). Controller ais already passed with the context argument, so you already has an access to cookies or request.

mdeutsch:
I've rebased onto master and extracted some of the duplicated code. Now the callbacks are closer to the Authorization class and I think the intent is clearer: that the callbacks are invoked when the request is successfully authorized.

app/controllers/doorkeeper/authorizations_controller.rb
```diff
@@ -84,7 +80,13 @@ def strategy
     end
 
     def authorize_response
-      @authorize_response ||= strategy.authorize
+      @authorize_response ||= begin
+        valid = pre_auth.authorizable?
```
nbulaj: Let's rename it to `authorizable` to be closer to implementation and the essentials of this method. What do you think?

mdeutsch: Good call -- I'll change it.

mdeutsch:
OK, I renamed the variable, added a changelog entry, and squashed/rebased.

nbulaj:
Got it, thank you for your work! :+1:

PR #1076, opened 2018-04-08
Add config to enforce content type

### LINKED ISSUE #1067: Access Token Request media type should be application/x-www-form-urlencoded
### Steps to reproduce

```shell
curl --include \
     --request POST \
     --header "Content-Type: application/json" \
     --data-binary '{"client_id":"9e48cc9491746c04ee5732b31e7cbb777352e066be149c878da955845e2f71c9", "grant_type": "password", "username": "user@example.com", "password": "secret"}' \
'http://localhost:3000/oauth/token'
```

I added Doorkeeper to an existing Rails 4.2 app and it's been working pretty well. Then I found out one of clients (a native mobile app) is sending auth requests in JSON format which is just by accident as other API endpoints are using the JSON API spec.

[According to the spec](https://tools.ietf.org/html/rfc6749#section-4.1.3), Access Token Request has parameters **using the "application/x-www-form-urlencoded"** format so I think it is better only allowing the correct media type for token requests.

### Expected behavior

Returning a 415 Unsupported media type

### Actual behavior

A successful authorization response is returned.

### System configuration

**Ruby version**: 4.2

baxang:
### Summary

Fixes #1067 

TODOs:

- [x] Fix on Rails 4.2
- [x] Improve test code so that HTTP action methods work with headers
- [x] Add the config to initializer template
- [x] Update NEWS


baxang:
Hi @nbulaj and @hallucinations,

Although this PR needs some more work, I think it is ready for someone else' review. Can you please take a look?

A few comments on my decisions are:

### Config API

I made the new config API `enforce_content_type` similar to `api_only`. The previous behaviour (allowing non-urlencoded-form media type) is preserved if not used.

Since many people, including myself, are already running Doorkeeper in production, I thought about more options than simply turning it on/off. For example, an option to log/collect metrics on content type so systems admins can understand their traffic. However I ended up thinking - ok let's just make it simple.

### Testing

I had some tricky problems in writing tests to support both Rails 4 and 5. There's some HTTP method shims for Rails 5 and the shims are written as if no headers are passed to them.

I ended up having some not pretty test code like below.

```ruby
it '200 for the correct media type' do
  @request.headers['Content-Type'] = 'application/x-www-form-urlencoded'
  get :index, {}, as: :url_encoded_form
  expect(response).to have_http_status 200
end
```

`@request.headers...` is for Rails 4 testing and `as: :url_encoded_form` is for Rails 5. Maybe I should spend some more time to make it more elegant. What do you think?

### `charset=utf-8`

As I said in #1067 I think it is reasonable to skip charset checking.

Regards,

nbulaj:
> I made the new config API `enforce_content_type`

This one good :+1: 

> For example, an option to log/collect metrics on content type so systems admins can understand their traffic. However I ended up thinking - ok let's just make it simple.

I think every production system has some additional services/tools for log analyze (graphana, ELK, NewRelic, etc). So they already have an ability to find useful info about content type in the requests. If not - well, they need to use it :)

> There's some HTTP method shims for Rails 5 and the shims are written as if no headers are passed to them.

So let's fix them to work with headers :) In this case tests will look good.

> As I said in #1067 I think it is reasonable to skip charset checking.

Absolutely agree.

---

Interesting thing: Hydra OAuth2 provider (Golang) supports both `application/json` and `application/x-www-form-urlencoded` for the requests https://github.com/ory/hydra/blob/366ed57d9c39d7601a40b5545f91361e6a2b9f5a/doc.go#L33 .

baxang:
@nbulaj Thanks for your review.

I will try to make test code nicer as well the rest of TODO items.

And regarding Hyda, it's interesting that I found an issue that rejected a feature request to support JSON: https://github.com/ory/hydra/issues/786. It might be the case JSON is used for other endpoints that are not part of OAuth 2.0 but I'm just guessing here.

baxang:
Hi @nbulaj I've added a couple commits to improve test code a bit.

What I've done is basically I inverted Rails 4 -> 5 syntax conversion: https://github.com/doorkeeper-gem/doorkeeper/pull/1076/commits/b6c22274bc8a5676712bdd6537e7895b25122229#diff-3b61ca35a7f47da3a2dbf21b5b9e3dd0

I think doing this is better for a few reasons:

- Rails 5 is the current stable so using its syntax makes sense
- Rails 5 syntax is more clearer with keyword arguments
- Importantly :-) it is easier to convert 5 -> 4

I'm still to fix CI failures, but before I spend more time I'd love to hear your thoughts.

nbulaj:
Hi @baxang . Totally agree: in future versions we will drop support of Rails 4.x, so it is reasonable to use actual syntax. Feel free to continue :+1:

baxang:
Test fails on all non-rails-5 versions and can't reproduce on local machine :-( Any advice would be appreciated. I will keep looking.

spec/support/http_method_shim.rb
```diff
@@ -3,34 +3,29 @@
...
-    super(path, params: params)
-  end
-
-  def put(path, params = {})
-    super(path, params: params)
+  def process(action, http_method = 'GET', **args)
+    if as = args.delete(:as)
+      @request.headers['Content-Type'] = Mime[as].to_s
+    end
+    super(action, http_method, args[:params], args[:session], args[:flash])
```
nbulaj: What about headers?

nbulaj:
I cloned your repo and executed rspec from the `rails_4_2` bundle and everything is OK O_o

spec/support/http_method_shim.rb
baxang: This is calling Rails 4's and in Rails 4 adding headers is not supported.

baxang:
Thanks for your effort and you're right 😕 that's what I'm experiencing too. What even further confuses me is CI errors.

Based on CI error log (from Ruby 2.4/Rails 4.2 https://travis-ci.org/doorkeeper-gem/doorkeeper/jobs/367131874 ) the first error reported is
https://travis-ci.org/doorkeeper-gem/doorkeeper/jobs/367131874#L576
https://travis-ci.org/doorkeeper-gem/doorkeeper/jobs/367131874#L637

which is

```
rspec ./spec/controllers/authorizations_controller_spec.rb:207 # Doorkeeper::AuthorizationsController implicit grant flow POST #create with callbacks when successful should call :before_successful_authorization callback
```

And the line is https://github.com/baxang/doorkeeper/blob/1067-enforce-strict-content-type/spec/controllers/authorizations_controller_spec.rb#L207 which does not look related to the failed spec I have in the log.

I'll keep digging

nbulaj:
Any news here?

baxang:
I could finally reproduce the errors with upstream master merged into my branch - which still does not explain failures on CI though. Maybe there's something that I don't know about how Travis CI runs tests.

Anyways the errors were caused by a couple specs recently added to master branch which obviously used Rails 4 `post` syntax. Easy fix and all looks good now.

baxang:
Let me know if you want me to squash commits

nbulaj:
Yes, could you please squash them? I think everything is ready to be merged

baxang:
No problem - I've squashed and pushed. Just waiting for Travis to turn green 💚 Thanks for your guidance. Can't wait to see 5.0 to out!

nbulaj:
Thank you @baxang for your work and patience! :+1:

nbulaj:
Hi @baxang . I've added format: :jsjon parameter for controller tests and it's failing for JSON API's. Could you please take a look? https://travis-ci.org/doorkeeper-gem/doorkeeper/jobs/385600167 Currently I'm looking how to pass this to original method

PR #1079, opened 2018-04-12
Quote surround I18n yes/no keys

trev:
### Summary

`yes` & `no` are reserved key names as part of the yaml spec: http://yaml.org/refcard.html
When left unquoted, they act as booleans and can't be referenced using the key name.


nbulaj:
Yep, you are right, thanx!

PR #1080, opened 2018-04-13
Only clean expired access tokens without a refresh token

NobodysNightmare:
The refresh token is usable even after the initial access token expired.
Therefore as soon as a refresh_token is present, we can't automatically
delete corresponding access tokens.

This does not affect revoked access tokens, their refresh_tokens are
supposed to be invalid, too.

### DIFF
--- a/lib/doorkeeper/orm/active_record/stale_records_cleaner.rb
+++ b/lib/doorkeeper/orm/active_record/stale_records_cleaner.rb
@@ -4,21 +4,21 @@ module Doorkeeper
   module Orm
     module ActiveRecord
       class StaleRecordsCleaner
-        def initialize(model)
-          @model = model
+        def initialize(base_scope)
+          @base_scope = base_scope
         end
 
         def clean_revoked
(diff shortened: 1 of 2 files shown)


lefty313:
@NobodysNightmare @nbulaj 

> The refresh token is usable even after the initial access token expired.

Could you tell me why deleting expired access token and still using previous refresh token to issue new access token cause issues with Doorkeeper ?

Is it related to doorkeeper internal or something from oauth spec enforce this logic ? I'm asking because I have a client which does not revoke refresh tokens which cause table to grow infinitely.

NobodysNightmare:
Both... So the implementation of doorkeeper stores issued refresh tokens in the table for access tokens (along with the access_token used to generate the refresh token).

By spec, refresh tokens do never expire, so technically it would be valid to use the refresh token of a 2 year old access token, to obtain a new one.

Depending on your use case, you should probably determine at which point it is safe to remove old refresh tokens.

PR #1084, opened 2018-04-15
Support OIDC 1.0's optional `prompt` parameter

### LINKED ISSUE #1073: Support Silent Authentication
In order to support renewing expired tokens for Single Page Applications using the implicit flow, Doorkeeper will need to support the [Silent Authentication](https://auth0.com/docs/api-auth/tutorials/silent-authentication) strategy.

Essentially, it's supporting a `prompt` parameter where when you pass `none` it will always redirect with either success or failure:

* A successful authentication response if the user was already logged in
* An error response if the user is not logged in and therefore cannot be silently authenticated

f3ndot:
Per http://openid.net/specs/openid-connect-core-1_0.html

Prompt accepts a few different values:

- `none`: Must not display consent or authorization pages. Either will redirect or error with `login_required` or `consent_required`. This param allows devs to use The Silent Authentication method as described by Auth0.
- `login`: Should prompt user for reauthentication. Doorkeeper will just fail with `login_required`.
- `consent`: Should prompt user for consent. Doorkeeper will not automatically redirect even if matching token.

Fixes #1073

### DIFF
--- a/app/controllers/doorkeeper/authorizations_controller.rb
+++ b/app/controllers/doorkeeper/authorizations_controller.rb
@@ -1,8 +1,15 @@
 module Doorkeeper
   class AuthorizationsController < Doorkeeper::ApplicationController
-    before_action :authenticate_resource_owner!
+    before_action do
+      render_prompt_error(:login_required) unless begin
+                                                    authenticate_resource_owner!
+                                                  rescue StandardError => e
+                                                    raise e unless no_prompt? || prompt_login?
+                                                  end
+    end
 
     def new
+      return render_prompt_error(:login_required) if prompt_login?
       if pre_auth.authorizable?
         render_success
       else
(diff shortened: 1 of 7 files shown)


app/controllers/doorkeeper/authorizations_controller.rb
```diff
@@ -1,8 +1,15 @@
 module Doorkeeper
   class AuthorizationsController < Doorkeeper::ApplicationController
-    before_action :authenticate_resource_owner!
+    before_action do
+      render_prompt_error(:login_required) unless begin
```
nbulaj: Let's extract this block into some method.

Also, what if some legacy projects will inherit from this controller - do they still get backward compatibility ?

app/controllers/doorkeeper/authorizations_controller.rb
```diff
@@ -1,8 +1,15 @@
...
+                                                    authenticate_resource_owner!
+                                                  rescue StandardError => e
```
nbulaj: Do we really need to handle `StandardError` ? :thinking:

f3ndot:
Also I see there is a project called [doorkeeper-openid_connect](https://github.com/doorkeeper-gem/doorkeeper-openid_connect). Is this not infringing upon that project? Should we keep this feature out of the "core" doorkeeper and instruct those (me) who wish to have certain OIDC features use that gem instead? 🤔

nbulaj:
Yep, `doorkeeper-openid_connect` is used for organizing OpenID Connect. I preffer to to have all the OpenID logic in this project if it is possible. The project is active and maintained by @toupeira and @nashby , so you could cooperate with them.

f3ndot:
Ah too bad I wasted my time! That project already supports `prompt` 😅. Will close this one out.

PR #1086, opened 2018-04-20
Use intersection of application and default server scopes #1002

rishabhsairawat:
### Summary
This PR aims to solve the issue #1002 .
### Reference
 Commit b5bf40e8d585ba61c205221c22cc2e3638533dcf by @nbulaj missed the following case:
 Default scopes: public, admin
 Application scopes: read, write, public
 And no scopes are passed.
 Expected output: public
 Output by b5bf40e8d585ba61c205221c22cc2e3638533dcf : invalid scope

### DIFF
--- a/lib/doorkeeper/oauth/password_access_token_request.rb
+++ b/lib/doorkeeper/oauth/password_access_token_request.rb
@@ -28,9 +28,11 @@ module Doorkeeper
 
       def validate_scopes
         application_scopes = client.try(:scopes)
-        return true if @original_scopes.blank? && application_scopes.blank?
-
-        ScopeChecker.valid?(scopes.to_s, server.scopes, application_scopes)
+        if @original_scopes.blank?
+          return true if application_scopes.blank?
+          @original_scopes = (application_scopes & server.default_scopes).to_s
+        end
+        ScopeChecker.valid? @original_scopes, server.scopes, application_scopes
       end
 
       def validate_resource_owner
--- a/spec/requests/flows/password_spec.rb
+++ b/spec/requests/flows/password_spec.rb
@@ -171,7 +171,7 @@ describe 'Resource Owner Password Credentials Flow' do
     end
 
     it 'should issue new token with one default scope that are present in application scopes' do
-      default_scopes_exist :public
+      default_scopes_exist :public, :admin
 
       expect do
         post password_token_endpoint_url(client: @client, resource_owner: @resource_owner)
@@ -195,7 +195,7 @@ describe 'Resource Owner Password Credentials Flow' do
 
       expect(token.application_id).to eq @client.id
       should_have_json 'access_token', token.token
-      should_have_json 'scope', 'public read'
+      should_have_json 'scope', 'read public'
     end
   end
 


nbulaj:
What about other grant flows, ie `client credentials`? I think they must work the same way

rishabhsairawat:
@nbulaj  Yes, Other grant flows must work the same way.

rishabhsairawat:
@nbulaj I think this can be done by changing scopes method in BaseRequest.rb file and then using scopes like you did in b5bf40e8d585ba61c205221c22cc2e3638533dcf .
```
     def scopes
        @scopes ||= if @original_scopes.present?
                      OAuth::Scopes.from_string(@original_scopes)
                    else
                      if @client.try(:scopes).present?
                        default_scopes & @client.scopes
                      else
                        default_scopes
                      end
                    end
      end
```

nbulaj:
@rishabhsairawat yep, this is a good strategy. Implemented in 976b235

rishabhsairawat:
@nbulaj Checkout following scenario:
**When there is no application scope and no default scope present, Doorkeeper issues token without any scope.**
Wouldn't it be a better to show error ?

nbulaj:
> When there is no application scope and no default scope present
> Currently, Doorkeeper issues AccessToken without any scope
> Wouldn't it be better to show error ?
> And Same is With AccessGrant

I don't sure if it must work this way. Some endpoints can require OAuth authentication, but without any particular scopes (just to be sure client is authenticated).

rishabhsairawat:
@nbulaj If you want this then I think In following case also, scopeless token and grant should be issued.

**If intersection of application scopes and default scopes is blank and No scopes are passed in request**
Intersection of application scopes and default  scopes will be blank in following cases
1. If default scopes are not present then intersection will automatically be blank 
2. If default scopes are present but differs from application scopes

nbulaj:
@rishabhsairawat of I understood you correctly:

> If intersection of application scopes and default scopes is blank and No scopes are passed in request
Intersection of application scopes and default scopes will be blank in following cases
> 
>     If default scopes are not present then intersection will automatically be blank
>     If default scopes are present but differs from application scopes

then I think it already works this way.

```ruby
context 'when application scopes contain some of the default scopes and no scope is passed' do
  before do
    client.update_attributes(scopes: 'value1 value2')
  end

  it 'issues new token with one default scope that are present in application scopes' do
    default_scopes_exist :value3

    headers = authorization client.uid, client.secret
    params  = { grant_type: 'client_credentials' }

    expect do
      post '/oauth/token', params: params, headers: headers
    end.to change { Doorkeeper::AccessToken.count }.by(1)

    token = Doorkeeper::AccessToken.first

    expect(token.application_id).to eq client.id
    expect(token.scopes).to be_empty

    should_have_json 'access_token', token.token
    should_not_have_json 'scope'
 end
end
```

rishabhsairawat:
@nbulaj Currently it works this way only for `client_credentials` flow as changes for `client_credentials` was done in your commit  976b235f468f33a177f1429fc7e7f6345847bc8e. But in my commits for other grant flows , I used other approach (showing error in place of token without any scope).

PR #1098, opened 2018-05-24
Fix ApplicationsController for API-only mode

### LINKED ISSUE #1096: Rails API-only mode and eager_load
### Steps to reproduce
- enable `api_only` in the Doorkeeper initializer
- set `config.eager_load = true` in your environment
- start the application

### Expected behavior
The application starts.

### Actual behavior
Eager loading tries to load `ApplicationsController` which invokes the non-existant `layout` method.

```
[bundle]/ruby/2.5.0/bundler/gems/doorkeeper-d98f6d0894d3/app/controllers/doorkeeper/applications_controller.rb:3:in `<class:ApplicationsController>': undefined method `layout' for Doorkeeper::ApplicationsController:Class (NoMethodError)
	from [bundle]/ruby/2.5.0/bundler/gems/doorkeeper-d98f6d0894d3/app/controllers/doorkeeper/applications_controller.rb:2:in `<module:Doorkeeper>'
	from [bundle]/ruby/2.5.0/bundler/gems/doorkeeper-d98f6d0894d3/app/controllers/doorkeeper/applications_controller.rb:1:in `<main>'
	from [bundle]/ruby/2.5.0/gems/bootsnap-1.3.0/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:100:in `load'
	from [bundle]/ruby/2.5.0/gems/bootsnap-1.3.0/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:100:in `load'
	from [bundle]/ruby/2.5.0/gems/activesupport-5.2.0/lib/active_support/dependencies.rb:468:in `block in load_file'
```

### System configuration
Ruby 2.5.1, Doorkeeper from current `master`. `d98f6d0894d3`.

milgner:
There's no `layout` method when in Rails API mode.

Fixes #1096 .

### DIFF
--- a/app/controllers/doorkeeper/applications_controller.rb
+++ b/app/controllers/doorkeeper/applications_controller.rb
@@ -1,6 +1,6 @@
 module Doorkeeper
   class ApplicationsController < Doorkeeper::ApplicationController
-    layout 'doorkeeper/admin'
+    try(:layout, 'doorkeeper/admin') # there's no layout method in API-only mode
 
     before_action :authenticate_admin!
     before_action :set_application, only: %i[show edit update destroy]


nbulaj:
Hi @milgner . Thanks for your work! Currently this controller [is skipped in api mode](https://github.com/doorkeeper-gem/doorkeeper/blob/master/lib/doorkeeper/rails/routes.rb#L27).

milgner:
Hi @nbulaj . The problem is that even though the controller is never used, the eager loading mechanism of Rails tries to parse the file and with the `layout` statement being class-level code, it will try to immediately parse it and stumble over it. I was quite surprised myself, too, when the application couldn't be deployed in `production` mode anymore even though all my specs were green 😉

nbulaj:
Ok, but I think in this case we must provide JSON API instead of simple Rails controller. So just to `try` - is not a proper solution.

nbulaj:
Hi @milgner . Could you please check my latest commit? I introduced API compatible version of this controllers (it doesn't make sense because this controllers are skipped in the API mode, but why not :) )

milgner:
Perfect! :+1: much appreciated. I'll give it a try immediately and deploy it to a test system...

Issue #1099, opened 2018-05-28
Unnecessary SQL query to get the refresh token, even if it's not enabled in the doorkeeper.rb


Erowlin:
### Steps to reproduce

Api::V1::HelloController: 
```ruby
class Api::V1::HelloController < ActionController::Base
  before_action :doorkeeper_authorize! 

  def hi
    render json: {hello: 'world'}
  end
end
```

Upon requests, rails logs give me this: 
```
Started GET "/api/v1/hello/hi" for 127.0.0.1 at 2018-05-28 18:58:43 +0200
Processing by Api::V1::HelloController#hi as */*
  Doorkeeper::AccessToken Load (0.3ms)  SELECT  "oauth_access_tokens".* FROM "oauth_access_tokens" WHERE "oauth_access_tokens"."token" = $1 LIMIT $2  [["token", "123456789blablablaMyAccessToken"], ["LIMIT", 1]]
  Doorkeeper::AccessToken Load (0.2ms)  SELECT  "oauth_access_tokens".* FROM "oauth_access_tokens" WHERE "oauth_access_tokens"."refresh_token" = $1 LIMIT $2  [["refresh_token", ""], ["LIMIT", 1]]
Completed 200 OK in 4ms (Views: 0.1ms | ActiveRecord: 0.5ms)
```

Note the second SQL `Doorkeeper::AccessToken Load (0.2ms)  SELECT  "oauth_access_tokens".* FROM "oauth_access_tokens" WHERE "oauth_access_tokens"."refresh_token" = $1 LIMIT $2  [["refresh_token", ""], ["LIMIT", 1]]` that I don't get since I specifically disabled refresh tokens. 

### Expected behavior
Only one SQL Query to match the access_token with the one registered in the `oauth_access_tokens` table

### Actual behavior
Another query is made to match the refresh token while it is not needed. 

### System configuration
You can help us to understand your problem if you will share some very
useful information about your project environment (don't forget to
remove any confidential data if it exists).

**Doorkeeper initializer**:
```ruby
Doorkeeper.configure do
  # Change the ORM that doorkeeper will use (needs plugins)
  orm :active_record

  # This block will be called to check whether the resource owner is authenticated or not.
  resource_owner_authenticator do
    sign_out(current_user) unless current_user&.is_a?(Admin)
    session[:user_return_to] = request.fullpath
    current_user&.is_a?(Admin) ? current_user : redirect_to(new_user_session_url)
  end

[...]

PR #1100, opened 2018-05-29
Erowlin/fix default config values and do not trigger double sql queries when refresh token is disabled

Erowlin:
### Summary

This solves #1099 : Whenever authenticating on the server, we tried to revoke a refresh token, even if it was not enabled in the configuration (initializer). 
It removes the unnecessary call to the ORM (or the storage). 

I also updated the `config` builder because it were returning tri-state booleans (`nil`, `true`, `false`) instead of `true` or `false`. Methods finishing with `question mark ?` should never return nil. 

### Other Information

Should I update `NEWS.md`?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,7 +6,10 @@ upgrade guides.
 User-visible changes worth mentioning.
 
 ## master
-
+- [#1099] All the configuration variables in `Doorkeeper.configuration` now
+          always return a non-nil value (`true` or `false`)
+- [#1099] ORM / Query optimization: Do not revoke the refresh token if it is not enabled
+          in `doorkeeper.rb`
 - [#996] Expiration Time Base On Grant Type
 - [#997] Allow PKCE authorization_code flow as specified in RFC7636
 - [#907] Fix lookup for matching tokens in certain edge-cases
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -292,24 +292,32 @@ module Doorkeeper
     attr_reader :api_only
     attr_reader :enforce_content_type
 
+    def api_only
+      @api_only ||= false
+    end
+
+    def enforce_content_type
+      @enforce_content_type ||= false
+    end
+
     def refresh_token_enabled?
-      defined?(@refresh_token_enabled) && @refresh_token_enabled
+      !!(defined?(@refresh_token_enabled) && @refresh_token_enabled)
     end
 
     def pkce_without_secret_enabled?
-      defined?(@pkce_without_secret_enabled) && @pkce_without_secret_enabled
+      !!(defined?(@pkce_without_secret_enabled) && @pkce_without_secret_enabled)
     end
 
     def enforce_configured_scopes?
-      defined?(@enforce_configured_scopes) && @enforce_configured_scopes
+      !!(defined?(@enforce_configured_scopes) && @enforce_configured_scopes)
     end
 
     def enable_application_owner?
-      defined?(@enable_application_owner) && @enable_application_owner
+      !!(defined?(@enable_application_owner) && @enable_application_owner)
     end
 
     def confirm_application_owner?
-      defined?(@confirm_application_owner) && @confirm_application_owner
+      !!(defined?(@confirm_application_owner) && @confirm_application_owner)
     end
 
     def default_scopes
--- a/lib/doorkeeper/oauth/token.rb
+++ b/lib/doorkeeper/oauth/token.rb
@@ -13,7 +13,10 @@ module Doorkeeper
         def authenticate(request, *methods)
           if (token = from_request(request, *methods))
             access_token = AccessToken.by_token(token)
-            access_token.revoke_previous_refresh_token! if access_token
+            refresh_token_enabled = Doorkeeper.configuration.refresh_token_enabled?
+            if access_token.present? && refresh_token_enabled
+              access_token.revoke_previous_refresh_token!
+            end
             access_token
           end
         end
--- a/spec/lib/config_spec.rb
+++ b/spec/lib/config_spec.rb
@@ -132,7 +132,7 @@ describe Doorkeeper, 'configuration' do
 
   describe 'use_refresh_token' do
     it 'is false by default' do
-      expect(subject.refresh_token_enabled?).to be_falsey
+      expect(subject.refresh_token_enabled?).to eq(false)
     end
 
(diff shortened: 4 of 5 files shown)


lib/doorkeeper/config.rb
```diff
@@ -292,24 +292,32 @@ def option(name, options = {})
     attr_reader :api_only
     attr_reader :enforce_content_type
 
+    def api_only
+      @api_only ||= false
+    end
+
+    def enforce_content_type
```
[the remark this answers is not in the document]
Erowlin: This one is fine, it is to set a default value to the `enforce_content_type`. I t is conflicting with the attr_reader. 
How should I fix that / ignore the message?

lib/doorkeeper/config.rb
```diff
@@ -292,24 +292,32 @@ def option(name, options = {})
     attr_reader :api_only
     attr_reader :enforce_content_type
 
+    def api_only
```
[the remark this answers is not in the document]
Erowlin: This one is fine, it is to set a default value to the enforce_content_type. I t is conflicting with the attr_reader.
How should I fix that / ignore the message?

nbulaj:
Thanks for contribution! I will check it in a hour.

> Should I update NEWS.md?

Sure! And then please squash all the commits to a single one. Thanks!

Erowlin:
@nbulaj It's done. Let me know what's next :).

NEWS.md
```diff
@@ -6,7 +6,10 @@ upgrade guides.
 User-visible changes worth mentioning.
 
 ## master
-
+- [#1099] All the configuration variables in `Doorkeeper.configuration` now always return
```
nbulaj: `always return` duplicates :)

lib/doorkeeper/config.rb
```diff
@@ -292,24 +292,32 @@ def option(name, options = {})
...
+      @api_only ||= false
+    end
+
+    def enforce_content_type
+      @enforce_content_type ||= false
+    end
+
     def refresh_token_enabled?
-      defined?(@refresh_token_enabled) && @refresh_token_enabled
+      !!(defined?(@refresh_token_enabled) && @refresh_token_enabled)
```
nbulaj: I think in this case we can return to previous  `!!@refresh_token_enabled` solution, but what is the case? `nil`-values in ruby is always falsey, what is the purpose of this changes?

lib/doorkeeper/oauth/token.rb
```diff
@@ -4,16 +4,19 @@ class Token
       class << self
         def from_request(request, *methods)
           methods.inject(nil) do |_, method|
-            method = self.method(method) if method.is_a?(Symbol)
+            method      = self.method(method) if method.is_a?(Symbol)
```
nbulaj: No need in this indent

lib/doorkeeper/oauth/token.rb
```diff
@@ -4,16 +4,19 @@ class Token
...
             credentials = method.call(request)
             break credentials unless credentials.blank?
           end
         end
 
         def authenticate(request, *methods)
           if (token = from_request(request, *methods))
-            access_token = AccessToken.by_token(token)
-            access_token.revoke_previous_refresh_token! if access_token
+            access_token          = AccessToken.by_token(token)
```
nbulaj: Also no need in indenttion

lib/doorkeeper/config.rb
Erowlin: The point is to avoid tri-state booleans as described here: https://robots.thoughtbot.com/avoid-the-threestate-boolean-problem. (The article is about DB, but I apply the same rule for Ruby code. 
Also as a rule of thumb, all question-marks methods should return either `true` or `false`. Very opinionated though, I can revert it if you're not comfortable :).

lib/doorkeeper/oauth/token.rb
Erowlin: This has been reverted already.

lib/doorkeeper/oauth/token.rb
Erowlin: Reverted as well.

lib/doorkeeper/config.rb
nbulaj: No no, I think it's a reasonable one. So maybe we can just return `!!@var` syntax for the options?

Erowlin: I'm not sure I get what you are suggesting. 

Can you provide an example?

nbulaj: Previously it was just

```ruby
!!@refresh_token_enabled
```

instead of

```ruby
defined?(@refresh_token_enabled) && @refresh_token_enabled
```

nbulaj: A little bit another way :)

https://github.com/doorkeeper-gem/doorkeeper/blob/ed95d941e5846ed5b8ef90ff2f6bac96f83e2fec/lib/doorkeeper/config.rb#L284-L285

So maybe we can save the code from your commit

Erowlin: Ok I get what you meant.

Erowlin: Ok I get what you meant. The issue with this approach is that the Doorkeeper.configuration object returned will contains the wrong value, it can be misleading to have the getter methods returning false and the Configuration object having `nil`.

lib/doorkeeper/config.rb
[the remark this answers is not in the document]
nbulaj: So maybe we need to change `attr_reader` to `attr_writer` because we already have a reader method?

lib/doorkeeper/config.rb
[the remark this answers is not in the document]
nbulaj: The same

lib/doorkeeper/config.rb
nbulaj: I think we can save you changes

Erowlin: So, do I have something to change or not? 
Sorry, I'm super confused on what you need me to do, and it's also the end of the day in France ;)

Erowlin:
I squashed the commits, I think it's fine now.

Erowlin:
Quick question, if I update my gem on my project, does it contains this fix yet?

nbulaj:
Only if it is set up from git. Stable release on RubyGems will be later.

Erowlin:
Ok, thank you!

PR #1102, opened 2018-05-31
Add custom expiration time based on scopes

thatsmydoing:
### Summary

This allows customizing the expiration time based on the requested scopes. This works much in the same way as https://github.com/doorkeeper-gem/doorkeeper/pull/1049

### Other Information

I personally think `custom_access_token_expires_in` is starting to get quite unwieldy with all the parameters. The method definition itself is getting quite long. Is this okay as it is now? Or would something like bundling everything into an object as a single parameter be something you'd like to happen?

Also, I'm not exactly sure how to write tests for this change as well. I'd like some direction on that.


nbulaj:
Hi @thatsmydoing . Thanks for your work! For sure it can be useful. I also think that this option (and not only it) becomes to bloated and it's definitely requires refactoring. We can extract it to a class, yep.

What about specs - you can check the sample in [the relevant PR](https://github.com/doorkeeper-gem/doorkeeper/commit/62fd33b232ddfc868ca70d6bbce9ceb68e8b754e)

thatsmydoing:
Alright, I'll try to extract it out into a class then. For testing, I'd like to know how to mock in the scopes since that's not quite obvious to me.

thatsmydoing:
I think this should be good. Can someone look over it?

thatsmydoing:
I've squashed the commits and added a news entry. I think this would also need an update in the migration guide as people will need to update their `custom_access_token_expires_in` in the config.

I'm not sure why coverage is failing since I did not change any code at all aside from squashing.

nbulaj:
>  I think this would also need an update in the migration guide as people will need to update their custom_access_token_expires_in in the config.

Completely agree. We can do it right now because we bump a new major release. Do you want to edit [https://github.com/doorkeeper-gem/doorkeeper/wiki/Migration-from-old-versions](Wiki page] yourself or I'll do it?

Issue #646, opened 2015-04-28
use_refresh_token customized


kiomac:
It should be possible to use or not use the refresh token accordingly.

Use case: In a service-to-service basis the client could use a JWT Assertion to request the access token. This token can have 1 hour TTL, with no refresh token. Meaning that he would need to use JWT assertion each time the token expires. At the same time, other clients using credentials may have a refresh token available.

Will it be ok to do something like `custom_access_token_expires_in` with a block call? Say, `custom_use_refresh_token`.

tute:
I'm open to that. Different than the existing case, I wouldn't add the `custom_` prefix; I would check in code if the object responds to `#call` and if so call it, otherwise use it directly.

thornbird:
@kiomac You raised this because you want to support concurrency of requesting refresh token?
E.g. multiple requests, with one same expired token, sent at the same time.

There will be case that server generated new access token already, but still, some requests are using old token?

Thanks!

tute:
https://github.com/doorkeeper-gem/doorkeeper/pull/769 is related with this. I think there are no further actionable items in this issue, so I'll close it. We can reopen whenever needed. Thanks!

thatsmydoing:
Hi, I'm interested in the feature originally proposed in this issue. Being able to customize whether or not to provide a refresh token given some conditions. I took a look at #769 and it wasn't entirely obvious to me how to use that for this case. Is it possible with the current doorkeeper?

PR #1103, opened 2018-06-07
Implement custom use_refresh_token

thatsmydoing:
### Summary

This is meant to address #646, it lets you customize when to give a refresh token or not. This uses the same `context` as `custom_access_token_expires_in` that was added in #1102

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,7 @@ upgrade guides.
 User-visible changes worth mentioning.
 
 ## master
+- [#1103] Allow customizing use_refresh_token
 - [#1089] Removed enable_pkce_without_secret configuration option
 - [#1102] Expiration time based on scopes
 - [#1099] All the configuration variables in `Doorkeeper.configuration` now
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -103,9 +103,12 @@ module Doorkeeper
         @config.instance_variable_set(:@access_token_methods, methods)
       end
 
-      # Issue access tokens with refresh token (disabled by default)
-      def use_refresh_token
-        @config.instance_variable_set(:@refresh_token_enabled, true)
+      # Issue access tokens with refresh token (disabled if not set)
+      def use_refresh_token(enabled = true, &block)
+        @config.instance_variable_set(
+          :@refresh_token_enabled,
+          block ? block : enabled
+        )
       end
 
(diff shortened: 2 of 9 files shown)


lib/doorkeeper/config.rb
```diff
@@ -115,9 +115,13 @@ def enable_pkce_without_secret
         @config.instance_variable_set(:@pkce_without_secret_enabled, true)
       end
 
-      # Issue access tokens with refresh token (disabled by default)
-      def use_refresh_token
-        @config.instance_variable_set(:@refresh_token_enabled, true)
+      # Issue access tokens with refresh token (disabled if not set)
+      def use_refresh_token(enabled = true, &custom)
```
thatsmydoing: I added the `enabled = true` parameter here in case people do `use_refresh_token false`. Since this takes a block parameter, I thought to make it a bit consistent.

NEWS.md
```diff
@@ -6,6 +6,7 @@ upgrade guides.
 User-visible changes worth mentioning.
 
 ## master
+- [#646] Allow customizing use_refresh_token
```
nbulaj: Fix it to #1103 please, it addresses the real PR that made changes

lib/doorkeeper/oauth/authorization/token.rb
```diff
@@ -5,30 +5,37 @@ class Token
...
-              expiration
-            else
-              server.access_token_expires_in
-            end
-          end
-
-          private
-
-          def custom_expiration(server, pre_auth_or_oauth_client, grant_type, scopes)
+          def make_context(pre_auth_or_oauth_client, grant_type, scopes)
```
nbulaj: Could you please rename it to `build_context` ?

lib/doorkeeper/config.rb
nbulaj: I prefer `&block` instead of `&custom` as a common practice, what do you think?

Also check out `force_ssl_in_redirect_uri` option and [how it works with block or boolean variables](https://github.com/doorkeeper-gem/doorkeeper/blob/master/app/validators/redirect_uri_validator.rb#L36-L43).

thatsmydoing:
Is this better?

nbulaj:
Yep, much better. Could you please resolve the conflicts?

nbulaj:
Thank you very much :+1:

PR #1108, opened 2018-06-15
Better format callback URLs on index page

rmm5t:
### Summary

Just a visual improvement to show each callback URL on it's own line per application

### Before

unnamed-3

### After

unnamed-4

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,7 @@ User-visible changes worth mentioning.
 
 - [#1106] Restrict access to AdminController with 'Forbidden 403' if admin_authenticator is not
   configured by developers..
+- [#1108] Simple formating of callback URLs when listing oauth applications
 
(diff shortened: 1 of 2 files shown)


nbulaj:
Hi @rmm5t . Could you please provide some screens over how it was look before and after? Currently I have no chance to see it myself, sorry. Also could you please add a changelog entry to NEWS.md and then squash commits to a single one? Something like "_[#pr number] Simple formating of callback URLs_". Also maybe it is required to do the same for detailed page (SHOW action) ?

rmm5t:
Hi @nbulaj. Sure thing. All done. I added the screenshots to the description of this PR above, and I added a line to the NEWS.

Thanks!

nbulaj:
Thanks @rmm5t  for this changes! :+1:

PR #1116, opened 2018-06-29
Jw revoke patch

jackvnimble:
### Summary

Hey folks,

Someone recently submitted an issue to my team about a security issue surrounding Doorkeeper access grants. Seems small but probably worth fixing.

The issue is that post-revocation of an access token, any access grants outstanding are still valid (until expiration which is default 10 mins). Since this code can be exchanged for an access token it should also be revoked with the other access tokens.

Notes are in the respective commits but to summarize changes:
- Update specs for `AccessToken.revoke_all_for` to ensure scope is respected
- Add `AccessGrant.revoke_all_for`
- Add `Application.revoke_all_for` which in turn calls `revoke_all_for` on both `AccessGrant` and `AccessToken`
- Call `Application.revoke_all_for` in `AuthorizedApplicationsController#destroy` action

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,8 @@ User-visible changes worth mentioning.
 - [#1106] Restrict access to AdminController with 'Forbidden 403' if admin_authenticator is not
   configured by developers..
 - [#1108] Simple formating of callback URLs when listing oauth applications
+- [#1116] `AccessGrant`s will now be revoked along with `AccessToken`s when
+  hitting the `AuthorizedApplicationController#destroy` route.
 
 ## 5.0.0.rc1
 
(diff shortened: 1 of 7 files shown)


nbulaj:
Hi @jackvnimble . Looks good to me except the method name for `Application` model. It sounds like "revoke application" and not like "revoke tokens and grants". Could we rename it to something more meaningfull? Also could you please add an entry to NEWS.md and squash all the commits to a single one? Thanks!

jackvnimble:
Done. Let me know if you'd like any other changes!

nbulaj:
Thanks @jackvnimble  for your work! :+1:

PR #1119, opened 2018-07-10
Use Application#confidential? to determine revocation auth eligibility

### LINKED ISSUE #891: Revoking token does not work for public clients
Hello,
I'm using grant `implicit` to allow authorizing an application from mobile applications, by initiating the flow as follows:

`https://example.com/oauth2/authorize?response_type=token&client_id={CLIENT_ID}&redirect_uri={REDIRECT_URI}&scope={REQUESTED_SCOPES}&state={STATE}
`

After a successful oauth flow, the generated token is saved in the `oauth_access_tokens` table jointly with the application referenced by the original `client_id` value:

`#<Doorkeeper::AccessToken id: 396, resource_owner_id: 2, application_id: 1, token: "49d9a5c6232d6880e9b8df1c9369edf6bfc00aceedd5bb8164...", refresh_token: nil, expires_in: 1209600, revoked_at: nil, created_at: "2016-10-05 12:37:11", scopes: "profile">
`

The `oauth_access_grants` table is empty, as expected since using `implicit`.

When I try to revoke this token using the [POST /oauth/revoke API endpoint](https://github.com/doorkeeper-gem/doorkeeper/wiki/API-endpoint-descriptions-and-examples#post-oauthrevoke) (the one at the bottom, which does not require `client_id` or `client_secret` to be passed), it does not get revoked.

Diving into the source code, I see in the `tokens_controller.rb`:

```
# OAuth 2.0 Section 2.1 defines two client types, "public" & "confidential".
# Public clients (as per RFC 7009) do not require authentication whereas
# confidential clients must be authenticated for their token revocation.
#
# Once a confidential client is authenticated, it must be authorized to
# revoke the provided access or refresh token. This ensures one client
# cannot revoke another's tokens.
#
# Doorkeeper determines the client type implicitly via the presence of the
# OAuth client associated with a given access or refresh token. Since public
# clients authenticate the resource owner via "password" or "implicit" grant
# types, they set the application_id as null (since the claim cannot be
# verified).
#
# https://tools.ietf.org/html/rfc6749#section-2.1
# https://tools.ietf.org/html/rfc7009
```

However, as per above the token _does_ specify the `application_id`, despite having used the `implicit` grant type.

Hence, when I try to revoke this token I land in the controller's [`authorized?` method](https://github.com/doorkeeper-gem/doorkeeper/blob/v4.2.0/app/controllers/doorkeeper/tokens_controller.rb#L52):

``` ruby
def authorized?
  if token.present?
    # Client is confidential, therefore client authentication & authorization
    # is required
    if token.application_id?
      # We authorize client by checking token's application
      server.client && server.client.application == token.application
    else
      # Client is public, authentication unnecessary
      true
    end
  end
end
```

Where `token` is present and specifies an `application_id`, however `server.client` is `nil` hence `authorized?` returns false and the token does not get revoked.

I'm using doorkeeper `4.2.0`.

Is this expected behavior or am I missing something?

Thank you.

f3ndot:
OAuth applications that obtain an access token using the "implicit" grant flow will have their ID set on the token record. Unfortunately this causes the revocation controller code to think it's as confidential application. Because of this, Doorkeeper enforces oauth client authentication and the revocation call fails.

Fixes #891

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -13,6 +13,7 @@ User-visible changes worth mentioning.
 - [#1116] `AccessGrant`s will now be revoked along with `AccessToken`s when
   hitting the `AuthorizedApplicationController#destroy` route.
 - [#1114] Make token info endpoint's attributes consistent with token creation
+- [#1119] Fix token revocation for OAuth apps using "implicit" grant flow
 
 ## 5.0.0.rc1
 
--- a/app/controllers/doorkeeper/tokens_controller.rb
+++ b/app/controllers/doorkeeper/tokens_controller.rb
@@ -56,16 +56,15 @@ module Doorkeeper
     # https://tools.ietf.org/html/rfc6749#section-2.1
     # https://tools.ietf.org/html/rfc7009
     def authorized?
-      if token.present?
-        # Client is confidential, therefore client authentication & authorization
-        # is required
-        if token.application_id?
-          # We authorize client by checking token's application
-          server.client && server.client.application == token.application
-        else
-          # Client is public, authentication unnecessary
-          true
-        end
+      return unless token.present?
+      # Client is confidential, therefore client authentication & authorization
+      # is required
+      if token.application_id? && token.application.confidential?
+        # We authorize client by checking token's application
+        server.client && server.client.application == token.application
+      else
+        # Client is public, authentication unnecessary
+        true
       end
     end
 
--- a/spec/controllers/tokens_controller_spec.rb
+++ b/spec/controllers/tokens_controller_spec.rb
@@ -56,15 +56,67 @@ describe Doorkeeper::TokensController do
     end
   end
 
-  describe 'when revoke authorization has failed' do
-    # http://tools.ietf.org/html/rfc7009#section-2.2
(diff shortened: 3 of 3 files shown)


f3ndot:
Since this is security related, is it worth backporting to 4.x? Would require #1031 to also be backported.

f3ndot:
@nbulaj this can be merged for 5.x fix.

Semver is getting in the way here, because I agree that we cannot "backport" to 4.x without requiring an upgrade path. #1031 defaults all apps as confidential, so it maintains backwards compatibility however developers must:

1. Run a migration to add the column.
2. Decide for themselves which apps are public and update them accordingly

I'm not familiar with the rules around stuff like this, but can't we autogenerate some "hey you're using 4.2.0-secfix1, please run `rails generate doorkeeper:install_client_confidentiality`" which would simply add the column? That's essentially a MINOR-level sort of change. Then they opt into using it by toggling which apps are public?

🤷‍♂️

nbulaj:
No problem for the 5.x release, yes,. What about 4.x - I'll think on it tomorow

PR #1120, opened 2018-07-11
[4.x] Backport CVE-2018-1000211 (no token revocation for public apps)

f3ndot:
### Summary

This aims to solve the security vulnerability outlined in #891 for the 4.x series without breaking changes.

I have cherry-picked the original code for 5.x that allows apps to be defined as either confidential or public. On top of that I've added migration generators, instructions, and warnings for developers on how to make use of the security fix.

If they upgrade and ignore, nothing breaks 👍 

### Other Information

I would love your input @brianp @ostinelli

### The Warnings

Post install:

```



  WARNING: This is a security release that addresses token revocation not working for public apps (CVE-2018-1000211)

  There is no breaking change in this release, however to take advantage of the security fix you must:

    1. Run `rails generate doorkeeper:add_client_confidentiality` for the migration
    2. Review your OAuth apps and determine which ones exclusively use public grant flows (eg implicit)
    3. Update their `confidential` column to `false` for those public apps

  This is a backported security release.

  For more information:

    * https://github.com/doorkeeper-gem/doorkeeper/pull/1119
    * https://github.com/doorkeeper-gem/doorkeeper/issues/891



```

Every time Doorkeeper::Application.confidential? is called without having added the neccessary column to fix the security issue:

```
2.5.1 :003 > Doorkeeper::Application.last.confidential?
DEPRECATION WARNING: You are susceptible to CVE-2018-1000211. Please follow instructions outlined in Doorkeeper::CVE_2018_1000211_WARNING (called from irb_binding at (irb):3)
 => true
```

### DIFF
--- a/lib/doorkeeper/orm/active_record/application.rb
+++ b/lib/doorkeeper/orm/active_record/application.rb
@@ -35,7 +35,7 @@ module Doorkeeper
     # Fallback to existing, default behaviour of assuming all apps to be
     # confidential if the migration hasn't been run
     def confidential
-      return super if self.class.column_names.include?('confidential')
+      return super if self.class.supports_confidentiality?
       ActiveSupport::Deprecation.warn 'You are susceptible to security bug ' \
         'CVE-2018-1000211. Please follow instructions outlined in ' \
         'Doorkeeper::CVE_2018_1000211_WARNING'
@@ -43,6 +43,10 @@ module Doorkeeper
     end
 
     alias_method :confidential?, :confidential
+
+    def self.supports_confidentiality?
+      column_names.include?('confidential')
+    end
 
     private
 
--- a/lib/doorkeeper/orm/active_record/application.rb
+++ b/lib/doorkeeper/orm/active_record/application.rb
@@ -35,7 +35,7 @@ module Doorkeeper
     # Fallback to existing, default behaviour of assuming all apps to be
     # confidential if the migration hasn't been run
     def confidential
-      return super if self.class.column_names.include?('confidentialz')
+      return super if self.class.column_names.include?('confidential')
       ActiveSupport::Deprecation.warn 'You are susceptible to security bug ' \
         'CVE-2018-1000211. Please follow instructions outlined in ' \
         'Doorkeeper::CVE_2018_1000211_WARNING'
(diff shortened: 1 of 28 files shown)


lib/generators/doorkeeper/templates/migration.rb.erb
```diff
@@ -6,6 +6,7 @@ class CreateDoorkeeperTables < ActiveRecord::Migration<%= migration_version %>
       t.string  :secret,       null: false
       t.text    :redirect_uri, null: false
       t.string  :scopes,       null: false, default: ''
+      t.boolean :confidential, null: false, default: true
```
f3ndot: Not sure if this should be removed, or remain 🤷‍♂️

nbulaj:
Use https://github.com/doorkeeper-gem/doorkeeper/tree/4.4.0

nbulaj:
Hi @f3ndot . This branch https://github.com/doorkeeper-gem/doorkeeper/tree/4.4.0 created from 4.3.2 version of Doorkeeper and you can use it.

brianp:
I'd be happy to check it out.
I'll pull it down and run through it tonight/tomorrow morning.

lib/doorkeeper/orm/active_record/application.rb
```diff
@@ -31,6 +32,18 @@ def self.authorized_for(resource_owner)
       where(id: resource_access_tokens.select(:application_id).distinct)
     end
 
+    # Fallback to existing, default behaviour of assuming all apps to be
+    # confidential if the migration hasn't been run
+    def confidential
+      return super if self.class.column_names.include?('confidentialz')
```
brianp: Is this a typo?

brianp:
I'm still seeing a bug where the column value for `confidential` is clearly `false`. The instance value shows `false` but the overloaded method `confidential` returns `true` because of the line `return super if self.class.column_names.include?('confidentialz')`

As my column name is `confidential`.

With the typo fixed things seem to work as expected.

lib/doorkeeper/orm/active_record/application.rb
f3ndot: Yeah. I goofed up there. Resolved it and so tests should pass

nbulaj:
Hi @f3ndot @brianp . Thanks for your work! Is it everything we need to release 4.4 version with security fixes?

f3ndot:
@nbulaj I believe it is ready for 4.4 release. You should also release a 5.0.0.rc2 at the same time. After that I'll write security bulletins for the CVEs

nbulaj:
Will do tomorrow, thanks!

nbulaj:
Hi @f3ndot . Released 4.4.0 and 5.0.0.rc2, thanks! I'll do GitHub releases also

Issue #1121, opened 2018-07-13
Authorization error response is not valid JSON when api_only


f3ndot:
### Steps to reproduce

1. Use Doorkeeper's `master` (presently `v5.0.0.rc1`)
2. Set initializer to use `api_only`
3. Make a GET request to `/oauth/authorize?client_id=badclient`

### Expected behavior

Response should be something like

```js
{
  "error": "The authorization server does not support this response type.",
  // potentially more metadata:
  "error_type": "unsupported_response_type",
  "status": 401
}
```

### Actual behavior

We get just raw text back, despite the response header `Content-Type: application/json`:

```
The authorization server does not support this response type.
```

![screen shot 2018-07-13 at 12 41 13 pm](https://user-images.githubusercontent.com/740289/42703722-61288e2c-869b-11e8-92de-9c018a71e7f7.png)

Here is a well-formed JSON response from a successful request:

![screen shot 2018-07-13 at 12 40 41 pm](https://user-images.githubusercontent.com/740289/42703744-73d10392-869b-11e8-80fc-35c8c5f8850d.png)

The issue appears to be https://github.com/doorkeeper-gem/doorkeeper/blob/bf36149a82693c92d7a3b4c3a3628ae3d4094dd1/app/controllers/doorkeeper/authorizations_controller.rb#L34-L41 where there is no root key or JSON object established. Just a string.

### Proposed solution

Prefix the error with a `error` key. Not sure if there is already an existing pattern elsewhere. If so, we should adopt that pattern.

PR #1122, opened 2018-07-13
[API mode] Respond with JSON-formatted error for Authorizations #new action

f3ndot:
See #1121


nbulaj:
This one is nice, thanx!

Issue #1110, opened 2018-06-20
Doorkeeper application controllers and Rails 5.2 protect_from_forgery default behavior


enrico:
As of Rails 5.2 ```protect_from_forgery``` is set by default on ```ActionController::Base``` (rather than protecting from forgery in the generated ApplicationController) and a new method ```skip_forgery_protection``` is introduced to skip this default protection setting. 

See this PR: https://github.com/rails/rails/pull/29742

currently Doorkeeper ```ApplicationController``` invokes ```protect_from_forgery``` no matter what:

https://github.com/doorkeeper-gem/doorkeeper/blob/8bd96739d587cd6ecafe31b82a4a57cb925f0f5e/app/controllers/doorkeeper/application_controller.rb#L8

For rails 5.2, Doorkeeper should examine the the value of ```config.action_controller.default_protect_from_forgery``` and decide whether to request protection or not based on it.

nbulaj:
Hi @enrico .Good point! Would you like to produce a PR to fix this?

enrico:
@nbulaj , I'm leaving town for a week, so I won't have time to do this right away. 

When I started thinking about how to do this, I realized we would also have to consider  ```Doorkeeper.configuration.base_controller``` 's value when performing the check (i.e. if a user has overridden the value it's likely they wanted their own control/granularity with respect to enforcing or not the forgery check)

So now I'm starting to think Doorkeeper should not call ```protect_from_forgery``` at all and instead allow the containing app to enforce whichever policy they have chosen from the inheritance hierarchy.  

So if a developer leaves ```Doorkeeper.configuration.base_controller```'s value to its default, rails checks (or not depending on ```config.action_controller.default_protect_from_forgery```). and it they set it to a separate controller, the developer is free to enforce it there...

What do you think?

nbulaj:
I think we need to be opionated enough, because it is a question of the security. If user would disable forgery protection - OAuth2 endpoints could be compromised. We can't allow this. So if developer disabled forgery protection for his controllers - we must validate requests ourselves..

nbulaj:
Hi @enrico. Any thoughts on this issue?

enrico:
Well, enabling ```protect_from_forgery``` based on the value of 
```Rails.application.config.action_controller.default_protect_from_forgery``` would not cover 100% of the cases since a developer can create his/her own base class to inherit from. 

[...]

PR #1123, opened 2018-07-17
Add note about Rails CSRF protections

arsduo:
It would be helpful if the documentation included a reference to Doorkeeper providing CSRF protection by default -- I just spent a while debugging why my client was failing with CSRF protections even though my app was configured not to use it; hopefully having the term in the readme will help others as well.

Thanks for the great gem!

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,8 +179,7 @@ end
 ```
 
 Keep in mind, that in this mode you will not be able to access `Applications` or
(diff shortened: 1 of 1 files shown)


nbulaj:
Hi @arsduo . Do you think we can do more on this? #1110

nbulaj:
Currently merging this one, but feel free to help with #1110

PR #1127, opened 2018-07-27
 Change the token_type initials of the Banner Token to uppercase

borisding1994:
### Summary

Change the token_type initials of the Banner Token to uppercase to comply with the RFC6750 specification. 

The current version of the doorkeeper has compatibility issues with [AWS ALB's OIDC functionality](https://aws.amazon.com/blogs/aws/built-in-authentication-in-alb/). The reason this failure occurred is that the load balancer was expecting the access token type as "token_type":"Bearer" (uppercase B) as per the OpenID Connect Specification mentioned below: 

> "The OAuth 2.0 token_type response parameter value MUST be Bearer, as specified in OAuth 2.0 Bearer Token Usage [RFC6750](https://tools.ietf.org/html/rfc6750
) , unless another Token Type has been negotiated with the Client. Servers SHOULD support the Bearer Token Type; use of other Token Types is outside the scope of this specification."

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,7 +7,7 @@ User-visible changes worth mentioning.
 
 ## master
 
-- [#] Add description here
+- [#1127] Change the token_type initials of the Banner Token to uppercase to comply with the RFC6750 specification.
 
 ## 5.0.0.rc2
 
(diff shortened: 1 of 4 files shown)


nbulaj:
This one is nice! Could you please also squash you commits to a single one? :+1:

Issue #1143, opened 2018-09-13
 breaking api change in versions greater than v4.2.6


pakwfoley:
### The Issue

There is a breaking api change in versions greater than `v4.2.6`. This is in the native redirect uri authorization code flow. Up until then the code response route, after resource owner authentication, was `/oauth/authorize/<code>` . Now the response is `oauth/authorize/native?code=<code>`. This breaks the native oauth flows for clients that previously parsed the response from `/oauth/authorize/<code>`. This is what has kept us from upgrading doorkeeper. But now that we have to upgrade to `4.4.2` to get the backport for the implicit token revocation fix https://github.com/doorkeeper-gem/doorkeeper/pull/1120, this has become a more pressing issue. It would be nice to have a initializer option that either lets us opt out of the breaking api change that exists in the minor version bump from `v4.2.6` - `v4.4.2` or add the option to customize this route. I would be happy to put in a PR for either option if they seem reasonable.

### Expected behavior
In a minor version bump I would not expect a breaking api change, I would expect the authorization code response for a native auth code flow to be: `/oauth/authorize/<code>`.

### Actual behavior
The response is `oauth/authorize/native?code=<code>`.

nbulaj:
Hi @pakwfoley . Yes, you are absolutely right. This change was made to help applications that automatically authorizes ([take a look at original PR](https://github.com/doorkeeper-gem/doorkeeper/pull/1003)). I think we need to revert this for 4.x release, but save it for 5.x with notice in upgrade guides and changelog.

pakwfoley:
@nbulaj Thank you for the quick response! The only concern I have for reverting would be for those applications who have already bumped past 4.2.6 and asked clients to move over to the new api. I know it’s not easy to ask clients to move to a new api, and it’s especially not easy to ask them to revert after changing. Would it be unreasonable to add an option to opt out of the api change in the 4.x release, and remove that option for 5.x?

nbulaj:
@pakwfoley I think you are right. It would be great if you can send a PR with such an option

nbulaj:
Released with 4.4.3

nbulaj:
Hi @pakwfoley . Can we close this issue ?

pakwfoley:
Released with 4.4.3

starsolutions:
Hi @nbulaj @pakwfoley ,

[...]

PR #1144, opened 2018-09-14
Make breaking api change to native authorization code route optional

pakwfoley:
### Summary

This PR creates the option to opt out of the breaking api change introduced in https://github.com/doorkeeper-gem/doorkeeper/pull/1003 . For an authorization code flow with a native redirect uri the default response route after resource owner authentication is still `oauth/authorize/native?code=<code>`. However if you add `opt_out_native_route_change` to your `config/initializers/doorkeeper.rb` the route will be `oauth/authorize/<code>` as it was in versions < `v4.3.0`.

### Other Information

This PR stems from https://github.com/doorkeeper-gem/doorkeeper/issues/1143

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,11 @@ User-visible changes worth mentioning.
 
 ## master
 
+## 4.4.3
+- [#1143] Adds a config option opt_out_native_route_change to opt out of the
+  breaking api changed introduced in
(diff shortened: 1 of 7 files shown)


nbulaj:
Hi @pakwfoley . Could you please squash the commits to a single one? Thanks

Issue #882, opened 2016-09-17
Why is Doorkeeper returning expired authorization codes?


theycallmeswift:
I've got an OAuth API using Doorkeeper 3.1.0.  For a small number of users (~5%) when they request an authorization code, the API is returning an expired or revoked code.

For example, here's a line from today's server logs (2016-09-17):

```
2016-09-17T19:36:30.032406+00:00 heroku[router]: at=info method=GET path="/callback?code=e2a96d3d24621b22d11f5feae2b2234630a1785cc61f6e7391579e203d74c3b1"
```

When I look at that token in the database, I see that it was created yesterday and revoked yesterday as well:

``` 
#<Doorkeeper::AccessGrant id: 57260, resource_owner_id: 22196, application_id: 174, token: "e2a96d3d24621b22d11f5feae2b2234630a1785cc61f6e7391...", expires_in: 600, redirect_uri: "https://example.com/callback", created_at: "2016-09-16 23:53:05", revoked_at: "2016-09-16 23:53:06", scopes: "profile contact events">
```

Why would doorkeeper send users an expired or revoked token?

heavysixer:
We are seeing the same problem.

theycallmeswift:
Is there a reason that method isn't using the `find_or_create_for` method right above it? https://github.com/doorkeeper-gem/doorkeeper/blob/ddcf2edd6045f2d54935865aed0a668b7fd03bcf/lib/doorkeeper/models/access_token_mixin.rb#L71

heavysixer:
@theycallmeswift it actually starts there and then eventually winds up at the method i pointed  out I believe.

theycallmeswift:
@heavysixer are you seeing this with access tokens or auth codes?  We're seeing it for auth codes which uses a different mixin.

heavysixer:
@theycallmeswift Doorkeeper is returning expired `oauth_access_tokens` from that table. I think that means that they are access_tokens.

theycallmeswift:
@heavysixer yeah, sounds like we're encountering different issues then.  My issue is related to Authorization Codes being expired.  May be connected somehow though.

nbulaj:
Hi @theycallmeswift . Have you try to update Doorkeeper to a new version? 

[...]

PR #1148, opened 2018-09-21
[#1138] Revert matching token changes for Authorization

nbulaj:
Revert changes introduced in c87b13c363073d4a06bada2ee2efd979b6329090


nbulaj:
Bug from #882 is not reproducible

PR #1152, opened 2018-10-02
[migration templates fix] migration generator template, resource_owner integer to references

atomiyama:
### Summary

In over Rails5.1, Default primary key has been bigint.
If Rails version is below 5.1, resource_owner_id column will be set INT, above 5.1 it will be set BIGINT

### DIFF
--- a/lib/generators/doorkeeper/templates/migration.rb.erb
+++ b/lib/generators/doorkeeper/templates/migration.rb.erb
@@ -13,7 +13,7 @@ class CreateDoorkeeperTables < ActiveRecord::Migration<%= migration_version %>
     add_index :oauth_applications, :uid, unique: true
 
     create_table :oauth_access_grants do |t|
-      t.integer  :resource_owner_id, null: false
+      t.references :resource_owner,  null: false
       t.references :application,     null: false
       t.string   :token,             null: false
       t.integer  :expires_in,        null: false
@@ -31,7 +31,7 @@ class CreateDoorkeeperTables < ActiveRecord::Migration<%= migration_version %>
     )
 
     create_table :oauth_access_tokens do |t|
-      t.integer  :resource_owner_id
+      t.references :resource_owner
       t.references :application
 
       # If you use a custom token generator you may need to change this column
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,6 +12,7 @@ User-visible changes worth mentioning.
   triggers every time)
 - [#1149] Fix for `URIChecker#valid_for_authorization?` false negative when query is blank, but `?` present.
 - [#1151] Fix Refresh Token strategy: add proper validation of client credentials both for Public & Private clients.
+- [#1152] Fixmigration template: change resource owner data type from integer to references
 
 ## 5.0.0
 


lib/generators/doorkeeper/templates/migration.rb.erb
```diff
@@ -10,9 +10,9 @@ class CreateDoorkeeperTables < ActiveRecord::Migration<%= migration_version %>
       t.timestamps             null: false
     end
 
-    add_index :oauth_applications, :uid, unique: true
+    add_index :oauth_applications, :uid, unique: true, <%= primary_id_option %>
```
nbulaj: Why do we need to add this option for `add_index` ?

lib/generators/doorkeeper/templates/migration.rb.erb
```diff
@@ -1,6 +1,6 @@
 class CreateDoorkeeperTables < ActiveRecord::Migration<%= migration_version %>
   def change
-    create_table :oauth_applications do |t|
+    create_table :oauth_applications, <%= primary_id_option %> do |t|
```
nbulaj: Could not find in any of my Rails 5.2 projects this option, can you please share a link to PR or any article describing this?

Just run `rails g migration CreateUsers name:string` and got:

```ruby
class CreateUsers < ActiveRecord::Migration[5.2]
  def change
    create_table :users do |t|
      t.string :name
    end
  end
end
```

lib/generators/doorkeeper/templates/migration.rb.erb
```diff
@@ -10,9 +10,9 @@ class CreateDoorkeeperTables < ActiveRecord::Migration<%= migration_version %>
       t.timestamps             null: false
     end
 
-    add_index :oauth_applications, :uid, unique: true
+    add_index :oauth_applications, :uid, unique: true, <%= primary_id_option %>
 
-    create_table :oauth_access_grants do |t|
+    create_table :oauth_access_grants, <%= primary_id_option %> do |t|
```
nbulaj: I think `,` must be included in the template only if primary ID option is not null...

nbulaj:
Hi @atomiyama . As far as I know Rails generates bigint by default. And [this article](http://www.mccartie.com/2016/12/05/rails-5.1.html) just approves my thoughts. Fix me if I wrong..

lib/generators/doorkeeper/templates/migration.rb.erb
atomiyama: it is not needed, will fixed.

lib/generators/doorkeeper/templates/migration.rb.erb
atomiyama: included

atomiyama:
At rails 5.2, if you want to use User table as the resource owner and you add foreign keys to your migration, your "db:migrate" cause data type unmatch error, because User table's primary key would be bigint by default.

nbulaj:
@atomiyama is it something specific for MySQL or how? Never faced this problem.

Also not so much results found with this use case https://github.com/search?q=ActiveRecord%3A%3AMigration+create_table+id%3A+%3Abigint+language%3Aruby&type=Code

atomiyama:
create_table is not needed to be passed id option. it has been deleted.
I faced this migration problem data type unmatch in Rails5.2 w/ mysql

nbulaj:
Oh, now I see it from you last changes :) Makes sense

But I prefer to use `t.belongs_to :resource_owner` (or `t.references :resource_owner`) because it will automatically use bigint for Rails >= 5.1 and `integer` for previous releases

atomiyama:
it makes sense! this PR should be close. thank you for your reviews :)

nbulaj:
You @atomiyama can produce a new PR (or fix this one) to use `t.belongs_to` instead of plain `t.integer`, it would be great! Thanks

atomiyama:
@nbulaj thanks for your kind. it has been fixed.

atomiyama:
@nbulaj done! :)

PR #1162, opened 2018-11-02
Fix #1161: check content type of PUT/PATCH/POST

### LINKED ISSUE #1161: enforce_content_type causes GET requests to return 415 status code
### Steps to reproduce
- Enable `enforce_content_type` in `doorkeeper.rb`
- Attempt GET on `oauth/token/info` eg:
```
curl -X GET \
  http://localhost:3000/oauth/token/info \
  -H 'Authorization: Bearer 123'
```

### Expected behavior
The token information should be retrieved.

### Actual behavior
Responds with 415 status code.

Including `Content-Type: application/x-www-form-urlencoded` in the request allows the request to succeed.

Related to #1067. I don't think enforcing a content-type on GET requests is the correct behaviour since a GET request doesn't have a body. cc @baxang

nbulaj:
Fixes #1161: check content type only for PUT/PATCH/POST requests

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,7 +7,7 @@ User-visible changes worth mentioning.
 
 ## master
 
-- [#] Add your description here.
+- [#1162] Fix `enforce_content_type` for requests without body.
 
 ## 5.0.2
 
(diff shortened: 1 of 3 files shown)


baxang:
Thanks for working on it so quickly! I just opened the issue and a patch has written. Looks good to me 👍

nbulaj:
Thanks @baxang . I'll merge it :)

Issue #854, opened 2016-06-23
undefined local variable or method `root_path`


Morpheu5:
I'm getting the same as #821 when in `api_only` mode, and the following

    #<ActionView::Template::Error: undefined local variable or method `root_path' for #<#<Class:0x007fc9342dd340>:0x007fc9390285f0> Did you mean? font_path>

when I turn `api_only` to false. I generated the app with `--api` so initial conditions may be different from those of the OP in #821.

tute:
Can you show your routes file, please?

Morpheu5:
Sure, sorry.

```
Rails.application.routes.draw do
  use_doorkeeper
  resources :presets
  resources :users
  # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
end
```

tute:
You need to define a `root` path. See http://guides.rubyonrails.org/routing.html#using-root.

tute:
That's where the application goes when you visit `/`.

Morpheu5:
Ah, yes… I keep tripping over the fact that I'm making an API and Doorkeeper has a web interface. It would be really nice if we could disable it. I'd like to keep my API minimal and develop a front-end to manage it, including managing Doorkeeper.

tute:
> It would be really nice if we could disable it

I'm not against this feature. Thanks for reporting!

realvjy:
My routes here

root 'static_pages#home'
get 'help', to: 'static_pages#help'
get 'about', to: 'static_pages#about'

Error
![image](https://cloud.githubusercontent.com/assets/7387461/23338553/28f00fbe-fc34-11e6-92bc-70754540ae2b.png)

PR #1164, opened 2018-11-07
Remove issue#854 when root_path is not set.

snyff:
### Summary

I ran into the following issue#854 (https://github.com/doorkeeper-gem/doorkeeper/issues/854) while using doorkeeper  and thought about a quick fix that will prevent this issue all together. It seems a bit annoying to have to set a root_path just because of a view. This fixes the issue by checking that the method root_path is available before calling it.


nbulaj:
thanks @snyff !

snyff:
@nbulaj Thanks for the quick merge!

PR #1175, opened 2018-12-20
Use scopes_string internally

ghiculescu:
This allows you to override `scopes_string` if you want to set default scopes for a particular type of model (eg. for a `Doorkeeper::Application`). Without this patch you'd need to override `scopes_string` and `scopes`.


nbulaj:
Hi @ghiculescu. This is a good changes. Could you please add a NEWS entry as bot suggested? Thanks

ghiculescu:
done!

Issue #1043, opened 2018-02-27
FactoryBot doesn't see Doorkeeper classes


slim1979:
good day for everyone
when using doorkeeper-4.3.0, a have a problem with FactoryBot. He refused to see doorkeeper classes:
```
NameError:
  uninitialized constant Doorkeeper::AccessToken
NameError:
  uninitialized constant Doorkeeper::Application
```
and other
rollback to 4.2.6 solve this problem

nbulaj:
Doorkeeper 4.3 uses Factory**Bot**, and Doorkeeper 4.2.6 uses Factory**Girl**. What gem do you have?

Also, could you please provide your Gemfile.lock  and spec_helper.rb?

slim1979:
i'm now use 4.2.6 with FactoryBot.
```
Gemfile.lock
GIT
  remote: https://github.com/thoughtbot/shoulda-matchers.git
  revision: 4b160bd19ecca7f97d7ac22dccd5fde9b0da5a9f
  branch: rails-5
  specs:
    shoulda-matchers (3.1.2)
      activesupport (>= 4.2.0)

GEM
  remote: https://rubygems.org/
  specs:
    actioncable (5.1.4)
      actionpack (= 5.1.4)
      nio4r (~> 2.0)
      websocket-driver (~> 0.6.1)
    actionmailer (5.1.4)
      actionpack (= 5.1.4)
      actionview (= 5.1.4)
      activejob (= 5.1.4)
      mail (~> 2.5, >= 2.5.4)
      rails-dom-testing (~> 2.0)
    actionpack (5.1.4)
      actionview (= 5.1.4)
      activesupport (= 5.1.4)
      rack (~> 2.0)
      rack-test (>= 0.6.3)
      rails-dom-testing (~> 2.0)
      rails-html-sanitizer (~> 1.0, >= 1.0.2)
    actionview (5.1.4)
      activesupport (= 5.1.4)
      builder (~> 3.1)
      erubi (~> 1.4)
      rails-dom-testing (~> 2.0)
      rails-html-sanitizer (~> 1.0, >= 1.0.3)
    activejob (5.1.4)
      activesupport (= 5.1.4)
      globalid (>= 0.3.6)
    activemodel (5.1.4)
      activesupport (= 5.1.4)
    activerecord (5.1.4)
      activemodel (= 5.1.4)
      activesupport (= 5.1.4)
      arel (~> 8.0)
    activesupport (5.1.4)
[...]

PR #1176, opened 2019-01-02
Define test factory classes using stringified class names

sman591:
As described in https://github.com/doorkeeper-gem/doorkeeper/issues/1043, factories defined by doorkeeper are unable to be loaded by host apps using the `factory_bot_rails` gem (receiving errors such as `uninitialized constant Doorkeeper::AccessGrant`)

Documentation for factory_bot was later added to explain an easy fix for this: instead of defining the factory's `class: ` with a named constant, wrap the constant in a string. By doing so, the constant isn't evaluated until the factory actually runs, at which point the `Doorkeeper::` models have all been loaded.

Fixes https://github.com/thoughtbot/factory_bot/issues/1188

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,7 @@ User-visible changes worth mentioning.
 - [#1162] Fix `enforce_content_type` for requests without body.
 - [#1164] Fix error when `root_path` is not defined.
 - [#1175] Internal refactor: use `scopes_string` inside `scopes`.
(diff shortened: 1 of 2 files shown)


nbulaj:
Hi @sman591 . Thanks for your work! Could you please add an entry to changelog (NEWS.md) and squash all the commits to a single one?

sman591:
@nbulaj Added & squashed! Happy to reword news entry if needed.

Issue #1177, opened 2019-01-02
Limiting scopes for certain grant_types


rishabhsairawat:
**Feature:** To limit certain scopes to certain grant type.

**Why?:** We may not want all the scopes available for all the grant types. We may want certain scopes to be available only for certain grant types. For example, read-only scopes over the implicit grant.

@nbulaj What do you think about this?

nbulaj:
Hi @rishabhsairawat . I think it can be useful, but requires some changes that must not break compatibility with previous versions of Doorkeeper. Also I think it must be good-configurable. Would you like to try to produce a PR?

PR #1178, opened 2019-01-05
adds an option to limit scopes by grant_type

rishabhsairawat:
### Summary
 This PR aims to implement the feature from #1177 .

To limit scopes for any grant type, you can configure `scopes_by_grant_type` following way:
 `scopes_by_grant_type implicit: [:view_profile, :check_status]`
 Now only `view_profile` and `check_status` will be available for `implicit` `grant_type`.

By default, all the `scopes` will be available for all the `grant_type`.

### DIFF
--- a/lib/doorkeeper/oauth/helpers/scope_checker.rb
+++ b/lib/doorkeeper/oauth/helpers/scope_checker.rb
@@ -7,31 +7,46 @@ module Doorkeeper
         class Validator
           attr_reader :parsed_scopes, :scope_str
 
-          def initialize(scope_str, server_scopes, application_scopes)
+          def initialize(scope_str, server_scopes, app_scopes, grant_type)
             @parsed_scopes = OAuth::Scopes.from_string(scope_str)
             @scope_str = scope_str
-            @valid_scopes = valid_scopes(server_scopes, application_scopes)
+            @valid_scopes = valid_scopes(server_scopes, app_scopes)
+
+            if grant_type
+              @scopes_by_grant_type = Doorkeeper.configuration.scopes_by_grant_type[grant_type.to_sym]
+            end
           end
 
           def valid?
             scope_str.present? &&
               scope_str !~ /[\n\r\t]/ &&
-              @valid_scopes.has_scopes?(parsed_scopes)
+              @valid_scopes.has_scopes?(parsed_scopes) &&
+              permitted_to_grant_type?
           end
 
           private
 
-          def valid_scopes(server_scopes, application_scopes)
-            if application_scopes.present?
-              application_scopes
+          def valid_scopes(server_scopes, app_scopes)
+            if app_scopes.present?
+              app_scopes
             else
               server_scopes
             end
           end
+
+          def permitted_to_grant_type?
+            return true unless @scopes_by_grant_type
+
+            OAuth::Scopes.from_array(@scopes_by_grant_type)
+                         .has_scopes?(parsed_scopes)
+          end
         end
 
-        def self.valid?(scope_str, server_scopes, application_scopes = nil)
-          Validator.new(scope_str, server_scopes, application_scopes).valid?
+        def self.valid?(scope_str:, server_scopes:, app_scopes: nil, grant_type: nil)
+          Validator.new(scope_str,
+                        server_scopes,
+                        app_scopes,
+                        grant_type).valid?
         end
       end
     end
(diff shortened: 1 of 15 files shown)


lib/doorkeeper/oauth/helpers/scope_checker.rb
```diff
@@ -7,16 +7,21 @@ module ScopeChecker
         class Validator
           attr_reader :parsed_scopes, :scope_str
 
-          def initialize(scope_str, server_scopes, application_scopes)
+          def initialize(grant_type, scope_str, server_scopes, application_scopes)
```
nbulaj: I think it would be better to refactor the initializer to use kwargs as it makes it easier to understand what arguments / params do we use.

lib/doorkeeper/oauth/refresh_token_request.rb
```diff
@@ -99,7 +99,7 @@ def validate_client_match
 
       def validate_scope
         if @original_scopes.present?
-          ScopeChecker.valid?(@original_scopes, refresh_token.scopes)
+          ScopeChecker.valid?(nil, @original_scopes, refresh_token.scopes)
```
nbulaj: Also using kwargs will avoid such cases when we need to pass `nil` values as a first argument. And we can use default values

lib/generators/doorkeeper/templates/initializer.rb
```diff
@@ -108,6 +108,15 @@
   # default_scopes  :public
   # optional_scopes :write, :update
 
+  # Define scopes_by_grant_type to restrict only certain scopes for grant_type
```
nbulaj: Maybe it would be better to clarify that in case developer is using `scopes_by_grant_type` - other definitions of the scopes (optional and default) wouldn't be used? WDYT?

rishabhsairawat: What if the developer wants to use `scopes_by_grant_type` option but he doesn't want to limit scopes for all the grant_types.
I think it would be better to keep this `scopes_by_grant_type` option separate as It gives more flexibility to the developer.

nbulaj: Sorry, I just misunderstood your solution. It's OK, thanks for clarification.

lib/doorkeeper/oauth/helpers/scope_checker.rb
```diff
@@ -7,31 +7,46 @@ module ScopeChecker
...
             else
               server_scopes
             end
           end
+
+          def permitted_to_grant_type?
+            return true unless @scopes_by_grant_type
+
+            OAuth::Scopes.from_array(@scopes_by_grant_type)
+                         .has_scopes?(parsed_scopes)
```
nbulaj: `@scopes_by_grant_type` is a `Hash`, will it work with `OAuth::Scopes.from_array` ?

lib/doorkeeper/oauth/helpers/scope_checker.rb
```diff
@@ -7,31 +7,46 @@ module ScopeChecker
...
+          def valid_scopes(server_scopes, app_scopes)
+            if app_scopes.present?
+              app_scopes
             else
               server_scopes
             end
           end
+
+          def permitted_to_grant_type?
+            return true unless @scopes_by_grant_type
```
nbulaj: What if developer will forget to define scopes, like following:

```ruby
Doorkeeper.configure do
  scopes_by_grant_type # no grant types defined
end
```

It will be set to `{}`, and `unless` statement wouldn't work with it. I prefer to use `if @scopes_by_grant_type.blank?`

lib/doorkeeper/oauth/helpers/scope_checker.rb
rishabhsairawat: `@scopes_by_grant_type` here is not a `Hash`. It will be `Array` of scopes for the `grant_type`.

`@scopes_by_grant_type = Doorkeeper.configuration.scopes_by_grant_type[grant_type.to_sym]`

Here  `Doorkeeper.configuration.scopes_by_grant_type` will be a `Hash`.

lib/doorkeeper/oauth/helpers/scope_checker.rb
rishabhsairawat: Explained in the previous comment.

lib/doorkeeper/oauth/helpers/scope_checker.rb
nbulaj: Ah, OK, makes sense. Thanks

Issue #1179, opened 2019-01-09
Post /oauth/token returning 500


hszeto:
**Ruby version**: 2.5.1
**doorkeeper version**: 5.0.1

I'm getting 500 for POST to /oauth/token without the client_id. Is this an expected behavior?

How I obtain the auth code:
```
client = OAuth2::Client.new(client_id, client_secret, :site => site)
client.auth_code.authorize_url(:redirect_uri => redirect_uri)
visit=> http://localhost:3000/oauth/authorize?response_type=code&client_id=...&redirect_uri=...
```
Post token:
```
curl -X POST \
  http://localhost:3000/oauth/token \
  -H 'Content-Type: application/json' \
  -d '{
     "grant_type":"authorization_code",
     "code":"34c34012......40e28cb",
     "client_secret": "4a48e7653b......c0d4f6e3c62f54a",
     "redirect_uri": "http://localhost:3000/callback"
  }'
```

hszeto:
I also observe that, POST /oauth/token without the client secret will return the access token.

```
curl -X POST \
  http://localhost:3000/oauth/token \
  -H 'Content-Type: application/json' \
  -d '{
     "grant_type":"authorization_code",
     "code":"34c34012......40e28cb",
     "client_id":"2c4fd1ccc118c27......156c82a696e7",
     "redirect_uri": "http://localhost:3000/callback"
  }'
```

I  thought the client secret is Required when exchange the code for the access token.

ZilvinasKucinskas:
@hszeto I can verify this is an issue and it should return something like `invalid_client` instead of 500 error. I want to volunteer and fix this issue and open a PR to Doorkeeper.

In regards to the second comment, `client_secret` is not required when using `Authorization Code Grant Flow` as described in [RFC6749](https://tools.ietf.org/html/rfc6749#section-4.1.1).

* If your application is confidential, `client_secret` is required.
* If your application is public, `client_secret` is not required. Public applications are insecure by default.

[More information about client types](https://tools.ietf.org/html/rfc6749#section-2.1)

Check your application - probably `application.confidential` is set to `false`.

PR #1184, opened 2019-01-11
Post /oauth/token returning 500 #1179

hszeto:
This is regarding https://github.com/doorkeeper-gem/doorkeeper/issues/1179

When POST to /oauth/token in the Authorization Code flow without `client_id`, it is currently returning a 500. This fix will check for the `client` in the `validate_client` method and return invalid_grant instead.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,7 @@ User-visible changes worth mentioning.
 
 ## master
 
+- [#1179] Authorization Code flow without client id returns invalid_grant error.
 - [#1182] Fix loopback IP redirect URIs to conform with RFC8252, p. 7.3 (fixes #1170).
 - [#1177] Allow to limit `scopes` for certain `grant_types`
 - [#1162] Fix `enforce_content_type` for requests without body.
--- a/lib/doorkeeper/oauth/authorization_code_request.rb
+++ b/lib/doorkeeper/oauth/authorization_code_request.rb
@@ -53,7 +53,7 @@ module Doorkeeper
       end
 
       def validate_grant
-        return false unless grant && grant.application_id == client.id
+        return false unless grant && client && grant.application_id == client.id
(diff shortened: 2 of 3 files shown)


NEWS.md
```diff
@@ -15,7 +16,7 @@ User-visible changes worth mentioning.
 - [#1176] Fix test factory support for `factory_bot_rails`
 - [#1168]: Allow optional hashing of tokens and secrets.
 
-## 5.0.2
+## 5.0.2g
```
nbulaj: I think we don't need this change

hszeto: fixed.

hszeto:
@nbulaj Can you help with the error from Travis CI? It's complaining about ruby-2.3.7 not installed and Bundler could not find compatible version for gem "bundler".

nbulaj:
@hszeto just ignore it, it's Travis issue

lib/doorkeeper/oauth/authorization_code_request.rb
```diff
@@ -53,7 +53,7 @@ def validate_client
       end
 
       def validate_grant
-        return false unless grant && grant.application_id == client.id
+        return false unless grant && client && grant.application_id == client.id
```
nbulaj: We have `validate :client,       error: :invalid_client`, why it passes blank client?

hszeto: In the Auth Code flow, when I was trying to exchange a code for an access token without sending the client id, it returns 500. While this is an edge case, the server should return a invalid response instead of 500 when a required params is missing.

nbulaj:
Closing in favor of #1185 (it has better solution with validation of a client in the `validate_client`)

Anyway big thanks @hszeto for your work, I really appreciate it! :+1:

PR #1185, opened 2019-01-13
Fix client validation in authorization code flow

ZilvinasKucinskas:
### Summary

Fixes https://github.com/doorkeeper-gem/doorkeeper/issues/1179. 

### Problem

`POST /oauth/token` without client_id returns 500

### Root cause

[Doorkeeper::OAuth::Client::authenticate](https://github.com/doorkeeper-gem/doorkeeper/blob/master/lib/doorkeeper/oauth/client.rb#L21) method was returning `false`, `nil` or `new(application)`, but [Doorkeeper::OAuth::AuthorizationCodeRequest#validate_client](https://github.com/doorkeeper-gem/doorkeeper/blob/master/lib/doorkeeper/oauth/authorization_code_request.rb#L52) method wasn't aware of possible `false` value and relied on `!client.nil?`.

### Changes

* Return invalid_client error if client_id is missing
* Unify client validation interface across validate_client calls
* Add test for authorization code flow with missing client_id

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,7 @@ User-visible changes worth mentioning.
 
 ## master
 
+- [#1179] Authorization Code Grant Flow without client id returns invalid_client error.
 - [#1182] Fix loopback IP redirect URIs to conform with RFC8252, p. 7.3 (fixes #1170).
 - [#1177] Allow to limit `scopes` for certain `grant_types`
 - [#1162] Fix `enforce_content_type` for requests without body.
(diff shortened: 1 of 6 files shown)


ZilvinasKucinskas:
@nbulaj Sorry for potential duplicate PR, but the actual error returned should be `invalid_client` instead of `invalid_grant` as in #1184 .

nbulaj:
Thanks @ZilvinasKucinskas for your contribution!

PR #1200, opened 2019-02-14
Increase token complexity

### LINKED ISSUE #1199: Consider increasing complexity of default generated tokens
Here are the character ranges for a few of the `SecureRandom` generator methods:

method|characters|total available characters
--------|-----------|------------------------------
`SecureRandom.hex` (Doorkeeper default)|`0-9`, `a-f`|`16`
`SecureRandom.alphanumeric `|`A-Z`, `a-z`, `0-9`|`62`
`SecureRandom.urlsafe_base64 `|`A-Z`, `a-z`, `0-9`, `-`,`_`|`64`

The [default configuration for doorkeeper uses `SecureRandom.hex`](https://github.com/doorkeeper-gem/doorkeeper/blob/72425150a95c4c5874ac2926b0ede90e705bd736/lib/doorkeeper/oauth/helpers/unique_token.rb#L8).

`.hex` generates a token with a limited character set. One of the alternatives with a larger character set would increase the complexity of the generated tokens.

Would it be worthwhile setting the default to `urlsafe_base64` or `alphanumeric` for fresh doorkeeper installs?

nbulaj:
Fix #1199: increase complexity of default access token value

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -13,6 +13,16 @@ User-visible changes worth mentioning.
   their block implementation: if you are using `oauth_client.application` to get `Doorkeeper::Application`
   instance, then you need to replace it with just `oauth_client`.
 
+- [#1200] Increase default Doorkeeper access token value complexity (`urlsafe_base64` instead of just `hex`)
+  matching RFC6749/RFC6750.
+
+  **[IMPORTANT]**: this change have possible side-effects in case you have custom database constraints for
+  access token value, application secrets, refresh tokens or you patched Doorkeeper models and introduced
+  token value validations, or you are using database with case-insensitive WHERE clause like MySQL
+  (you can face some collisions). Before this change access token value matched `[a-f0-9]` regex, and now
+  it matches `[a-zA-Z0-9\-_]`. In case you have such restrictions and your don't use custom token generator
+  please change configuration option `default_generator_method ` to `:hex`.
+
 - [#1195] Allow to customize Token Introspection response (fixes #1194).
 - [#1189] Option to set `token_reuse_limit`.
 - [#1191] Try to load bcrypt for hashing of application secrets, but add fallback.
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,6 @@ gem "appraisal"
 gem "bcrypt", "~> 3.1", require: false
 
 gem "activerecord-jdbcsqlite3-adapter", platform: :jruby
-gem "sqlite3", platform: [:ruby, :mswin, :mingw, :x64_mingw]
+gem "sqlite3", "~> 1.3", "< 1.4", platform: [:ruby, :mswin, :mingw, :x64_mingw]
 gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw]
 gemspec
--- a/gemfiles/rails_4_2.gemfile
+++ b/gemfiles/rails_4_2.gemfile
@@ -6,7 +6,7 @@ gem "rails", "~> 4.2.0"
 gem "bcrypt", "~> 3.1", require: false
 gem "appraisal"
 gem "activerecord-jdbcsqlite3-adapter", platform: :jruby
-gem "sqlite3", platform: [:ruby, :mswin, :mingw, :x64_mingw]
+gem "sqlite3", "~> 1.3", "< 1.4", platform: [:ruby, :mswin, :mingw, :x64_mingw]
 gem "tzinfo-data", platforms: [:mingw, :mswin, :x64_mingw]
 # Older Grape requires Ruby >= 2.2.2
 gem "grape", '~> 0.16', '< 0.19.2'
--- a/gemfiles/rails_5_0.gemfile
+++ b/gemfiles/rails_5_0.gemfile
@@ -6,7 +6,7 @@ gem "rails", "~> 5.0.0"
 gem "bcrypt", "~> 3.1", require: false
 gem "appraisal"
 gem "activerecord-jdbcsqlite3-adapter", platforms: :jruby
-gem "sqlite3", platforms: [:ruby, :mswin, :mingw, :x64_mingw]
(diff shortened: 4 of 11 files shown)


eliotsykes:
@nbulaj I'm wondering if there's any risk of bugs here for existing installs of doorkeeper, when they upgrade, and what if anything can be done to help them catch issues earlier. 

Things that might cause bugs or perhaps I'm being overly cautious:

- If an install has any restrictions/validations/db-constraints on the characters allowed in the token
- If the token columns in their doorkeeper db tables are setup to be case-insensitive, e.g. through mysql case insensitive collation. The previous hex tokens didn't contain uppercase chars, these new base64 tokens will.
- If they use `-` or `_` as separators in strings containing their tokens, e.g. an authorization header that has a bearer token of the format `<AUTH_VERSION><SEPARATOR><USER_ID><SEPARATOR><ACCESS_TOKEN>` so `v2--user123456--0123456789abcdef0123456789abcdef`. Now the access token can contain `-` and `_`, this may have unintended consequences.

nbulaj:
I understand your objections, but we can't support any patches somebody added to their projects. We can't change dedault behavior without brakung changes. Instead we can introduce a configurable option that will switch to using urlsafe_base64, but nobody will use it.

eliotsykes:
If we wanted to address this one:

> * If they use `-` or `_` as separators in strings containing their tokens, e.g. an authorization header that has a bearer token of the format `<AUTH_VERSION><SEPARATOR><USER_ID><SEPARATOR><ACCESS_TOKEN>` so `v2-user123456-0123456789abcdef0123456789abcdef`. Now the access token can contain `-` and `_`, this may have unintended consequences.

We could switch to `alphanumeric` instead of `urlsafe_base64`. Is it worth it? I don't know doorkeeper and typical setups well enough to comment.

For the other 2 potential issues, we could ensure the release notes have a very clear warning how the generated tokens format has changed with before/after example tokens.

Overall, do we consider this a breaking change, should it be released in a new major version?

Also, what do you think about introducing a config option for the default method that will be set to `:urlsafe_base64`? Allows for upgraders to stick with their existing `hex` more easily if they need to.

nbulaj:
> Overall, do we consider this a potentially breaking change, should it be released in a new major version?

We can release it with 6.0, but don't sure when it will be.

> Also, what do you think about introducing a config option for the default method that will be set to `:urlsafe_base64`

Access token value generator already is an option and you can introduce your own with `urlsafe_base64` or whatever you want. This case is only about Doorkeeper defaults.

So I see two ways:

* We change it and adding good upgrading guide / changelog entry;
* We release it with a major version update

nbulaj:
Well, about this case:

> If they use `-` or `_` as separators in strings containing their tokens, e.g. an authorization header that has a bearer token of the format `<AUTH_VERSION><SEPARATOR><USER_ID><SEPARATOR><ACCESS_TOKEN>` so `v2-user123456-0123456789abcdef0123456789abcdef`. Now the access token can contain `-` and `_`, this may have unintended consequences.

Don't sure I understand what problem could be here, but as long as access token value corresponds RFC, project implementation details must also follow them.

NEWS.md
```diff
@@ -13,6 +13,15 @@ User-visible changes worth mentioning.
...
 
+- [#1200] Increase default Doorkeeper access token value complexity (`urlsafe_base64` instead of just `hex`)
+  matching RFC6749/RFC6750.
+
+  **[IMPORTANT]**: this change have possible side-effects in case you have custom database constraints for
+  access token value or patched Doorkeeper models with token value validations, or you are using database with
+  case-insensitive WHERE clause like MySQL (you can face some collisions). Before this change access token
+  value matched `[a-f0-9]` regex, and now it matches `[a-zA-Z0-9\-_]`. In case you have such restrictions and 
+  your don't use custom token generator please change configuration option `default_generator_method ` to `:hex`.
+
```
eliotsykes: `UniqueToken.generate` is used for more than access tokens. Should this also mention refresh tokens, application secret, and application uid?

nbulaj: I thought I pointed it, but you are right! I'll fix this, thanks!

nbulaj:
Well, @eliotsykes I checked some big projects oauth2 implementations:

* [Twitter](https://developer.twitter.com/en/docs/basics/authentication/guides/bearer-tokens) - `[a-zA-Z0-9%]`
* [Github](https://developer.github.com/apps/building-oauth-apps/authorizing-oauth-apps/) - `[a-f0-9]`
* [Google](https://developers.google.com/identity/protocols/OAuth2WebServer) - `[a-zA-Z0-9.-_]`
* [Amazon](https://developer.amazon.com/docs/login-with-amazon/access-token.html) - `[a-zA-Z0-9-|_]`

I definitely think that we can leave `urlsafe_base64` as a default implementation.

NEWS.md
nbulaj: Fixed

daveallie:
Hey @nbulaj and @eliotsykes, I'm pretty confused by this move. Without a change to the default size of the token, all this has done is made tokens shorter, included more complicated characters, and increased risk to existing users of doorkeeper without increasing any complexity.

Both the old and the new implementation are 32 bytes of content, just encoded differently. Was the lack of token size adjustment intentional? If base64 is the default, then perhaps the token length default should be longer. If you keep the string length at 64 characters, then you can increase the token complexity to ~46 bytes (~43% increase in complexity over the 32 bytes currently offered).

eliotsykes:
Thanks @daveallie, and yes, I think the shorter length was a mistake. Would this change resolve this:

```diff
-token_size       = options.delete(:size)      || 32
+token_size       = options.delete(:size)      || 48
```

```rb
SecureRandom.hex(32).size
=> 64
SecureRandom.urlsafe_base64(48).size
=> 64
```

daveallie:
Yeah I think you will need to be careful about that change too as it will make hex tokens jump to 96 chars. Preferably the size should be configurable, and if not overridden should be 32 for the `:hex` generator and 48 for the the `:urlsafe_base64` generator.

Issue #550, opened 2015-01-13
Minor: Error response HTTP status differs from OAuth Spec


ahegyi:
I just noticed that in the OAuth spec, when there is an error response upon issuing an access token, a `400 Bad Request` should be returned (in most cases): http://tools.ietf.org/html/rfc6749#section-5.2

However, it looks like the error code sent back from Doorkeeper is always a `401 Unauthorized`: https://github.com/doorkeeper-gem/doorkeeper/blob/master/lib/doorkeeper/oauth/error_response.rb#L29

Does Doorkeeper intentionally always use a `401` or am I misunderstanding something? Thanks!

tute:
We should conform to spec, PR with tests welcome this inconsistency! Thanks for reporting.

ahegyi:
Cool, thanks for confirming! If I get a chance, I'll submit a PR.

tute:
ping @ahegyi

tute:
Quoting the HTTP spec:

> 400 Bad Request
> The server cannot or will not process the request due to something that is perceived to be a client error (e.g., malformed request syntax, invalid request message framing, or deceptive request routing).[14]

These errors seem not fit into a 400. Anyhow, which flavor of 40x we return might be practically irrelevant because if clients follow the HTTP spec they will treat any 4XX as a 400 (3rd paragraph of http://tools.ietf.org/html/rfc7231#section-6).

Closing for now, thanks for reporting!

slross16:
Vanilla HTTP spec doesn't include the Oauth context. Oauth spec which was linked above does place these errors in a 400:  http://tools.ietf.org/html/rfc6749#section-5.2

401 vs 400 does matter in reality - iOS 8 safari for example will pop up a basic authentication dialog (and doesn't allow client JS to handle the response) on receiving a 401 - which may be an error on their part but this is a specific example of an advantage to having doorkeeper follow oauth spec.

tute:
@slross16, indeed we can do better and clients can be smarter if we provide more specific responses. I'll be glad to review and merge a PR that improves this. Thanks for your input.

toupeira:
@nbulaj can we reopen this issue?

If I understand https://tools.ietf.org/html/rfc6749#section-5.2 correctly:
- a 400 status should be returned for most error conditions
- a 401 MAY be be returned for an `invalid_client` error, and MUST be returned if the `Authorization` request header is used.

[...]

PR #1202, opened 2019-02-15
Use correct HTTP status codes for error responses

toupeira:
### Summary

Fixes https://github.com/doorkeeper-gem/doorkeeper/issues/550

Changes the HTTP status codes for error responses to conform with the OAuth specs, as discussed in the issue.

https://tools.ietf.org/html/rfc6749#section-5.2:

- a 400 status should be returned for most error conditions
  - implemented
- a 401 MAY be be returned for an `invalid_client` error, and MUST be returned if the `Authorization` request header is used.
  - implemented

https://tools.ietf.org/html/rfc6750#section-3.1:

- a 401 SHOULD be returned for an `invalid_token` error
  - implemented
- a 403 SHOULD be returned for an `insufficient_scope` error
  - not implemented, Doorkeeper doesn't seem to use this error currently

### DIFF
--- a/app/controllers/doorkeeper/token_info_controller.rb
+++ b/app/controllers/doorkeeper/token_info_controller.rb
@@ -6,7 +6,7 @@ module Doorkeeper
       if doorkeeper_token && doorkeeper_token.accessible?
         render json: doorkeeper_token, status: :ok
       else
-        error = OAuth::ErrorResponse.new(name: :invalid_request)
+        error = OAuth::InvalidTokenResponse.new
         response.headers.merge!(error.headers)
         render json: error.body, status: error.status
       end
--- a/spec/controllers/authorizations_controller_spec.rb
+++ b/spec/controllers/authorizations_controller_spec.rb
@@ -165,7 +165,7 @@ describe Doorkeeper::AuthorizationsController, 'implicit grant flow' do
     let(:redirect_uri) { response_json_body['redirect_uri'] }
 
     it 'renders 400 error' do
-      expect(response.status).to eq 401
+      expect(response.status).to eq 400
     end
 
     it 'includes correct redirect URI' do
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,11 @@ User-visible changes worth mentioning.
 
 ## master
 
+- [#1202] Use correct HTTP status codes for error responses
+
+  **[IMPORTANT]**: this change might break your application if you were relying on the previous
+  401 status codes, this is now a 400 by default, or a 401 for `invalid_client` and `invalid_token` errors.
+
 - [#1201] Fix custom TTL block `client` parameter to always be an `Doorkeeper::Application` instance.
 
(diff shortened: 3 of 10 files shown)


app/controllers/doorkeeper/token_info_controller.rb
```diff
@@ -6,7 +6,7 @@ def show
       if doorkeeper_token && doorkeeper_token.accessible?
         render json: doorkeeper_token, status: :ok
       else
-        error = OAuth::ErrorResponse.new(name: :invalid_request)
+        error = OAuth::InvalidTokenResponse.new
```
toupeira: I'm not entirely sure if `invalid_token` instead of `invalid_request` is valid here, since the `invalid_token` error is only mentioned in [RFC 6750](https://tools.ietf.org/html/rfc6750#section-3.1).

But it seems appropriate since the code above validates the token, also compare with a similar check in `lib/doorkeeper/rails/helpers.rb`

nbulaj:
Hey @toupeira , thanks for this work!

Could you please rebase to current master? SQlite updated this month and there are some issues on CI with it, it currently fixed in latest master.

toupeira:
@nbulaj yep encountered the same issue in `doorkeeper-openid_connect` :) Rebased now!

nbulaj:
Thanks @toupeira . I'm wondering if we can break somebody project that depends on this status code? :thinking: Maybe we can add some entry to upgrade guides, WDYT ?

app/controllers/doorkeeper/token_info_controller.rb
nbulaj: Checked the document, agree

spec/controllers/authorizations_controller_spec.rb
```diff
@@ -165,7 +165,7 @@ def translated_error_message(key)
     let(:redirect_uri) { response_json_body['redirect_uri'] }
 
     it 'renders 400 error' do
-      expect(response.status).to eq 401
+      expect(response.status).to eq 400
```
nbulaj: :+1: see example description :)

toupeira:
> Thanks @toupeira . I'm wondering if we can break somebody project that depends on this status code? :thinking: Maybe we can add some entry to upgrade guides, WDYT ?

@nbulaj yeah I guess this could be considered a breaking API change :) I'd be happy to add something, you mean [this upgrade guide](https://github.com/doorkeeper-gem/doorkeeper/wiki/Migration-from-old-versions) right?

nbulaj:
@toupeira yep. Also we can add a note in NEWS.md for those who reads changelogs :)
I'm planning to release it with 5.1, don't think that such change can bring a lot of problems to end users.

nbulaj:
Any news here @toupeira ?

toupeira:
@nbulaj sorry, dropped off the face of the earth there for a bit ;-)

I added some details to the `NEWS.md` entry now, and also copied it to the wiki.

PR #1210, opened 2019-03-10
Extract documentation to https://doorkeeper.gitbook.io/guides/

felipeelias:
### Summary

I noticed that is becoming really hard to get started with Doorkeeper, there are lots of different information mixed up with how-to steps.

For that, I decided to introduce https://doorkeeper.gitbook.io/guides/

This should make easier to separate how-to information from the rest. It is also relatively easier to use than the Wiki, which I'm considering moving some of the content there as well.

The guides are still available and editable on https://github.com/doorkeeper-gem/guides and pull-requests are welcome there.

Besides, I want to better acknowledge current/past contributors and open a section for sponsors. I also deduplicated some of the links.


nbulaj:
This one looks great! Wiki became too big and we definitely need to decompose it

PR #1220, opened 2019-03-22
Drop Rails 4.2 & Ruby < 2.4 support

jasl:
### Summary

I always enjoy removing codes, here's a PoC to see what if we remove Ruby < 2.4 & Rails 4.2 support

Rails 4.2 will EOL when 6.0 release, and Ruby 2.3 will EOL in the end of March.

Besides, Rails 5 and 6 has the same internal architecture,
so set minimum Rails to 5.0 can help follow edge Rails

this PR includes:

- Drop support Ruby < 2.4
- Drop support Rails < 5.0
- `ApplicationMetalController` inherit `ActionController::API` now, no breaking, but follow Rails standard
- `rubocop -a`


nbulaj:
Hi @jasl . I reviewed the current state of Rubies and Rails and agree with you about dropping support of legacy dependencies. So this PR would be a great start.

Also it would be great if we don't introduce breaking changes for next (5.1.x) release.

jasl:
@nbulaj I'll rebase & restore MetalController this weekend.

 I also consider improve Doorkeeper's Rubocop configuration, but you've done this, Great work

jasl:
Seems Hounds doesn't follow 

```
Style/StringLiterals:
  EnforcedStyle: double_quotes
```

nbulaj:
You can ingnore Hound errors because it wotks weird, I already reported them

jasl:
I reverted changes about ApiController, so there are no breaking now.

jasl:
@nbulaj 

Rebased & reverted breaking & CI passed

Issue #1221, opened 2019-03-25
Puzzled by native use cases


montdidier:
When attempting to use the authentication_code flow with a native iOS application there seems to be a use case mismatch with the way doorkeeper is designed. 

As far as I can tell with doorkeeper one should 

- set the redirect_uri in the native app
- setup doorkeeper to recognise this url (application redirect_url)
- set the native_redirect_uri to match in the doorkeeper initialiser

In an iOS application it seems almost expected (there may be other use cases I am not aware of) that one would use a registered custom scheme. 

However doorkeeper does not redirect upon successful authentication to the expected custom scheme URI but instead redirects to the hardcoded URI of the style:

https://server.domain.tld/oauth/authorize/native?code=[CODE]

Which then presents the CODE in a view. This behaviour seems similar as what one would expect using the `urn:ietf:wg:oauth:2.0:oob` URI as per Google's documentation. The OOB request_uri hint never became a standard but Google still uses it.

As far as I understand it this approach has been superseded by the approach(es) documented in [rfc8252](https://tools.ietf.org/html/rfc8252#section-7). Doorkeeper as it currently is, doesn't seem to support any of these.

Short story. I would expect doorkeeper to return a redirect to the native app's custom scheme.

e.g. `appname://oauth/callback?code=[CODE]` or similar.

nbulaj:
Hi @montdidier . I think we can close this issue, right?

montdidier:
@nbulaj Yes, we should close it.

montdidier:
Addressed in referenced issues and PRs

PR #1229, opened 2019-03-27
Allow for custom scheme in native redirect

montdidier:
### Summary

Doorkeeper doesn't seem to currently allow custom scheme redirection in 'native redirect' only Google's non-standard OOB methodology. This PR attempts to address this gap.

Relates to [this issue](https://github.com/doorkeeper-gem/doorkeeper/issues/1221) 

I don't expect that this will be the final form but I wanted start the discussion on how best get this functionality into doorkeeper.

### DIFF
--- a/lib/doorkeeper/oauth/authorization/code.rb
+++ b/lib/doorkeeper/oauth/authorization/code.rb
@@ -20,7 +20,7 @@ module Doorkeeper
         end
 
         def native_redirect
-          if 'urn:ietf:wg:oauth:2.0:oob' == pre_auth.redirect_uri
+          if pre_auth.redirect_uri == 'urn:ietf:wg:oauth:2.0:oob'
             { action: :show, code: token.token }
           else
             "#{pre_auth.redirect_uri}?code=#{token.token}"
--- a/lib/doorkeeper/oauth/authorization/code.rb
+++ b/lib/doorkeeper/oauth/authorization/code.rb
@@ -20,7 +20,11 @@ module Doorkeeper
         end
 
         def native_redirect
-          { action: :show, code: token.token }
+          if 'urn:ietf:wg:oauth:2.0:oob' == pre_auth.redirect_uri
+            { action: :show, code: token.token }
+          else
+            "#{pre_auth.redirect_uri}?code=#{token.token}"
+          end
         end
 
         def configuration


nbulaj:
Hi @montdidier . This feature seems reasonable to me. The only thing I'm wondering is doesn't it need to be a configurable option ? (like `native_redirect` or smth else).

montdidier:
Hi @nbulaj thanks for the quick response. From my understanding of the code path my snippet will only apply if the native_redirect_url is set to the corresponding desired native redirect url in the initializer as one can see [here](https://github.com/doorkeeper-gem/doorkeeper/blob/master/lib/generators/doorkeeper/templates/initializer.rb) 

The suggested default setting appears to be `urn:ietf:wg:oauth:2.0:oob` anyway which will mirror the behaviour of the Google OAuth2 provider. 

Perhaps slightly off topic but I kind of question the value of having the configuration option at all really.

It might be useful in the case where you have a service and want to lock it down only one kind of custom redirect scheme. 

However, you might have a use case where you really don't care what the application is and you want to allow multiple different types of applications that do not use standard https redirect to be able to use OAuth2. 

Some might use a custom scheme, some might want to use localhost redirection, some native application may also work with https redirection via deeplinking type methods. 

I think what I am saying is that the distinction between native and web is somewhat artificial in doorkeeper. The current "non-native" path in doorkeeper appears to prevent the use of a custom scheme because the validation checks for the existence of a host, which doesn't appear in a custom scheme redirect.

This small change gets my use case over the line but since I am here I am happy to accept some direction on how the doorkeeper maintainers would like this improved. Some possible suggestions: 

1. Just this small change. Use cases wanting custom scheme redirection just assign `native_redirect_url = appname:/path` in the initializer. OOB use cases assign  `urn:ietf:wg:oauth:2.0:oob`. Simple and works for now.

2. Remove the idea of native redirect as a special case all together from doorkeeper. Loosen the validation to understand custom schemes and localhost redirection sending everything through the "normal" codepath. Maintain compatibility with the OOB use cases by checking the redirect_url and rendering the correct view if `urn:ietf:wg:oauth:2.0:oob` is detected. I'd be tempted to add some constants like `REDIRECT_URL_OOB =  urn:ietf:wg:oauth:2.0:oob` and `REDIRECT_URL_OOB_AUTO =  urn:ietf:wg:oauth:2.0:oob:auto` into the codebase to facilitate supporting this non-standard Google behaviour.

nbulaj:
Well,I read this https://www.oauth.com/oauth2-servers/oauth-native-apps/redirect-urls-for-native-apps/ and seems like we need your changes to support a custom URL schemas. 

@felipeelias WDYT?

jasl:
I have question about what is the best flow for native app, previously I used `urn:ietf:wg:oauth:2.0:oob` for our own mobile app (our mobile app is registered as a doorkeeper application), but I'm not sure this is the right way.

And I guess for 3rd app I should use Authorization Flow with PKCE, and embed a WebView to show login page of my site then tip to authorize & redirect to app scheme, am I right? does it secure enough?

montdidier:
@jasl from my recent reading on this topic I believe you are correct. Authorization flow with PKCE is the best choice for this use case currently.

A normal webview is not the most secure choice considering the possibility of a rogue app developer in your threat model. 

The current recommendation is to defer the login to the system browser returned to the app thereafter. 

However, Apple have more recently provided secure webviews specifically for this purpose (SFAuthenticationSession, ASWebAuthenticationSession). Apple state these webviews run “out of process” and a rogue app developer or low level hacker cannot use the app itself as a means to snoop on a user password. 

I don’t know if Android has a similar concept, but certainly a system browser is available on all platforms.

felipeelias:
I've been going through old PRs, trying to wrap up my mind around the issue.

Originally this option was called `test_redirect_uri` and then renamed on #409 and #351. While there is a merit to have the option renamed, I firmly believe that this setting should be used **only** for testing purposes.

What also appears on Google documentation ([here](https://developers.google.com/identity/protocols/OAuth2InstalledApp#enable-apis))

> **Important**: The custom URI scheme and loopback IP address options now provide more reliable, secure, and user-friendly ways to obtain user authorization. This option may be discontinued in the future and should only be used if the choices above are not viable.

And for programatic extraction:

> **Important**: This method is identical to the manual copy/paste method described above except the confirmation page does not instruct the user to copy the authorization code. Instead, the page just asks the user to close the window.
> This option is **deprecated** for OAuth 2.0 to Google. It was designed for embedded browsers, or web-views. The custom URI scheme and loopback IP address options now provide more reliable, secure, and user-friendly ways to obtain user authorization.

It basically confirms what @montdidier said.

In hindsight, this option should not exist, but at that time (6y ago) there was not much consensus on how to support native apps.

IMO the `native_redirect_uri` feature should be dropped and we should implement [the RFC](https://tools.ietf.org/html/rfc8252#section-7) in regards to support for native redirection. I think it's better to go the recommended way than try to hack around this. Of course we should consider how this affects existing applications.

jasl:
@felipeelias 

`urn:ietf:wg:oauth:2.0:oob` is just a workaround, because not all platform support custom scheme at 4 years before

For now, custom scheme should be the right and the best practice now, I believe that some native app such as Atlassian Sourcetree already use this for years.

jasl:
@felipeelias 

`urn:ietf:wg:oauth:2.0:oob` is just a workaround, because not all platform support custom scheme at 4 years before

For now, custom scheme should be the right and the best practice now, I believe that some native app such as Atlassian Sourcetree already use this for years.

felipeelias:
Thanks all for the comments. So do we agree with this high-level action plan:
1. Introduce deprecation for `native_redirect_uri` option, release a new version
2. Allow applications to register "Private-Use URI Scheme", release another version

@nbulaj?

jasl:
Maybe `Allow applications to register "Private-Use URI Scheme"`  could be included in the same release?

Technically we just need to loosen URI validate in `redirect_uris` thats allows non http(s) scheme

jasl:
BTW, There's another (not important) improvement is refactor `scopes` and `redirect_uris` with [`serialize`](https://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/Serialization/ClassMethods.html)

montdidier:
Closing this PR and let's move to https://github.com/doorkeeper-gem/doorkeeper/pull/1238

PR #1231, opened 2019-03-27
Update .hound.yml to unify RuboCop config

gylaz:
### Summary

Update `.hound.yml` to fix Hound and RuboCop behavior.

### Other Information

Hound looks for either `ruby` or `rubocop` keys in the config file.
`rubocop` is the new preferred way.


nbulaj:
Thanks @gylaz !

Issue #1224, opened 2019-03-26
NoMethodError raised when fallback access token lookup fails


JGrishey:
### Steps to reproduce
Create an application that uses access token hashing with an active hashing algorithm and a fallback hashing algorithm.

### Expected behavior
In theory, the application should return a `401 Unauthorized` when making a request to a protected resource with an invalid access token.

### Actual behavior
Instead, the application returns a `500 Internal Server Error` when the fallback lookup fails. Here is the backtrace, which fails when trying to use `#public_send` for `:token=`:

```
11:20:20 web.1    | Completed 500 Internal Server Error in 187ms
11:20:20 web.1    |
11:20:20 web.1    |
11:20:20 web.1    |
11:20:20 web.1    | NoMethodError (undefined method `token=' for nil:NilClass):
11:20:20 web.1    |
11:20:20 web.1    | doorkeeper (5.1.0.rc2) lib/doorkeeper/secret_storing/base.rb:23:in `public_send'
11:20:20 web.1    | doorkeeper (5.1.0.rc2) lib/doorkeeper/secret_storing/base.rb:23:in `store_secret'
11:20:20 web.1    | doorkeeper (5.1.0.rc2) lib/doorkeeper/models/concerns/secret_storable.rb:86:in `upgrade_fallback_value'
11:20:20 web.1    | doorkeeper (5.1.0.rc2) lib/doorkeeper/models/concerns/secret_storable.rb:69:in `block in find_by_fallback_token'
11:20:20 web.1    | doorkeeper (5.1.0.rc2) lib/doorkeeper/models/concerns/secret_storable.rb:68:in `tap'
11:20:20 web.1    | doorkeeper (5.1.0.rc2) lib/doorkeeper/models/concerns/secret_storable.rb:68:in `find_by_fallback_token'
11:20:20 web.1    | doorkeeper (5.1.0.rc2) lib/doorkeeper/models/concerns/secret_storable.rb:48:in `find_by_plaintext_token'
11:20:20 web.1    | doorkeeper (5.1.0.rc2) lib/doorkeeper/models/access_token_mixin.rb:27:in `by_token'
11:20:20 web.1    | doorkeeper (5.1.0.rc2) lib/doorkeeper/oauth/token.rb:17:in `authenticate'
11:20:20 web.1    | doorkeeper (5.1.0.rc2) lib/doorkeeper/helpers/controller.rb:35:in `doorkeeper_token'
[...]

PR #1236, opened 2019-03-28
Do not try to store token if not found by fallback

JGrishey:
### Summary

This fixes the bug outlined in #1224. There is an added guard clause in `#find_by_fallback_token` which ensures that the token is found before attempting to store it.

### DIFF
--- a/lib/doorkeeper/models/concerns/secret_storable.rb
+++ b/lib/doorkeeper/models/concerns/secret_storable.rb
@@ -66,6 +66,8 @@ module Doorkeeper
           # Use the previous strategy to look up
           stored_token = fallback_secret_strategy.transform_secret(plain_secret)
           find_by(attr => stored_token).tap do |resource|
+            return nil unless resource
+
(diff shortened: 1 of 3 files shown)


lib/doorkeeper/models/concerns/secret_storable.rb
```diff
@@ -66,6 +66,8 @@ def find_by_fallback_token(attr, plain_secret)
           # Use the previous strategy to look up
           stored_token = fallback_secret_strategy.transform_secret(plain_secret)
           find_by(attr => stored_token).tap do |resource|
```
nbulaj: Why not `return` instead of break?

JGrishey: image

@nbulaj Got this lint warning from rubocop. I'll gladly change it to `return`, which is what I wrote in the first place :)

JGrishey:
@nbulaj This should be ready to review again. Thanks for the input! 😄

nbulaj:
Thanks @JGrishey !

PR #1238, opened 2019-04-02
Better support for native apps

montdidier:
This PR relates to the discussion [here](https://github.com/doorkeeper-gem/doorkeeper/issues/1221) and [here](https://github.com/doorkeeper-gem/doorkeeper/pull/1229)

I have loosened the validation to allow native apps to present a custom scheme. The old Google style OOB method is still supported (backwards compatible). 

I have removed the `native_redirect_url` functionality but I don't know how you would like to deprecate it formally. My opinion is to keep the config for a while but print a deprecation warning (so upgraders do not break their app keeping their configuration from previously). The configuration has no impact on the workings of doorkeeper in this PR.

### DIFF
--- a/app/validators/redirect_uri_validator.rb
+++ b/app/validators/redirect_uri_validator.rb
@@ -6,10 +6,6 @@ require "uri"
 # to OAuth standards and Doorkeeper configuration.
 #
 class RedirectUriValidator < ActiveModel::EachValidator
-  def self.native_redirect_uri
-    Doorkeeper.configuration.native_redirect_uri
-  end
-
   def validate_each(record, attribute, value)
     if value.blank?
       return if Doorkeeper.configuration.allow_blank_redirect_uri?(record)
@@ -17,12 +13,13 @@ class RedirectUriValidator < ActiveModel::EachValidator
       record.errors.add(attribute, :blank)
     else
       value.split.each do |val|
-        uri = ::URI.parse(val)
-        next if native_redirect_uri?(uri)
+        next if oob_redirect_uri?(val)
 
+        uri = ::URI.parse(val)
         record.errors.add(attribute, :forbidden_uri) if forbidden_uri?(uri)
         record.errors.add(attribute, :fragment_present) unless uri.fragment.nil?
-        record.errors.add(attribute, :relative_uri) if uri.scheme.nil? || uri.host.nil?
+        record.errors.add(attribute, :unspecified_scheme) if unspecified_scheme?(uri)
+        record.errors.add(attribute, :relative_uri) if relative_uri?(uri)
         record.errors.add(attribute, :secured_uri) if invalid_ssl_uri?(uri)
       end
     end
@@ -32,12 +29,22 @@ class RedirectUriValidator < ActiveModel::EachValidator
 
   private
 
-  def native_redirect_uri?(uri)
-    self.class.native_redirect_uri.present? && uri.to_s == self.class.native_redirect_uri.to_s
+  def oob_redirect_uri?(uri)
+    Doorkeeper::OAuth::NonStandard::IETF_WG_OAUTH2_OOB_METHODS.include?(uri)
   end
 
   def forbidden_uri?(uri)
     Doorkeeper.configuration.forbid_redirect_uri.call(uri)
+  end
+
+  def unspecified_scheme?(uri)
+    return true if uri.opaque.present?
+
+    %w[localhost].include?(uri.try(:scheme))
+  end
+
+  def relative_uri?(uri)
+    uri.scheme.nil? && uri.host.nil?
   end
 
   def invalid_ssl_uri?(uri)
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -11,6 +11,7 @@ en:
             redirect_uri:
               fragment_present: 'cannot contain a fragment.'
               invalid_uri: 'must be a valid URI.'
+              unspecified_scheme: 'must specify a scheme.'
               relative_uri: 'must be an absolute URI.'
               secured_uri: 'must be an HTTPS/SSL URI.'
               forbidden_uri: 'is forbidden by the server.'
@@ -33,7 +34,6 @@ en:
         confidential: 'Application will be used where the client secret can be kept confidential. Native mobile apps and Single Page Apps are considered non-confidential.'
         redirect_uri: 'Use one line per URI'
         blank_redirect_uri: "Leave it blank if you configured your provider to use Client Credentials, Resource Owner Password Credentials or any other grant type that doesn't require redirect URI."
-        native_redirect_uri: 'Use %{native_redirect_uri} if you want to add localhost URIs for development purposes'
         scopes: 'Separate scopes with spaces. Leave blank to use the default scopes.'
       edit:
         title: 'Edit application'
--- a/lib/doorkeeper/config/option.rb
+++ b/lib/doorkeeper/config/option.rb
@@ -38,14 +38,20 @@ module Doorkeeper
 
         Builder.instance_eval do
           remove_method name if method_defined?(name)
-          define_method name do |*args, &block|
-            value = if attribute_builder
-                      attribute_builder.new(&block).build
-                    else
-                      block || args.first
-                    end
+          if options[:deprecated]
+            define_method name do |*_, &_|
+              Kernel.warn "[DOORKEEPER] #{name} has been deprecated and will soon be removed"
+            end
+          else
+            define_method name do |*args, &block|
+              value = if attribute_builder
+                        attribute_builder.new(&block).build
+                      else
+                        block || args.first
+                      end
 
-            @config.instance_variable_set(:"@#{attribute}", value)
+              @config.instance_variable_set(:"@#{attribute}", value)
+            end
           end
         end
 
--- a/lib/doorkeeper/oauth/authorization/code.rb
+++ b/lib/doorkeeper/oauth/authorization/code.rb
@@ -15,7 +15,7 @@ module Doorkeeper
           @token ||= AccessGrant.create!(access_grant_attributes)
         end
 
-        def native_redirect
+        def oob_redirect
           { action: :show, code: token.plaintext_token }
         end
 
--- a/lib/doorkeeper/oauth/helpers/uri_checker.rb
+++ b/lib/doorkeeper/oauth/helpers/uri_checker.rb
@@ -25,10 +25,10 @@ module Doorkeeper
     module Helpers
       module URIChecker
         def self.valid?(url)
-          return true if native_uri?(url)
+          return true if oob_uri?(url)
 
           uri = as_uri(url)
-          uri.fragment.nil? && !uri.host.nil? && !uri.scheme.nil?
+          valid_scheme?(uri) && iff_host?(uri) && uri.fragment.nil? && uri.opaque.nil?
         rescue URI::InvalidURIError
           false
         end
@@ -78,8 +78,22 @@ module Doorkeeper
           client_query.split("&").sort == query.split("&").sort
         end
 
-        def self.native_uri?(url)
-          url == Doorkeeper.configuration.native_redirect_uri
+        def self.valid_scheme?(uri)
+          return false if uri.scheme.nil?
+
+          %w[localhost].include?(uri.scheme) == false
+        end
+
+        def self.hypertext_scheme?(uri)
+          %w[http https].include?(uri.scheme)
+        end
+
+        def self.iff_host?(uri)
+          !(hypertext_scheme?(uri) && uri.host.nil?)
+        end
+
+        def self.oob_uri?(uri)
+          NonStandard::IETF_WG_OAUTH2_OOB_METHODS.include?(uri)
         end
       end
     end
--- a/spec/requests/flows/authorization_code_spec.rb
+++ b/spec/requests/flows/authorization_code_spec.rb
@@ -28,8 +28,8 @@ feature "Authorization Code Flow" do
       config_is_set(:token_secret_strategy, ::Doorkeeper::SecretStoring::Sha256Hash)
     end
 
-    scenario "Authorization Code Flow with hashing" do
-      @client.redirect_uri = Doorkeeper.configuration.native_redirect_uri
+    def authorize(redirect_url)
+      @client.redirect_uri = redirect_url
       @client.save!
       visit authorization_endpoint_url(client: @client)
       click_on "Authorize"
@@ -42,16 +42,28 @@ feature "Authorization Code Flow" do
       hashed_code = Doorkeeper::AccessGrant.secret_strategy.transform_secret code
       expect(hashed_code).to eq Doorkeeper::AccessGrant.first.token
 
-      expect(code).not_to eq(hashed_code)
+      [code, hashed_code]
+    end
 
+    scenario "using redirect_url urn:ietf:wg:oauth:2.0:oob" do
+      code, hashed_code = authorize("urn:ietf:wg:oauth:2.0:oob")
+      expect(code).not_to eq(hashed_code)
+      i_should_see "Authorization code:"
+      i_should_see code
+      i_should_not_see hashed_code
+    end
+
+    scenario "using redirect_url urn:ietf:wg:oauth:2.0:oob:auto" do
+      code, hashed_code = authorize("urn:ietf:wg:oauth:2.0:oob:auto")
+      expect(code).not_to eq(hashed_code)
       i_should_see "Authorization code:"
       i_should_see code
       i_should_not_see hashed_code
     end
   end
 
-  scenario "resource owner authorizes using test url" do
-    @client.redirect_uri = Doorkeeper.configuration.native_redirect_uri
+  scenario "resource owner authorizes using oob url" do
+    @client.redirect_uri = "urn:ietf:wg:oauth:2.0:oob"
     @client.save!
     visit authorization_endpoint_url(client: @client)
     click_on "Authorize"
--- a/spec/validators/redirect_uri_validator_spec.rb
+++ b/spec/validators/redirect_uri_validator_spec.rb
@@ -18,7 +18,7 @@ describe RedirectUriValidator do
   #
   # @see https://www.oauth.com/oauth2-servers/redirect-uris/redirect-uris-native-apps/
   it "is valid when the uri is custom native URI" do
-    subject.redirect_uri = "myapp://callback"
+    subject.redirect_uri = "myapp:/callback"
     expect(subject).to be_valid
   end
 
@@ -27,33 +27,48 @@ describe RedirectUriValidator do
     expect(subject).to be_valid
   end
 
-  it "accepts native redirect uri" do
+  it "accepts nonstandard oob redirect uri" do
     subject.redirect_uri = "urn:ietf:wg:oauth:2.0:oob"
     expect(subject).to be_valid
   end
 
-  it "rejects if test uri is disabled" do
-    allow(RedirectUriValidator).to receive(:native_redirect_uri).and_return(nil)
-    subject.redirect_uri = "urn:some:test"
-    expect(subject).not_to be_valid
+  it "accepts nonstandard oob:auto redirect uri" do
+    subject.redirect_uri = "urn:ietf:wg:oauth:2.0:oob:auto"
+    expect(subject).to be_valid
   end
 
   it "is invalid when the uri is not a uri" do
     subject.redirect_uri = "]"
     expect(subject).not_to be_valid
-    expect(subject.errors[:redirect_uri].first).to eq("must be a valid URI.")
+    expect(subject.errors[:redirect_uri].first).to eq(I18n.t("activerecord.errors.models.doorkeeper/application.attributes.redirect_uri.invalid_uri"))
   end
 
   it "is invalid when the uri is relative" do
     subject.redirect_uri = "/abcd"
     expect(subject).not_to be_valid
-    expect(subject.errors[:redirect_uri].first).to eq("must be an absolute URI.")
+    expect(subject.errors[:redirect_uri].first).to eq(I18n.t("activerecord.errors.models.doorkeeper/application.attributes.redirect_uri.relative_uri"))
   end
 
   it "is invalid when the uri has a fragment" do
     subject.redirect_uri = "https://example.com/abcd#xyz"
     expect(subject).not_to be_valid
-    expect(subject.errors[:redirect_uri].first).to eq("cannot contain a fragment.")
+    expect(subject.errors[:redirect_uri].first).to eq(I18n.t("activerecord.errors.models.doorkeeper/application.attributes.redirect_uri.fragment_present"))
+  end
+
+  it "is invalid when scheme resolves to localhost (needs an explict scheme)" do
+    subject.redirect_uri = "localhost:80"
+    expect(subject).to be_invalid
+    expect(subject.errors[:redirect_uri].first).to eq(I18n.t("activerecord.errors.models.doorkeeper/application.attributes.redirect_uri.unspecified_scheme"))
+  end
+
+  it "is invalid if an ip address" do
+    subject.redirect_uri = "127.0.0.1:8080"
+    expect(subject).to be_invalid
+  end
+
+  it "accepts an ip address based URI if a scheme is specified" do
+    subject.redirect_uri = "https://127.0.0.1:8080"
+    expect(subject).to be_valid
   end
 
   context "force secured uri" do
@@ -62,13 +77,23 @@ describe RedirectUriValidator do
       expect(subject).to be_valid
     end
 
-    it "accepts native redirect uri" do
-      subject.redirect_uri = "urn:ietf:wg:oauth:2.0:oob"
+    it "accepts custom scheme redirect uri (as per rfc8252 section 7.1)" do
+      subject.redirect_uri = "com.example.app:/oauth/callback"
       expect(subject).to be_valid
     end
 
-    it "accepts app redirect uri" do
-      subject.redirect_uri = "some-awesome-app://oauth/callback"
+    it "accepts custom scheme redirect uri (as per rfc8252 section 7.1) #2" do
+      subject.redirect_uri = "com.example.app:/test"
+      expect(subject).to be_valid
+    end
+
+    it "accepts custom scheme redirect uri (common misconfiguration we have decided to allow)" do
+      subject.redirect_uri = "com.example.app://oauth/callback"
+      expect(subject).to be_valid
+    end
+
+    it "accepts custom scheme redirect uri (common misconfiguration we have decided to allow) #2" do
+      subject.redirect_uri = "com.example.app://test"
       expect(subject).to be_valid
     end
 
@@ -118,7 +143,7 @@ describe RedirectUriValidator do
       subject.redirect_uri = "http://example.com/callback"
       expect(subject).not_to be_valid
       error = subject.errors[:redirect_uri].first
-      expect(error).to eq("must be an HTTPS/SSL URI.")
+      expect(error).to eq(I18n.t("activerecord.errors.models.doorkeeper/application.attributes.redirect_uri.secured_uri"))
     end
   end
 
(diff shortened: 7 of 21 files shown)


jasl:
You can run `rubocop -a` to fix all style issues

app/validators/redirect_uri_validator.rb
```diff
@@ -27,14 +23,22 @@ def validate_each(record, attribute, value)
...
   def forbidden_uri?(uri)
     Doorkeeper.configuration.forbid_redirect_uri.call(uri)
   end
 
+  def relative_uri?(uri)
+    uri.scheme.nil? || (host_based_scheme?(uri) && uri.host.nil?)
+  end
+
+  def host_based_scheme?(uri)
+    %w[http https localhost].include?(uri.try(:scheme))
```
montdidier: The localhost here is because URI.parse('localhost:3000') interprets localhost as a scheme when presented without a scheme 🤷‍♂️ perhaps we want to handle this another way.

lib/doorkeeper/version.rb
```diff
@@ -10,7 +10,7 @@ module VERSION
     MAJOR = 5
     MINOR = 1
     TINY = 0
-    PRE = "rc2"
+    PRE = "rc2".freeze
```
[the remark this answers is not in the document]
montdidier: @jasl hmm. `rubcop -a` did that, might need to update the configuration.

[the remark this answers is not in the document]
nbulaj: @montdidier seems it's OK with this change that was forgotten somewhere in previous PRs :)

felipeelias:
I think a simple `warn "[DOORKEEPER] ...` if the setting is present would be enough to notify the users. We should then remove it on a major version (?)

spec/validators/redirect_uri_validator_spec.rb
```diff
@@ -62,13 +61,18 @@
       expect(subject).to be_valid
     end
 
-    it "accepts native redirect uri" do
+    it "accepts nonstandard oob redirect uri" do
       subject.redirect_uri = "urn:ietf:wg:oauth:2.0:oob"
       expect(subject).to be_valid
     end
 
+    it "accepts nonstandard oob:auto redirect uri" do
```
felipeelias: this test seems duplicated? At least i don't see any diff from the one above

spec/validators/redirect_uri_validator_spec.rb
```diff
@@ -62,13 +61,18 @@
...
+      subject.redirect_uri = "urn:ietf:wg:oauth:2.0:oob:auto"
+      expect(subject).to be_valid
+    end
+
     it "accepts app redirect uri" do
-      subject.redirect_uri = "some-awesome-app://oauth/callback"
+      subject.redirect_uri = "some-awesome-app:/oauth/callback"
```
felipeelias: Does it have to be a single slash? I was [looking around](https://www.oauth.com/oauth2-servers/redirect-uris/redirect-uris-native-apps/) and it seems that two slashes are used for the URI

montdidier: I think it is the recommendation in the RFC. I will confirm.

montdidier: In [rfc8252 section 7.1](https://tools.ietf.org/html/rfc8252#section-7) it states:

>Following the requirements of Section 3.2 of [RFC3986], as there is no naming authority for private-use URI scheme redirects, only a single slash ("/") appears after the scheme component.  A complete example of a redirect URI utilizing a private-use URI scheme is:

> com.example.app:/oauth2redirect/example-provider

montdidier: Where authority in rfc3986 is defined by:

authority  = [ userinfo "@" ] host [ ":" port ]

jasl: I've test this in `irb`

```
[5] pry(main)> uri = URI.parse("some-awesome-app:/oauth/callback")
=> #<URI::Generic some-awesome-app:/oauth/callback>
[6] pry(main)> uri.scheme
=> "some-awesome-app"
[7] pry(main)> uri.path
=> "/oauth/callback"
```

It looks OK.

But I suggests to also test more normal case (which most users using for, I mean `://` style) ensuring that compatible with it and we leave a comment here about the RFC standard

spec/validators/redirect_uri_validator_spec.rb
montdidier: It is within the "force secured uri" context. Want me to remove it?

jasl: since we support custom scheme, I think we can also deprecate "force secured uri", @felipeelias

felipeelias: I found the commit that introduced the `context` in question: 5ce2e5d. To be honest I'm not sure why there is a special `context` when nothing the the setup of the test really changed. 

@montdidier I think you should remove it since the test is duplicated.

> since we support custom scheme, I think we can also deprecate "force secured uri"
@jasl I'm not sure what you mean by that

jasl: @felipeelias I mean we can deprecate <https://github.com/doorkeeper-gem/doorkeeper/blob/master/lib/doorkeeper/config.rb#L288>

felipeelias: may I ask why?

jasl: - I think `force_ssl_in_redirect_uri` only disable `http`, we will support custom scheme and
 `force_ssl_in_redirect_uri` may become confuse to users
- As a provider, it allows 3rd apps connect to, `http` redirect uri usually for their test or dev usage, so using `http` redirect uri relates to 3rd app, not provider's Rails env,  one example is Gitlab

felipeelias: Thanks for the explanation

> I think force_ssl_in_redirect_uri only disable http, we will support custom scheme and
force_ssl_in_redirect_uri may become confuse to users

If I understood correctly, you would like to drop the `force_ssl_in_redirect_uri` option, since `http` can still be registered as a native url?

> As a provider, it allows 3rd apps connect to, http redirect uri usually for their test or dev usage, so using http redirect uri relates to 3rd app, not provider's Rails env

I agree that `redirect_uri` is 3rd party's responsibility, but shouldn't we try to be secure by default but still let developers change it if needed?

jasl: > If I understood correctly, you would like to drop the force_ssl_in_redirect_uri option, since http can still be registered as a native url?

Yes

> I agree that redirect_uri is 3rd party's responsibility, but shouldn't we try to be secure by default but still let developers change it if needed?

I think there's no site (Gitlab, Github, Twitter, Google, etc.) disable `http` for OAuth redirect URI, if they use `doorkeeper`, the configure must be set to `false`.

Consider you are doing an app with Github sign in, you probably using `http` in dev and test, if Github force you use `https`, it make your work difficulty.

felipeelias: Absolutely, but allowing insecure transport must be a conscious decision by the developer. Example:

```ruby
force_ssl_in_redirect_uri { |uri| uri.host != 'localhost' }
```

Maybe one of the args for this block should also include the `application`, so this could be defined on a per application basis. 

```ruby
force_ssl_in_redirect_uri { |context| !context.application.development? }
```

(assuming development is a flag added by the developer)

What do you think?

jasl: I think it's better than current.

I forgot there's another reason is deprecating `force_ssl_in_redirect_uri` will help to simplify doorkeeper's code (because we only need to validate schemes are valid form)

Combine this reason and your propose, what about this?

```ruby
uri_scheme_checker do |uri|
  if uri.host != "localhost"
    uri.scheme != "http"
  else
    true
  end
end
```

nbulaj: Looks complicated to me, it's hard to understand what does this option do..

jasl: I'm thinking separate two layers to validate redirect URIs, the first layer is to validate it is an URI, and the second layer is validate its pattern, and the second layer is configurable.

for example, we want to strict not to use `http`, so we configure second layer like my previous comment.

I think this more generic and can be use to other purpose, such as scheme & host whitelists

felipeelias: I agree w/ @nbulaj, and for me, it's becoming complicated.

Goal of this PR is to improve native app support while keeping backwards-compatibility so let's stick to the topic. If you want to propose a new way of validating URIs, it deserves a new issue.

app/validators/redirect_uri_validator.rb
felipeelias: I'm a bit concerned with this and I think we need to handle differently.

From what I see, URI.parse creates different objects for known schemes:

```
[9] pry(main)> URI.parse('localhost:3000')
=> #<URI::Generic localhost:3000>
[10] pry(main)> URI.parse('http://localhost:3000')
=> #<URI::HTTP http://localhost:3000>
[11] pry(main)> URI.parse('https://localhost:3000')
=> #<URI::HTTPS https://localhost:3000>
[12] pry(main)> URI.parse('myapp://localhost:3000')
=> #<URI::Generic myapp://localhost:3000>
[13] pry(main)> URI.parse('myapp:/localhost:3000')
=> #<URI::Generic myapp:/localhost:3000>
[14] pry(main)> URI.parse('ftp://localhost:3000')
=> #<URI::FTP ftp://localhost:3000/>
```

I'm not sure it will help, but maybe splitting the validation between HTTP, HTTPS, Generic (and maybe not allowing other types of URIs like FTP) would simplify a bit

spec/validators/redirect_uri_validator_spec.rb
jasl: Understood

montdidier:
Sorry was just making sure these changes actually work against my app. They do seem to work.

app/validators/redirect_uri_validator.rb
montdidier: I have been thinking about this a bit more. I was tempted to just remove the localhost and change the test. The way the URI library works seems consistent with other URI libraries in other languages and perhaps it's just a misconfiguration by the developer we don't actually need to conscientiously prevent. 

I can switch to checking the type returned by URI.parse but do we consider dynamic type checking (instance_of? etc) a code smell in ruby? (in other languages it would be considered so).

After those thoughts I started to actually wonder if we need the relative_uri check at all but it does indeed state [here in rfc6749](https://tools.ietf.org/html/rfc6749#section-3.1.2) that that is the case. However, it can only conceivably apply in the case of http and https URI.

I suggest that scheme is always explicitly stated, as my interpretation is that that must always be that case anyway.

so maybe

```

def relative_uri?(uri)
  scheme_requires_authority?(uri) && uri.host.nil?
end 

def scheme_requires_authority?(uri)
  ['http', 'https'].include?(uri.scheme)
end

```

Do we want to explicitly forbid uri without schemes? [here in rfc6749 section 3.1.2](https://tools.ietf.org/html/rfc6749#section-3.1.2) in conjunction with [rfc3886 section 4.3](https://tools.ietf.org/html/rfc3986#section-4.3) implies this in my view.

``` 
record.errors.add(attribute, :fragment_present) unless uri.fragment.nil?
record.errors.add(attribute, :scheme_missing) if uri.scheme.nil?
```

montdidier: Actually perhaps all this goes away if we just do:

```
record.errors.add(attribute, :relative_uri) if uri.scheme.nil? && uri.host.nil?
``

**Comment by montdidier** [comment]
Alright, how do we feel about what we see here now? Any more feedback or desires?

**Comment by felipeelias** [comment]
I will double check this today evening, thanks for the work so far!

File: app/validators/redirect_uri_validator.rb (hunk 1 above)
Comment (felipeelias) [reply app/validators/redirect_uri_validator.rb]: I'm not very confident on the latest approach. This would let pass things like:
```ruby
uri = URI.parse('app.com:80')
uri.scheme.nil? && uri.host.nil? # => false, scheme is there
```
I noticed though, that when such URI is passed, the value of `opaque` is set (instead of port):
```ruby
URI.parse('app.com:80').opaque
# => "80"
```

So I'm thinking about extending what you suggested:
```ruby
def validate(uri)
  uri = URI.parse(uri)
  !uri.scheme.nil? && !uri.host.nil? && uri.fragment.nil? && uri.opaque.nil?
end

[28] pry(main)> validate('app.com:80')
=> false
[29] pry(main)> validate('app.com:/80')
=> false
[30] pry(main)> validate('app.com://80')
=> true
[31] pry(main)> validate('https://80')
=> true
[32] pry(main)> validate('http://80')
=> true
[33] pry(main)> validate('http://80?test=1')
=> true
[34] pry(main)> validate('http://80?test=1#oops')
=> false
[35] pry(main)> validate('app.com://80?test=1#oops')
=> false
[36] pry(main)> validate('app.com://80?test=1')
=> true
[37] pry(main)> validate('locahost:9000')
=> false
```

The only caveat is that this rejects `app.com:/80`, which I think it makes sense. WDYT?

montdidier: The most immediate problem in my view with this validation is that the _most_ rfc correct presentation for apps with native scheme will also be rejected. 

```
validate('com.domain.app:/oauth2')
=> false
```

`com.domain.app://oauth2` is not technically correct; it's a scheme and host not a scheme and path but we have decided to allow it as some native apps do this incorrectly but we cannot expect app developers to use this incorrect presentation out of necessity.

With the rules of rfc alone it is difficult to resolve an interpretation in every case or such a simple presentation. It simply isn't enough information. So as humans we  might use the appearance as a hint as to what it might actually mean. We see the number and interpret it as a port, we see localhost and assume a host over a scheme. 

We are not the first with this [problem](https://bitbucket.org/openid/connect/issues/625/discovery-223-examplecom-8080-has-a-scheme)

I think since we are only validating if it is relative here I think it is ok, do we really want to validate here that somebody has incorrectly presented a host based URI without authority marker '//' here?
 
We could attempt to pick-up this misconfiguration up earlier. Warn if redirect_uri look like host based uri without schemes.

I will continue to think on this.

felipeelias: Thanks for digging deeper on this 👏 

I have been misinterpreting this all along 🤦‍♂️! In my mind such URI `com.domain.app:/oauth2` didn't make sense because I was used to thinking that URIs should look like `scheme://something` - with `://` always there.

But indeed, like you said, my interpretation is `scheme` and `host` and not `scheme` and `path`.

> I think since we are only validating if it is relative here I think it is ok, do we really want to validate here that somebody has incorrectly presented a host based URI without authority marker '//' here?

If I understand correctly your question, after your explanation I think that no, we should not require the presence/absence of the authority marker for **generic URIs**. In short, `com.domain.app://oauth2` and `com.domain.app:/oauth2` are both valid.

Do you agree that we should at least require a `path` to avoid confusion? In other words, a path-rootless URI like`com.domain.app:oauth2` should be invalid (due to potential confusion with `domain:port`)?

Note: I hope we also agree that we are talking only about generic URIs and we should treat differently in the code the validation of http/https/generic schemes, for maintainability purposes.

montdidier: It might be that looking for the existence of a path works for most cases but I don't see anything that explicitly excludes a pathless URI for use as a redirect_url. It seems to me to be a plausible use case. Especially in the case of _Loopback Interface Redirection_ as discussed in [rfc8252#section-7.3](https://tools.ietf.org/html/rfc8252#section-7.3) where a redirect uri with a form like `http://127.0.0.1:9876/` seems highly plausible since the port will only be opened briefly on the client system in order to receive the callback and the closed again, it simply doesn't need a path.

I think what I am saying is that it makes it very hard to protect integrators from the myriad of misconfigurations possible in that validation without also closing off perfectly valid use cases in this validation. 

doorkeeper has historically allowed simply `localhost:80` and similar but I think we it should be discouraged, it should be presented as `http://localhost:80`.

I am coming around to the view that we should, on loading the configuration heuristically map common misconfigurations: 

`localhost:8080` => `http://localhost:8080` 

_and_ print a visible warning explaining that this happened and implementors should use the fully qualified URI to avoid problems and undefined behaviour. We can also update the document to make it clear integrators need to be explicit. 

In this way we can try to support existing deployed misconfigurations if that is deemed important.

felipeelias: Thanks for your reply.

> doorkeeper has historically allowed simply localhost:80 and similar but I think we it should be discouraged, it should be presented as http://localhost:80.

I am unable to create an application with simply `localhost:80`, on 5.1 I get `Redirect URI must be an absolute URI.`. It requires `https` too by default, and I know that this can be changed via setting. We should keep that.

> In this way we can try to support existing deployed misconfigurations if that is deemed important.

Do you have an example of such misconfiguration? I might be missing something here.

Additionally, I'm also able to register redirect URI as `com.example.app://test` but not `com.example.app:/test`. Now that I have full understanding of what the latter means, I do believe we should allow such validation, but that should not come at the cost of allowing `localhost:80` for two reasons:
1. backwards compatibility, in other words no surprises when updating doorkeeper.
2. it creates questions when redirecting, like which scheme should it use.

I'd like to keep that constraint, if you agree

montdidier: > Do you have an example of such misconfiguration? I might be missing something here. 

I just meant that I consider the simple `localhost:80` etc a misconfiguration by strict interpretation of the rfcs but I understand that the reality is that there are likely many instances in use like this already. 

OK. I'll have another run at this part. I'm currently on holiday but I am back on the 29 April 2019 so it will be around then.

lib/doorkeeper/oauth/helpers/uri_checker.rb
```diff
@@ -25,10 +25,9 @@ module OAuth
     module Helpers
       module URIChecker
         def self.valid?(url)
-          return true if native_uri?(url)
-
+          return true if oob_uri?(url)
```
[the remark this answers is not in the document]
montdidier: I don't know why this didn't update.

montdidier:
@felipeelias Ok. How we do we feel about this current state? I think this does everything we want?

felipeelias:
@montdidier thanks for the follow up, indeed looks exactly like we spoke.

One question though: why the logic between registration and code validation now differs? I mean, I was expecting this validation to be the same on both registration and authorization code flow. Am I missing something?

One problem I see with the above is: 

Doorkeeper Provider Example 2019-05-01 09-25-25

I tested your branch against the provider app and I was able to register with the non-scheme uris, but then, as I attempt the authorization, I receive the  error page saying `The requested redirect uri is malformed or doesn't match client redirect URI.`

I believe that can be confusing for users, what do you think?

montdidier:
@felipeelias I must have missed something. I'll try to fix it. What was the URI you tried to use?

NEWS.md
```diff
@@ -12,7 +12,8 @@ User-visible changes worth mentioning.
 ## 5.1.0
 
 - [#1243]: Add nil check operator in token checking at token introspection.
-- [#1241] Explaining foreign key options for resource owner in a single place
+- [#1241] Explaining foreign key options for resource owner in a single place.
```
nbulaj: Don't forget to move it under `master` section please

config/locales/en.yml
```diff
@@ -11,6 +11,7 @@ en:
             redirect_uri:
               fragment_present: 'cannot contain a fragment.'
               invalid_uri: 'must be a valid URI.'
+              unspecified_scheme: 'a scheme must be specified.'
```
nbulaj: Humanized errors will look strange, like `Redirect URI a scheme must be specified.`

Let's make it `must specify a scheme` so that result error message will be `Redirect URI must specify a scheme`

spec/requests/flows/authorization_code_spec.rb
```diff
@@ -28,8 +28,8 @@
       config_is_set(:token_secret_strategy, ::Doorkeeper::SecretStoring::Sha256Hash)
     end
 
-    scenario "Authorization Code Flow with hashing" do
-      @client.redirect_uri = Doorkeeper.configuration.native_redirect_uri
+    def authorize(redirect_url)
```
nbulaj: Why do we removed `scenario "Authorization Code Flow with hashing` ?

felipeelias:
@montdidier `app.co:80`

spec/requests/flows/authorization_code_spec.rb
montdidier: I didn't really remove the scenario per se, more added two variants and changed the name(s), testing `oob` and `oob:auto`. Since these scenarios happen within the feature 'Authorization Code Flow' and context 'with grant hashing enabled' I didn't think the scenario as named was saying much. The two new scenarios I guess would read 

1. Authorization Code Flow with grant hashing enabled using redirect_url `urn:ietf:wg:oauth:2.0:oob`.
2. Authorization Code Flow with grant hashing enabled using redirect_url `urn:ietf:wg:oauth:2.0:oob:auto`.

If expanded out.

Happy to change to whatever folks would prefer.

montdidier:
> @montdidier `app.co:80` and variations of it

Ok. I've changed the validation code, so now it should be consistent across both validators. Let me know what you think.

felipeelias:
@montdidier this looks good! I tested against the provider app and we're good to go, just make sure that the specs are passing and from my side it's good to merge

spec/validators/redirect_uri_validator_spec.rb
```diff
@@ -56,19 +55,45 @@
     expect(subject.errors[:redirect_uri].first).to eq("cannot contain a fragment.")
   end
 
+  it "is invalid when scheme resolves to localhost (needs an explict scheme)" do
+    subject.redirect_uri = "localhost:80"
+    expect(subject).to be_invalid
+    expect(subject.errors[:redirect_uri].first).to eq("a scheme must be specified.")
```
nbulaj: Let's use I18n key to test an actual error in order not to change the text everytime :)

montdidier:
@nbulaj Are you sure you want me to squash or will somebody squash when it comes to merge time in the UI?

felipeelias:
@montdidier please do, I think there is a reason why it's there

montdidier:
@felipeelias there we go. let me know if anything else needs to happen.

app/validators/redirect_uri_validator.rb
```diff
@@ -29,14 +26,24 @@ def validate_each(record, attribute, value)
...
+  def oob_redirect_uri?(uri)
+    Doorkeeper::OAuth::NonStandard::IETF_WG_OAUTH2_OOB_METHODS.include?(uri)
   end
 
   def forbidden_uri?(uri)
     Doorkeeper.configuration.forbid_redirect_uri.call(uri)
   end
 
+  def unspecified_scheme?(uri)
+    return true unless uri.opaque.nil?
```
nbulaj: I personally prefer `if uri.opaque` (or `opaque.present?`) for simplicity

lib/doorkeeper/config/option.rb
```diff
@@ -35,17 +35,24 @@ module Option
...
 
         Builder.instance_eval do
           remove_method name if method_defined?(name)
-          define_method name do |*args, &block|
-            value = if attribute_builder
-                      attribute_builder.new(&block).build
-                    else
-                      block || args.first
-                    end
+          if deprecated
```
nbulaj: Don't see any reason to store options value in a local var here because we use it only one time, so maybe directly here? (`if options[:deprecated]`)

lib/doorkeeper/oauth/authorization/code.rb
```diff
@@ -15,8 +15,13 @@ def issue_token
           @token ||= AccessGrant.create! access_grant_attributes
         end
 
-        def native_redirect
-          { action: :show, code: token.plaintext_token }
+        def oob_redirect
+          case pre_auth.redirect_uri
```
nbulaj: What is the reason to have `case` statement here? I don't see any diff in result, also I see `nil` for `else` branch.. It's scares me :)

montdidier: It is a just vestigial thinko on my part. I thought it was going be different but really it's the just the presentation/view can be slightly different. I remove it.

For the now the presentation remains the same for both.

lib/doorkeeper/config/option.rb
montdidier: True. Fixed.

app/validators/redirect_uri_validator.rb
montdidier: ok. easy improvement. done.

nbulaj:
@montdidier could you please rebase with the current master? And fix conflicts please?

I think this PR is ready to be merged, I have no objections. So you can squash it and we'll continue.

montdidier:
@nbulaj done.

nbulaj:
Last @montdidier . Do we need to specify in our upgrade guides any breaking changes for this PR? I mean if something previously worked and now is broken - we need to define it.

https://github.com/doorkeeper-gem/doorkeeper/wiki/Migration-from-old-versions

montdidier:
@nbulaj It should not have changed for people using the old method of native support. I might be worth noting in the changes guide that the configuration setting `native_redirect_url` is deprecated. It just does nothing now but having it won't break anything having it.

felipeelias:
@montdidier thank you for your contribution ❤️

montdidier:
@felipeelias No problem! Thank you, for your feedback and patience.

ghiculescu:
@montdidier we currently support creating apps with a local redirect URI (using `native_redirect_uri`)

Am I right in assuming doorkeeper no longer supports this functionality? Should we remove our support for it?

I don't mind removing it but I'm confused by your comments that "It should not have changed for people using the old method of native support." As far as I can tell, support for that method was entirely removed [here](https://github.com/doorkeeper-gem/doorkeeper/pull/1238/files#diff-1de26e1e806e732aaaf8b2010ee70af7R16).

montdidier:
@ghiculescu if I remember correctly it still works but when you request the redirect url in the admin page you use `urn:ietf:wg:oauth:2.0:oob` to signal you want to use this method. It was a method that was never formally ratified and I would recommend moving to a better method if you can, _but_ it should still work if used in this way. What did you have in your native_redirect_uri ?

ghiculescu:
@montdidier this is what our `/oauth/applications/new` page looks like atm:

image

montdidier:
My memory is foggy on this topic because I have not looked at it since; so definitely double check what I am saying. 

So depending on how your app works you have a few options. 

1. You can use the original method which was a never ratified proposal from the working group which you can use by putting `urn:ietf:wg:oauth:2.0:oob` or urn:ietf:wg:oauth:2.0:oob:auto` in the redirect URL. I believe `urn:ietf:wg:oauth:2.0:oob` originally was intended for a copy and paste by the user and `urn:ietf:wg:oauth:2.0:oob:auto` used what ever programatic extraction from within the app was possible.  If you are using this currently and it works you can probably stick with it, but I would recommend switching to the one of the options below. 

This basically puts the authorization token directly into a web view or the browser title view and the mobile app retrieves it from there. The `native_redirect_uri` was some belt and suspenders way for checking what was put in the redirect_uri. When the app is registered the redirect URL is registered there so the `native_redirect_uri` was not flexible enough to support all method simultaneously. 

2. You can use  custom scheme redirection. Which looks like `com.example.app:redirect_uri_path` and you build a special custom scheme handler in the app. This is better. Use authorization code flow with PKCE extension for maximum best practice.

3. You can use loopback redirection which looks like `http://127.0.0.1:9871`. In this case you need to open the a port to receive an HTTP request when waiting for the result to redirect back from the local webview or securewebview or whatever.

What did you app have for `native_redirect_uri` currently?

ghiculescu:
We had this:

image

I guess we were using it as a nicer looking version of `urn:ietf:wg:oauth:2.0:oob`.

> This basically puts the authorization token directly into a web view or the browser title view and the mobile app retrieves it from there. 

Yep, that's what we were after, except I think it was mostly useful for generating auth tokens for testing quickly without necessarily implementing an Oauth flow. As far as I can tell it wasn't used in any meaningful way in production. Definitely a nice to have but not critical.

PR #1241, opened 2019-04-06
Explaining foreign key options for resource owner in a single place

felipeelias:
### Summary

1. Add bit better explanation on how to add foreign keys to users on initial migration
2. Change resource_owner to `references`
3. Add non-null application reference in access_tokens

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,7 @@ User-visible changes worth mentioning.
 
 ## master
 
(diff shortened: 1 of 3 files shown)


felipeelias:
I wanted to take advantage of `foreign_key` option on migration, but it turns out that doesn't work since the table name (`oauth_applications`) does not match with `references :application`.

Due to that, I just updated comments + spacing

Issue #1242, opened 2019-04-10
No method for nil value exception in token introspection


linhdangduy:
### Steps to reproduce
After update code with #1209 pull request (on master branch or tags v5.1.0rc2), send a request to token introspection endpoint with following format:
```
POST /oauth/introspect HTTP/1.1
Header: Authorization = Bearer bearer_access_token
Body: token=Wrong_token (just set a random token)
```

### Expected behavior
Because body's token was wrong, expected response is
```
{ "active": false }
```

### Actual behavior
Exception occurred: No method `.token` for nil value. At the following line https://github.com/doorkeeper-gem/doorkeeper/blob/9bb03f86583b4b71bd3632191abd93b1f7978021/lib/doorkeeper/oauth/token_introspection.rb#L166

### Reason
`@token` is gotten from request's param at these line https://github.com/doorkeeper-gem/doorkeeper/blob/9bb03f86583b4b71bd3632191abd93b1f7978021/app/controllers/doorkeeper/tokens_controller.rb#L77-L80
In case token is wrong, it will be nil. So the exception will occurred.

### System configuration
Using doorkeeper on master branch or tags v5.1.0rc2.

nbulaj:
Fixed by #1243

PR #1243, opened 2019-04-10
Add nil check operator in token checking at token introspection

linhdangduy:
### Summary
#1242 fix

### DIFF
--- a/lib/doorkeeper/oauth/token_introspection.rb
+++ b/lib/doorkeeper/oauth/token_introspection.rb
@@ -163,7 +163,7 @@ module Doorkeeper
 
       # RFC7662 Section 2.1
       def authorized_token_matches_introspected?
-        authorized_token.token == @token.token
+        authorized_token.token == @token&.token
       end
 
       # If token doesn't belong to some client, then it is public.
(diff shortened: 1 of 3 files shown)


linhdangduy:
@nbulaj 
I have just add NEWS.md and fix typo of one test. Is that ok?
Should I write test for this change?

nbulaj:
@linhdangduy it would be great to have a test for such case, so yes if you can

linhdangduy:
@nbulaj I added the test for this case. Is that clear?

lib/doorkeeper/oauth/token_introspection.rb
```diff
@@ -163,7 +163,7 @@ def valid_token?
 
       # RFC7662 Section 2.1
       def authorized_token_matches_introspected?
-        authorized_token.token == @token.token
+        authorized_token.token == @token&.token
```
manhdaovan: Cause `&.` operator is support from ruby ver. >= 2.3, so is this ok if ruby version is < 2.3?

nbulaj:
@linhdangduy yep, thanks!

PR #1252, opened 2019-05-15
Returning forbidden when an unauthorised user tries to revoke a token

cristofer:
### Summary

When digging into an issue we are currently facing with Revoked tokens, I found out that the test `should not revoke the token as its unauthorized` leads me to see that if the user is not authorised to perform the `revoke`, then it was not receiving the correct status for the operation, which is `forbidden`

### DIFF
--- a/app/controllers/doorkeeper/tokens_controller.rb
+++ b/app/controllers/doorkeeper/tokens_controller.rb
@@ -18,12 +18,13 @@ module Doorkeeper
       # Doorkeeper does not use the token_type_hint logic described in the
       # RFC 7009 due to the refresh token implementation that is a field in
       # the access token model.
-      revoke_token if authorized?
 
-      # The authorization server responds with HTTP status code 200 if the token
-      # has been revoked successfully or if the client submitted an invalid
-      # token
-      render json: {}, status: 200
+      if authorized?
+        revoke_token
+        render json: {}, status: 200
+      else
+        render json: { error: :unauthorized_client, error_description: error_description }, status: :forbidden
+      end
     end
 
     def introspect
@@ -71,7 +72,10 @@ module Doorkeeper
     end
 
     def revoke_token
-      token.revoke if token.accessible?
+      # The authorization server responds with HTTP status code 200 if the token
+      # has been revoked successfully or if the client submitted an invalid
+      # token
+      token.revoke if token&.accessible?
     end
 
     def token
@@ -85,6 +89,10 @@ module Doorkeeper
 
     def authorize_response
       @authorize_response ||= strategy.authorize
+    end
+
+    def error_description
+      I18n.t(:unauthorized, scope: %i[doorkeeper errors messages revoke])
     end
   end
 end
--- a/spec/requests/flows/revoke_token_spec.rb
+++ b/spec/requests/flows/revoke_token_spec.rb
@@ -40,16 +40,14 @@ describe "Revoke Token Flow" do
       end
 
       context "with invalid token to revoke" do
-        it "should not revoke any tokens and respond successfully" do
+        it "should not revoke any tokens and respond with forbidden" do
           expect do
             post revocation_token_endpoint_url,
                  params: { token: "I_AM_AN_INVALID_TOKEN" },
                  headers: headers
           end.not_to(change { Doorkeeper::AccessToken.where(revoked_at: nil).count })
 
-          # The authorization server responds with HTTP status code 200 even if
-          # token is invalid
-          expect(response).to be_successful
+          expect(response).to be_forbidden
         end
       end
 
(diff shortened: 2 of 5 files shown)


cristofer:
Something weird with the errors in Travis, I will take a look. However in the meantime I would appreciate if you see this PR and can confirm that the new logic is correct.

nbulaj:
Hi @cristofer . I took a look at `RFC 7009` (OAuth 2.0 Token Revocation), section `2.1.  Revocation Request`:

>    The authorization server first validates the client credentials (in
>    case of a confidential client) and then verifies whether the token
>    was issued to the client making the revocation request.  If this
>    validation fails, the request is refused and the client is informed
>    of the error by the authorization server as described below.

And then checked section `5.  Security Considerations` and don't see that we must return 200 in case client unauthorized. So seems like this PR is totally reasonable.

Also I think we need to change [this line](https://github.com/doorkeeper-gem/doorkeeper/blob/9d28ef171db536b3736f73963086031db45f7e13/app/controllers/doorkeeper/tokens_controller.rb#L78):

```ruby
#   token.revoke if token.accessible?
#   to:
token.revoke if token&.accessible? # in case invalid access token value was passed and we didn't find the token
```

as of:

> The authorization server responds with HTTP status code 200 if the
> token has been revoked successfully or **if the client submitted an**
>  **invalid token**.

Also this point I think we don't check:

> A malicious client may attempt to guess valid tokens on this endpoint
> by making revocation requests against potential token strings.
> According to this specification, **a client's request must contain a**
>  **valid client_id**, in the case of a public client, or valid client
>  credentials, in the case of a confidential client.

Could you please add it? If no - it's OK, I'll do it myself when I have time.
Thanks for your contribution!

nbulaj:
Well, seems like your changes broke the specs :) You need to double-check them and fix I think

cristofer:
thank you @nbulaj ! I will take care of what you mentioned before as soon as I have time. And about the  travis errors, any clue about this error? `uninitialized constant RSpec::Rails::ChannelExampleGroup::ActionCable` it looks super weird :-/

nbulaj:
Oh, seems like test app (spec/dummy) loads all Rails engines including ActionCable. Wondering how we didn't face it before :thinking:  Need to check it out

nbulaj:
I fixed the specs, so you can rebase with current master

cristofer:
@nbulaj please let me know if the changes I pushed in 72f84d1 are good enough (added the error description and the guard, also fixed the tests). I was not sure how to add the logic here https://github.com/doorkeeper-gem/doorkeeper/blob/72f84d1b052b164afd20b9215360d7d50af25751/app/controllers/doorkeeper/tokens_controller.rb#L74 for the valid `client_id`, should I compare it with the server application? thanks!

app/controllers/doorkeeper/tokens_controller.rb
```diff
@@ -18,12 +18,16 @@ def revoke
...
-      # token
-      render json: {}, status: 200
+      if authorized?
+        revoke_token
+        # The authorization server responds with HTTP status code 200 if the token
+        # has been revoked successfully or if the client submitted an invalid
+        # token
+        render json: {}, status: 200
+      else
+        render json: { error: :unauthorized_client }, status: :forbidden
```
nbulaj: I think it will be good to have one more OAuth error response field, `error_description`, like:

https://github.com/doorkeeper-gem/doorkeeper/blob/72425150a95c4c5874ac2926b0ede90e705bd736/lib/doorkeeper/oauth/error.rb#L6-L12

nbulaj:
> . I was not sure how to add the logic here 

@cristofer you don't need to add anything here. This method check if token (that must be revoked) is present. If not - we have an error (and it's OK). It token value present in the request, then server check's if token issues to some application. If this application is confidential, then server must (and do it) check if somebody tries to revoke it's own token. In case application not confidential (i.e _public_) - it's OK, token can be revoked by anybody.

cristofer:
@nbulaj I hope f221f12 includes what is needed in order to merge this PR :) if not let me know and I will do the necessary changes.

Thank you!

app/controllers/doorkeeper/tokens_controller.rb
```diff
@@ -18,12 +18,14 @@ def revoke
...
 
-      # The authorization server responds with HTTP status code 200 if the token
-      # has been revoked successfully or if the client submitted an invalid
-      # token
-      render json: {}, status: 200
+      if authorized?
+        revoke_token
+        render json: {}, status: 200
+      else
+        error_description = "You are not authorized to revoke this token"
```
nbulaj: Could you please use i18n? As in my previous example

spec/requests/flows/revoke_token_spec.rb
```diff
@@ -127,14 +131,18 @@
         it "should not revoke the access token provided" do
           post revocation_token_endpoint_url, params: { token: access_token.token }
 
-          expect(response).to be_successful
+          expect(response).to be_forbidden
+          expect(response.body).to include("unauthorized_client")
+          expect(response.body).to include("You are not authorized to revoke this token")
```
nbulaj: Here we can also check using i18n. Doorkeeper is a multilanguage solution so we must support internationalization :)

cristofer:
All good now @nbulaj ! Thanks for the time you took to give feedback, I am very excited about contributing to this amazing gem :) 

Chao!

nbulaj:
Thanks @cristofer for your patience and work! :tada:

PR #1255, opened 2019-05-16
Update Rubocop, fix specs with ActionCable

nbulaj:
* Update Rubocop, add rubocop-performance
* Refactor code with new code styles
* Fix specs with ActionCable

### DIFF
--- a/spec/dummy/config/application.rb
+++ b/spec/dummy/config/application.rb
@@ -5,11 +5,13 @@ require "rails"
 %w[
   action_controller/railtie
   action_view/railtie
+  action_cable/engine
   sprockets/railtie
 ].each do |railtie|
   begin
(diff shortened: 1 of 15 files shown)


spec/dummy/config/application.rb
```diff
@@ -5,11 +5,13 @@
 %w[
   action_controller/railtie
   action_view/railtie
+  action_cable/engine
```
cristofer: 👍

nbulaj: New Rails/RSpec includes ActionCable tests and waits for it in example groups without checking if it used in the project. So we need to include it for test app in order not to get an exception. Maybe it wil be fixed in stable 4.0 release

PR #1257, opened 2019-05-17
Add constraint configurations when using client authentication on introspection endpoint

linhdangduy:
### Summary
As discuss by @ZilvinasKucinskas at #1195, this pull request add configuration when using Client Credentials to authenticated for introspection endpoint. 

#### Problem
Current implementation of doorkeeper only allows `client credentials that token is issued for` introspect token, which make any other client can not introspect that token.
Protected resource does not own the token, but as [RFC7662](https://tools.ietf.org/html/rfc7662): `introspection endpoint is used by protected resource to introspect token information.`
#### Solving
Adding constraint configuration when using client authentication on introspection endpoint.
```ruby
# config.rb
option :allow_token_introspection,
       default: (lambda do |token, client|
         if token.application
           token.application == client
         else
           true
         end
       end)
```
By default this configuration allow a client can only introspect the token issued to it, OR token doesn't belong to any client (public token). Otherwise disallow.

#### Customize
Configuration customize example:  adding protected_resource check or checking application scopes
```ruby
# adding protected_resource check (add `protected_resource` using migrations)
Doorkeeper.configure do
  allow_token_introspection do |token, client|
    if token.application
      token.application == client || client.protected_resource?
    else
      true
    end
  end
end
```

### DIFF
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -321,6 +321,20 @@ module Doorkeeper
                grant_flows.exclude?("implicit")
            end)
 
+    # When using client credentials to authenticate at introspection endpoint,
+    # Allow introspection if
+    #   the inspected token to be connected to the same client,
+    #   OR token doesn't belong to any client (public token).
+    #
+    option :allow_token_introspection,
+           default: (lambda do |token, client|
+             if token.application
+               token.application == client
+             else
+               true
+             end
+           end)
+
     attr_reader :api_only,
                 :enforce_content_type,
                 :reuse_access_token,
--- a/lib/doorkeeper/oauth/token_introspection.rb
+++ b/lib/doorkeeper/oauth/token_introspection.rb
@@ -150,7 +150,7 @@ module Doorkeeper
       #
       def active?
         if authorized_client
-          valid_token? && authorized_for_client?
+          valid_token? && allow_token_introspection?
         else
           valid_token?
         end
@@ -166,14 +166,12 @@ module Doorkeeper
         authorized_token.token == @token&.token
       end
 
-      # If token doesn't belong to some client, then it is public.
-      # Otherwise in it required for token to be connected to the same client.
-      def authorized_for_client?
-        if @token.application
-          @token.application == authorized_client.application
-        else
-          true
-        end
+      # config constraints for introspection in Doorkeeper.configuration.allow_token_introspection
+      def allow_token_introspection?
+        !!Doorkeeper.configuration.allow_token_introspection.call(
+          @token,
+          authorized_client.application
+        )
       end
 
       # Allows to customize introspection response.
--- a/lib/generators/doorkeeper/templates/initializer.rb
+++ b/lib/generators/doorkeeper/templates/initializer.rb
@@ -314,6 +314,33 @@ Doorkeeper.configure do
   #   client.superapp? or resource_owner.admin?
   # end
 
+  # Implement constraints in case you use Client Credentials to authenticate
+  # the introspection endpoint.
+  # By default allow introspection if the introspected token belongs to authorized client,
+  # OR token doesn't belong to any client (public token). Otherwise disallow.
+  #
+  # Params:
+  # `token` - the token to be introspected (see Doorkeeper::AccessToken)
+  # `client` - the client application authorized for the endpoint (see Doorkeeper::Application)
+  #
+  # You can completely ignore it:
+  # allow_token_introspection do |_token, _client|
+  #   false
+  # end
+  #
+  # Or you can define your custom check:
+  #   Adding `protected_resource` boolean column to applications table
+  #   to allow protected_resource client introspect the token of normal client.
+  #   In this case, protected_resource client must be confidential.
+  #
+  # allow_token_introspection do |token, client|
+  #   if token.application
+  #     token.application == client || client.protected_resource?
+  #   else
+  #     true
+  #   end
+  # end
+
   # WWW-Authenticate Realm (default "Doorkeeper").
   #
   # realm "Doorkeeper"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ User-visible changes worth mentioning.
 
 ## master
 
+- [#PR] Add your PR description here.
+- [#1257] Add constraint configuration when using client authentication on introspection endpoint.
 - [#1252] Returning `unauthorized` when the revocation of the token should not be performed due to wrong permissions.
 - [#1249]: Specify case sensitive uniqueness to remove Rails 6 deprecation message
 - [#1248] Display the Application Secret in HTML after creating a new application even when `hash_application_secrets` is used.
--- a/spec/controllers/tokens_controller_spec.rb
+++ b/spec/controllers/tokens_controller_spec.rb
@@ -147,7 +147,7 @@ describe Doorkeeper::TokensController do
       end
     end
 
-    context "authorized using valid Client Authentication" do
+    context "authorized using Client Credentials of the client that token is issued to" do
       it "responds with full token introspection" do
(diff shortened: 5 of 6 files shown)


nbulaj:
Hi @linhdangduy . I don't think that such feature must be implemented using new database field. When I thought about it I was thinking about to implement configuration options with block that must be resolved to allow token introspection, smth like:

```ruby
# config/doorkeeper.rb
Doorkeeper.configure do
  allow_token_introspection do |token, client|
     client.protected_resource? # you can add `protected_resource` using migrations
  end
end
```

linhdangduy:
@nbulaj Thank for your suggestion, I will do more research on configuration options! That seem more reasonable.

linhdangduy:
@nbulaj 
 > ```ruby
> # config/doorkeeper.rb
> Doorkeeper.configure do
>   allow_token_introspection do |token, client|
>     client.protected_resource? # you can add `protected_resource` using migrations
>   end
> end
> ```
I want to discuss more about the params (token, client) pass to block. 

- `client` param: Using `client` param to check if client has authority to introspect given token, as suggestion above.

- `token` param: I think the token pass to block is `authorized_token` in Doorkeeper::OAuth::TokenIntrospection (is the token sent in header of request). But I'm not sure what is use-case for this param?

And should I create another pull request for configuration options or just force-push this pull-request?

nbulaj:
It was just an example, I don't sure what data this endpoint need.

My thoughts were like the following:

* When Token introspection requested, we know **the token** (that must be introspected) and, if authorized, **the client** (`nil` in case request has no auth).
* We can pass this data to configuration option so developers can implement their own constraints for token introspection (check if token belongs to client, check if client has `introspect` scope, etc)

> And should I create another pull request for configuration options or just force-push this pull-request?

Just force-push and rename this PR, but you can create a new one if you prefer.

linhdangduy:
@nbulaj 
I force-pushed my fixing.
> check if token belongs to client

I made this checking to default. And user could add their config like:
`token.application == client.application || client.application.protected_resource?`

This configuration is only used when using client authentication for introspection endpoint.
(There still problem with authorized by bearer token: haven't checking scope)

lib/doorkeeper/config.rb
```diff
@@ -321,6 +321,14 @@ def configure_secrets_for(type, using:, fallback:)
                grant_flows.exclude?("implicit")
            end)
 
+    # When using client credentials to authenticate at introspection endpoint,
+    # Allow introspection if the inspected token to be connected to the same client.
+    #
+    option :allow_token_introspection,
+           default: (lambda do |token, client|
+             token.application == client.application
```
nbulaj: Actually, token can be issued without reference to application, so I think we need to add a guard clause like:

```ruby
return true unless token.application # Public token

token.application == client.application # Private token
```

lib/doorkeeper/oauth/token_introspection.rb
```diff
@@ -191,6 +190,10 @@ def customize_response(response)
 
         response.merge(customized_response)
       end
+
+      def allow_token_introspection?
+        !!Doorkeeper.configuration.allow_token_introspection.call(@token, authorized_client)
```
nbulaj: I recommend to send actual application here (`authorized_client&.appliucation`) instead of `Doorkeeper::OAuth::Client` instance. It will simplify configuration block and will be more transparent to developers.

lib/generators/doorkeeper/templates/initializer.rb
```diff
@@ -314,6 +314,23 @@
   #   client.superapp? or resource_owner.admin?
   # end
 
+  # Implement constraints in case you use Client Credentials to authenticate
+  # the introspection endpoint. By default allow a client can only introspect
+  # the token issued to it, otherwise disallow.
+  #
+  # Params:
+  # `token` - the token to be introspected
+  # `client` - the client authorized for introspection endpoint
```
nbulaj: Let's add here that `client` is an instance of `Doorkeeper::Application` entity so developers wouldn't need to check what type is it

lib/generators/doorkeeper/templates/initializer.rb
```diff
@@ -314,6 +314,23 @@
   #   client.superapp? or resource_owner.admin?
   # end
 
+  # Implement constraints in case you use Client Credentials to authenticate
+  # the introspection endpoint. By default allow a client can only introspect
+  # the token issued to it, otherwise disallow.
+  #
+  # Params:
+  # `token` - the token to be introspected
```
nbulaj: an instance of `Doorkeeper::AccessToken` entity, like the below :point_down:

lib/doorkeeper/oauth/token_introspection.rb
```diff
@@ -167,13 +167,12 @@ def authorized_token_matches_introspected?
...
-      # Otherwise in it required for token to be connected to the same client.
+      # Otherwise in it required for token to be connected to the same client by default,
+      # or by other configuration in Doorkeeper.configuration.allow_token_introspection.
       def authorized_for_client?
-        if @token.application
-          @token.application == authorized_client.application
-        else
-          true
-        end
+        return true unless @token.application
```
nbulaj: I think it must be in configuration block default body because some developers may want to forbid introspection for **any** access token regardless it's publicity

lib/doorkeeper/oauth/token_introspection.rb
linhdangduy: > authorized_client&.application

Because it is called after `authorized_client` checking at [here](https://github.com/doorkeeper-gem/doorkeeper/blob/59a1227a963b1034bb9df61ad377bb745c34f297/lib/doorkeeper/oauth/token_introspection.rb#L152), so we can write as `authorized_client.application` (no need `&`)

lib/generators/doorkeeper/templates/initializer.rb
linhdangduy: If using `Doorkeeper::Application` entity, should I use the param name to `application` instead of `client`? Because other configuration options is use `client` param name as `Doorkeeper::OAuth::Client` entity.

lib/doorkeeper/oauth/token_introspection.rb
linhdangduy: Yes, I will put original constraints of `authorized_for_client ?` method to block default body.

lib/doorkeeper/oauth/token_introspection.rb
nbulaj: Well, somebody can try to introspect token without auth, so I think authorized_client can be nil. Could you please check it?

lib/generators/doorkeeper/templates/initializer.rb
nbulaj: It's OAuth term that we widely use (client). Application is just the concrete entity in our solution. I recommend client as we use it in other places

lib/doorkeeper/oauth/token_introspection.rb
linhdangduy: As last pushed code:
```ruby
# token_introspection.rb
def active?
  if authorized_client
    valid_token? && allow_token_introspection?
  else
    valid_token?
  end
end
```
Because of this checking, if request is authorized by `bearer_token` (which is without `client_credentials`), `authorized_client` will be nil, and `allow_token_introspection?` won't be executed. 
This configuration is only used for request **authorized by client credentials**.

lib/generators/doorkeeper/templates/initializer.rb
linhdangduy: I kept the `client` term and added comments to these params.

lib/doorkeeper/oauth/token_introspection.rb
nbulaj: Yep, now I checked it. It's OK, thanks!

PR #1258, opened 2019-05-21
Add comment to explain that the application_id constraint in the oauth_access_tokens table should be optional

elioncho:
The application_id constraint on the oauth_access_tokens table should be optional. If a developer is trying to set up a Password Credentials Grant flow, the application will trigger a:

`
ActiveRecord::NotNullViolation (PG::NotNullViolation: ERROR:  null value in column "application_id" violates not-null constraint
`

I added a comment to inform developers of the need to comment the foreign key constraint.

### DIFF
--- a/lib/generators/doorkeeper/templates/migration.rb.erb
+++ b/lib/generators/doorkeeper/templates/migration.rb.erb
@@ -36,6 +36,9 @@ class CreateDoorkeeperTables < ActiveRecord::Migration<%= migration_version %>
 
     create_table :oauth_access_tokens do |t|
       t.references :resource_owner, index: true
+
+      # Remove `null: false` if you are planning to use Password
+      # Credentials Grant flow that doesn't require an application.
       t.references :application,    null: false
 
       # If you use a custom token generator you may need to change this column


nbulaj:
Hi @elioncho . Doesn't it the same for just access token? :thinking:

elioncho:
> Hi @elioncho . Doesn't it the same for just access token? 🤔

Sorry, I made a mistake before. I just edited the PR and the description. If you're using Password Credentials Grant flow, you must comment on the application_id reference in the oauth_access_tokens table alone, not the oauth_access_grants.

nbulaj:
Hi @elioncho . I think it also applicable for Client Credentials flow, right?

elioncho:
> Hi @elioncho . I think it also applicable for Client Credentials flow, don't you think so? We also just need key and secret to authorize the client

If you're using Client Credentials it is assumed that you have created an application, so you don't need to remove the constraint. It really happens only when you're using Password Credentials and you're not sending a client_id and client_secret to the **/oauth/token** path, because you don't need them.

nbulaj:
Ahh, sorry, you are right, my fault! It's OK!

Issue #1204, opened 2019-02-22
Token Introspection failed with a protected resource's credential


kurodashota-mf:
[RFC7622 OAuth 2.0 Token Introspection](https://tools.ietf.org/html/rfc7662#section-2.1) says,
```
   To prevent token scanning attacks, the endpoint MUST also require
   some form of authorization to access this endpoint, such as client
   authentication as described in OAuth 2.0 [RFC6749] or a separate
   OAuth 2.0 access token such as the bearer token described in OAuth
   2.0 Bearer Token Usage [RFC6750].

   ...

   In this example, the protected resource uses a client identifier and
   client secret to authenticate itself to the introspection endpoint.
```
So Token Introspection Endpoint should require a protected resource's credential.
But doorkeeper's implementation requires a credential of the client the access token was issued to.
https://github.com/doorkeeper-gem/doorkeeper/blob/a729e80248031c1030c4dc0f89474932fe0d6603/lib/doorkeeper/oauth/token_introspection.rb#L104-L109

### Steps to reproduce

Send a request to Token Introspection Endpoint with a protected resource's client_id/client_secret.

### Expected behavior
Token Introspection Endpoint returns a success response.

### Actual behavior
Token Introspection Endpoint returns a failure response.
```
{ active: false }
```

### System configuration

**Doorkeeper initializer**:
no customize

**Ruby version**:
`2.5.3`

**Gemfile.lock**:
```
doorkeeper (5.0.2)                                                                                                                                                                                                                                                            
  railties (>= 4.2)
```

nbulaj:
Hi @kurodashota-mf . Thanks for reporting this issue!

Let's look:

> such as client authentication as described in OAuth 2.0 [RFC6749]
> **or a** separate OAuth 2.0 access token such as the bearer token  
> described in OAuth 2.0 Bearer Token Usage [RFC6750].

So Doorkeeper can implement either first or second type of authorization (not **both**), right?

Btw, I agree that we can implement the second solution from the spec. Would you like to submit a PR?

kurodashota-mf:
Hi @nbulaj. Thank you for your quick response.

There may be a little misunderstanding.
I don't want to the second solution.

Doorkeeper returns failure response in case `The token was issued to a different client than is making this request`. But I think it's wrong. 

From my understanding, in Token Introspection Endpoint, client_id and client_secret are not for authenticating a client an access_token issued to, but for authenticating a protected resource.

nov:
[...]

PR #1262, opened 2019-05-23
#1204: Improve token introspection configuration

nbulaj:
Improve token introspection configuration (provide access to authorized client and authorized token)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,17 +7,17 @@ User-visible changes worth mentioning.
 
 ## master
 
-- [#PR] Add your PR description here.
+- [#1260], [#1261] Improve Token Introspection configuration option (access to tokens, client).
 - [#1257] Add constraint configuration when using client authentication on introspection endpoint.
 - [#1252] Returning `unauthorized` when the revocation of the token should not be performed due to wrong permissions.
-- [#1249]: Specify case sensitive uniqueness to remove Rails 6 deprecation message
+- [#1249] Specify case sensitive uniqueness to remove Rails 6 deprecation message
 - [#1248] Display the Application Secret in HTML after creating a new application even when `hash_application_secrets` is used.
 - [#1248] Return the unhashed Application Secret in the JSON response after creating new application even when `hash_application_secrets` is used.
 - [#1238] Better support for native app with support for custom scheme and localhost redirection.
 
 ## 5.1.0
 
-- [#1243]: Add nil check operator in token checking at token introspection.
+- [#1243] Add nil check operator in token checking at token introspection.
 - [#1241] Explaining foreign key options for resource owner in a single place
 - [#1237] Allow to set blank redirect URI if Doorkeeper configured to use redirect URI-less grant flows.
 - [#1234] Fix `StaleRecordsCleaner` to properly work with big amount of records.
--- a/lib/doorkeeper.rb
+++ b/lib/doorkeeper.rb
@@ -81,6 +81,8 @@ require "doorkeeper/stale_records_cleaner"
 
 require "doorkeeper/orm/active_record"
 
+# Main Doorkeeper namespace.
+#
 module Doorkeeper
   def self.authenticate(request, methods = Doorkeeper.configuration.access_token_methods)
     OAuth::Token.authenticate(request, *methods)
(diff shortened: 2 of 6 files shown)


linhdangduy:
@nbulaj 
Here is some thought about current token introspection configuration:

1. NoMethodError for nil class:
I see that `authorized_client` and `authorized_token` parameters could be **nil** depend on the request send to introspection endpoint.
Suggestion: Doorkeeper should _remind_ users when using these instance at this configuration (quite confuse to user!?)
  - check nil before using these two params
  - Or using `authorized_client`,  `authorized_token` with `&`, for examples:
`authorized_client&.protected_resource?` and `authorized_token&.some_method?`
If not using `&`, `NoMethodError - undefined method for nil:NilClass` exception can occur.

2. Introspection Response:
You may think with this configuration, user could config to check scopes of `authorized_token` params. So if `authorized_token` doesn't have sufficient scopes (does not contain sufficient privileges), the response will be 200 status and body is `active: false`. 
But in this case, [here in rfc7662](https://tools.ietf.org/html/rfc7662#section-2.3) states that it returns 401 status code.

`
If the protected resource uses an OAuth 2.0 bearer token to authorize
   its call to the introspection endpoint and the token used for
   authorization does not contain sufficient privileges or is otherwise
   invalid for this request, the authorization server responds with an
   HTTP 401 code...
`

nbulaj:
That's are great points, @linhdangduy ! Would you like to propose a PR to fix them? I can do it myself if not.

1) I'm for `Or using authorized_client, authorized_token with &`

2) Introspection Response: yep, we need to return 401 in case user authorized with token and `allow_token_introspection` isn't passed.

linhdangduy:
@nbulaj I learned so much while reading doorkeeper and your feedback at my pull request. 
I would like to create a pull request. I'm thinking more about (2)

linhdangduy:
@nbulaj 
Following is my proposing for token introspection endpoint, looking forward to hearing from you.

### Proposing
The request can be authorized by using two method `client credentials` or `token`:
- `client credentials`:  (solved)
https://github.com/doorkeeper-gem/doorkeeper/blob/4965eed838cc5bef3f794439fce754bfe22d4bab/lib/doorkeeper/oauth/token_introspection.rb#L150-L153
after authorizing credentials, 👆the config will be run and if it's fail, response is 200 status and {active: false}

- `token`: at authorizing step: (proposing)
https://github.com/doorkeeper-gem/doorkeeper/blob/4965eed838cc5bef3f794439fce754bfe22d4bab/lib/doorkeeper/oauth/token_introspection.rb#L60-L70
after checking `authorized_token_matches_introspected?` and `accessible?`, the config should be run at this step. If it's fail, request is unauthorized, response will be 401.

As upon describing, response is different with different authorization method, I think we could **separating two config** for these two method:
```ruby
# for `client credentials`
allow_token_introspection_by_seprated_token do |token, authorized_token|
end
# for `token`
allow_token_introspection_by_client do |token, authorized_client|
end
```

Pros:
- Right responding (as RFC7662), and providing two authorization method.
- Don't need `&` anymore, the arguments pass to configs always `present`.
- Transparent to user, configuration will be simply to understand and setup.

Cons:
- Two configurations.

### Some thought about current Code:
https://github.com/doorkeeper-gem/doorkeeper/blob/4965eed838cc5bef3f794439fce754bfe22d4bab/lib/doorkeeper/oauth/token_introspection.rb#L153-L155
I'm not for the `authorized_token&.application` as the `client` send to config block. `client` param is should only used for authorized_client, not application of authorized_token.
Because at here, the request is authorized by `authorized_token` method, so if using only one configuration, `authorized_token` method will accidentally read the config of `client_credential` method (for example: `token.application == authorized_client`)

linhdangduy:
@nbulaj 
After some update of `allow_token_introspection` configuration, my team really want to use this configuration to enable introspection function correctly.
when does doorkeeper's team plan to release v5.2.0? Hope it could be release soon.

nbulaj:
Hi @linhdangduy . There are no concrete release date, we are watching for opened / critical issues / PRs and make a decision on top of that data. Very approximately - mid of June if we wouldn't catch any big issues.

nbulaj:
Hi @linhdangduy . You can try to use doorkeeper `5.2.0rc2` from rubygems and report any issues you faced :tada:

linhdangduy:
@nbulaj Sure, I will try it. Thank you very much :heart:

PR #1263, opened 2019-05-24
Response properly when introspection fails and fix configurations's user guide

linhdangduy:
### Summary

As discuss on https://github.com/doorkeeper-gem/doorkeeper/pull/1262, there are problems with [the use](https://github.com/doorkeeper-gem/doorkeeper/blob/4965eed838cc5bef3f794439fce754bfe22d4bab/lib/doorkeeper/oauth/token_introspection.rb#L169) of current token introspection configuration, so this pull request fix them.

1. NoMethodError for nil class:
I see that `authorized_client` and `authorized_token` parameters could be **nil** depend on the request send to introspection endpoint.
Suggestion: Doorkeeper should _remind_ users when using these instance at this configuration (quite confuse to user!?)
    - check nil before using these two params
    - Or using `authorized_client`,  `authorized_token` with `&`, for examples:
`authorized_client&.protected_resource?` and `authorized_token&.some_method?`
If not using `&`, `NoMethodError - undefined method for nil:NilClass` exception can occur.

2. Introspection Response:
With this configuration, user could config to check scopes of `authorized_token` params. So if `authorized_token` doesn't have sufficient scopes (does not contain sufficient privileges), the response will be 200 status and body is `active: false`. 
But in this case, [here in rfc7662](https://tools.ietf.org/html/rfc7662#section-2.3) states that it returns 401 status code.

 `
 If the protected resource uses an OAuth 2.0 bearer token to authorize
   its call to the introspection endpoint and the token used for
   authorization does not contain sufficient privileges or is otherwise
   invalid for this request, the authorization server responds with an
   HTTP 401 code...
 `

I think we need docs to guide user using this configuration properly. (incase they use both `authorized_client` and `authorized_token`)

### DIFF
--- a/lib/doorkeeper/oauth/token_introspection.rb
+++ b/lib/doorkeeper/oauth/token_introspection.rb
@@ -67,7 +67,7 @@ module Doorkeeper
           #  HTTP 401 code as described in Section 3 of OAuth 2.0 Bearer Token
           #  Usage [RFC6750].
           #
-          @error = :invalid_token if authorized_token_matches_introspected? || !authorized_token.accessible?
+          @error = :invalid_token unless valid_authorized_token?
         else
           @error = :invalid_request
         end
@@ -149,9 +149,9 @@ module Doorkeeper
       #
       def active?
         if authorized_client
-          valid_token? && token_introspection_allowed?(authorized_client.application)
+          valid_token? && token_introspection_allowed?(auth_client: authorized_client.application)
         else
-          valid_token? && token_introspection_allowed?(authorized_token&.application)
+          valid_token?
         end
       end
 
@@ -160,20 +160,26 @@ module Doorkeeper
         @token&.accessible?
       end
 
+      def valid_authorized_token?
+        !authorized_token_matches_introspected? &&
+          authorized_token.accessible? &&
+          token_introspection_allowed?(auth_token: authorized_token)
+      end
+
       # RFC7662 Section 2.1
       def authorized_token_matches_introspected?
         authorized_token.token == @token&.token
       end
 
       # config constraints for introspection in Doorkeeper.configuration.allow_token_introspection
-      def token_introspection_allowed?(client)
+      def token_introspection_allowed?(auth_client: nil, auth_token: nil)
         allow_introspection = Doorkeeper.configuration.allow_token_introspection
         return allow_introspection unless allow_introspection.respond_to?(:call)
 
         allow_introspection.call(
           @token,
-          client,
-          authorized_token
+          auth_client,
+          auth_token
         )
       end
 
--- a/lib/generators/doorkeeper/templates/initializer.rb
+++ b/lib/generators/doorkeeper/templates/initializer.rb
@@ -314,22 +314,12 @@ Doorkeeper.configure do
   #   client.superapp? or resource_owner.admin?
   # end
 
-  # Configure custom constraints for the Token Introspection request. By default
-  # this configuration option allows to introspect the token that belongs to authorized
-  # client (from Bearer token or from authenticated client) OR when token doesn't
+  # Configure custom constraints for the Token Introspection request.
+  # By default this configuration option allows to introspect a token by another
+  # token of the same application, OR to introspect the token that belongs to
+  # authorized client (from authenticated client) OR when token doesn't
   # belong to any client (public token). Otherwise requester has no access to the
   # introspection and it will return response as stated in the RFC.
-  #
-  # You can define your custom check:
-  #
-  # allow_token_introspection do |token, authorized_client, _authorized_token|
-  #   if token.application
-  #     # `protected_resource` is a new database boolean column, for example
-  #     token.application == authorized_client || authorized_client.protected_resource?
-  #   else
-  #     true
-  #   end
-  # end
   #
   # Block arguments:
   #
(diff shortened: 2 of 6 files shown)


lib/doorkeeper/oauth/token_introspection.rb
```diff
@@ -149,9 +149,9 @@ def failure_response
       #
       def active?
         if authorized_client
-          valid_token? && token_introspection_allowed?(authorized_client.application)
+          valid_token? && token_introspection_allowed?
         else
-          valid_token? && token_introspection_allowed?(authorized_token&.application)
```
linhdangduy: I'm not for the `authorized_token&.application` as the client send to config block. client param is should only used for `authorized_client`, not `application` of `authorized_token`.
Because at here, the request is authorized by authorized_token method, so the request authorized by authorized_token method will accidentally read the config of client_credential method (for example: token.application == authorized_client)

linhdangduy: I'm not agree the using of `authorized_token&.application` as the `client` send to introspection config block. `client` param is should only used for `authorized_client`, not `application` of `authorized_token`.
Because at here, the request is authorized by authorized_token method, so the request authorized by authorized_token method will accidentally read the config of client_credential method (for example: token.application == authorized_client)

nbulaj: Hi @linhdangduy . Thanks for your work ❤️ 

Actually this check was introduced for the one main reason: previous implementation allowed to introspect any token if you are authorizing using bearer token (without any check). So instead of client credentials that scopes introspected token by application everybody can just use bearer token for auth and see *ANY* token.

So I decided to add the same guard as for client credentials auth - validate if app of the token been introspected is the same as bearer token one. If developer need other rules - he can implement it himself using configuration option.

Im just thinking about the best and security solution by default..

nbulaj:
I'm a little busy on weekend, so I'll check it tomorrow. Thanks for the contribution!

lib/doorkeeper/oauth/token_introspection.rb
linhdangduy: @nbulaj For above security reason by default, we can _set that check at default configuration_.
Because of the flexibility this configuration provides, misuse of configuration can lead to security problem. I think it's better:
- One parameter represents only one role.
- Adding to doorkeeper doc: examples for properly use of config.

Should I write the the config as the following:
```ruby
# config.rb
option :allow_token_introspection,
       default: (lambda do |token, authorized_client, authorized_token|
         if !token.application
           true  # public token
         elsif authorized_token
           # by default: authorized_token and introspected token need to belong to one application
           authorized_token.application == token.application
         elsif authorized_client
           authorized_client == token.application
         end
       end)  
```

lib/doorkeeper/oauth/token_introspection.rb
```diff
@@ -160,19 +160,25 @@ def valid_token?
...
+      end
+
       # RFC7662 Section 2.1
       def authorized_token_matches_introspected?
         authorized_token.token == @token&.token
       end
 
       # config constraints for introspection in Doorkeeper.configuration.allow_token_introspection
-      def token_introspection_allowed?(client)
+      def token_introspection_allowed?
```
linhdangduy:
```ruby
# token_introspection.rb
def token_introspection_allowed?
  (..)
  allow_introspection.call(
    @token,
    authorized_client&.application,
    authorized_token
  )
end
```
This 👆 method is called at one of following two scenarios: when `authorized_token` or when `authorized_client`
1. when `authorized_token`: `authorized_client` should be `nil`
So it's _needless_ to call `allow_introspection` block with `authorized_client` method as parameter.
2. when `authorized_client`: `authorized_token` should be `nil`. The same as 👆

Should I pass `authorized_client` and `authorized_token` params clearly to this method, as below:
```ruby
# token_introspection.rb
def token_introspection_allowed?(client, auth_token) 
  (..)
  allow_introspection.call(@token, client, auth_token)
end
```
At (1), just call `token_introspection_allowed?(nil, authorized_token)`
At (2), `token_introspection_allowed?(authorized_client&.application, nil)`

lib/doorkeeper/oauth/token_introspection.rb
nbulaj: Yep, looks good for me. his is exactly what I meant :)

lib/generators/doorkeeper/templates/initializer.rb
```diff
@@ -343,15 +333,38 @@
...
-  # Keep in mind, that in case the block returns `nil` or `false` introspection response
-  # doesn't have 401 status code and some descriptive body, you'll get 200 with
-  # { "active": false } body as stated in the RFC 7662 section 2.2. Introspection Response.
+  # In case the block returns `nil` or `false` introspection responses with 401 status code
+  # when using authorized token to introspect, or you'll get 200 with { "active": false } body
+  # when using authorized client to introspect as stated in the
+  # RFC 7662 section 2.2. Introspection Response.
+  #
+  # Using with caution:
+  # Keep in mind that these three parameters pass to block can be nil as following case:
```
linhdangduy: Adding comments to explain when `allow_token_introspection` block's parameters get `nil`.

linhdangduy:
I pushed update of the pull request. 
At this PR, `allow_token_introspection` is used at two different places of token_introspection.rb, that make three parameters pass to configuration block can be nil: 
- one at `authorize!` method which responsibility is authorizing the `authorized token`: introspected token haven't be checking yet (by `valid_token?` method), so at here `@token` is nil. `authorized_client` is nil, too.
- and one at `active?` method: `valid_token?` is called, so it makes sure `@token` is present. In this case, `authorized_token` is nil.

linhdangduy:
Sorry, I just added the CHANGELOG.md entry.

nbulaj:
Thanks @linhdangduy for your contribution :heart: :+1:

PR #1264, opened 2019-05-31
Add #{before,after}_successful_authorization hooks in Doorkeeper::TokensController 

fotos:
### Summary

Added `#{before,after}_successful_authorization` hooks in `Doorkeeper::TokensController#create`, similar to #1064, as amended in #1075).

### Other Information

For the `#before_successful_authorization` I am not sure how to (pre) authorize like [`Doorkeeper::AuthorizationsController#pre_auth`](https://github.com/doorkeeper-gem/doorkeeper/blob/614ec7d/app/controllers/doorkeeper/authorizations_controller.rb#L84).

The `pre_auth.authorizable?` seems to contradict the hook name. If the `#pre_auth` is `#authorizable?` does it mean that the strategy is going to be successful? 🤔

In any case I decided to keep the names the same. I would appreciate some feedback or a pointer towards the right direction.

Additionally I added specs for `Doorkeeper::TokensController#create`. For some reason they have been disabled in 7f54ac250.

### DIFF
--- a/app/controllers/doorkeeper/tokens_controller.rb
+++ b/app/controllers/doorkeeper/tokens_controller.rb
@@ -88,7 +88,20 @@ module Doorkeeper
     end
 
     def authorize_response
-      @authorize_response ||= strategy.authorize
+      @authorize_response ||= begin
+        before_successful_authorization
+        auth = strategy.authorize
+        after_successful_authorization unless auth.is_a?(Doorkeeper::OAuth::ErrorResponse)
+        auth
+      end
+    end
+
+    def after_successful_authorization
+      Doorkeeper.configuration.after_successful_authorization.call(self)
+    end
+
+    def before_successful_authorization
+      Doorkeeper.configuration.before_successful_authorization.call(self)
     end
 
     def revocation_error_response
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ User-visible changes worth mentioning.
 ## master
 
 - [#PR ID] Add your description here.
+- [#1264] Add :before_successful_authorization and :after_successful_authorization hooks in TokensController
 - [#1263]: Response properly when introspection fails and fix configurations's user guide.
 
 ## 5.2.0.rc1
--- a/spec/controllers/tokens_controller_spec.rb
+++ b/spec/controllers/tokens_controller_spec.rb
@@ -3,31 +3,139 @@
 require "spec_helper"
 
 describe Doorkeeper::TokensController do
-  describe "when authorization has succeeded" do
-    let(:token) { double(:token, authorize: true) }
+  let(:client) { FactoryBot.create :application }
+  let!(:user)  { User.create!(name: "Joe", password: "sekret") }
 
-    it "returns the authorization" do
-      skip "verify need of these specs"
+  before do
+    Doorkeeper.configure do
+      resource_owner_from_credentials do
+        User.first
+      end
+    end
 
-      expect(token).to receive(:authorization)
+    allow(Doorkeeper.configuration).to receive(:grant_flows).and_return(["password"])
+  end
 
-      post :create
+  subject { JSON.parse(response.body) }
+
+  describe "POST #create" do
+    before do
+      post :create, params: {
+        client_id: client.uid,
+        client_secret: client.secret,
+        grant_type: "password",
+      }
+    end
+
+    it "responds after authorization" do
+      expect(response).to be_successful
+    end
+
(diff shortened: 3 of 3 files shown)


app/controllers/doorkeeper/tokens_controller.rb
```diff
@@ -88,7 +88,20 @@ def strategy
     end
 
     def authorize_response
-      @authorize_response ||= strategy.authorize
+      @authorize_response ||= begin
+        before_successful_authorization
+        auth = strategy.authorize
+        after_successful_authorization if auth.is_a?(Doorkeeper::OAuth::TokenResponse)
```
nbulaj: Can you please point me why we call this hook only for `TokenResponse` ? Not `CodeResponse`, for example

**UPD**: ahh, seems like because of your question about pre authorization, right?

fotos: Yes. That (pre authorization) and my meagre understanding of how it's supposed to work. 😄 

Initially I had this as:

```Ruby
after_successful_authorization unless auth.is_a?(Doorkeeper::OAuth::ErrorResponse)
```

☝️ would that work?

PS. I just noticed that `Doorkeeper::OAuth::TokenResponse` in not inheriting from `BaseResponse` like `Doorkeeper::OAuth::CodeResponse`. 😱

nbulaj: Actually I don't remember. Need some time to check it :)

fotos: I changed it to what I proposed above which makes more sense anyway, since there might be a `Doorkeeper::OAuth::CodeResponse` as well (for `grant_type=authorization_code`).

I'd ❤️ some feedback for the `pre_auth` part! 🙏

nbulaj:
Actually I'm thinking about how valid is this hook here. Initially `before_successful_authorization` hook was added for Authorization flow, and tokens controller used for other flows..

fotos:
I guess I need to explain the use case: similar to `Doorkeeper::AuthorizationsController` we need to be able to monitor and emit token creation events.

The `password` flow (`resource_owner_from_credentials`) goes straight into the `Doorkeeper::TokensController`. The existing strategy hooks (`{before,after}_successful_strategy_response`) do not yield the controller or the params.

Would adding different hooks (e.g. `{before,after}_successful_token`) be a better approach?

nbulaj:
Hi @fotos . Need to think a little bit more about it. Seems like any other grant flow also must have such hooks and it's also an authorization process, but something confuses me ... Maybe we can find an author of the original PR introduced this hooks and asked him why this hooks works as they works? 

I don't remember, maybe it was @f3ndot (sorry if I'm wrong, I remember you refactored or introduced this configuration options).

fotos:
👋 @nbulaj. No worries, although I'd love to get this merged there is no rush.

I linked to the PRs that introduced the feature in the description (summary). The authors that introduced the callbacks were @nattfodd and @mdeutsch (both PRs have been merged by you 🙌).

>  Seems like any other grant flow also must have such hooks and it's also an authorization process, but something confuses me […]

I lost you here. You mean other flows should have such hooks and they are currently missing them or other flows have such hooks already (except the `password` flow in `TokensController`)?

In any case let me know if I can help, in any way, to get this landed! 🛬

nbulaj:
> . You mean other flows should have such hooks and they are currently missing them

this one

app/controllers/doorkeeper/tokens_controller.rb
```diff
@@ -88,7 +88,20 @@ def strategy
     end
 
     def authorize_response
-      @authorize_response ||= strategy.authorize
+      @authorize_response ||= begin
+        before_successful_authorization
+        auth = strategy.authorize
+        after_successful_authorization unless auth.is_a?(Doorkeeper::OAuth::ErrorResponse)
```
nbulaj: Took a look at the code, yep, seems like `ErrorResponse` is a right solution (all the flows uses it for non-successful responses)

fotos:
Thanks @nbulaj!

Is there anything else that needs to be changed or shall I squash the commits and then you can merge? 🚀

nbulaj:
Yep, squash them and I think it's ready for merging. Sorry, I'm very busy right now, can't find much time for open-source

fotos:
@nbulaj commits squashed. 🗜

No worries, I'm used to waiting weeks or sometimes even months. 😁 

A single day turnaround is pretty cool in my book. 😎

nbulaj:
Thanks, @fotos , for your work! :heart:

PR #1266, opened 2019-06-05
Use strong parameters within pre-authorization

stokarenko:
### Summary

It uses strong parameters within `Doorkeeper::OAuth::PreAuthorization`, in order to gracefully ignore non-scalar values passed as `scope` parameter for instance, instead of raising the things like:

> NoMethodError: undefined method `split' for #<ActionController::Parameters:0x...>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ User-visible changes worth mentioning.
 ## master
 
 - [#PR ID] Add your description here.
+- [#1266]: Use strong parameters within pre-authorization.
 - [#1264] Add :before_successful_authorization and :after_successful_authorization hooks in TokensController
 - [#1263]: Response properly when introspection fails and fix configurations's user guide.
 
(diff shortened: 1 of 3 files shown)


lib/doorkeeper/oauth/pre_authorization.rb
```diff
@@ -15,7 +15,10 @@ class PreAuthorization
                     :code_challenge, :code_challenge_method
       attr_writer   :scope
 
-      def initialize(server, client, attrs = {})
+      def initialize(server, client, params)
```
nbulaj: Actually, I think it must be done on a controller lever, not in the library classes itself.

stokarenko: @nbulaj  I feel so, just tried to keep the knowledge about the list of params in single place.

It's not critical at all for me to move `.permit` call to controller, please confirm that I should )

nbulaj: For sure, it would be great, thanks!

stokarenko: Done! )

PR #1269, opened 2019-06-07
Update initializer template documentation

fotos:
### Summary

* Fix the `before_successful_authorization` example code
* Fix a broken link towards the Custom Access Token Generator docs
* Update the link for scopes to `gitbook.io`

### DIFF
--- a/lib/generators/doorkeeper/templates/initializer.rb
+++ b/lib/generators/doorkeeper/templates/initializer.rb
@@ -65,7 +65,7 @@ Doorkeeper.configure do
   # end
 
   # Use a custom class for generating the access token.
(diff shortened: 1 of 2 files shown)


lib/generators/doorkeeper/templates/initializer.rb
```diff
@@ -296,7 +296,7 @@
   # or add any other functionality.
   #
   # before_successful_authorization do |controller|
-  #   Rails.logger.info(params.inspect)
+  #   Rails.logger.info(controller.request.params.inspect)
```
nbulaj: :+1:

PR #1270, opened 2019-06-10
Fix #1193: batch matching tokens for reuse_access_token

### LINKED ISSUE #1193: Client Credentials reuse token query is unbounded, entire result set loaded into memory
### Steps to reproduce
1. Enable `client_credentials` flow.
2. Use client credentials flow with external clients that don't cache tokens but rather create a new token for every request.
3. Realize this is happening and enable `reuse_access_token`.
4. Observe ~3s response times and random OOM errors from the frontend rails containers.
5. Realize that each token request returns 96k results and loads them all into memory.

### Expected behavior
Doorkeeper should not match _all_ client credentials tokens for an application when querying the database.  It should use either `revoked_at` or date math on `expires_in`.

### Actual behavior
Doorkeeper matched all 96k tokens for one application.  Due to the use of `detect` in `matching_token_for`, doorkeeper loads the entire result set into memory.

### System configuration

**Doorkeeper initializer**:
```ruby
Doorkeeper.configure do
  # Change the ORM that doorkeeper will use (needs plugins)
  orm :active_record

  # This block will be called to check whether the resource owner is authenticated or not.
    # Put your resource owner authentication logic here.
    # Example implementation:
    #   User.find_by_id(session[:user_id]) || redirect_to(new_user_session_url)
    current_user || authenticate_user!
  end

  admin_authenticator do
    current_admin_user || authenticate_admin_user!
  end

  # See https://github.com/doorkeeper-gem/doorkeeper#custom-access-token-generator
  #
  access_token_generator '::Doorkeeper::JWT'

  reuse_access_token
  use_refresh_token

  enforce_configured_scopes

  default_scopes :identity, :read, :write
  optional_scopes :provision_users

  access_token_methods :from_bearer_authorization, :from_access_token_param

  force_ssl_in_redirect_uri !Rails.env.development? && !Rails.env.test?

  grant_flows %w[authorization_code client_credentials]

  skip_authorization do |resource_owner, client|
    resource_owner.is_a?(User) &&
      resource_owner.white_label.oauth_application_id == client.id
  end
end
```

**Ruby version**:
2.4.5
**Gemfile.lock**:
```
    doorkeeper (5.0.2)
      railties (>= 4.2)
    doorkeeper-jwt (0.3.0)
      jwt (~> 2.1.0, >= 2.1.0)
```

nbulaj:
Problem described in #1193

Instead of loading full set of records we'll try to run a batch requests in order to find matching token. It must save the application from memory bloat.

@ryansch could you please take a look at it? Maybe also some profiling?

**[NOTE]**: any ORM now must **override** `find_access_token_in_batches` interface in order to run batching enumeration (if `find_in_batches` method isn't supported by the ORM).


lib/doorkeeper/models/access_token_mixin.rb
```diff
@@ -76,9 +76,50 @@ def matching_token_for(application, resource_owner_or_id, scopes)
...
+      # @param application [Doorkeeper::Application]
+      #   Application instance
+      # @param scopes [String, Doorkeeper::OAuth::Scopes]
+      #   set of scopes
+      #
+      # @return [Doorkeeper::AccessToken, nil] Access Token instance or
+      #   nil if matching record was not found
+      #
+      def find_matching_token(relation, application, scopes)
+        return nil if relation.blank?
```
ryansch: For ActiveRecord, this calls `to_a` internally and will load the result into memory.

ryansch:
@nbulaj I see what you mean about `find_in_batches` getting in the way in the general case here. Normally it seems that there are so few records returned that it shouldn't have an effect on performance and now we have a defence against memory bloat in case of a bug.

I'll give this a go in production and report back.

I also wrote #1271 while digging into this today. Please have a look.

PR #1271, opened 2019-06-10
Reintroduce existing token revocation for client credentials

ryansch:
### Summary

This is related to #1193 and #1270.

Doorkeeper assumes that old tokens will have a revoked_at time set on them. This is usually enforced via `RefreshTokenRequest#before_successful_response`.  `ClientCredentialsRequest` did have revocation behaviour in it but it was removed in https://github.com/doorkeeper-gem/doorkeeper/commit/23bc9a0a0a93be35e40dd44d87417b16339efde5. This pull request adds that behaviour back in while also being careful to honour reusability settings.

### Other Information

I ended up needing to copy some of the implementation of `AccessToken.find_or_create_for` as I needed to run code when the token was found vs created. I don't love the duplication but the code is all using public interfaces.  Feedback welcome!

### DIFF
--- a/lib/doorkeeper/oauth/client_credentials/creator.rb
+++ b/lib/doorkeeper/oauth/client_credentials/creator.rb
@@ -5,10 +5,24 @@ module Doorkeeper
     class ClientCredentialsRequest < BaseRequest
       class Creator
         def call(client, scopes, attributes = {})
+          existing_token = existing_token_for(client, scopes)
+
+          if Doorkeeper.configuration.reuse_access_token && existing_token&.reusable?
+            return existing_token
+          end
+
+          existing_token&.revoke
+
           AccessToken.find_or_create_for(
             client, nil, scopes, attributes[:expires_in],
             attributes[:use_refresh_token]
           )
+        end
+
+        private
+
+        def existing_token_for(client, scopes)
+          Doorkeeper::AccessToken.matching_token_for client, nil, scopes
         end
       end
     end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ User-visible changes worth mentioning.
 ## master
 
 - [#PR ID] Add your description here.
+- [#1271] Reintroduce existing token revocation for client credentials.
 - [#1269] Update initializer template documentation.
 - [#1266]: Use strong parameters within pre-authorization.
 - [#1264] Add :before_successful_authorization and :after_successful_authorization hooks in TokensController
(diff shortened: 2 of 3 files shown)


lib/doorkeeper/oauth/client_credentials/creator.rb
```diff
@@ -5,11 +5,25 @@ module OAuth
     class ClientCredentialsRequest < BaseRequest
       class Creator
         def call(client, scopes, attributes = {})
+          existing_token = existing_token_for(client, scopes)
+
+          if Doorkeeper.configuration.reuse_access_token && existing_token&.reusable?
+            return existing_token
+          end
+
+          existing_token&.revoke
```
nbulaj: Remind me please why we are revoking the token ?

lib/doorkeeper/oauth/client_credentials/creator.rb
```diff
@@ -5,11 +5,25 @@ module OAuth
...
           AccessToken.find_or_create_for(
             client, nil, scopes, attributes[:expires_in],
             attributes[:use_refresh_token]
           )
         end
+
+        private
+
+        def existing_token_for(client, scopes)
+          Doorkeeper::AccessToken.matching_token_for client, nil, scopes
```
nbulaj: I think it's OK to reuse this here, I thought we have more problems than you described :D

lib/doorkeeper/oauth/client_credentials/creator.rb
ryansch: The `authorized_tokens_for` scope assumes that old tokens will be revoked: https://github.com/doorkeeper-gem/doorkeeper/blob/master/lib/doorkeeper/models/access_token_mixin.rb#L156

Making this change causes client credentials to be performant in this case.  Since we only revoke the token when the previous one is either expired or can't be reused, we're not actually revoking a token that could otherwise be used.

NobodysNightmare: For reference, I created this ticket in response to this comment: https://github.com/doorkeeper-gem/doorkeeper/issues/1314

nbulaj: Some fixes for those who will find this topic: #1353 (Doorkeeper 5.3)

nbulaj:
Hi @ryansch . This PR actually is a regression and would be made optional in Doorkeeper 5.3

1318

ryansch:
For posterity: I only wrote this for the performance benefits.

PR #1274, opened 2019-06-17
Fix typo in initializer template

fotos:
### Summary

Fix a small typo in the doorkeeper initializer template, spotted while updating my local config.

The typo was introduced in #1178.


nbulaj:
Thanks @fotos :)

PR #1277, opened 2019-06-19
Prevent requested scope be empty on authorization request, handle and add description for invalid request error

### LINKED ISSUE #1268: Enforce Scope On Access Token Request
Per [the spec](https://tools.ietf.org/html/rfc6749#section-3.3), the authorization server should have the option to reject authorization token requests that don't supply a scope:

>  If the client omits the scope parameter when requesting
   authorization, the authorization server MUST either process the
   request using a pre-defined default value or fail the request
   indicating an invalid scope.

It looks like Doorkeeper currently doesn't support rejecting requests that lack a scope out of the box. Is this something that should be added?

linhdangduy:
### Summary

- Resolves #1268 issue, I think that no need more configure for this issue, following changes satisfy [rfc6749#section-3.3](https://tools.ietf.org/html/rfc6749#section-3.3):
```ruby
# pre_authorization.rb
def scope
  @scope.presence || (server.default_scopes.presence && build_scopes)
end
# and this: after try to get scope at scope method, fail the request if scope is blank
return false if scope.blank?
```
- At authorization endpoint: In case `scope` param is not sent, server will check if `default_scopes` configure is presence, then use that `default_scopes`. If server can not get any scope, it will **fail the request indicating an invalid scope**

- I also added test for this change. Because `default_scopes` is necessary if `scope` param is empty, I add `default_scopes` configure to test to make the test passed.

### Notices for changes at this pull:

**1. Database change:**
- As [rfc6749#section-3.3](https://tools.ietf.org/html/rfc6749#section-3.3), `scope` param need to be provided at authorization request incase server doesn't defined `default_scope` --> Change column attribute: `scopes` of `oauth_access_grants` (table that saving authorization_code) MUST NOT be null.

- Change made at: `lib/generators/doorkeeper/templates/migration.rb.erb`
Of the `oauth_access_grants` table, change `scopes` to [null: false, default: ''](https://github.com/doorkeeper-gem/doorkeeper/pull/1277/files#diff-30a229e46cde1768502bf85189f63440R27)

**2. Return error message properly:** Adding descriptive `error_description` for `invalid_request` error.
```yml
# invalid_request error message
# config/locales/en.yml
invalid_request:
  unknown: 'The request is missing a required parameter, includes an unsupported parameter value, or is otherwise malformed.'
  missing_param: 'Missing required parameter: #{value}.'
  not_support_pkce: 'Invalid code_verifier parameter. Server does not support pkce.'
  request_not_authorized: 'Request need to be authorized. Required parameter for authorizing request is missing or invalid.'
```

**2.1 return `error=invalid_request` and adding descriptive `error_description` when _missing required param_.**
 At authorization request and token request, when required parameter is  _missing_(not be sent, or _blank_ ), doorkeeper will return error as following:

```json
status: 400 Bad Request
{
 "error": "invalid_request",
 "error_discription": "Missing required parameter: #{required_parameter_name}."
}
```
- For authorizations request (authorizations_controller):
  - [response_type](https://github.com/doorkeeper-gem/doorkeeper/pull/1277/files#diff-5c19299fbbf253b0b137bb78daa96299R77)
  - [client_id](https://github.com/doorkeeper-gem/doorkeeper/pull/1277/files#diff-5c19299fbbf253b0b137bb78daa96299R74-R78)
  - [scope](https://github.com/doorkeeper-gem/doorkeeper/pull/1277/files#diff-5c19299fbbf253b0b137bb78daa96299R79) (and if server doesn't configure the `default_scopes`)

- For tokens endpoint request (tokens_controller), with `authorization_code` flow:
    - [code](https://github.com/doorkeeper-gem/doorkeeper/pull/1277/files#diff-1801f141b314d7f208c264723b05b1d2R20)

**2.2 Adding descriptive `error_description` for other reasons:** 
- [error_description = :not_support_pkce](https://github.com/doorkeeper-gem/doorkeeper/pull/1277/files#diff-5c1f9daf0b81d01ceeeb79eaea1cbbe2R44) (_Invalid code_verifier parameter. Server does not support pkce._)
At tokens_controller, with `authorization_code` flow (token strategy): return `invalid_request` error when server doesn't support pkce but the request includes `code_verifier` param.

- [error_description = :request_not_authorized](https://github.com/doorkeeper-gem/doorkeeper/pull/1277/files#diff-f53182008be9d6b6035db59e213adb90R75) (_Request need to be authorized. Required parameter for authorizing request is missing or invalid._)
At token introspection, when the request failed with `invalid_request` error.

**3. When refresh_token is revoked:** return `error=invalid_grant` instead of `error=invalid_request`.

At `refresh_token.transaction` of `refresh_token_request.rb`, change `raise Errors::InvalidTokenReuse` (which return `error=invalid_request`) to [raise Errors::InvalidGrantReuse](https://github.com/doorkeeper-gem/doorkeeper/pull/1277/files#diff-aa3b637335d18183d5e2717f3b25689aL35-R36) (which return `error=invalid_grant`)

_Reference_:`invalid_grant` at https://tools.ietf.org/html/rfc6749#section-5.2


lib/doorkeeper/oauth/pre_authorization.rb
```diff
@@ -35,7 +35,7 @@ def scopes
       end
 
       def scope
-        @scope.presence || build_scopes
+        @scope.presence || (server.default_scopes.presence && build_scopes)
```
nbulaj: Actually, `build_scopes` returns `server.default_scopes.to_s`. So maybe we can just remove `#to_s` from it and do it on other place after we make sure scopes present?

linhdangduy: I think we don't have to change this because current code  makes sure scopes present. 
`@scope` is set by request's `params[:scope]` by this line: https://github.com/doorkeeper-gem/doorkeeper/blob/62a304dfb6b31df0201a1b29ac3a568004dbc24d/lib/doorkeeper/oauth/pre_authorization.rb#L23 so it's type is `string`, 
`build_scopes` also be `string` by https://github.com/doorkeeper-gem/doorkeeper/blob/62a304dfb6b31df0201a1b29ac3a568004dbc24d/lib/doorkeeper/oauth/scopes.rb#L41-L43
Then rails will check string empty by `blank?` method.

linhdangduy:
I'm revising some spec test code.
This line `attr_writer` for `scope` attribute of `PreAuthorization`
https://github.com/doorkeeper-gem/doorkeeper/blob/62a304dfb6b31df0201a1b29ac3a568004dbc24d/lib/doorkeeper/oauth/pre_authorization.rb#L16
seem it is used only for test at here:
https://github.com/doorkeeper-gem/doorkeeper/blob/62a304dfb6b31df0201a1b29ac3a568004dbc24d/spec/lib/oauth/pre_authorization_spec.rb#L88
For this test I think just change assignment from `subject.scope =` to this: `attributes[:scope] = ` (set attribute of request's param), test will be passed.

nbulaj:
What if developer wouldn't configure default or optional scopes? 

Also I remember the question or smth else from somebody about just giving a token without permissions.. But maybe it's not related to Doorkeeper actually :)

linhdangduy:
@nbulaj 
> What if developer wouldn't configure default or optional scopes?

At authorization request, if `scope` param is blank, this PR's implementation only checks if `default_scopes` exists or not in order to set the `scope` value. If developer doesn't configure `default_scopes`, client must provide at least one scope to be able to request authorization endpoint, otherwise, server will `failed the client's request indicating an invalid scope`.
(`optional_scopes` is used to check scope validity after scope value be set)

If so, here are 2 things to consider:

1. Because scope param of request is not nil, Should the scope of `oauth_access_grants` model (token is built at authorization request) would never be `nil`, too !?
Current scope column at schema: 

2. `server fails the request indicating an invalid scope`. How should error be returned?
Searched about how error be return at authorization endpoint, As described at https://tools.ietf.org/html/rfc6749#section-4.1.2.1:
 ```
If the resource owner denies the access request or if the request
fails for reasons other than a missing or invalid redirection URI,
the authorization server informs the client by adding the following
parameters to the query component of the redirection URI using the
"application/x-www-form-urlencoded" format
 ```
I see that `create` method (controller method of POST request) at authorization controller followed the above instruction because it use `redirect_or_render` method (which knows how to response error properly)
https://github.com/doorkeeper-gem/doorkeeper/blob/1a68d15f9392707dd5e0ab3bb1da900fef3494e6/app/controllers/doorkeeper/authorizations_controller.rb#L16-L18

But with `new` method (controller method of GET request), the above instruction **is not followed**. Because it's using `render_error` method which doesn't have `redirect` method in it, only `render` -> Maybe this is *new issue*?
https://github.com/doorkeeper-gem/doorkeeper/blob/1a68d15f9392707dd5e0ab3bb1da900fef3494e6/app/controllers/doorkeeper/authorizations_controller.rb#L36-L43

linhdangduy:
> Also I remember the question or smth else from somebody about just giving a token without permissions

The [rfc6749-3.3](https://tools.ietf.org/html/rfc6749#section-3.3) said that 
```
If the client omits the scope parameter when requesting
   authorization, the authorization server MUST either process the
   request using a pre-defined default value or fail the request
   indicating an invalid scope
```
It is **MUST**, so at `authorization endpoint`, server cannot issue `authorization code` without scope.

Because above quote is `when requesting authorization` context, seems server can issue `access token` without scope, but by requesting `access token` using `token endpoint`(access token request context) in `client credentials` flow: https://tools.ietf.org/html/rfc6749#section-4.4.2
But it seems little confusing, I'm not sure, just followed RFC document.

nbulaj:
> Because scope param of request is not nil, Should the scope of oauth_access_grants model (token is built at authorization request) would never be nil, too !?

Yep, I think you are right. But now with this breaking change I afraid to break existing clients, actually. But `scopes` for database must be NOT NULL.

lib/doorkeeper/oauth/pre_authorization.rb
nbulaj: OK, it's ok for me

nbulaj:
> server fails the request indicating an invalid scope. How should error be returned?
Searched about how error be return at authorization endpoint, As described at https://tools.ietf.org/html/rfc6749#section-4.1.2.1:

I think it must be an error response with `invalid_request` with error description about missed `scope` parameter, as Google do:

image

Could you please add it ?

linhdangduy:
I am sorry to have confused you.

Want to describe about (2):  `Maybe this is new issue?`
At **authorization endpoint**, all the request's params are checked for validity. If any params is invalid, server will handle the error as following instruction at [Authorization Code Grant - 4.1.2.1 Error Response](https://tools.ietf.org/html/rfc6749#section-4.1.2.1) OR [Implicit Grant - 4.2.2.1 Error Response](https://tools.ietf.org/html/rfc6749#section-4.2.2.1):
```
If the request fails due to a missing, invalid, or mismatching
redirection URI, or if the client identifier is missing or invalid,
the authorization server SHOULD inform the resource owner of the
error and MUST NOT automatically redirect the user-agent to the
invalid redirection URI.

If the resource owner denies the access request or if the request
fails for reasons other than a missing or invalid redirection URI,
the authorization server informs the client by adding the following
parameters to the query component of the redirection URI using the
"application/x-www-form-urlencoded" format, per Appendix B:
```
We could see that generally there are 2 types of error:
#### 1. Errors that be informed to resource owner:
- missing, invalid, or mismatching redirection URI
- the client identifier is missing or invalid
#### 2. Errors that be informed to client:
- resource owner denies the access request
- errors that different than (1). For example: `invalid_scope`, `unsupported_response_type`

For (1), server could render a page that resource owner could see.
For (2), server informs client by redirect to client's `redirect path` adding error parameters like: `host:port/callback_path?error=invalid_scope...`

### Current doorkeeper implementation:
At [authorizations_controller.rb](https://github.com/doorkeeper-gem/doorkeeper/blob/1a68d15f9392707dd5e0ab3bb1da900fef3494e6/app/controllers/doorkeeper/authorizations_controller.rb#L53):
- At `def create`, when error happened (validity check failed), the `redirect_or_render` method will **handle error properly** like describe above.
- **BUT** at `def new`, when validity check failed (for example `invalid_scope`, which is type 2), `new` will `render :error` screen (inform to resource owner), not by redirect to `redirect_url` with an `error` parameter to inform to client. (client can read the error param and handle it)

linhdangduy:
> I think it must be an error response with invalid_request with error description about missed scope parameter

Currently it easily to render `invalid_code` error like this commit: https://github.com/doorkeeper-gem/doorkeeper/pull/1277/commits/db434a3417470e8df244de2e2d96dca539e7b313

If return `invalid_request`, other param validity error (check at [pre_authorization.rb](https://github.com/doorkeeper-gem/doorkeeper/blob/1a68d15f9392707dd5e0ab3bb1da900fef3494e6/lib/doorkeeper/oauth/pre_authorization.rb)) should be returned by `invalid_request`, too (of course with equivalent error description)

About the `(2) May be this is new issue`, we could discuss this in another thread.

nbulaj:
`invalid_scope` or `invalid_code` ? :thinking: 

As far as I remember `invalid_scope` returns only in case scope is present (so it has a value, `""` or any else string) and it's invalid invalid  or unknown. In case when we **require** scope parameter and user doesn't send it - it's `invalid_request`.

From 4.1.2.1.  Error Response:

> **invalid_request**
>               The request is missing a required parameter, includes an
>               invalid parameter value, includes a parameter more than
>               once, or is otherwise malformed.

linhdangduy:
> In case when we require scope parameter and user doesn't send it - it's invalid_request

Authorization endpoint by default requires `client_id` and `response_type`.

[here at This Test](https://github.com/doorkeeper-gem/doorkeeper/blob/1a68d15f9392707dd5e0ab3bb1da900fef3494e6/spec/controllers/authorizations_controller_spec.rb#L440-L448), the authorization request's `response_type` params is missing, but it return error as `validate` method defined: `unsupported_response_type`, not `invalid_request`.

If change to `invalid_request`, I think *the request that absent error of `client_id` or `response_type`* should be returned by `invalid_request` error, too.

nbulaj:
> here at This Test, the authorization request's response_type params is missing, but it return error as validate method defined: unsupported_response_type, not invalid_request.

It's because RFC has additional error response for such case, `unsupported_response_type`.

linhdangduy:
Yeah, server will return `unsupported_response_type` error response if `response_type` param is present.

As you said for `scope` param, if `scope` is not present, server will return `invalid request`.
So what if user doesn't send `response_type` or `client` param? (Currently server return `unsupported_response_type` or `invalid_client`)

nbulaj:
For Authorization endpoint `response_type` is a required parameter and if it's blank - it must be invalid request (see 3.1.1.  Response Type.  https://tools.ietf.org/html/rfc6749#section-3.1.1 )

> If an authorization request is missing the "response_type" parameter,
> or if the response type is not understood, the authorization server
> MUST return an error response as described in Section 4.1.2.1.

Section 4.1.2.1:

> invalid_request
>              The request is **missing a required parameter**, includes an
>              invalid parameter value, includes a parameter more than
>              once, or is otherwise malformed.
>

If you'll take a look at 4.1.3. Access Token Request section, you'll find that it requires only `grant_type`, `code`, `redirect_uri` and `client_id`  (https://tools.ietf.org/html/rfc6749#section-4.1.3 ).

In such questions I always look at how big popular projects processes OAuth requests, like mentioned Google above.

linhdangduy:
Thank you.
I will try to push the commit that returning `invalid_request` error when required parameter is not sent.

> But scopes for database must be NOT NULL.

Just want to ask should I add this constrain in this PR?

nbulaj:
Yep, let's add it :+1:

linhdangduy:
@nbulaj I have checked error handle of authorization endpoint request and token endpoint request. Then fix them at https://github.com/doorkeeper-gem/doorkeeper/pull/1277/commits/eedfcc8cc6fe5c31ae612a5fe33009d4d5d8a2d8

- For authorization request: 
return `invalid_request` error for missing require param: `response_type`, `client_id`, and `scope` (if server doesn't define `default_scopes` config)

- For token endpoint request:  
  - Three grant flows and their required params: `password`(grant_type, username, password), `client_credentials`(grant_type), `refresh_token`(grant_type, refresh_token) are ok, I do not change anything.
  - `authorization_code` flow (grant_type, code, redirect_uri, client_id): 
    - fix `code`: raise `MissingAuthorizationCode` and return `invalid_request`
    - `client_id`: this param is required only when the client that sent request is `unauthenticated client`, currently Doorkeeper returns `invalid_client` if it does not find any `client` from request param. I think it'ok,

Currently I only return `invalid_request` error, with [**default error_description**](https://github.com/doorkeeper-gem/doorkeeper/blob/1a68d15f9392707dd5e0ab3bb1da900fef3494e6/config/locales/en.yml#L91) of doorkeeper, which doesn't specify what param is missing like Google does (`Missing required parameter: scope`)
if ok, now I will try to return `error_description` like Google does, specify `what required param is missing`.

app/controllers/doorkeeper/authorizations_controller.rb
```diff
@@ -10,11 +10,14 @@ def new
       else
         render_error
       end
+    rescue Errors::DoorkeeperError => e
```
linhdangduy: adding `rescue` for `new` and `create` method as at tokens_controller to handle raised error (invalid_request type error): https://github.com/doorkeeper-gem/doorkeeper/blob/1a68d15f9392707dd5e0ab3bb1da900fef3494e6/app/controllers/doorkeeper/tokens_controller.rb#L9

lib/doorkeeper/oauth/pre_authorization.rb
```diff
@@ -5,6 +5,7 @@ module OAuth
     class PreAuthorization
       include Validations
 
+      validate :attributes, error: :invalid_request
```
nbulaj: Maybe `validate_params` ? Like HTTP request params. Just a "sugar"

lib/doorkeeper/errors.rb
```diff
@@ -38,6 +38,18 @@ def type
       end
     end
 
+    class MissingClientId < DoorkeeperError
```
nbulaj: What if introduce one class like `MissedRequiredParameter` with overridden initializer that receives missed parameter? I think it will be easier to introduce error description and will be more DRY, WDYT ?

```ruby
class MissedRequiredParameter < DoorkeeperError
  def initialize(message, parameter)
    super(message)
    @parameter = parameter
  end

  def type
     :invalid_request
  end
 
  # ...

**Comment by nbulaj** [comment]
> now I will try to return error_description like Google does at next commit, specify what required param is missing.

I think it will be a great addition and improvement for user/developers experience :+1:

File: lib/doorkeeper/oauth/pre_authorization.rb (hunk 3 above)
Comment (linhdangduy) [reply lib/doorkeeper/oauth/pre_authorization.rb]: Indeed at that time, I followed this line
https://github.com/doorkeeper-gem/doorkeeper/blob/25c5bda05df1c03770b8c1e999858942eb8f0eb0/lib/doorkeeper/oauth/authorization_code_request.rb#L6

Comment (nbulaj) [reply lib/doorkeeper/oauth/pre_authorization.rb]: Oh, we can replace it too :)

File: lib/doorkeeper/errors.rb (hunk 4 above)
Comment (linhdangduy) [reply lib/doorkeeper/errors.rb]: Great suggestion, I will try this to inform missing parameter.

File: lib/doorkeeper/oauth/pre_authorization.rb [hunk 5]
```diff
@@ -5,14 +5,16 @@ module OAuth
     class PreAuthorization
       include Validations
 
-      validate :response_type, error: :unsupported_response_type
       validate :client, error: :invalid_client
-      validate :scopes, error: :invalid_scope
       validate :redirect_uri, error: :invalid_redirect_uri
+      validate :params, error: :invalid_request
```
linhdangduy: Change the order of `validation` to make sure that `AuthorizationsController` response properly when error happens, check that: `Does that ErrorResponse redirectable?`

- Because with missing or invalid `client` and `redirect_uri` param, `AuthorizationsController` must not redirect client to `redirect_uri`, that is checked at [here](https://github.com/doorkeeper-gem/doorkeeper/blob/99f83f0dd88a215eb5b5f815f4b2a9e07caf9b5d/lib/doorkeeper/oauth/error_response.rb#L42) and called at [here](https://github.com/doorkeeper-gem/doorkeeper/blob/99f83f0dd88a215eb5b5f815f4b2a9e07caf9b5d/app/controllers/doorkeeper/authorizations_controller.rb#L54). So this order make sure that we have checked `client` and `redirect_uri` first.

lib/doorkeeper/oauth/invalid_request_response.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+module Doorkeeper
+  module OAuth
+    class InvalidRequestResponse < ErrorResponse
```
linhdangduy: Introduce `InvalidRequestResponse` to handle `invalid_request`
- I added explanation for `invalid_request` by two following variable: `missing_parameter` (for missing required parameter) and `invalid_request_reason` (for reason other than missing required parameter: happen at token_introspection endpoint and, authorization_code_request)

lib/doorkeeper/request.rb
```diff
@@ -4,19 +4,22 @@ module Doorkeeper
   module Request
     class << self
       def authorization_strategy(response_type)
+        raise Errors::MissingRequiredParameter, :response_type if response_type.blank?
+
         get_strategy(response_type, authorization_response_types)
-      rescue NameError
-        raise Errors::InvalidAuthorizationStrategy
+      rescue StandardError
+        build_strategy_class("error_response_type")
```
linhdangduy: Introduce `ErrorResponseType` request for `authorization request` in `authorization_code` & `implicit` flow.

Because when `Errors::InvalidAuthorizationStrategy` happens, response is `ErrorResponse`, will be handle at [here](https://github.com/doorkeeper-gem/doorkeeper/blob/99f83f0dd88a215eb5b5f815f4b2a9e07caf9b5d/lib/doorkeeper/helpers/controller.rb#L48). AuthorizationController will check if that response is redirectable? by https://github.com/doorkeeper-gem/doorkeeper/blob/99f83f0dd88a215eb5b5f815f4b2a9e07caf9b5d/lib/doorkeeper/oauth/error_response.rb#L42-L45

So in this case ErrorResponse is always redirectable even `redirect_uri` is invalid because the `name` of error is always be `unsupported_response_type`

nbulaj:
Sorry @linhdangduy , too busy, can't check it. I'll return to this PR ASAP.
P.S. many changes, will be hard to review :)

lib/doorkeeper/oauth/refresh_token_request.rb
```diff
@@ -32,7 +33,7 @@ def load_client(credentials)
       def before_successful_response
         refresh_token.transaction do
           refresh_token.lock!
-          raise Errors::InvalidTokenReuse if refresh_token.revoked?
+          raise Errors::MissingRequiredParameter, :refresh_token if refresh_token.revoked?
```
nbulaj: Actually I don't sure it's a valid behavior to return `MissingRequiredParameter` in case refresh token exist, but revoked. Does RFC say something like that? I think `MissingRequiredParameter` must indicate only errors when parameter is **really** missed

lib/doorkeeper/oauth/error_response_type_request.rb
```diff
@@ -0,0 +1,21 @@
...
+  module OAuth
+    class ErrorResponseTypeRequest
+      attr_accessor :pre_auth
+
+      def initialize(pre_auth)
+        @pre_auth = pre_auth
+      end
+
+      def authorize
+        @response = if pre_auth.error == :invalid_request
```
nbulaj: I see a lot of similar checks in the PR. My opinion is not to ask, but tell (*tell, don't ask*).

So it'd be better to have something like

```ruby
pre_auth.build_error_response
```

And hide this logic under this method or to introduce a `ErrorResponseBuilder` that will encapsulate the logic of building error response in it (this way I like more btw)

lib/doorkeeper/oauth/refresh_token_request.rb
linhdangduy: I fixed it here: https://github.com/doorkeeper-gem/doorkeeper/pull/1277/commits/8fbf46ca61ed7e8f2fa237129e4be31976875ee1

lib/doorkeeper/oauth/authorization_code_request.rb
```diff
@@ -42,11 +39,17 @@ def before_successful_response
         super
       end
 
-      def validate_attributes
-        return false if grant&.uses_pkce? && code_verifier.blank?
-        return false if grant && !grant.pkce_supported? && !code_verifier.blank?
+      def validate_params
+        if grant && !grant.pkce_supported? && code_verifier.present?
```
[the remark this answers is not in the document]
linhdangduy: Refactored: https://github.com/doorkeeper-gem/doorkeeper/pull/1277/commits/f291c2620ec952b025984ba900eef1a048dc3868

lib/doorkeeper/oauth/pre_authorization.rb
```diff
@@ -5,15 +5,17 @@ module OAuth
     class PreAuthorization
       include Validations
 
-      validate :response_type, error: :unsupported_response_type
-      validate :client, error: :invalid_client
-      validate :scopes, error: :invalid_scope
-      validate :redirect_uri, error: :invalid_redirect_uri
+      validate :client,                error: :invalid_client
+      validate :redirect_uri,          error: :invalid_redirect_uri
```
linhdangduy: Because with missing or invalid `client` and `redirect_uri` param, `AuthorizationsController` must not redirect client to `redirect_uri`. So I change the order of `validate`:  **Firstly, checked `client` and `redirect_uri` param**. This order makes sure that ErrorResponse is redirectable or not.

lib/doorkeeper/oauth/error_response_type_request.rb
linhdangduy: Thank for your suggestion. Seem I can decrease the some complicate parts of this pull  
 request.

linhdangduy: @nbulaj 
Indeed `pre_auth` itself has `error_response` method to return ErrorResponse. So I fix and use it to return error_response for above case. 
Besides, I see that current logic of `authorize_response` method at authorizations_controller.rb is a bit confusing, because it continues processing even `pre_auth` is not authorizable.
https://github.com/doorkeeper-gem/doorkeeper/blob/6eda02d7b0e9971a71850aff49737e1cf141d2be/app/controllers/doorkeeper/authorizations_controller.rb#L87-L95
So I just return error_response immediately if pre_auth is not authorizable, so it make sure that the followed processing is always authorize. (not have error)
```ruby
def authorize_response
  @authorize_response ||= begin
    authorizable = pre_auth.authorizable?
    return pre_auth.error_response unless authorizable

    before_successful_authorization
    auth = strategy.authorize
    after_successful_authorization
    auth
  end
end
```
All the fixes are in this commit: https://github.com/doorkeeper-gem/doorkeeper/pull/1277/commits/9eeb70c9ba6b1f12e1caae6863b2eeaa64529205

nbulaj: I like it :)

nbulaj:
I like the changes, @linhdangduy ! Great work :+1: 

But we have a lot of changes here, and I think breaking changes... So I need more time to check them. Also I'm afraid we can break many apps... So I must to make a hard decision..

lib/doorkeeper/request.rb
```diff
@@ -4,30 +4,25 @@ module Doorkeeper
   module Request
     class << self
       def authorization_strategy(response_type)
-        get_strategy(response_type, authorization_response_types)
-      rescue NameError
-        raise Errors::InvalidAuthorizationStrategy
+        build_strategy_class(response_type)
```
nbulaj: Why we replaced it? :thinking: 

Also I see that `token_strategy` do a check for available types, but authorization - not.. I'm kind of confused ...

linhdangduy:
@nbulaj I'm sorry to make you confused with these a lot of changes. 
Maybe some changes should be moved to another pull request, it will be easier to review. If ok, I will do that.

lib/doorkeeper/request.rb
linhdangduy: Because:
- At authorizations_controller **return error immediately** when pre_auth is not authorized:
https://github.com/doorkeeper-gem/doorkeeper/pull/1277/files#diff-1282d211312f992c2a614d1f4a6302feR94-R97
- `authorization_strategy` is called from `authorizations_controller.rb` **only when pre_auth is passed all the validation check (all parameters is valid)**:
    call `authorization_request`
https://github.com/doorkeeper-gem/doorkeeper/blob/6eda02d7b0e9971a71850aff49737e1cf141d2be/app/controllers/doorkeeper/authorizations_controller.rb#L83-L85
    then call `authorization_strategy ` at https://github.com/doorkeeper-gem/doorkeeper/blob/6eda02d7b0e9971a71850aff49737e1cf141d2be/lib/doorkeeper/server.rb#L12

- So I replaced all the error checking and just do `build_strategy_class`.

linhdangduy: `token_strategy` is otherwise called from `tokens_controller.rb`, which doesn't have object like `pre_authorization` to check whether the params are valid or not. So we need to check if token strategy is valid at here.

nbulaj: It's totally OK, thanks for extended clarification :bowing_man:

nbulaj:
Hi @linhdangduy , can we also collect list of changes for the end developers? I mean like _error response changed in case XXX parameter missed_,   &nbsp;&nbsp;  or  &nbsp;&nbsp;  _now authorization is denied on case pre-auth is not authorizable_. I need to prepare such list of changes for upgrade guides for 5.2 release.

linhdangduy:
Hi @nbulaj, I added the change list at [the description](https://github.com/doorkeeper-gem/doorkeeper/pull/1277#issue-289719338) of this pull request.

nbulaj:
Maaan, you awesome :+1: 
I'll check it on weekend. Big thanks! :handshake:

linhdangduy:
@nbulaj After rebasing and fixing conflict based on #1296 , we don't need to handle exception at `app/controllers/doorkeeper/authorizations_controller.rb` anymore.

https://github.com/doorkeeper-gem/doorkeeper/pull/1277/files#diff-1282d211312f992c2a614d1f4a6302feL15

lib/doorkeeper/oauth/refresh_token_request.rb
```diff
@@ -32,7 +33,7 @@ def load_client(credentials)
       def before_successful_response
         refresh_token.transaction do
           refresh_token.lock!
-          raise Errors::InvalidTokenReuse if refresh_token.revoked?
+          raise Errors::InvalidGrantReuse if refresh_token.revoked?
```
nbulaj: What's the reason to change this class? :thinking:

linhdangduy: As I describe at PR's description: `3. When refresh_token is revoked`
because `refresh_token` is revoked, so we should return `invalid_grant` error as [rfc's defination](https://tools.ietf.org/html/rfc6749#section-5.2)
> invalid_grant
The provided authorization grant (e.g., authorization
code, resource owner credentials) or **refresh token is
invalid, expired, revoked,** does not match the redirection
URI used in the authorization request, or was issued to
another client.

we can see that [Errors::InvalidGrantReuse](https://github.com/doorkeeper-gem/doorkeeper/pull/1277/files#diff-f1ea26f732c79d7938f919b79036e55fL23-L25) returns `invalid_grant`, while [InvalidTokenReuse](https://github.com/doorkeeper-gem/doorkeeper/pull/1277/files#diff-f1ea26f732c79d7938f919b79036e55fL17-L21) returns `invalid_request`.

nbulaj: Ahh, sorry man, too much I need to check and do after vacation :bowing_man: 
:+1:

nbulaj:
Thanks @linhdangduy for your awesome contribution <3
I plan to release an RC3 so we can test the gem on existing projects before main release

JalisoCSP:
> > Because scope param of request is not nil, Should the scope of oauth_access_grants model (token is built at authorization request) would never be nil, too !?
> 
> Yep, I think you are right. But now with this breaking change I afraid to break existing clients, actually. But `scopes` for database must be NOT NULL.
> 
> > But with new method (controller method of GET request), the above instruction is not followed. Because it's using render_error method which doesn't have redirect method in it, only render -> Maybe this is new issue?
> 
> Don't sure I understand the issue here

This is a breaking change and is preventing a Rails 6 upgrade for us.

v5.2.0.rc2 appears to be working and fine, but this issue is blocking us from upgrading further.

linhdangduy:
> This issue is a breaking change and is preventing a Rails 6 upgrade for us.
v5.2.0.rc2 appears to be working and fine, but this issue is blocking us from upgrading further.

@JalisoCSP I'm sorry to hear that. Could you describe further what prevent you upgrading Doorkeeper?

1. Is that because of **afraid to break existing clients**? as below:
> > Because scope param of request is not nil, Should the scope of oauth_access_grants model (token is built at authorization request) would never be nil, too !?
> 
> Yep, I think you are right. But now with this breaking change I afraid to break existing clients, actually. But `scopes` for database must be NOT NULL.

With above change:
- if *server configure `default_scopes`*, client's authorization request can have none `scope` parameter.
- if *server doesn't configure `default_scopes`*, client's authorization request is forced to includes `scope` parameter.

2. As nbulaj said: "Don't sure I understand the issue here". The following is not considered a problem to change at this pull request.
> > But with new method (controller method of GET request), the above instruction is not followed. Because it's using render_error method which doesn't have redirect method in it, only render -> Maybe this is new issue?
> 
> Don't sure I understand the issue here 😞

PR #1279, opened 2019-06-24
Use `find_by` instead of `find_by_id` in initializer template

okuramasafumi:
`find_by_id` style is deprecated.


nbulaj:
Thanks @okuramasafumi !

Issue #1245, opened 2019-04-19
Customize grant flows per application: config doorkeeper grant_flows passing a block.


phlegx:
Hi!

Can I pass a block to:

```ruby
Doorkeeper.configure do
  ...
  grant_flows %w[authorization_code client_credentials]
end
```

like in `custom_access_token_expires_in`:

```ruby
Doorkeeper.configure do
  ...
  custom_access_token_expires_in do |context|
    context.client.some_method
  end
end
```

Passing a block to `grant_flows` can help setting up different grant flows per application. Example:

```ruby
Doorkeeper.configure do
  ...
  grant_flows do |client|
    client.available_grant_flows
  end
end
```

nbulaj:
Hi @phlegx . It's impossible at the moment. Also this configuration option is designed for the entire server and knows nothing about aoolications (clients). It defines grant flows supported by your server

phlegx:
Hi @nbulaj. But can be added as feature? 

Yes, the server is one, but in Doorkeeper I can add many applications. Why should every application have all the default grant flows? Single application, single configuration, multiple applications, multiple configurations. I mean, why there are context blocks for `custom_access_token_expires_in` and `use_refresh_token`? Because they are useful to have better config granularity. The same could be very useful for `grant_flows`.

phlegx:
Found issue #1207 with the same feature request!

nbulaj:
Landed in `master` with #1286

Issue #1207, opened 2019-03-01
Feature request: Customize grant types per application


alexpil:
Hello,
We want to define different grant types for each oauth application we have. 
Some solutions for that could be having `config.custom_grant_flows` like in the case of  `config.custom_access_token_expires_in`
 or maybe allowing to accept proc in existing `config.grant_flows`.

Thank you.

nbulaj:
I'm closing this in favor of #1245, @alexpil

PR #1286, opened 2019-07-11
Add support of grant flows per application

nbulaj:
Allows to forbid (or allow) OAuth2 grant flows per client (application).

Implements #1245 , #1207 

```ruby
# Allows to customize OAuth grant flows that +each+ application support.
# You can configure a custom block (or use a class respond to `#call`) that must
# return `true` in case Application instance supports requested OAuth grant flow
# during the authorization request to the server. This configuration +doesn't+
# set flows per application, it only allows to check if application supports
# specific grant flow.
#
# For example you can add an additional database column to `oauth_applications` table,
# say `t.array :grant_flows, default: []`, and store allowed grant flows that can
# be used with this application there. Then when authorization requested Doorkeeper
# will call this block to check if specific Application (passed with client_id and/or
# client_secret) is allowed to perform the request for the specific grant type
# (authorization, password, client_credentials, etc).
#
# Example of the block:
#
#   ->(flow, client) { client.grant_flows.include?(flow) }
#
# In case this option invocation result is `false`, Doorkeeper server returns
# :unauthorized_client error and stops the request.
#
# @param allow_grant_flow_for_client [Proc] Block or any object respond to #call
# @return [Boolean] `true if allow or `false` if forbid the request
#
option :allow_grant_flow_for_client,    default: ->(_grant_flow, _client) { }
```

/cc @phlegx , @alexpil

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ User-visible changes worth mentioning.
 
 - [#1288] Allow to pass attributes to the `Doorkeeper::OAuth::PreAuthorization#as_json` method to customize
   the PreAuthorization response.
+- [#1286] Add ability to customize grant flows per application (OAuth client) (#1245 , #1207)
 - [#1283] Allow to customize base class for `Doorkeeper::ApplicationMetalController` (new configuration
   option called `base_metal_controller` (fix #1273).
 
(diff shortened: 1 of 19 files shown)


taylor-a-barnette:
is `client.grant_flows` correct? it doesn't work for me but `client.application.grant_flows` does.

nbulaj:
Which version of Doorkeeper you're using @taylor-a-barnette ?

taylor-a-barnette:
Sorry, @nbulaj, I was out on a long weekend! Gemfile says `doorkeeper (5.2.1)`.

taylor-a-barnette:
bump :)

PR #1288, opened 2019-07-17
Allow PreAuthorization#as_json method to take optional attributes

eulbat:
### Summary

This PR renames `PreAuthorization#as_json`'s `_options` param to `attributes`.
If a hash is passed in this param, it will merge its data to the pre authorization payload.
If no param is passed, the value defaults to `{}` which allows the call of `as_json` without params.
If a non-hash param is passed, only the original pre authorization payload is returned

_Original message:_
>The `PreAuthorization#as_json` method takes a single param (`_options`) which is unused. Yet, this param is required. This PR sets a default value to this `_options` param: an empty hash.
>
>Is there a reason to not add this default value ?

### DIFF
--- a/lib/doorkeeper/oauth/pre_authorization.rb
+++ b/lib/doorkeeper/oauth/pre_authorization.rb
@@ -42,16 +42,10 @@ module Doorkeeper
         OAuth::ErrorResponse.from_request(self)
       end
 
-      def as_json(_options)
-        {
-          client_id: client.uid,
-          redirect_uri: redirect_uri,
-          state: state,
-          response_type: response_type,
-          scope: scope,
-          client_name: client.name,
-          status: I18n.t("doorkeeper.pre_authorization.status"),
-        }
+      def as_json(attributes = {})
+        return pre_auth_hash.merge(attributes.to_h) if attributes.respond_to?(:to_h)
+
+        pre_auth_hash
       end
 
       private
@@ -100,6 +94,18 @@ module Doorkeeper
       def validate_code_challenge_method
         code_challenge.blank? ||
           (code_challenge_method.present? && code_challenge_method =~ /^plain$|^S256$/)
+      end
+
+      def pre_auth_hash
+        {
+          client_id: client.uid,
+          redirect_uri: redirect_uri,
+          state: state,
+          response_type: response_type,
+          scope: scope,
+          client_name: client.name,
+          status: I18n.t("doorkeeper.pre_authorization.status"),
+        }
       end
     end
   end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ User-visible changes worth mentioning.
 
 ## master
 
+- [#1288] Allow to pass attributes to the `Doorkeeper::OAuth::PreAuthorization#as_json` method to customize
+  the PreAuthorization response.
 - [#1283] Allow to customize base class for `Doorkeeper::ApplicationMetalController` (new configuration
   option called `base_metal_controller` (fix #1273).
 
--- a/spec/lib/oauth/pre_authorization_spec.rb
+++ b/spec/lib/oauth/pre_authorization_spec.rb
@@ -182,18 +182,44 @@ module Doorkeeper::OAuth
         allow(client).to receive(:name).and_return client_name
       end
 
-      let(:json) { subject.as_json({}) }
-
       it { is_expected.to respond_to :as_json }
 
-      it "returns correct values" do
-        expect(json[:client_id]).to eq client_id
-        expect(json[:redirect_uri]).to eq subject.redirect_uri
-        expect(json[:state]).to eq subject.state
-        expect(json[:response_type]).to eq subject.response_type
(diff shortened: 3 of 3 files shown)


nbulaj:
Hi @eulbat . First of all thanks for your interest to the project!

I need to understand what's the reason to add the default value? Do you use this method somewhere outside ot Doorkeeper internals?

eulbat:
Hey @nbulaj, thanks for your fast answer!
We're heavily using Doorkeeper at Aircall, so first thank you for this great library.

We have a custom field (an external id) on our OAuth applications that we'd like to return alongside the PreAuth. In order to do that, we're overriding the AuthorizationsController to merge that custom field to the PreAuth payload, so we're first converting the PreAuth to a hash via the `as_json` method.

The fact that the parameter is not optional is not really a pain, but not passing an empty hash would make the code more readable, so I thought I'd make a PR for it.

Let me know what you think! Happy to discuss it, maybe there's a better way to do that, similar to the custom token response?

nbulaj:
Actually we can add a default value, why not ? 
The only thing I'm wondering is maybe we need to merge the hash in the method body with this `options = {}` ? And actually it's not an options, we don't pass this hash anywhere, so maybe we can just drop the arguments at all?

eulbat:
I'm not sure what the purpose of the `options` param was in the first place. Maybe we can indeed merge it with the PreAuth hash ? We would need to enforce that it's a hash though.
Removing the argument altogether would be good if it's not used or planned to be used, however it would be a breaking change for people already calling it with an argument.
Adding the default value is kind of the "easy fix" IMO.

nbulaj:
It was introduced 2 years ago here: https://github.com/doorkeeper-gem/doorkeeper/commit/00fc9549109ca9a2401b2e0af8046b24cd280b02#diff-5c19299fbbf253b0b137bb78daa96299R39

I don't actually remember if we discussed it, so maybe it was required for some purposes at first, then just left as we see it now. 

I'm for renaming it to something more meaningful, like `attributes`, and merging with original hash. It can safe a few lines of code for those like you who adds some additional data to the pre-auth JSON hash. Just a default value will do nothing except we can call `.as_json` without any arguments. If we introduce this change, it wouldn't break anything because `_options` never used in the method. And for those who monkey-patched the `def as_json` method of `PreAuthorization` - well, just don't do a :monkey: -patches :)

eulbat:
Alright, I can do that. I'll make the changes and let you know when it's ready.
Thanks for your help!

eulbat:
@nbulaj I added an implementation, let me know what you think

lib/doorkeeper/oauth/pre_authorization.rb
```diff
@@ -52,6 +52,7 @@ def as_json(_options)
           client_name: client.name,
           status: I18n.t("doorkeeper.pre_authorization.status"),
         }
+        attributes.is_a?(Hash) ? pre_auth.merge(attributes) : pre_auth
```
nbulaj: We can impove it in the way like:

`pre_auth.merge(attributes.to_h) if attributes.respond_to?(:to_h)`

This will allow any object to be passed that can be converted to a hash, WDYT?

eulbat: I can change it to that, sure. However we still need to return the pre_auth if the attributes don't respond to `to_h` right ?

nbulaj: Sure! Just a tiny improvement to allow any "hashable" object to be passed :)

nbulaj:
Thanks @eulbat for your contribution! <3

nbulaj:
Hi @eulbat . I think we need to revert this PR due to https://github.com/doorkeeper-gem/doorkeeper/issues/1312 :( At least we need to just support `options = nil` without additional merging of extra attrbutes (let's quack like AM Serialization - https://api.rubyonrails.org/classes/ActiveModel/Serializers/JSON.html#method-i-as_json :duck: ). Maybe you wanna to propose a fix PR?

eulbat:
Hi @nbulaj, sure thing, I'll work on a fix PR and let you know when it's done.
Thanks for the ping :)

eulbat:
Hi @nbulaj 
I took a look and managed to reproduce the issue https://github.com/doorkeeper-gem/doorkeeper/issues/1312.
The issue is that the `as_json` method is always called with a hash in params and it's trying to merge it (https://github.com/rails/rails/blob/master/activesupport/lib/active_support/json/encoding.rb#L35). And since `as_json` is called recursively (https://github.com/rails/rails/blob/master/activesupport/lib/active_support/json/encoding.rb#L79-L80), it causes the stack overflow.

My proposal is:
1. Remove the attributes merging in the PreAuthorization's `as_json` method. Instead, simply introduce a default value for the `_options` parameter.
2. Make another PR that introduces a new named parameter to the PreAuthorization's `as_json` method that would be merged with the `pre_auth_hash` payload if possible.

WDYT ?

nbulaj:
The simplest is better for me, so I'm actually fort 1 option. Anybody who need additional data can just merge it to #as_json result

PR #1289, opened 2019-07-23
add rethinkdb to ORM list

stakach:
adds rethinkdb to the ORM list


nbulaj:
Maybe you also wanna to add a link here? https://github.com/doorkeeper-gem/guides

PR #1295, opened 2019-08-01
Fix issue where redirect_uri is nil

cincospenguinos:
Co-authored-by: Kevin Lai <kevin@mavenlink.com>

### Summary

We bumped into an issue with how the `redirect_uri` validation is related to other validations in `PreAuth`. In short, an error response could have been considered redirectable while `@redirect_uri` is nil. This PR patches that issue.

We discovered this issue by writing a test for one of our own Oauth flow endpoints that submitted a request with bad params--specifically, no `state`, no `scope`, no `response_type` and no `redirect_uri`. We discovered that when `PreAuth` is validated, `Doorkeeper::Validations` selects the first error it discovers and submits an `ErrorResponse` object with that validation error as its reason for being invalid. Since it does not check any other case, it assumes that the error response is redirectable even though `redirect_uri` is `nil`.

We have included a spec that exposes this issue, and a simple fix. If there is more that is needed, please let us know and we will be more than happy to contribute.

### DIFF
--- a/lib/doorkeeper/oauth/error_response.rb
+++ b/lib/doorkeeper/oauth/error_response.rb
@@ -41,7 +41,7 @@ module Doorkeeper
 
       def redirectable?
         name != :invalid_redirect_uri && name != :invalid_client &&
-          !URIChecker.oob_uri?(@redirect_uri)
+          !@redirect_uri.nil? && !URIChecker.oob_uri?(@redirect_uri)
       end
 
       def redirect_uri
--- a/spec/lib/oauth/error_response_spec.rb
+++ b/spec/lib/oauth/error_response_spec.rb
@@ -16,6 +16,13 @@ module Doorkeeper::OAuth
       end
     end
 
+    describe 'the redirect uri' do
+      it 'is not redirectable when redirect uri is nil' do
+        response = ErrorResponse.new(error: :some_error, state: nil)
+        expect(response.redirectable?).to eq(false)
+      end
+    end
+
     describe :from_request do
       it "has the error from request" do
         error = ErrorResponse.from_request double(error: :some_error)
--- a/spec/lib/oauth/error_response_spec.rb
+++ b/spec/lib/oauth/error_response_spec.rb
@@ -16,8 +16,8 @@ module Doorkeeper::OAuth
       end
     end
 
-    describe 'the redirect uri' do
-      it 'is not redirectable when redirect uri is nil' do
+    describe "the redirect uri" do
+      it "is not redirectable when redirect uri is nil" do
         response = ErrorResponse.new(error: :some_error, state: nil)
         expect(response.redirectable?).to eq(false)
       end


lib/doorkeeper/oauth/error_response.rb
```diff
@@ -41,7 +41,7 @@ def status
 
       def redirectable?
         name != :invalid_redirect_uri && name != :invalid_client &&
-          !URIChecker.oob_uri?(@redirect_uri)
+          !@redirect_uri.nil? && !URIChecker.oob_uri?(@redirect_uri)
```
jasl: `!URIChecker&.oob_uri?(@redirect_uri)`

nbulaj: Why we afraid that `URIChecker` doesn't have this method and need use safe-navigation operator? :thinking:

nbulaj: I think it's safe enough to call this method with `nil` redirect uri @jasl , because of:

https://github.com/doorkeeper-gem/doorkeeper/blob/59a1227a963b1034bb9df61ad377bb745c34f297/lib/doorkeeper/oauth/helpers/uri_checker.rb#L95-L97

jasl: Oh I just comment that `!@redirect_uri.nil? && !URIChecker.oob_uri?(@redirect_uri)` can refactor to `!URIChecker&.oob_uri?(@redirect_uri)` which more simplify.

nbulaj: Once again - I don't understand why we do a safe-call of `oob_uri?` for `URIChecker`? Safe-navigation operator purposes is to do a safe calls against `nil` objects (_Maybe_ monad for poor people ©), but here `URIChecker` isn't a `nil:NilClass` :thinking: 

Also  @cincospenguinos I prefer `@redirect_uri.present?` over `!@redirect_uri.nil?`

linhdangduy: >  it assumes that the error response is redirectable even though redirect_uri is nil

@cincopenguinos you are right, current doorkeeper's code base will `redirectable even though redirect_uri is nil`. But not only here, it happens with another case, too.

So, I am going to fix it at here by re-ordering validation of pre_auth: https://github.com/doorkeeper-gem/doorkeeper/pull/1277/files#r310095549

cincospenguinos: Awesome. Thank you guys so much for the quick response time! I look forward to that change

cincospenguinos:
Closing this PR because [this change](https://github.com/doorkeeper-gem/doorkeeper/pull/1277/files#r310095549) handles the same issue.

nbulaj:
Thanks @cincospenguinos for your work anyway! We appreciate it :+1:

Issue #1287, opened 2019-07-12
Problem with params on master


georgepalmer:
### Steps to reproduce

I think I've found a problem with the latest, unreleased, version of the gem which results in a strong params error.  We are using the latest version of the gem because we need the base_metal_controller option.

To eliminate our client code we recreated the problem using these steps:
https://github.com/doorkeeper-gem/doorkeeper/wiki/Authorization-Code-Flow  It works when on v5.1 of the gem but not when on master.

When requesting authorisation in the first step of that document we get a strong params error:

    ActionController::UnpermittedParameters (found unpermitted parameter: :client_id)

We believe that's because a client_id is passed through but this commit doesn't permit that param: https://github.com/doorkeeper-gem/doorkeeper/commit/03b1437c6bda10ec74c256b4f8c5c3d98b4e5227  I know that parameter isn't used there but it is elsewhere and that strong param check will still flag it if present.  To test our theory we monkeypatched the gem to allow a client_id through but that then resulted in another error:

    ActionView::Template::Error (undefined method `client' for nil:NilClass):
    <main role="main">
       <p class="h4">
         <%= raw t('.prompt', client_name: content_tag(:strong, class: 'text-info') { @pre_auth.client.name }) %>
       </p>
    
      <% if @pre_auth.scopes.count > 0 %>

At this point we got a bit lost as we don't know the codebase very well and couldn't track down the issue.

Any insight/pointers/sanity checks that could help us track it down?

nbulaj:
Hi @georgepalmer . Thanks for reporting the issue!
I don't have a quick answer, I'm also need to check what's going on in the controller and with params. I let you know if I find something

nbulaj:
Hi @georgepalmer . Could you please check it with rc1 and rc2 version of 5.2 Doorkeeper ? I don't sure it's actually a strong params issue, maybe some other change introduced the problem.

georgepalmer:
@nbulaj I've just tested and rc1 is fine but rc2 has the issue

georgepalmer:
@nbulaj can you reproduce from the steps above?  Or do you nee more info?  It's just entering the authorization url as outlined in step 2 of the wiki page into a browser and it'll hit the issue

nbulaj:
It would be great to have a feature test (cucumber) if it's possible. I don't have a time to setup a project and test it right now :(

nbulaj:
[...]

PR #1296, opened 2019-08-11
Adding client id to strong parameters

linhdangduy:
### Summary

Adding `client_id` parameter to permitted strong parameters.

As the needs of #1287 , in development environment, setting:
```ruby
```
to raise error if unpermitted parameters are sent to controller.
But currently, at `authorizations_controller`, the required `client_id` parameter is used at `server.client_via_uid`, not included in strong parameter list. So we cannot use above setting.

Solving by:
- passing`client_id` to permitted parameters. Then get client at pre_authorization, not by using `server.client_via_uid`.

Comment:
- This changes makes code base easier to be understand.
- By passing `client_id` to pre_authorization, this PR facilitate the changes make in #1277 : easier to handle when `client_id` parameter is missing (not have to adding more annoying `raise` and `rescue` couple -> decrease the codes change)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ User-visible changes worth mentioning.
 
 ## master
 
+- [#1296] Adding client_id to strong parameters.
 - [#1293] Move ar specific redirect uri validator to ar orm directory.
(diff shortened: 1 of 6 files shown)


linhdangduy:
Done squash commits to a single 1️⃣commit.

nbulaj:
Hi @linhdangduy . Thanks for the contribution <3

PR #1297, opened 2019-08-12
Strong params fix for Rails forms

georgepalmer:
Fix for part of https://github.com/doorkeeper-gem/doorkeeper/issues/1287

As a Rails form will include other params such as a commit, utf_8 and authenticity_token we need to pull out only the params we require or we'll get a strong params exception when running Rails with the setting:


georgepalmer:
Closing this off for a PR with squashed commits

Issue #1290, opened 2019-07-26
device code draft


bpieck:
This is not about an issue in doorkeeper. We were requested to implement the Device Authorization Grant draft for oauth: https://tools.ietf.org/html/draft-ietf-oauth-device-flow-15

We currently are in a mode of refactoring. Now I would be interested if you are interested in this flow and would like to see a PR for doorkeeper. If so, we would consider this in our refactoring work. If not - work on our side is less. But of course we'd like to share this work, if anyone is interested in it.

nbulaj:
Looks interesting, @bpieck ! It would be great to have this in Doorkeeper :heart:

bpieck:
Ok, thanks, my colleague Dan Shi will propose a PR in near future :)

Blume73:
I would love to integrate this feature in my application as well @bpieck :)

bpieck:
@Blume73: We have already some implementation here: https://github.com/x-ion-de/doorkeeper
Maybe you like to have a look into how to use it and give some feedback?

ddanshi:
@Blume73 

here is a brief walkthrough for the flow.  Looking forward to your feedback :)

### Migrations

Use generator to create all necessary tables:

```ruby
rails generate doorkeeper:migration
```

To enable device flow, run:

```ruby
rails generate doorkeeper:device_code_grant
```
This generator will modify some existing columns and add some new ones that comply with the device flow.  

### Configurations

In Doorkeeper configuration initializer, enable the device flow:
```ruby
grant_flows %w[urn:ietf:params:oauth:grant-type:device_code]
```

In addition, one can configure the polling interval and format of the use code in initializer.

``` ruby
  # Configurations for device grant flow
  #
  # The polling interval configures how often the client should poll the access token from
  # authorization server. The default value is 5 seconds
  #
  device_code_polling_interval 5
  
  #
  # The user code can be composed of letters, digits and hyphens.
  # For example: "4d-6w" means a user code format with 4 digits and 6 letters joined by a hyphen.
  # In case you need both digits and letters for the user code and hyphen is not used to separate
  # letters and digits, which mismatches the accepted format, the default "4w-4w" will be applied.
  #
  user_code_format '4w-4w'

```

### Usage

1. create applications without Confidential on page /oauth/applications.
2. Send a post to /oauth/authorize_device to obtain an access grant by using the application uid from step 1:

```ruby
POST /oauth/authorize_device
[...]

PR #1299, opened 2019-08-13
add device grant flow

bpieck:
### Summary

As suggested here: https://github.com/doorkeeper-gem/doorkeeper/issues/1290 we implemented the device code grant from draft: https://tools.ietf.org/html/draft-ietf-oauth-device-flow-15

### Other Information

You find documentation about how this grant is used on https://github.com/doorkeeper-gem/doorkeeper/issues/1290 provided by Dan Shi. There are some differences to the suggestions in the draft. We did not want to leave the /oauth namespace in path, for instance. So our device_code authorization url is not /device but /oauth/device. But on the draft the suggestion of /device is just an example.

With: `WITHOUT_DEVICE_CODE=true rspec spec` you can check, that tests are still green, without device code migration (so without running the generator). We wanted to ensure this way, we do not break anything on the standard doorkeeper functionality.

Thanks for looking into our PR. If there is anything to improve, we are glad to improve our implementation.


lib/doorkeeper/config.rb
```diff
@@ -510,6 +530,9 @@ def calculate_authorization_response_types
       types = []
       types << "code"  if grant_flows.include? "authorization_code"
       types << "token" if grant_flows.include? "implicit"
+      if grant_flows.include? "urn:ietf:params:oauth:grant-type:device_code"
```
nbulaj: I believe we have some module with such constants, like OOB NATIVE REDIRECT URI or smth like this. Could you please check and move it there?

lib/doorkeeper/config.rb
```diff
@@ -495,6 +499,22 @@ def option_defined?(name)
...
+    # The method device_code_supported? checks, if the migration to enable device code grant
+    # was generated. Normally you use Doorkeeper::AccessGrant.device_code_supported?, since that
+    # just checks the column existing on Database.
+    # This here is the fallback for cases, where database could not be loaded or migrated already
+    # - as in some cases on routes.
+    # Sorry - I don't know, how to check this on Grape. But since lib/doorkeeper/rails/routes.rb
+    # is for rails exclusively - we currently do not use this method outside of Rails.
+
+    def device_code_supported?
+      return false unless defined?(::Rails)
```
nbulaj: Looks very strange for me. Why do we need to search for migration instead of checking the model for the column? Like it is done for refresh token and some other config options :thinking:

lib/doorkeeper/orm/active_record/access_grant.rb
```diff
@@ -9,12 +9,12 @@ class AccessGrant < ActiveRecord::Base
...
                              optional: true, inverse_of: :access_grants
 
-    validates :resource_owner_id,
-              :application_id,
+    validates :application_id,
               :token,
               :expires_in,
-              :redirect_uri,
               presence: true
+    validates :resource_owner_id,
```
nbulaj: This change looks dangerous for me :skull: And like a breaking change...

lib/doorkeeper/rails/routes.rb
```diff
@@ -40,11 +40,19 @@ def generate_routes!(options)
           map_route(:applications, :application_routes)
           map_route(:authorized_applications, :authorized_applications_routes)
           map_route(:token_info, :token_info_routes)
+          device_flow_routes
```
nbulaj: Could we please rename the method to `map_device_flow_routes` ? I think it more semantically fit here

lib/doorkeeper/rails/routes.rb
```diff
@@ -84,12 +112,10 @@ def introspect_routes(mapping)
       end
 
       def token_info_routes(mapping)
-        routes.resource(
-          :token_info,
-          path: "token/info",
-          only: [:show], as: mapping[:as],
-          controller: mapping[:controllers]
-        )
+        routes.resource :token_info,
```
nbulaj: Any reason to change this? Also the code style of changes above doesn't look the same as for this changed piece of code

lib/doorkeeper/request.rb
```diff
@@ -33,9 +33,17 @@ def token_grant_types
       end
 
       def build_strategy_class(grant_or_request_type)
-        strategy_class_name = grant_or_request_type.to_s.tr(" ", "_").camelize
+        strategy_class_name = if grant_or_request_type.to_s.start_with?(extension_grant_start)
```
nbulaj: I need to think more here... :hourglass:

lib/doorkeeper/config.rb
bpieck: The other issues Dan will respond to. But this was workaround was done by me. I will need to reproduce the issue again. But I am quite sure, that there is no database connection yet, when routes are preloaded. So the big difference between refresh_token and device_code - are the extra routes, that refresh token does not need.

lib/doorkeeper/orm/active_record/access_grant.rb
bpieck: what is removed here - is the requirement of redirect_uri in an Access::Grant. I agree, that it could be changed to
````
validates :redirect_uri, unless: -> { respond_to?(:user_code) && !user_code.blank?}
````
So it only should be required, when it's not a device code grant.

Another change is, that resource_owner_id is only required, when there is no user_code. In any other case, there should never be a user_code. So it should not change any behavior. But maybe you see something else, it could break?

lib/doorkeeper/rails/routes.rb
bpieck: agreed - no reason for this change. Maybe it was something happening on fixing newly introduced rubocop issues.

lib/doorkeeper/request.rb
bpieck: I must admit, that I don't like the specs of oauth about extension grant types - but that's not coming from the draft. So I assume, we cannot easily break this down. Else I would like to use grant_type=device_code. On the other hand, maybe this enables others to add extension grant types, like the saml2 bearer in oauth standard grant_type descriptor.
Just for reference:
https://tools.ietf.org/html/draft-ietf-oauth-v2-22#section-4.5

lib/doorkeeper/config.rb
nbulaj: We definitely need to think about other way to check for supported device_code. 

Also I'm thinking about can we move all this PR in some gem extension? I don't really like the idea of having all in one. It would be great if you can check the `doorkeeper-openid-connect` gem and we could move the logic to a standalone extension. I can create a repo for it and add you as a collaborators for it (if you want to).

ddanshi: that sounds a good idea. We would like to work on it further.

Ashleyanne041585:
> ### Summary
> As suggested here: #1290 we implemented the device code grant from draft: https://tools.ietf.org/html/draft-ietf-oauth-device-flow-15
> 
> ### Other Information
> You find documentation about how this grant is used on #1290 provided by Dan Shi. There are some differences to the suggestions in the draft. We did not want to leave the /oauth namespace in path, for instance. So our device_code authorization url is not /device but /oauth/device. But on the draft the suggestion of /device is just an example.
> 
> With: `WITHOUT_DEVICE_CODE=true rspec spec` you can check, that tests are still green, without device code migration (so without running the generator). We wanted to ensure this way, we do not break anything on the standard doorkeeper functionality.
> 
> Thanks for looking into our PR. If there is anything to improve, we are glad to improve our implementation.
npm install workbox-webpack-plugin --save-dev

pelcasandra:
Any plans to merge this pr anytime soon? Thanks!

nbulaj:
As I already said somewhere above, I think it would be great to move this PR into a standalone project (extension) like doorkeeper-openid_connect in order not to bloat the core OAuth implementation

bpieck:
Sorry, as far as I heard last time, there was to much to do for Dan to work on this. I switched companies - but maybe I still could be of some assistance to build that extension.
Dan: Anything new on your side? Can you get some time to work on this? Or shall I take over?

DanasDeveloper:
Sorry for overlooking your messages. My mailbox was flooded with too much stuff :( 
@bpieck Feel free to take it over. I would like to work on it, but I don't think I can have time for it in November. Hopefully things will get better next Month. Let me know if I can be helpful. Thanks!

bpieck:
Ok, I will have some time this weekend to work on the extension. I will come back to you about the repository @nbulaj, if I have some working state.

bpieck:
Sorry, I will need some longer. I just recognized, the draft this was implemented for, is expired and there is now https://tools.ietf.org/html/rfc8628

I will have to check, how much has changed and maybe change the implementation accordingly :)

nbulaj:
I'll close this for now. If you wanna provide any extension using Device grant flow or need help to make it easier to implement Doorkeeper extensions - please contact me. Thanks for your work!

PR #1300, opened 2019-08-13
Limiting access to attributes of pre_authorization

linhdangduy:
#### Problem
At `pre_authorization.rb`, I see that current code use `attr_accessor` and `attr_writer` with attributes to enable writer (attr_writer) for those attributes. These writers are for using at `pre_authorization_spec.rb` test only. I think this exposure is unnecessary.

#### What does this PR do?
Because `pre_authorization` is built from parameters passed to it, we can use `attributes[:param_name] =` instead of `subject.param_name =` writer at `pre_authorization_spec.rb` test. 
And I also limit the access to pre_authorization's attributes: `attr_accessor` -> `attr_reader`, delete `attr_writer`.

#### (UDPDATE) Additionally
I have noticed that there are bunch of unnecessary exposures at other class in project, not only at pre_authorization. Some of those exposures are used for test, other used for nothing (just change to `attr_reader` and code work normally)

### DIFF
--- a/spec/lib/oauth/pre_authorization_spec.rb
+++ b/spec/lib/oauth/pre_authorization_spec.rb
@@ -6,8 +6,7 @@ module Doorkeeper::OAuth
   describe PreAuthorization do
     let(:server) do
       server = Doorkeeper.configuration
-      allow(server).to receive(:default_scopes).and_return(Scopes.new)
-      allow(server).to receive(:scopes).and_return(Scopes.from_string("public profile"))
+      allow(server).to receive(:default_scopes).and_return(Scopes.from_string("public profile"))
       server
     end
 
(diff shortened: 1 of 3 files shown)


spec/lib/oauth/pre_authorization_spec.rb
```diff
@@ -162,12 +161,12 @@ module Doorkeeper::OAuth
...
 
     it "requires a redirect uri" do
-      subject.redirect_uri = nil
+      attributes[:redirect_uri] = nil
       expect(subject).not_to be_authorizable
     end
 
     describe "as_json" do
-      before { subject.client = client }
+      before { subject.authorizable? }
```
linhdangduy: `pre_authorization` need to run validates to assign client attributes

spec/lib/oauth/pre_authorization_spec.rb
```diff
@@ -107,43 +106,43 @@ module Doorkeeper::OAuth
...
         expect(subject).not_to be_authorizable
       end
     end
 
     it "uses default scopes when none is required" do
       allow(server).to receive(:default_scopes).and_return(Scopes.from_string("default"))
-      subject.scope = nil
-      subject.client = client
+      attributes[:scope] = nil
+      expect(subject).to be_authorizable
```
linhdangduy: The same with https://github.com/doorkeeper-gem/doorkeeper/pull/1300/files#r313330230

linhdangduy:
@nbulaj Thank you <3
How about:
> (UDPDATE) Additionally
I have noticed that there are bunch of unnecessary exposures at other class in project, not only at pre_authorization. Some of those exposures are used for test, other used for nothing (just change to attr_reader and code work normally) (e.g: code_response, issuer.rb...)

May be we can see that as a refactor works later.

PR #1305, opened 2019-08-29
Fix Doorkeeper::ApplicationController inheritance

### LINKED ISSUE #1302: When setting api_only mode to true, ActionController::Base will cause CSRF issues
I've spent several days trying to figure out why my api_only Rails 5 app and api_only Doorkeeper configuration would be throwing CSRF issues.

With some assistance I finally figured out that it was because Doorkeeper, by default, inherits from `ActionController::Base` which will re-add the check for CSRF protections (which are attempted to be disabled in `app/controller/doorkeeper/application_controller.rb`)

I believe it would be useful to either:
1. Change the `base_controller` to `ActionController::API` when `api_only` is set to `true
2. Add a note in the documentation/guide (perhaps adding on to the notes added in #1123 )
3. Add a note to the generated `config/initializers/doorkeeper.rb`, specifically the `api_only` block mentioning that adjustments to the `base_controller` need to be made.

or some combination of the above.

### Steps to reproduce
Create an API only rails application (I'm on 5.2.3). Add doorkeeper in and uncomment `api_only` in the initializer.

After creating an application, POST to /oauth/authorize as laid out in [the API docs](https://github.com/doorkeeper-gem/doorkeeper/wiki/API-endpoint-descriptions-and-examples#post---oauthauthorize)

### Expected behavior
POSTing to /oauth/authorize (with appropriate client_id, client_secret, etc.) should return a code and redirect

Setting api_only mode to true should either change the base_controller or tell me I have to in documetnation.

### Actual behavior
POSTing to /oauth/authorize `ActionController::InvalidAuthenticityToken (ActionController::InvalidAuthenticityToken)` is raised

### System configuration

**Ruby version**: 2.6.3

**Doorkeeper version**: 5.1.0

I think this is somewhat related to the discussions happening in #1110

nbulaj:
Make `Doorkeeper::ApplicationController` to inherit from `ActionController::API` in cases
 when `api_mode` enabled (fixes #1302).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ User-visible changes worth mentioning.
 
 ## master
 
-- [#PR ID] Add your description here.
+- [#1305] Make `Doorkeeper::ApplicationController` to inherit from `ActionController::API` in cases
+  when `api_mode` enabled (fixes #1302).
 
 ## 5.2.0.rc3
 
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,7 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
 gemspec
 
-gem "rails", "~> 6.0.0.rc2"
+gem "rails", "~> 6.0.0"
 
 # TODO: Remove when rspec-rails 4.0 released
 gem "rspec-core", github: "rspec/rspec-core"
--- a/app/controllers/doorkeeper/application_controller.rb
+++ b/app/controllers/doorkeeper/application_controller.rb
@@ -2,7 +2,7 @@
 
 module Doorkeeper
   class ApplicationController <
-    Doorkeeper.configuration.base_controller.constantize
+    Doorkeeper.configuration.resolve_controller(:base)
     include Helpers::Controller
 
     unless Doorkeeper.configuration.api_only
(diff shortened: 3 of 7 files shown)


nbulaj:
@0xtobit could you please take a look?

0xtobit:
This looks great! Thanks for doing this, I was struggling to grok the option DSL to get this working myself.

I would suggest adding this

```diff
diff --git a/lib/generators/doorkeeper/templates/initializer.rb b/lib/generators/doorkeeper/templates/initializer.rb
index 0aa76930..c469a137 100644
--- a/lib/generators/doorkeeper/templates/initializer.rb
+++ b/lib/generators/doorkeeper/templates/initializer.rb
@@ -70,7 +70,8 @@ Doorkeeper.configure do
   # access_token_generator '::Doorkeeper::JWT'

   # The controller Doorkeeper::ApplicationController inherits from.
-  # Defaults to ActionController::Base.
+  # Defaults to ActionController::Base unless api_only is set, which changes the default to
+  # ActionController::API
```
or some similar note to the generated config.

Thanks again! 🙌

nbulaj:
That's a nice point, @0xtobit ! I forgot about it, thanks. I'll add it

Issue #770, opened 2015-12-29
Regenerating secrete tokens


theycallmeswift:
I want to implement a way for users to regenerate their secret keys if they have been exposed.  Here's some pseudo code of what I'm thinking:

```ruby
def regenerate
  @application = Doorkeeper::Application.by_uid(params[:uid])
  @applicaton.secret = Doorkeeper::OAuth::Helpers::UniqueToken.generate

   if @application.save
      # blah
   else
      # blah
   end
end
```

Are there any issues around doing it this way?  Would there be interest in a pull request that added something like this to doorkeeper?

christopherhein:
I like the idea would make implementing something like what Github has where you can easily refresh your application very easy.

Only thing I would do differently would be implementing a `.reset_secret` method on the `Doorkeeper::ApplicationMixin` module and then call `.generate` in there and call `.save` to clean up the controller a bit.

I'm not on Doorkeeper core, but I personally like the idea. :+1:

theycallmeswift:
+1.  I think that's a much better way to implement it in Doorkeeper core.

tute:
> I want to implement a way for users to regenerate their secret keys if they have been exposed.

Your code looks good.

If it were to make it into doorkeeper, it should probably belong in the admin interface, but then I'd worry that less competent developers would needlessly (or without understading the consequences) regenerate tokens, making their client apps fail.

I think this is simple enough (certainly the simplest of the tasks you have to do once keys have been compromised) to not deserve its implementation in doorkeeper.

Thank you!

PR #1315, opened 2019-10-11
Make it easier to reset an application secret

stefanwork:
### Summary

I wanted the ability to regenerate my application secrets when needed (e.g. recurring rotation or in case of compromise).

Currently, it's not easy, even in a `rails console`. You have to run the following:

```ruby
app = Doorkeeper::Application.find_by(uid: "abcdef...")
app.secret = nil
app.send(:generate_secret)
app.save
```

With this change, you can do it like so:

```ruby
app = Doorkeeper::Application.find_by(uid: "abcdef...")
app.generate_secret(true)
app.save
```

By making this method public, application authors can now more easily expose this functionality in controllers, without having to worry as much that the interface will change.

Let me know if you think that exposing this in a controller should be included in this PR. In my opinion, that could be done in a future PR. At least this change removes the need to use `send` to invoke a private method.

Fixes https://github.com/doorkeeper-gem/doorkeeper/issues/770.

### Other Information

I'm considering changing the code to the following. Please let me know what you think the interface should be.

```ruby
    # Resets the secret for this application.
    #
    def reset_secret!
      @raw_secret = UniqueToken.generate
      secret_strategy.store_secret(self, :secret, @raw_secret)
    end

    [...]

    private

    def generate_secret
      return unless secret.blank?
      reset_secret!
    end

```

### DIFF
--- a/lib/doorkeeper/orm/active_record/application.rb
+++ b/lib/doorkeeper/orm/active_record/application.rb
@@ -46,6 +46,18 @@ module Doorkeeper
       AccessGrant.revoke_all_for(id, resource_owner)
     end
 
+    # Generates (or regenerates) the secret for this application.
+    #
+    # @param allow_overwrite [Boolean]
+    #   whether to allow overwriting an existing secret
+    #
+    def generate_secret(allow_overwrite = false)
+      return unless secret.blank? || allow_overwrite
+
+      @raw_secret = UniqueToken.generate
+      secret_strategy.store_secret(self, :secret, @raw_secret)
+    end
+
     # We keep a volatile copy of the raw secret for initial communication
     # The stored refresh_token may be mapped and not available in cleartext.
     #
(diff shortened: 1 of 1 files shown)


nbulaj:
Hi @stefanwork . This sounds totally OK to me. But from my POV it would be better not to touch existing private `generate_secret` method and just introduce a new one called `renew_secret`.

`generate_secret` sounds for me like "hey, return me some string with the secret key", not like "hey, update this AR instance attribute with the new secret key value". So something like `renew_secret` must do exactly this thing. 

Also it would be great if you'll add a specs for Doorkeeper Application model to test this new method:

```ruby

describe "#renew_secret" do
  # ...
end
```

And add a changelog entry.

stefanwork:
Thanks for the review. I am now calling `renew_secret` in `generate_secret` to keep the code DRY, but let me know if you want it to be done differently.

PR #1317, opened 2019-10-15
Fix application references

### LINKED ISSUE #1316: AccessToken.find_or_create_for passed incorrect argument types
[Docs for `AccessToken.find_or_create_for`](https://github.com/doorkeeper-gem/doorkeeper/blob/master/lib/doorkeeper/models/access_token_mixin.rb#L154-L155) point to this method expecting a `Doorkeep::Application` instance. Internal invocations appears to instead pass a `Doorkeeper::OAuth::Client`.

See:
- https://github.com/doorkeeper-gem/doorkeeper/blob/41ab5b20a1aeab237d349ea0aa64023dee3d0ded/lib/doorkeeper/oauth/client_credentials/creator.rb#L16-L19
- https://github.com/doorkeeper-gem/doorkeeper/blob/59a1227a963b1034bb9df61ad377bb745c34f297/lib/doorkeeper/oauth/authorization/token.rb#L57-L63
- https://github.com/doorkeeper-gem/doorkeeper/blob/6a9b09142b2e89ab87751322170ecf6f5dd4fe51/lib/doorkeeper/oauth/base_request.rb#L39-L45

[Delegations exist](https://github.com/doorkeeper-gem/doorkeeper/blob/59a1227a963b1034bb9df61ad377bb745c34f297/lib/doorkeeper/oauth/client.rb#L8) which masks this under most usage scenarios. However, issues arise downstream when attempting to access application attributes that are not included in these delegations. An example of this is [extracting `owner_id` in the couchbase orm](https://github.com/acaprojects/doorkeeper-couchbase/blob/468949de7e6582e922b072b6db02e5d372795a2c/lib/support/orm/couchbase/access_token.rb#L176).

Happy to submit a PR to fix this up, but before doing so - I don't have super deep knowledge of this lib - is this a bug, or is there a reason that client instances are passed instead of applications?

KimBurgess:
Resolves #1316


KimBurgess:
Oh yeah wow, sorry about that. I was a little lazy in this and had only checked functionality in a very narrow scope (outside of full specs). Ruby's lack of type system strikes again :)

I've fixed up a dependency issue needed for the specs to run, but can see the wall of red 'F's you're referring to. Will take a look and see what's needed to resolve.

nbulaj:
Could you please also rebase with the master and squash the commits? Thabks!

KimBurgess:
The issues this introduced aren't quite fixed yet. 3a29ee2 was just for some of the rspec dependencies.

There looks to be a lot of crossover between where `Doorkeeper::OAuth::Client` and `Doorkeep::Application` application are used so need to spend some time untangling this. A quick fix would be to use a nasty hack like checking if `respond_to?(:application)` in calls to `AccessToken.find_or_create_for`, but that feels dirty.

nbulaj:
Any update here?

KimBurgess:
Unfortunately no - I have been extremely short on time lately. Likely to remain that way for the next 2 - 4 weeks.

I'll close this for now, then resubmit when everything is ready to go.

PR #1320, opened 2019-10-18
Call configured authenticate_resource_owner block once per request.

stevehodges:
### Summary

The block configured for `authenticate_resource_owner` is called 2-3 times per request. If that block performs more complex logic or database queries, it unnecessarily reexecutes that logic.

This PR memoizes to execute only one time.


nbulaj:
Thanks @stevehodges !

PR #1326, opened 2019-11-13
Refactor: move response_type check in pre_authorization to a method to be easily to override

linhdangduy:
### Summary

This pull request refactors `error_response` method of `pre_authorization` in authorization request.

In Oauth2, `response_on_fragment=true` when `response_type=token` (which is implicit flow).
But OIDC introduces other `response_type` which also define that `response_on_fragment=true` (response_type = `id_token` or `id_token token`...)

I move the response_type check to `response_on_fragment?` method, so it could be easily override this behavior like:
```ruby
def response_on_fragment?
  super || response_type == "id_token"
end
```

### References:
https://openid.net/specs/openid-connect-core-1_0.html#ImplicitAuthRequest

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ User-visible changes worth mentioning.
 ## master
 
 - [#PR ID] Your PR short description.
+- [#1326] Move response_type check in pre_authorization to a method to be easily to override
 
 ## 5.2.2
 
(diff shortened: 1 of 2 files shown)


linhdangduy:
Haven't find out where to fix to avoid `coverage decreased` error. 👀 
I think that this pull is refactoring only.

nbulaj:
Agree @linhdangduy

PR #1329, opened 2019-11-22
Fix find_in_batches order warning

vinc:
### Summary

This PR fix the following warning that appears in the log when using Doorkeeper with Rails 6: 
```
Doorkeeper::AccessToken : Scoped order is ignored, it's forced to be batch order.
```

As mentioned in the [existing comment](https://github.com/doorkeeper-gem/doorkeeper/blob/v5.2.2/lib/doorkeeper/models/access_token_mixin.rb#L93) of that class, ActiveRecord 5.x - 6.x ignores custom ordering so the gem can't perform a database sort by `created_at`.

The fix is to use the sort only where it's needed to avoid the warning in [find_in_batches](https://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_in_batches).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ User-visible changes worth mentioning.
 
 - [#PR ID] Your PR short description.
 - [#1326] Move response_type check in pre_authorization to a method to be easily to override
+- [#1329] Fix find_in_batches order warning.
 
 ## 5.2.2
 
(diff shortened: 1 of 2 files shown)


nbulaj:
Hi @vinc . Could you please do what bot suggested above?

vinc:
I added the fix in the changelog. Does it warrant a test? I'm not sure how to test that.

nbulaj:
Nono, I tjink it's OK, thanks!

PR #1334, opened 2019-11-28
avoid setting flash type so it does not define extra methods/unexpect…

grosser:
…ed redirect keys

 - gem is not using the redirect keyword argument
 - gem is not using the defined application_secret method ... which has a very misleading name

partially reverts http://github.com/doorkeeper-gem/doorkeeper/pull/1248 ... which added the flash without needing to afaik

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ User-visible changes worth mentioning.
 ## master
 
 - [#PR ID] Your PR short description.
+- [#1334] Remove `application_secret` flash helper and `redirect_to` keyword
 - [#1331] Move redirect_uri_validator to where it is used
 - [#1326] Move response_type check in pre_authorization to a method to be easily to override
 
--- a/app/controllers/doorkeeper/applications_controller.rb
+++ b/app/controllers/doorkeeper/applications_controller.rb
@@ -4,7 +4,6 @@ module Doorkeeper
   class ApplicationsController < Doorkeeper::ApplicationController
     layout "doorkeeper/admin" unless Doorkeeper.configuration.api_only
 
-    add_flash_types :application_secret unless Doorkeeper.configuration.api_only
     before_action :authenticate_admin!
     before_action :set_application, only: %i[show edit update destroy]
 


nbulaj:
I remeber we added it for some reason, try to do git blame to find original PR

grosser:
tests should fail if it was needed ... I'll dig in if they do

grosser:
all good 🎉

grosser:
... also FYI get rid of coverall and use [single_cov](https://github.com/grosser/single_cov) ;)

nbulaj:
Hi @grosser . I don't really understand your concerns

We use this flask (`application_secret`) for showing unencrypted secret after creating application. It is used here on the view: https://github.com/doorkeeper-gem/doorkeeper/blob/f1be81891c3d54a42928c1b9e03c5d6833b8af87/app/views/doorkeeper/applications/show.html.erb#L11

Also it passed using `redirect_to` here:
https://github.com/doorkeeper-gem/doorkeeper/blob/ae5f86ac7aaa10797393c4be79da4296caa7282e/app/controllers/doorkeeper/applications_controller.rb#L39 and for `update` action too. We can't reuse Rails-default one so we need to define our own flash type (check [Rails docs](https://api.rubyonrails.org/classes/ActionController/Flash/ClassMethods.html)).

So I don't see any reason to remove this helpful method

grosser:
both these work without defining an extra `application_secret`
controller-method + helper, so I'd rather remove that call

nbulaj:
Let's try to call @iv-mexx here to discuss

iv-mexx:
As long as the flash with the application secret still works without this (and the tests should cover that), then sure, remove it. 
I guess I did not look deep enough into what `add_flash_types` does and thought it was necessary

nbulaj:
Hi @grosser , could you please check if we have a spec for this case in our test suite? I'm currently on vacation and don't sure when I'd havw a good internet

grosser:
secret is tested at https://github.com/doorkeeper-gem/doorkeeper/blob/59a1227a963b1034bb9df61ad377bb745c34f297/spec/controllers/applications_controller_spec.rb#L170 and that still works

PR #1341, opened 2020-01-13
Fix revoked_on_use with hashed secrets

tritrg:
### Summary

Revocation of tokens on first use doesn't work with enabled `hash_token_secrets`.

### DIFF
--- a/lib/doorkeeper/models/access_token_mixin.rb
+++ b/lib/doorkeeper/models/access_token_mixin.rb
@@ -40,6 +40,20 @@ module Doorkeeper
         find_by_plaintext_token(:refresh_token, refresh_token)
       end
 
+      # Returns an instance of the Doorkeeper::AccessToken
+      # with specific token value
+      #
+      # @param previous_refresh_token [#to_s]
+      #   previous refresh token value (any object that reponds to `#to_s`)
+      #
+      # @return [Doorkeeper::AccessToken, nil] AccessToken object or nil
+      #   if there is no record with such refresh token
+      def find_previous_refresh_token(previous_refresh_token)
+        token = previous_refresh_token.to_s
+
+        find_by(:refresh_token => token) 
+      end
+
       # Revokes AccessToken records that have not been revoked and associated
       # with the specific Application and Resource Owner.
       #
--- a/lib/doorkeeper/models/access_token_mixin.rb
+++ b/lib/doorkeeper/models/access_token_mixin.rb
@@ -40,20 +40,6 @@ module Doorkeeper
         find_by_plaintext_token(:refresh_token, refresh_token)
       end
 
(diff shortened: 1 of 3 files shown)


nbulaj:
Hi! We can't use ActiveRecord method here, we must use `#by_refresh_token` as it ORM-dependent and wouldn't break any doorkeeper ORM extension. Also it uses strategy class to find by refresh token, so maybe the issue actually somewhere in hashing strategy. Could you please check it?

nbulaj:
@oliverguentheroliverguenther  @donbobka do you remember if we have plain refresh tokens even if you use some secrets hashing strategy? I mean do changes from this PR is possible if we use some hashing?

lib/doorkeeper/models/access_token_mixin.rb
```diff
@@ -40,6 +40,20 @@ def by_refresh_token(refresh_token)
...
 
+      # Returns an instance of the Doorkeeper::AccessToken
+      # with specific token value
+      #
+      # @param previous_refresh_token [#to_s]
+      #   previous refresh token value (any object that reponds to `#to_s`)
+      #
+      # @return [Doorkeeper::AccessToken, nil] AccessToken object or nil
+      #   if there is no record with such refresh token
+      def find_previous_refresh_token(previous_refresh_token)
```
nbulaj: Let's call it `by_previous_refresh_token` to be convenient with other methods

donbobka:
@nbulaj If you enabled hashing secrets strategy, then `refresh_token` field stores digest of refresh_token. Why not id?

https://github.com/doorkeeper-gem/doorkeeper/blob/master/lib/doorkeeper/oauth/refresh_token_request.rb#L65
So, means `previous_refresh_token` column will keep digest when hashing enabled and plaintext token when hashing disabled

tritrg:
@nbulaj the naming should not be 'by_previous_refresh_token' since this is searching for token by refresh_token, disregarding hashed/plaintext. Also, @donbobka and I propose to move revoke_previous_refresh_token! from revocable to access_token_mixin.

nbulaj:
>  the naming should not be 'by_previous_refresh_token' since this is searching for token by refresh_token, disregarding hashed/plaintext.

Honestly I didn't understand you. I see that you call `find_by(refresh_token: token)` inside the newly added method method. We use `by_xxx` style across the project for ORM-dependent models API (`by_token`, `by_refresh_token`, `by_uid_and_secret` and so on).

nbulaj:
> Also, @donbobka and I propose to move revoke_previous_refresh_token! from revocable to access_token_mixin.

This makes sense since it related only to tokens, not grants :+1:

nbulaj:
OK, I think I can do some changes myself, so I'll merge this. Thanks @tritrg !

PR #1343, opened 2020-01-16
Fix Ruby 2.7 kwargs warning in InvalidTokenResponse

oshow:
I Run specs and found the following warning in Doorkeeper's code itself.

```
lib/doorkeeper/oauth/invalid_token_response.rb:31:
warning: Using the last argument as keyword parameters is deprecated;
maybe ** should be added to the call
```

Instead of expanding the hash to keyword arguments with
double splat, I have rewritten the hash into keyword arguments.
It's the same as `Doorkeeper::OAuth::InvalidRequestResponse#description`.


nbulaj:
Tested localy and seems like some weird gem broke the specs :(

PR #1345, opened 2020-01-17
Allow to configure Doorkeeper models

### LINKED ISSUE #1327: [Discussion] Revisit approach to let users extend models
In many scenarios an app needs to extend a Doorkeeper::Application  and Doorkeeper::AccessToken model.

I have an app running in production where we needed to extend those for different features, and I know I'm not the first.

It currently feels very hacky that the gem's users have to research proper ways to extending - some will just try monkey-patching and find issues. Others will try module prepend, and then will face issues with Rails reloader.

I think this gem is great, but I think the path that it took to define its models in an extensible way is just not mature as many apps need.

I think the maintainers should consider accepting a proposal for a more extensible approach, found in gems like Devise - you're responsible for providing the model and them we could just include the Doorkeeper's module/concern in there. That would let the users do whatever the want with the model without any hacks.

nbulaj:
Allow Doorkeeper models to be configurable in order to make it easy to extend or override them.

* Extract AR mixins for reuse
* Add config options to define own models
* Define specs
* Add description to initializer template

Fixes #1327

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -50,7 +50,7 @@ Style/DoubleNegation:
 
 Layout/MultilineMethodCallIndentation:
   EnforcedStyle: indented
-Layout/TrailingBlankLines:
+Layout/TrailingEmptyLines:
   Enabled: true
 Layout/DotPosition:
   EnforcedStyle: leading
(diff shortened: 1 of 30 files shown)


rafaelsales:
@nbulaj this is huge! I really appreciate it!

PR #1349, opened 2020-01-20
Ensure `Doorkeeper::Application` AR associations always use `application_id` as their foreign key

moiristo:
# Summary

This should not be implicit as the application model name can be configured via `Doorkeeper.configuration.application_class`.

### Other Information

Connects to doorkeeper-gem/doorkeeper#1327


moiristo:
@nbulaj I tried to create a spec for this, but I couldn't really find a way to set a custom Application class and reload all models to pick up on it. Maybe you know something? However, I have tested that this will correctly query the DB now when using a custom application class.

nbulaj:
@moiristo could you please add a changelog entry and squash commits to a single one? Thanks!

moiristo:
Done!

PR #1354, opened 2020-01-29
Add option to authorize the calling user to access an Application

JoshNorthrup:
### Summary

I've added an option to define whether a user can authorize an application.
My specific use case is the following:
- All users belong to an organization
- Each Application belongs to an organization (via a foreign key on Application)
- I don't want users to be able to authorize an application that belongs to a different organization or see that it even exists for privacy reasons.

This is now achievable with the new option:
```
application_access do |resource_owner, application|
  resource_owner.org_id == application.org_id
end
```

When this validation fails the `invalid_client` error is used so it's indistinguishable from a nonexistent Application

### DIFF
--- a/app/controllers/doorkeeper/authorizations_controller.rb
+++ b/app/controllers/doorkeeper/authorizations_controller.rb
@@ -65,7 +65,11 @@ module Doorkeeper
     end
 
     def pre_auth
-      @pre_auth ||= OAuth::PreAuthorization.new(Doorkeeper.configuration, pre_auth_params)
+      @pre_auth ||= OAuth::PreAuthorization.new(
+        Doorkeeper.configuration,
+        pre_auth_params,
+        current_resource_owner,
+      )
     end
 
     def pre_auth_params
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -280,6 +280,10 @@ module Doorkeeper
            default: {},
            deprecated: { message: "Customize Doorkeeper models instead" }
 
+    # Hook to allow arbitrary user-client authorization
+    option :authorize_resource_owner_for_client,
+           default: ->(_client, _resource_owner) { true }
+
     # Allows to customize OAuth grant flows that +each+ application support.
     # You can configure a custom block (or use a class respond to `#call`) that must
     # return `true` in case Application instance supports requested OAuth grant flow
--- a/lib/doorkeeper/oauth/pre_authorization.rb
+++ b/lib/doorkeeper/oauth/pre_authorization.rb
@@ -7,6 +7,8 @@ module Doorkeeper
 
       validate :client_id,             error: :invalid_request
       validate :client,                error: :invalid_client
+      # The authorize_resource_owner_for_client config option is used for this validation
+      validate :access_to_client,      error: :invalid_client
       validate :redirect_uri,          error: :invalid_redirect_uri
       validate :params,                error: :invalid_request
       validate :response_type,         error: :unsupported_response_type
@@ -15,9 +17,9 @@ module Doorkeeper
       validate :client_supports_grant_flow, error: :unauthorized_client
 
       attr_reader :server, :client_id, :client, :redirect_uri, :response_type, :state,
-                  :code_challenge, :code_challenge_method, :missing_param
+                  :code_challenge, :code_challenge_method, :missing_param, :resource_owner
 
-      def initialize(server, attrs = {})
+      def initialize(server, attrs = {}, resource_owner = nil)
         @server                = server
         @client_id             = attrs[:client_id]
         @response_type         = attrs[:response_type]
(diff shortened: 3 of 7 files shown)


app/controllers/doorkeeper/authorizations_controller.rb
```diff
@@ -70,6 +70,7 @@ def pre_auth
 
     def pre_auth_params
       params.slice(*pre_auth_param_fields).permit(*pre_auth_param_fields)
+        .merge(current_resource_owner: current_resource_owner)
```
nbulaj: I don't like this change just because we merge current user to params. Couldn't we move it in pre_authorization initializer somewhere?

lib/doorkeeper/config.rb
```diff
@@ -269,6 +269,9 @@ def configure_secrets_for(type, using:, fallback:)
     option :grant_flows,                    default: %w[authorization_code client_credentials]
     option :handle_auth_errors,             default: :render
 
+    # Hook to allow arbitrary user-application access enforcement
+    option :application_access,             default: ->(_application, _resource_owner) { true }
```
nbulaj: I would like to rename the option. Maybe something like `allow_resource_owner_to_access_application`, but shorter :grimacing: Can't same exactly, need to think about it

lib/doorkeeper/oauth/pre_authorization.rb
```diff
@@ -7,6 +7,7 @@ class PreAuthorization
 
       validate :client_id,             error: :invalid_request
       validate :client,                error: :invalid_client
+      validate :resource_can_access_client, error: :invalid_client
```
nbulaj: Some other name too, maybe `access_to_client` or similar

app/controllers/doorkeeper/authorizations_controller.rb
JoshNorthrup: I did it this way so I didn't have to change the pre authorization initializer signature. I can make it an optional param to not affect other usages

JoshNorthrup: I don't really like multiple optional parameters though. Perhaps a named parameter? I'll let you decide

lib/doorkeeper/config.rb
JoshNorthrup: I changed it to `resource_owner_application_access`, let me know if you still want to change it

app/controllers/doorkeeper/authorizations_controller.rb
JoshNorthrup: I went with another optional param, the named param doesn't play nice with the `attrs` hash

lib/doorkeeper/oauth/pre_authorization.rb
```diff
@@ -17,7 +18,7 @@ class PreAuthorization
       attr_reader :server, :client_id, :client, :redirect_uri, :response_type, :state,
                   :code_challenge, :code_challenge_method, :missing_param
 
-      def initialize(server, attrs = {})
+      def initialize(server, attrs = {}, resource_owner = nil)
```
nbulaj: It's OK for me just because `attrs` not options, so we can add one more argument

lib/doorkeeper/oauth/pre_authorization.rb
```diff
@@ -26,6 +27,7 @@ def initialize(server, attrs = {})
         @state                 = attrs[:state]
         @code_challenge        = attrs[:code_challenge]
         @code_challenge_method = attrs[:code_challenge_method]
+        @resource_owner        = resource_owner
```
nbulaj: Let's add it also to attr_readers, maybe we wanna to access it in the future for some reason

lib/doorkeeper/orm/active_record/mixins/application.rb
```diff
@@ -72,6 +72,10 @@ def as_json(options = {})
         hash
       end
 
+      def viewable_to_resource_owner?(resource_owner)
```
nbulaj: `authorized_for_resource_owner?`

app/controllers/doorkeeper/authorizations_controller.rb
nbulaj: >  Perhaps a named parameter? I'll let you decide

For retro-compatibility we _must_ use optional parameter here. Maybe in some major version update I'll refactor this

lib/doorkeeper/config.rb
```diff
@@ -274,6 +274,10 @@ def configure_secrets_for(type, using:, fallback:)
            default: {},
            deprecated: { message: "Customize Doorkeeper models instead" }
 
+    # Hook to allow arbitrary user-application access enforcement
+    option :resource_owner_application_access,
```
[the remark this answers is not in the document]
JoshNorthrup: I'm all for consistency 👍

PR #1355, opened 2020-01-31
Introduce polymorphic resource owner

### LINKED ISSUE #233: Resource owner in AccessToken
Why isn't there a belongs_to declaration on AccessToken resource owner? I'm assuming this is to support polymorphic resource owners, but why isn't the field declared as such? (along with a resource_owner_type column). This would allow to define a real AR assoc on top of the relation and make accessing the object a little less painful.

### LINKED ISSUE #456: Multiple User Models
Hello,

I have a centralized user application with doorkeeper and devise. There are two user tables in this application. Tenant specific users and Admin users that can manage the tenant specific users.

Both should be able to authenticate client apps. I see that the resource_owner_id is set in the access_grants and access_tokens.

The problem is that for multiple user models you get conflicts because my admin user has id = 1 and my tenant user has id = 1 and they are both registered for the same client app.

the solution is to make the resource_owner_id polymorphic so multiple user models can be used or at least a resource_owner_type should be set.

### LINKED ISSUE #651: Polymorphic ResourceOwner support
Hey,

My App needs to support authentication of both `User` and `Operator` model (we use two different Devise scopes). 

It would be great if we could add a `resource_owner_type` field to `Doorkeeper::AccessToken` for supporting a polymorphic association.
Owner type could be passed as an optional argument on the `POST /oauth/token?owner_type=operator` request if multiple owner type are supported.

It could be related to https://github.com/doorkeeper-gem/doorkeeper/pull/461, but I would gladly fire up a new PR for that.

Thoughts? 

PS: Thanks for this awesome gems :heart:!

### LINKED ISSUE #1069: Polymorphic user model associations
Just like in this [this SO question](https://stackoverflow.com/questions/25834404/doorkeeper-resource-owner-credential-flow-for-multiple-models), I'm using doorkeeper scopes to hand out access tokens for multiple models.

The problem with the accepted answer there is that doorkeeper doesn't persist the information of the model on the token, causing problems with scoping of the `resource_owner_id` column in methods like [`AccessToken.revoke_all_for`](https://github.com/doorkeeper-gem/doorkeeper/blob/master/lib/doorkeeper/models/access_token_mixin.rb#L47).

Are there any plans to support a polymorphic user model association within doorkeeper? Would you accept a pull request towards to accomplish this? My plan would be to just add a `resource_owner_type` column, adapt the access token mixin and set `polymorphic: true` on the ActiveRecord association. I'd need to investigate the other ORM integrations separately on how to accomplish this but I assume that they have similar features.

nbulaj:
Adds long-awaited ability to set polymorphic Resource Owner for Access Token & Grant (and not just it's ID).

Fixes #1069
Fixes #651
Fixes #456
Fixes #233

## TODO

* [x] Add configuration option to enable polymorphic resource owner
* [x] Refactor existing code base to use resource owner instance and no just it's ID
  * [x] Make it backward compatible as much as possible
* [x] Add generator for new migration with required columns
* [x] Document changes, add to initializer
* [ ] Cover it all with specs

## How to use it

Enable configuration option:

```ruby
Doorkeeper.configure do
  orm :active_record

  use_polymorphic_resource_owner
end
```

Generate migration to add polymorphic columns to tables:

```bash
$ bundle exec rails g doorkeeper:enable_polymorphic_resource_owner
```

If you previously set foreign key for `:resource_owner_id` column - you need to drop it as now it would be a polymorphic association. Add it to the migration generated above.

Run the migration:

```bash
$ bundle exec rails db:migrate
```

Use it! :cowboy_hat_face: 

:warning: :warning:  **[ATTENTION]**: If you enabled this option on existing project - you need to manually set polymorphic type column with the class of the Resource Owner. Otherwise your existing users/clients could face issues and spam you logs with errors.

:warning: :warning:  **[ATTENTION]**: Also this feature could be a breaking change for those users that patched Doorkeeper internals, specially models. You need to fix your patches to user `resource_owner` instance instead of it's ID (like it was before).

## Known projects affected

* GitLab (https://gitlab.com/gitlab-org/gitlab-foss/blob/master/config/initializers/doorkeeper.rb#L129)
* Doorkeeper ORM extensions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ User-visible changes worth mentioning.
 
 ## master
 
+- [#1355] Allow to enable polymorphic Resource Owner association for Access Tokens/Grants
+  (`polymorphic_resource_owner` configuration option).
 - [#1356] Invalidate duplicated scopes for Access Tokens and Grants in database.
 - [#1357] Fix `Doorkeeper::OAuth::PreAuthorization#as_json` method causing 
   `Stack level too deep` error with AMS (fix #1312).
--- a/app/controllers/doorkeeper/authorizations_controller.rb
+++ b/app/controllers/doorkeeper/authorizations_controller.rb
@@ -44,7 +44,7 @@ module Doorkeeper
     def matching_token?
       AccessToken.matching_token_for(
         pre_auth.client,
-        current_resource_owner.id,
+        current_resource_owner,
         pre_auth.scopes,
       )
     end
--- a/lib/doorkeeper.rb
+++ b/lib/doorkeeper.rb
@@ -66,6 +66,7 @@ require "doorkeeper/models/concerns/reusable"
 require "doorkeeper/models/concerns/revocable"
 require "doorkeeper/models/concerns/accessible"
 require "doorkeeper/models/concerns/secret_storable"
+require "doorkeeper/models/concerns/resource_ownerable"
 
 require "doorkeeper/models/access_grant_mixin"
 require "doorkeeper/models/access_token_mixin"
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -157,6 +157,12 @@ module Doorkeeper
         @config.instance_variable_set(:@api_only, true)
       end
 
+      # Enables polymorphic Resource Owner association for Access Grant and
+      # Access Token models. Requires additional database columns to be setup.
+      def use_polymorphic_resource_owner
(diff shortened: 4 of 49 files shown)


jturkel:
It's so awesome to see this supported out of the box! Are there any plans to support restricting oauth applications by resource owner type e.g. one set of oauth applications for "users" and a disjoint set for "operators"? I guess #1354 would make this possible with an appropriately configured block.

nbulaj:
OK, I'll try to add some specs later, need to think more about how to reload mixins and models that depends on Doorkeeper config (and polymcorphic resource owner option).

PR #1365, opened 2020-02-27
use unshift instead of push for updating LOAD_PATH

pravi:
When another version is already installed the current implementation will prefer that instead of local version. This means a wrong version of the gem gets added to gemspec file when building from github source repo (for example when creating a debian package).


nbulaj:
Thanks @pravi !

PR #1368, opened 2020-03-04
pass auth on after_successful_authorization

ericsullivan:
### Summary

I needed to correlate the session_id with an AccessGrant and then copy it to the AccessToken (to support back-channel logout). I added an sid column to both tables, but could not set it's value in 'after_successful_authorization', as the auth objects were not available. This change passes those auth objects to the callback so they can be modified.

### Other Information

An example of using this code to set the sid:

```ruby
after_successful_authorization do |controller, auth|
  case auth
  when Doorkeeper::OAuth::CodeResponse
    access_grant = auth.auth.token
    access_grant.sid = controller.session[:session_id]
    access_grant.save!

  when Doorkeeper::OAuth::TokenResponse
    case grant_type = controller.params[:grant_type]
    when 'authorization_code'
      # Copy the session id from the Doorkeeper::AccessGrant to the
      # Doorkeeper::AccessToken (to support single-sign-out)
      access_grant = controller.send(:strategy).request.grant

      access_token = auth.token
      access_token.sid = access_grant.sid
      access_token.save!
    when 'refresh_token'
      raise 'TODO'
    else
      raise "Not Implemented for grant_type: #{grant_type}"
    end
  else
    raise "Not Implemented for class: #{auth.class}"
  end
end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,18 +10,19 @@ User-visible changes worth mentioning.
 - [#1354] Add option to authorize the calling user to access an application.
 - [#1355] Allow to enable polymorphic Resource Owner association for Access Token & Grant
   models (`use_polymorphic_resource_owner` configuration option).
-  
+
   **[IMPORTANT]** Review your custom patches or extensions for Doorkeeper internals if you
   have such - since now Doorkeeper passes Resource Owner instance to every objects and not
   just it's ID. More info could be found in GitHub PR description.
-  
+
 - [#1356] Remove duplicated scopes from Access Tokens and Grants on attribute assignment.
-- [#1357] Fix `Doorkeeper::OAuth::PreAuthorization#as_json` method causing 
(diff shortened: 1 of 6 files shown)


nbulaj:
Hi @ericsullivan . The problem with such changes is that always somebody needs more info from the hooks. But we can't increase amount of arguments everytime :( We need to balance between codebase and developers needs.

I have some thoughts on how to make it more generic so everybody will be happy, but it's hard to describe it here. If you don't mind I would be happy to propose a PR instead of this one that will allow to use the hook you need

nbulaj:
Hi @ericsullivan . I've added desired behavior in #1384. Now you have a generic context object with attributes you need in hooks:

```ruby
after_successful_authorization do |controller, context|
  case context.auth
    # ...
end
```

Thanks for your idea and contribution :handshake:

ericsullivan:
@nbulaj That PR is exactly what I was going to suggest after reading your feedback. Thank you for the quick response.

PR #1370, opened 2020-03-06
Fix token revocation endpoint behavior to conform with RFC 7009

### LINKED ISSUE #1362: Attempting to revoke an invalid token responds with an error
### Steps to reproduce
When making a properly formed revocation request (`POST /oauth/revoke`) with an invalid token, a 403 is returned.  This seems to disagree with https://tools.ietf.org/html/rfc7009#section-2.2 which states 
```
The authorization server responds with HTTP status code 200 if the
token has been revoked successfully or if the client submitted an
invalid token.
```

### Expected behavior
Given a properly formed request with an invalid token, `TokensController#revoke` should respond with 200

nbulaj:
Fix token revocation endpoint behavior & response for invalid token and unauthorized requests
to conform with RFC 7009.

Fixes #1362.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,11 @@ User-visible changes worth mentioning.
   in external extensions.
 - [#1360] Increase `matching_token_for` lookup size to 10 000 and make it configurable.
 - [#1371] Fix controllers to use valid classes in case Doorkeeper has custom models configured.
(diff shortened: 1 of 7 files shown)


JoshNorthrup:
The spec changes make sense to me 👍 . Thanks for this!

PR #1374, opened 2020-03-12
Revoke and issue client credentials token in a transaction with a row lock

ryansch:
### Summary

It's possible for concurrent requests for client credentials tokens to result in undefined behaviour when using `revoke_previous_client_credentials_token`. I've adapted the work in #712 to apply to client credentials tokens as well. This wraps the revocation and subsequent token creation in a transaction with a row lock on the old token.

### DIFF
--- a/spec/lib/oauth/client_credentials/creator_spec.rb
+++ b/spec/lib/oauth/client_credentials/creator_spec.rb
@@ -18,9 +18,12 @@ class Doorkeeper::OAuth::ClientCredentialsRequest
     end
 
     context "when reuse_access_token is true" do
+      before do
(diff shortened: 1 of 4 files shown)


spec/lib/oauth/client_credentials/creator_spec.rb
```diff
@@ -46,7 +48,7 @@ class Doorkeeper::OAuth::ClientCredentialsRequest
           expect(result).to eq(existing_token)
         end
 
-        context "and when revoke_previous_client_credentials_token is true" do
+        context "and when revoke_previous_client_credentials_token is false" do
```
morganick: 👍

PR #1386, opened 2020-03-30
Feat/expires in token on resource

Paradoxor:
feat: enable resource based expiring tokens

### Summary

We extended access_token_expires_in to also accept the resource owner id. This enables the Gem to have expiring timings based on each resource. For example admins can be logged in for x days.


nbulaj:
Hi @Paradoxor . Actually for such purposes we have `custom_access_token_expires_in` config option.

Also the problem with such changes is that every developer always finds a use case where he/she need something co super customize. But we (a gem) can't consider every developers need, so it's always hard to bloat the gem with such features.

nbulaj:
Hi @Paradoxor . If you're still interested in this issue, I recommend you to extend `build_context` method and use `custom_access_token_expires_in` as I mentioned above. You will need to add a resource owner instance (not ID) here https://github.com/doorkeeper-gem/doorkeeper/blob/c20947c562637f5681e628e2210561f8b157ad9d/lib/doorkeeper/oauth/authorization/context.rb#L6

nbulaj:
Implemented in #1421

PR #1391, opened 2020-04-02
Add a note for backward incompatible change to CHANGELOG [ci skip]

giraffate:
### Summary

In #1296, `Doorkeeper::Server#client_via_uid` was removed. However,
this method looks public. If this method is used, `NoMethodError`
happens when upgrading.


nbulaj:
Actually it's internal class so I don't sure why somebody can use it. I understand when it's models and controllers, but monkey-patches for gem internal tools always is a bad idea :see_no_evil:  

But why not - let's add it. Thanks!

PR #1397, opened 2020-04-11
Require params using configured application model's param key

s-andringa:
### Summary

In https://github.com/doorkeeper-gem/doorkeeper/pull/1371 controllers are updated to use the configured models. In my app I have configured a custom application model named `Api::Application`. This causes the application forms to render their input fields with names such as `api_application[name]` instead of the default `doorkeeper_application[name]`. However, the application controller still tries to require the `:doorkeeper_application` param, which in my case causes an ActionController::ParameterMissing exception to be raised when I try to submit an application form.

This PR fixes this by fixing the application form's input field prefix using form_for's `:as` option.  

### Other Information

I haven't updated CHANGELOG.md because I guess this change really belongs to https://github.com/doorkeeper-gem/doorkeeper/pull/1371, which already adds a proper CHANGELOG entry.

### Workaround

I worked around the issue by overriding my application model's `.model_name` class method to return `Doorkeeper::Application.model_name`. But this might not be an option for everyone.

Your review would be appreciated, thanks!

### DIFF
--- a/app/views/doorkeeper/applications/_form.html.erb
+++ b/app/views/doorkeeper/applications/_form.html.erb
@@ -1,4 +1,4 @@
-<%= form_for application, url: doorkeeper_submit_path(application), html: { role: 'form' } do |f| %>
+<%= form_for application, url: doorkeeper_submit_path(application), as: :doorkeeper_application, html: { role: 'form' } do |f| %>
   <% if application.errors.any? %>
     <div class="alert alert-danger" data-alert><p><%= t('doorkeeper.applications.form.error') %></p></div>
   <% end %>


app/controllers/doorkeeper/applications_controller.rb
```diff
@@ -88,10 +88,14 @@ def set_application
     end
 
     def application_params
-      params.require(:doorkeeper_application)
+      params.require(application_param_key)
         .permit(:name, :redirect_uri, :scopes, :confidential)
     end
 
+    def application_param_key
+      Doorkeeper.config.application_model.model_name.param_key
```
nbulaj, line 96: This actually too strict to ActiveRecord / ActiveModel and could require additional changes for any Doorkeeper ORM (Mongoid, Sequel, etc).

Maybe it's better to use `:as` option of `form_for` helper on the application form view to set a strict parameter name (`doorkeeper_application`) ?

https://api.rubyonrails.org/v6.0.2.2/classes/ActionView/Helpers/FormHelper.html#method-i-form_for

s-andringa: Yes, that's a better option. Thanks, I'll update the code.

s-andringa: ✔️

nbulaj:
Thanks @s-andringa !

PR #1400, opened 2020-04-15
Yield application to allow_grant_flow_for_client? client credentials …

### LINKED ISSUE #1398: `client` yielded to `allow_grant_flow_for_client` config can either be a client or an application object
### Steps to reproduce

When setting `allow_grant_flow_for_client` in the doorkeeper config, the `client` that is yielded can be of two different types.

* https://github.com/doorkeeper-gem/doorkeeper/blob/v.5.3.1/lib/doorkeeper/oauth/pre_authorization.rb#L36 yields the application.
* https://github.com/doorkeeper-gem/doorkeeper/blob/v.5.3.1/lib/doorkeeper/oauth/client_credentials/validator.rb#L29 yields the client.
* https://github.com/doorkeeper-gem/doorkeeper/blob/v.5.3.1/lib/doorkeeper/oauth/password_access_token_request.rb#L53 yields the client.

Is this intentional? It means when configuring the lamba, we must deal with the differing types.

This is using the latest released version 5.3.1. (links above to the code in 5.3.1)

Thanks

aclemons:
…validation

Closes #1398

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ User-visible changes worth mentioning.
 - [#1395] Fix `NameError: uninitialized constant Doorkeeper::AccessToken` for Rake tasks.
 - [#1397] Add `as: :doorkeeper_application` on Doorkeeper application form in order to support
   custom configured application model.
+- [#1400] Correctly yield the application to allow_grant_flow_for_client? (Fixes #1398)
 
 ## 5.4.0.rc1
 - [#1366] Sets expiry of token generated using `refresh_token` to that of original token. (Fixes #1364) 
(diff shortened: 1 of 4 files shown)


nbulaj:
Something wrong with the specs.

We had a case in the past when hooks could have Application instance or Doorkeeper::OAuth::Client, so maybe it makes sense to use this approach too

Maybe this one or somewhere else:

https://github.com/doorkeeper-gem/doorkeeper/blob/50efdb039110a6d8c2730896c619f31ded9942bc/lib/doorkeeper/oauth/authorization/token.rb#L11-L17

nbulaj:
Ah, OK, I've checked ./spec/lib/oauth/client_credentials_integration_spec.rb:11 and seems like it's incorrect.

You could just use this one:

```ruby
let(:client) { Doorkeeper::OAuth::Client.new(FactoryBot.build_stubbed :application) }
```

And spec must pass

aclemons:
Resolved the changelog merge conflict and updated the integration spec now.

nbulaj:
Thanks @aclemons !

PR #1404, opened 2020-05-06
Make Doorkeeper::Application#read_attribute_for_serialization public.

irphilli:
### Summary

Making ` Doorkeeper::Application#read_attribute_for_serialization` public allows for gem users to create their own `ActiveModel::Serializer` implementations.

Otherwise, you run into something like:

```
NoMethodError (private method `read_attribute_for_serialization' called for #<Doorkeeper::Application:0x00007fcdc2844468>
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ User-visible changes worth mentioning.
 ## master
 
 - [#PR ID] Your PR description.
+- [#1404] Make `Doorkeeper::Application#read_attribute_for_serialization` public
 
 ## 5.4.0.rc2
 
--- a/lib/doorkeeper/orm/active_record/mixins/application.rb
+++ b/lib/doorkeeper/orm/active_record/mixins/application.rb
@@ -88,6 +88,17 @@ module Doorkeeper::Orm::ActiveRecord::Mixins
         Doorkeeper.configuration.authorize_resource_owner_for_client.call(self, resource_owner)
       end
 
+      # We need to hook into this method to allow serializing plan-text secrets
+      # when secrets hashing enabled.
+      #
+      # @param key [String] attribute name
+      #
+      def read_attribute_for_serialization(key)
+        return super unless key.to_s == "secret"
+
+        plaintext_secret || secret
+      end
+
       private
 
       def generate_uid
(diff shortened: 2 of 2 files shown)


irphilli:
I have my own serializer. Can reproduce even with something simple like:
```
module Doorkeeper
  class ApplicationSerializer < ActiveModel::Serializer
    attributes :id
  end
end
```

Now, when using this serializer, I get an error along the lines of:
```
NoMethodError (private method `read_attribute_for_serialization' called for #<Doorkeeper::Application:0x00007f8d40086d60>
Did you mean?  read_attribute_for_validation):
    /Users/phil/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/activemodel-5.2.4.2/lib/active_model/attribute_methods.rb:427:in `method_missing'
    /Users/phil/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/active_model_serializers-0.8.3/lib/active_model/serializer.rb:99:in `block in attribute'
    (eval):3:in `_fast_attributes'
```

Seemed like the addition of the private method `read_attribute_for_serialization` was the culprit: https://github.com/doorkeeper-gem/doorkeeper/commit/25d038022c2fcad45af5b73f9d003cf38ff491f6#diff-db8e1aa900c1af4541224aa0127ae853R143 Before, `ActiveModel::Serializer` was calling its own private method, so it was OK. Interestingly enough, in the source (https://github.com/rails/rails/blob/064338998fbe5374b7757e37065f2717a1fbfecd/activemodel/lib/active_model/serialization.rb#L149-L166), the commented code example does not declare the overridden function as private.

Alternatively (short of monkey patching), I can refactor to use `as_json` instead of a custom serializer.

nbulaj:
The strange thing is that AMS calling `read_attribute_for_serialization` as a public method while it is actually private in AM::Serialization:

https://www.rubydoc.info/gems/active_model_serializers/ActiveModel%2FSerializer:read_attribute_for_serialization

I don't really like to expose a lot of API to the public, especially those intended for internal usage, but here I think it's reall makes sense

nightpool:
can we get this rolled into a release? it's preventing us from fixing the doorkeeper security bug in Mastodon.

nbulaj:
Released with 5.3.3 / 5.2.6 / 5.1.2

Issue #1411, opened 2020-05-19
Ignore code_verifier for non-PKCE grants


toupeira:
### Summary

Doorkeeper rejects non-PKCE authorization requests if they include a `code_verifier` param:

```
invalid_request (The request is missing a required parameter, includes an unsupported parameter value, or is otherwise malformed.)
```

This does not conform with the spec, see the very last paragraph at https://tools.ietf.org/html/rfc6749#section-3.2 :

> The authorization server MUST ignore unrecognized request parameters.

### Background

This was originally reported at https://github.com/panva/node-openid-client/issues/247 and https://gitlab.com/gitlab-org/gitlab/-/issues/213643.

### Proposal

Only check the presence of `code_verifier` for PKCE grants:

```patch
diff --git i/lib/doorkeeper/oauth/authorization_code_request.rb w/lib/doorkeeper/oauth/authorization_code_request.rb
index 7511dee5..124f823f 100644
--- i/lib/doorkeeper/oauth/authorization_code_request.rb
+++ w/lib/doorkeeper/oauth/authorization_code_request.rb
@@ -56,7 +56,7 @@ module Doorkeeper
       end
 
       def validate_pkce_support
-        @invalid_request_reason = :not_support_pkce if grant &&
+        @invalid_request_reason = :not_support_pkce if grant&.uses_pkce? &&
                                                        !pkce_supported? &&
                                                        code_verifier.present?
 ```

I'm not sure if we still need the last condition (`code_verifier.present?`), it seems we should reject all PKCE grants if `pkce_supported?` is false?

Happy to submit a PR for this, but wanted to hear your thoughts first.

nbulaj:
Hi @toupeira ! Thanks for reporting this!

Yep, makes sense to me . As well as your point about last condition. I remember @bpieck done a lot of work on PKCE so I'll ping him just in case

bpieck:
Looks totally reasonable to me.

toupeira:
@nbulaj @bpieck thanks for your input!

I wanted to implement this but then saw that `uses_pkce?` also checks `pkce_supported?`, so my suggested condition would actually never be true: https://github.com/doorkeeper-gem/doorkeeper/blob/4dd7c514e40eb8aea196f1bcf219ab234fc2e89e/lib/doorkeeper/models/access_grant_mixin.rb#L18

[...]

PR #1415, opened 2020-05-21
Ignore PKCE params for non-PKCE grants

toupeira:
# Summary

This aligns the behavior with sections 3.1 and 3.2 from RFC 6749:

> The authorization server MUST ignore unrecognized request parameters.

https://tools.ietf.org/html/rfc6749#section-3.1
https://tools.ietf.org/html/rfc6749#section-3.2

### Other Information

Closes https://github.com/doorkeeper-gem/doorkeeper/issues/1411

### DIFF
--- a/lib/doorkeeper/oauth/authorization_code_request.rb
+++ b/lib/doorkeeper/oauth/authorization_code_request.rb
@@ -3,7 +3,6 @@
 module Doorkeeper
   module OAuth
     class AuthorizationCodeRequest < BaseRequest
-      validate :pkce_support, error: :invalid_request
       validate :params,       error: :invalid_request
       validate :client,       error: :invalid_client
       validate :grant,        error: :invalid_grant
@@ -55,14 +54,6 @@ module Doorkeeper
         Doorkeeper.config.access_grant_model.pkce_supported?
       end
 
-      def validate_pkce_support
-        @invalid_request_reason = :not_support_pkce if grant &&
-                                                       !pkce_supported? &&
-                                                       code_verifier.present?
-
-        @invalid_request_reason.nil?
-      end
-
       def validate_params
         @missing_param = if grant&.uses_pkce? && code_verifier.blank?
                            :code_verifier
@@ -93,8 +84,8 @@ module Doorkeeper
       # if either side (server or client) request PKCE, check the verifier
       # against the DB - if PKCE is supported
       def validate_code_verifier
-        return true unless grant.uses_pkce? || code_verifier
-        return false unless pkce_supported?
+        return true unless pkce_supported?
+        return grant.code_challenge.blank? if code_verifier.blank?
 
         if grant.code_challenge_method == "S256"
           grant.code_challenge == generate_code_challenge(code_verifier)
--- a/spec/requests/flows/authorization_code_spec.rb
+++ b/spec/requests/flows/authorization_code_spec.rb
@@ -223,13 +223,7 @@ feature "Authorization Code Flow" do
         visit authorization_endpoint_url(client: @client)
         click_on "Authorize"
 
-        authorization_code = current_params["code"]
-        create_access_token authorization_code, @client, code_verifier
-
-        expect(json_response).to match(
-          "error" => "invalid_grant",
-          "error_description" => translated_error_message(:invalid_grant),
-        )
+        url_should_have_param("code", Doorkeeper::AccessGrant.first.token)
       end
 
       scenario "mobile app requests an access token with authorization code and plain code challenge method" do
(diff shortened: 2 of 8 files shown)


lib/doorkeeper/oauth/authorization_code_request.rb
```diff
@@ -93,7 +84,7 @@ def validate_redirect_uri
       # if either side (server or client) request PKCE, check the verifier
       # against the DB - if PKCE is supported
       def validate_code_verifier
-        return true unless grant.uses_pkce? || code_verifier
+        return true unless grant.uses_pkce?
```
toupeira, line 87: This check had to be removed as well, and caused the changed spec in `spec/requests/flows/authorization_code_spec.rb` to fail.

spec/requests/flows/authorization_code_spec.rb
```diff
@@ -223,13 +223,7 @@ def authorize(redirect_url)
...
         click_on "Authorize"
 
-        authorization_code = current_params["code"]
-        create_access_token authorization_code, @client, code_verifier
-
-        expect(json_response).to match(
-          "error" => "invalid_grant",
-          "error_description" => translated_error_message(:invalid_grant),
-        )
+        url_should_have_param("code", Doorkeeper::AccessGrant.first.token)
```
toupeira, line 226: I'm a bit confused by the PKCE specs in this file because it seems we don't mock `Doorkeeper::AccessGrant).pkce_supported?`, or set it up through other ways.

Should we add this?

nbulaj: I think yes

lib/doorkeeper/oauth/authorization_code_request.rb
```diff
@@ -55,14 +54,6 @@ def pkce_supported?
         Doorkeeper.config.access_grant_model.pkce_supported?
       end
 
-      def validate_pkce_support
-        @invalid_request_reason = :not_support_pkce if grant &&
```
linhdangduy, line 59: Because `not_support_pkce` is used only here, so should remove from other files, too. (en.yml, invalid_request_response.rb)

lib/doorkeeper/oauth/authorization_code_request.rb
linhdangduy: I think we can thinking about `validate_code_verifier` as following:
- `pkce non supported` (pkce_supported? == false) -> return true

- `pkce is supported` (pkce_supported? == true)

||code_challenge.present?|code_challenge.blank?|
|---|---|---|
|code_verifier.present?| (1) run validate | (2) run validate |
|code_verifier.blank?| (4) (already validated at validate_params)return false | (3) return true |

So, as the table, currently changes ([Line 87 and 88](https://github.com/doorkeeper-gem/doorkeeper/pull/1415/files#diff-5c1f9daf0b81d01ceeeb79eaea1cbbe2R87-R88)) will not run validated on (2), when code_verifier is present, and code_challenge is not sent from client at authorization request, this can causing [Downgrade Attack on PKCE](https://danielfett.de/2020/05/16/pkce-vs-nonce-equivalent-or-not/#downgrade-attack-on-pkce)

So I think this validation can write as below:

```ruby
return true unless pkce_supported?
return false if grant.code_challenge.present? && code_verifier.blank? # (*)
# (*) is currently validated at `validate_params`, but I still wrote it because
# I think we should consider this is independent from `validate_params`, 
# any change to `validate_params` should not break `validate_code_verifier` 
# (cause `validate_params` is to informs that client missing necessary parameter)

if grant.code_challenge_method == "S256"
.....
```

spec/requests/flows/authorization_code_spec.rb
toupeira: @nbulaj I saw now that we indirectly set this up by adding the relevant columns: https://github.com/toupeira/doorkeeper/blob/4a7a106dc03411d9ada0a627f6b46e9e8022c393/spec/dummy/db/schema.rb#L25-L28

Running these specs with `WITHOUT_PKCE=1` makes them fail, so I think we're good here and I retract my concern :grin:

lib/doorkeeper/oauth/authorization_code_request.rb
toupeira: @linhdangduy thanks, missed those and removed them now :+1:

lib/doorkeeper/oauth/authorization_code_request.rb
toupeira: @linhdangduy thanks, this makes sense to me. :+1: 

I updated the method now as per your suggestion, though I simplified the 4/5 case into `return grant.code_challenge.blank? if code_verifier.blank?` and added a spec for the case where PKCE is enabled but both `code_challenge` and `code_verifier` are blank. I think the remaining cases should already be covered by the specs in `spec/lib/oauth/authorization_code_request_spec.rb`.

nbulaj:
Thanks @toupeira ! :handshake:

Issue #764, opened 2015-12-21
Integrating SAML SSO


KevinColemanInc:
Is there any documentation for creating custom grant types?  

Basically, I want to be able to issue oAuth tokens if the client provides valid SAML parameters.  I am following this draft https://tools.ietf.org/html/draft-ietf-oauth-saml2-bearer-03 and it proposes a saml sso grant type.

KevinColemanInc:
looking at the code it looks like I need to create my own strategy and it figures out what klass to use for a grant type [based on camelized grant_type](https://github.com/doorkeeper-gem/doorkeeper/blob/master/lib/doorkeeper/request.rb#L27).

So if the spec wants the grant type to be: 'http://oauth.net/grant_type/assertion/saml/2.0/bearer' I am thinking this code needs to be changed to something with more flexible grant_type options.

tute:
Is this aa bug in doorkeeper? Is there anything we can do in the main repository for you to more easily implement the SAML strategy you are after? I suggest you ask in StackOverflow too, as someone there might have done something similar to this and might be able to better help.

KevinColemanInc:
@tute, its not so much a bug in doorkeeper in that, the draft spec wants this string to be the grant type: http://oauth.net/grant_type/assertion/saml/2.0/bearer

But, the way doorkeeper is currently designed, it camelizes and constantizes the grant type into a class name.  This works for grant types like 'password' or 'implicit_grant', but it won't work for a grant type that looks like a URI.

I propose that grant types to classes are mapped via a hash instead of inferring the class from the name.  This would give more flexibility in the future to support grant types that are not formatted with lowercase_separated_by_underscores.

tute:
> I propose that grant types to classes are mapped via a hash instead of inferring the class from the name. This would give more flexibility in the future to support grant types that are not formatted with lowercase_separated_by_underscores.

Here is an implementation that addresses this issue: https://github.com/doorkeeper-gem/doorkeeper/pull/733

tute:
Also related with: https://github.com/doorkeeper-gem/doorkeeper-grants_assertion/issues/9#issuecomment-166970162

KevinColemanInc:
https://tools.ietf.org/html/rfc7522

The draft is finalized.

nbulaj:
@KevinColemanInc it's now possible yo register custom grant types, one step forward to implement this ticket :)

emptyflask:
[...]

PR #1418, opened 2020-05-26
Add ability to register custom Grant Flows

nbulaj:
Allow flexible extending of Doorkeeper with a new OAuth Grant Flows without adding them to the gem itself.

A new flow is registered like so:

```ruby
Doorkeeper::GrantFlow.register :authorization_code,
  response_type_matches: 'code',
  response_type_strategy: Doorkeeper::Request::Code,
  grant_type_matches: 'authorization_code',
  grant_type_strategy: Doorkeeper::Request::AuthorizationCode
```

Based on #733 with additions & retro-compatibility for next extensions:
* doorkeeper_openid-connect
* doorkeeper-grant_assertions

This changes allows to fix / implement:

* #764 (SAML)
* https://github.com/doorkeeper-gem/doorkeeper-grants_assertion/issues/9 (Assertions)

## TODO:

- [x] Implement custom grant flows registration
- [x] Allow multiple flows aliased with a one name in `grant_flows` (like in openid_connect)
- [x] Add specs

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -29,6 +29,9 @@ Metrics/BlockLength:
 Metrics/MethodLength:
   Exclude:
     - spec/dummy/db/**/*
+Style/CaseEquality:
+  Exclude:
+    - lib/doorkeeper/grant_flow/flow.rb
 Style/StringLiterals:
   EnforcedStyle: double_quotes
 Style/StringLiteralsInInterpolation:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -20,7 +20,7 @@ Layout/DotPosition:
 Layout/LineLength:
   Exclude:
     - spec/**/*
-  Max: 100
+  Max: 120 # Remove after we migration to 0.84
 Metrics/BlockLength:
   Exclude:
     - spec/**/*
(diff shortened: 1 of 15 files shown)


joeljunstrom:
@nbulaj this is pretty neat. I had missed that this was a new thing when I fixed a bug in the openid connect lib (a bug that I think exists here in the main repo as well).

The bug consists of when a response type that requires (by spec) that the fragment response mode is used but throws an error in the controller layer will use the default (query) mode instead of the fragment for the error redirect. This is due to no checking of the response type at the controller level raise / rescue handling.

Looking at this I think it might make sense to let the grant flow strategy own the information about which response modes are available / default or MUST be used.

Would you be open to such a patch? Or are there alternative ideas on how response modes should be handled going forward?

My fix is here https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/118 but that is just fixing the already brittle monkey patching.

rafaelsales:
@nbulaj We're about to use this feature in prod - would be nice to have this released in `5.5.0`

k-rudy:
@nbulaj thank you so much for the job.

What's the preferred place to put `Doorkeeper::GrantFlow.register` now? Also should we still indicate the name of the custom grant in `grant_flows %w[authorization_code ...]` config setting?

I'm asking as I tried moving the custom grant registration to new way using:
```
Doorkeeper::GrantFlow.register :jwt_bearer,
  grant_type_matches: 'jwt_bearer',
  grant_type_strategy: Doorkeeper::Request::JwtBearer
```
and keep getting `The authorization grant type is not supported by the authorization server`.

Chucheen:
Is there any documentation about how to use the custom grant flows?

Issue #1412, opened 2020-05-20
Recent Change in Token Revocation Endpoint Behavior Breaks Revocation of Tokens for Public Client


edmondchui:
### Steps to reproduce

I have a mobile app using a Rails server, which in turn depends on Doorkeeper to generate access tokens for API calls. I used to be able to logout successfully from the client using a call to `Doorkeeper::TokensController.revoke`. However, after upgrading to version Doorkeeper 5.4.0, I got HTTP response 401 during logout.

### Expected behavior

Token revocation using a valid access token from Resource Owner Password Credentials Flow should work.

### Actual behavior

Token revocation fails with a valid access token from Resource Owner Password Credentials Flow.

### System configuration

In version 5.3.3, before this change https://github.com/doorkeeper-gem/doorkeeper/pull/1370/commits/d194bc40e335464e67bc36dfd7e2e61247547545, a token granted to a public client (https://tools.ietf.org/html/rfc6749#section-2.1) via Resource Owner Password Credentials Flow works fine as the method `Doorkeeper::TokensController.authorized?` is used for authentication.

In version 5.4.0, the method `Doorkeeper::TokensController.revoke` is modified to require a `server.client`, which is a confidential client (https://tools.ietf.org/html/rfc6749#section-2.1).

**Ruby version**: `2.6.5`

nbulaj:
Hi @edmondchui .

Did you do this from your mobile app?

> The client also includes its authentication credentials as described in Section 2.3. of [RFC6749].

`server.client` stores client instance that performs a request. And if it's blank - no auth passed - request is blocked as stated in the the RFC.

You need to call /token/revoke with ?token value of access token you wanna to revoke and auth headers to inform the server that this request is authorized to perform an action.

AndKiel:
I've got a similar problem. I'm trying to upgrade from Doorkeeper 5.1 to latest.

Token revocation stopped working for me. I'm getting 403 with `{"error"=>"unauthorized_client", "error_description"=>"You are not authorized to revoke this token"}` in my spec. 

Previously `post oauth_revoke_path` spec was passing, returning 200 with an empty response. I tried changing to `post oauth_revoke_path, headers: { 'Authorization': "Bearer #{access_token.token}" }` but this does not pass either (this is basically what I was doing in the UI when "logging out").

Current upgrade PR: https://github.com/AndKiel/tournaments-api/pull/8

[...]

PR #1420, opened 2020-05-29
Make ROPG flow to require client auth as stated in the RFC

nbulaj:
Require client authentication for Resource Owner Password Grant as stated in OAuth RFC

  **[IMPORTANT]** you need to create a new OAuth client (`Doorkeeper::Application`) if yoo didn't
    have it before and use client credentials in HTTP Basic auth if you previously used this grant
    flow without client authentication. For migration purposes you could enable
    `skip_client_authentication_for_password_grant` configuration option to `true`, but such behavior
    (as well as configuration option) would be completely removed in a future version of Doorkeeper.
    All the users of your provider application now need to include client credentials when they use
    this grant flow.

Fixes https://github.com/doorkeeper-gem/doorkeeper/issues/1412#issuecomment-632750422

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,15 @@ User-visible changes worth mentioning.
 - [#1410] Properly memoize `current_resource_owner` value (consider `nil` and `false` values).
 - [#1415] Ignore PKCE params for non-PKCE grants.
 - [#1418] Add ability to register custom OAuth Grant Flows.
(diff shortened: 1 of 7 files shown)


ghiculescu:
@nbulaj should https://github.com/doorkeeper-gem/doorkeeper/wiki/Using-Resource-Owner-Password-Credentials-flow be updated? The current example doesn't mention this at all.

PR #1430, opened 2020-07-17
Add grant types to access tokens

AGarrow:
### Summary

We've found it useful in our doorkeeper integration to record an access tokens grant type.

There are a few pretty specific instances where we factor grant_type into authorization. It's also helpful to have for auditability.

This PR adds the `grant_type` field to the `oauth_access_tokens` table, and makes it's presence a requirement on creation only.

Tokens created from refresh tokens will take on the grant_type of it's predecessor, or `refresh_token` to ensure backwards compatibility.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ User-visible changes worth mentioning.
 
 ## master
 
+- [#1430] Add grant types to access tokens. run `rails generate doorkeeper:grant_type_access_tokens` to generate necessary migration. If you create your own `Doorkeeper::AccessToken`s, be sure to add the relevant `grant_type` string param.
 - [#1426] Ensure ActiveRecord callbacks are executed on token revocation.
 - [#1407] Remove redundant and complex to support helpers froms tests (`should_have_json`, etc).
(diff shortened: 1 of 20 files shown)


nbulaj:
Hi @AGarrow . I think such functionality is possible [using hooks](https://github.com/doorkeeper-gem/doorkeeper/blob/master/lib/generators/doorkeeper/templates/initializer.rb#L396-L424), WDYT ?

AGarrow:
hi @nbulaj ,

thanks for taking a look! Could you elaborate on what that approach would look like? I'm having a hard time thinking of the benefit over inserting it as part of the normal record creation flow.

nbulaj:
My main point is not to add a features that are required by a few number of developers (aka not to bloat the gem). Maybe if I'll see a lot of interest in it - this could be added to the gem.

That's why I suggested to use hooks like the following (need exploration, I don't remember actual arguments in the hooks, you could check the docs and specs inside the gem itself):

```ruby

after_successful_authorization do |controller, context|
  context.issued_token.update_columns(grant_type: controller.request.params[:grant_type])
end

after_successful_response do |context|
  context.access_token.update_columns(grant_type: %some magic%)
end
```

PR #1434, opened 2020-07-29
Capitalize token_type in specs

stahor:
https://github.com/doorkeeper-gem/doorkeeper/blob/v5.4.0/lib/doorkeeper/models/access_token_mixin.rb#L286
`Bearer` token type starts with capital B since v4.4.something. But In specs it's still `bearer` and it's a little bit confusing.


nbulaj:
Thanks @stahor !

PR #1440, opened 2020-09-02
Validate empty host in redirect_uri

igor-drozdov:
### Summary

When `http://` or `https://` is specified as `redirect_uri`, we should expect it to be invalid

We had this behavior until these changes have been specified: https://github.com/doorkeeper-gem/doorkeeper/commit/ebedb3c8fde06ad377f33ba11958e32dc7ff614e#diff-1de26e1e806e732aaaf8b2010ee70af7L25

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ User-visible changes worth mentioning.
 ## master
 
 - [#PR ID] Add your PR description here.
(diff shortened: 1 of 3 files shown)


nbulaj (approved):
Nice catch :+1:

nbulaj:
Thanks @igor-drozdov !

Issue #1442, opened 2020-09-10
Rake task doorkeeper:db:cleanup:expired_tokens fetches expired tokens with an inconsistent expiration time calculation


pjmartorell:
The rake task `doorkeeper:db:cleanup:expired_tokens` uses `clean_expired` method to clean stale tokens, but the query that fetches expired tokens is not consistent with what we consider "expired", since it uses `Doorkeeper.config.access_token_expires_in` as a reference instead of the `expires_in` attribute of the access token. Every token may have a different `expires_in` value depending on the moment the global configuration `Doorkeeper.config.access_token_expires_in` was changed and is the `expires_in` the one that indicates if a token is expired or not.
Compare:

https://github.com/doorkeeper-gem/doorkeeper/blob/c54e73115738edb5eb819147516592253198beeb/lib/doorkeeper/rake/db.rake#L20-L25

https://github.com/doorkeeper-gem/doorkeeper/blob/c54e73115738edb5eb819147516592253198beeb/lib/doorkeeper/orm/active_record/stale_records_cleaner.rb#L23-L29

With:
https://github.com/doorkeeper-gem/doorkeeper/blob/c54e73115738edb5eb819147516592253198beeb/lib/doorkeeper/models/concerns/expirable.rb#L26-L33

So in my opinion `clean_expired` should be: 
```ruby
 # Clears expired records 
 def clean_expired
   @base_scope.where("created_at + expires_in * interval '1 second' < ?", Time.current).in_batches(&:delete_all) 
 end 
```
Note that the syntax is in PostgreSQL, I couldn't come up with a more agnostic query. `ttl` argument would be no longer needed.

nbulaj:
H @pjmartorell . This was discussed somewhere in other issue, you could search by `interval '1 second'`.

The problem with the last solution is that we must support all the databases at least for ActiveRecord: MySQL / MSSQL / Postgres / SQLIte / etc. And there were also more issues as I remember

pjmartorell:
@nbulaj I don't have much more ideas apart from creating adapters for every database. An alternative would be to store an `expires_at` date in a new column, which would simplify the query and could be written in `ActiveRecord` easily, but I don't think it's worth to introduce this change just to fix this rake task

Yoshiji:
I came across this rake task and had the same questioning. So far, I can think of 3 different approaches:

1. allow an optional argument for the rake task, overwriting the `ttl` given to the cleaner:
```
[...]

PR #1452, opened 2020-11-16
Empty previous_refresh_token if present?

Yoshiji:
### Summary

The call to `update_attribute` in `revoke_previous_refresh_token!` is always performed, regardless of the current state. This means a DB transaction is opened, even if there are no changes to apply (i.e. the `previous_refresh_token` was already emptied).

This PR simply empty the `previous_refresh_token` if it is `present?` to avoid an unnecessary database transaction, which can happen at each `authenticate` call if `Doorkeeper.config.refresh_token_enabled?` ([source](https://github.com/doorkeeper-gem/doorkeeper/blob/034d87eff801ef6576fa1af31ba995eefa435924/lib/doorkeeper/oauth/token.rb#L19))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ User-visible changes worth mentioning.
 ## master
 
 - [#PR ID] Add your PR description here.
+- [#1452] Empty previous_refresh_token only if present
 - [#1440] Validate empty host in redirect_uri
 - [#1438] Add form post response mode.
 
--- a/lib/doorkeeper/models/access_token_mixin.rb
+++ b/lib/doorkeeper/models/access_token_mixin.rb
@@ -377,7 +377,7 @@ module Doorkeeper
       return unless self.class.refresh_token_revoked_on_use?
 
       old_refresh_token&.revoke
-      update_attribute(:previous_refresh_token, "")
+      update_attribute(:previous_refresh_token, "") if previous_refresh_token.present?
     end
 
     private


nbulaj (approved):
LGTM :+1: 

Just add a changelog entry please as bot suggested

nbulaj:
Thanks @Yoshiji !

maxencehenneron:
Can we release this? This is running the query at every single requests made on my API, and my table is large so it's expensive and it takes a while to run it.

nbulaj:
If we will release this it could be only RC2 version because 5.5 has some possible breaking changes @maxencehenneron 
If you're OK with 5.5RC2 then we can release, yes

maxencehenneron:
@nbulaj Thanks for the quick update. It would be nice to release this in 5.4.1 as a bug fix. But if that's not possible I can spend some time integrating 5.5. The issue is that at every single requests made on my API, it runs the following query:

```
D, [2020-12-16T12:10:43.610231 #14] DEBUG -- : [89819f7d-5c7c-495d-b28a-5b21cc5af25e]   Doorkeeper::AccessToken Update (1161.5ms)  UPDATE "oauth_access_tokens" SET "previous_refresh_token" = $1 WHERE "oauth_access_tokens"."id" = $2  [["previous_refresh_token", ""], ["id", "xxx"]]
```

The query is quite slow as my table has million of records. Also, I'm gettings an average of 10 requests per second on my API so this runs 10 times per second.

Yoshiji:
@maxencehenneron The issue you're describing does not fit the problem solved by this PR: `UPDATE` statements are legitimate when they _reset_ the `previous_refresh_token` to an empty string. Rails' ActiveRecord will not trigger an `UPDATE` if there are no changes. For example: `user = User.last ; user.update_attribute(:name, user.name)` will not trigger an `UPDATE` **(1)**

So in your case, it looks like these statements are legitimate: the value of `previous_refresh_token` was already different from an empty string, hence Rails triggered the actual `UPDATE`. You can verify this by doing the following in your `rails console`:

```ruby
# no UPDATE
Doorkeeper::AccessToken.where(previous_refresh_token: "").last.update_attribute(:previous_refresh_token, "")
# an UPDATE is triggered
Doorkeeper::AccessToken.where.not(previous_refresh_token: "").last.update_attribute(:previous_refresh_token, "")
```

Your problem probably lies somewhere else, most likely a combination of the table's significant size + indexes:
- reduce size: you can schedule a task to delete expired tokens to avoid having more and more tokens. Doorkeeper comes with a Rake task for this (see [documentation](https://doorkeeper.gitbook.io/guides/internals/rake#cleaning-up-the-database)) but beware it does not respect each token's `expires_in` value (see #1442 for more details)
- optimize indexes: indexes are great for _reading_ the content, but you pay the price when you _write_ in that table (writing in the table implies writing in the indexes as well). Make sure you only have relevant indexes in there.

---

**(1)** Even though Rails will not trigger an `UPDATE`, it will still open a DB transaction, do nothing, then close it. This can cause systems using a read-only replica + main read&write databases to switch to the main DB because of the transaction, even if there are no actual mutating statement in there. The useless empty transactions  issue has been fixed in https://github.com/rails/rails/pull/32647 (Rails 6) thanks to _lazy transactions_.

If you are working with the [`distribute_reads`](https://github.com/ankane/distribute_reads) gem, the lazy transactions did not fix the problem because behind the scene, `makara` is relying on the method names to determine which DB (main/replica) it should be using, not the actual SQL. So in any cases, `user.update_attribute(:name, user.name)` will still make `makara` _think_ it should switch to the main database even though it should not (see issue https://github.com/instacart/makara/issues/218)

maxencehenneron:
@Yoshiji I'm using Rails 6.0.2 so it shouldn't be related to what you linked, but thank you for looking into this, it might have been the issue!

I actually tried to apply your PR on the 5.4.0 tag, and it resolved the issue, but while applying it, I noticed something.

The current 5.4.0 tag uses update_column and not update_attribute. 
update_column will always run the SQL query even if the column does not change.

The change was actually applied in this PR, available in 5.5.0-rc1:

https://github.com/doorkeeper-gem/doorkeeper/commit/37ef580fc1110002dacaaafababc9d05caddeb1e

You're right that your PR does not solve the issue, it seems that it was fixed already in 5.5.0-rc1.

Thank you again for investigating!

PR #1457, opened 2020-12-04
Make oauth_applications' owner_id column a bigint

### LINKED ISSUE #1456: owner_id should be an int8 column
In the `add_owner_to_application_migration.rb` migration template, the `owner_id` is just an `integer`: https://github.com/doorkeeper-gem/doorkeeper/blob/b8abbe6bb8646232b9d5065c27de9e1e78e974b9/lib/generators/doorkeeper/templates/add_owner_to_application_migration.rb.erb#L5

In Postgres (and presumably MySQL/sqlite), this creates an int4, which means it maxes out at around 2.1 billion records (`2,147,483,647` specifically).

Since Rails 5.1, the primary key on all tables has been a bigint (int8): https://www.mccartie.com/tech/2016/12/05/rails-5.1.html

While it's unlikely anyone will ever actually hit anything this large, I figure it's worth updating the generated migration to use int8 just in case.

connorshea:
### Summary

For newly-generated owner migrations, this will make owner_id a bigint just in case anyone ever hits the limit for int4 values on their users table.

Fixes #1456.

### Other Information

I've tested this change in a Rails app I maintain and it works fine, and I don't think it should require any changes besides the migration itself since Rails treats int4 and int8 values the same on its side of things.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ User-visible changes worth mentioning.
 ## master
 
 - [#PR ID] Add your PR description here.
+- [#1457] Make owner_id a bigint for newly-generated owner migrations
 - [#1452] Empty previous_refresh_token only if present
 - [#1440] Validate empty host in redirect_uri
 - [#1438] Add form post response mode.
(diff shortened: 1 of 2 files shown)


nbulaj (approved):

nbulaj:
Thankks @connorshea !

PR #1458, opened 2020-12-07
Refactor `skip_client_authentication_for_password_grant` option to make it available for the long term

ghiculescu:
I appreciate the work done in https://github.com/doorkeeper-gem/doorkeeper/pull/1370 and https://github.com/doorkeeper-gem/doorkeeper/pull/1420 to make Doorkeeper more secure by default. I think strictly enforcing the RFC by default is exactly right.

The changes are causing us a bit of concern on existing apps that have been running with this authentication strategy for many years.

Our example use case is here: https://my.tanda.co/api/v2/documentation#header-authentication-(password) - we have hundreds of consumers relying on this approach in the wild (many of them whom we have never directly talked to about it), in addition we have built native apps that depend on this behavior for user login, and we believe other third parties have too.

An unaffiliated example of a similar workflow can be found here https://developer.twitter.com/en/docs/authentication/basic-auth

For this reason, I would like to propose keeping `config.skip_client_authentication_for_password_grant` as an option for the long term, but heavily discouraging it.

In this PR I've:

- Added a big warning around the `config.skip_client_authentication_for_password_grant` config, but removed the deprecation notice.
- Lightly refactored `Doorkeeper::TokensController`, and made the behavior in https://github.com/doorkeeper-gem/doorkeeper/pull/1370 conditional on `config.skip_client_authentication_for_password_grant`.
- Added tests for https://github.com/doorkeeper-gem/doorkeeper/pull/1370 and https://github.com/doorkeeper-gem/doorkeeper/pull/1420 when `config.skip_client_authentication_for_password_grant` is `true`.
- Updated the `CHANGELOG` - noting this change, but also tweaking the warning for https://github.com/doorkeeper-gem/doorkeeper/pull/1420 slightly. (If this PR is approved I'd also update the [wiki](https://github.com/doorkeeper-gem/doorkeeper/wiki/Migration-from-old-versions#from-5x-to-54x))

Thank you for considering!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ User-visible changes worth mentioning.
 - [#1452] Empty previous_refresh_token only if present
 - [#1440] Validate empty host in redirect_uri
 - [#1438] Add form post response mode.
+- [#1458] Make `config.skip_client_authentication_for_password_grant` a long term configuration option
 
 ## 5.5.0.rc1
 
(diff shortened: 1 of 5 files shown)


nbulaj (approved):
LGTM
:+1:

nbulaj:
Thanks @ghiculescu . I see a lot of requests to leave this option so I think I have no choice :)

ghiculescu:
I appreciate it @nbulaj. For new projects it's great to follow the spec, but I think there is a lot of old code (often not controlled by the API server) that depends on this out there.

chewy22:
[](url)

PR #1459, opened 2020-12-10
Use built-in option to remove padding

dee-see:
### Summary

I was reading the code and came across how the padding is stripped from the `code_challenge`. There's a builtin option in `Base64.urlsafe_encode64(..., padding: false)` to do that. Tiny tiny change but we're saving a split. :)


dee-see:
I think we can skip the changelog for this?

nbulaj:
You don't know @dee-see how I love such changes :+1: 
Thanks! Yeah, no need in changelog I think

nbulaj (approved):

PR #1465, opened 2020-12-29
Memoize nil doorkeeper_token

kitop:
### Summary

In some applications extending `AccessToken` class with custom
functionality, it's some times required to call `doorkeeper_token`
method directly. If this method is called multiple times, it can cause
repeated DB calls when there's an invalid token that returns `nil` from
the `OAuth::Token.authenticate` call.

This could be improved by doing similar as in `current_resource_owner`,
where it memoizes `nil` return values for
`instance_eval(&Doorkeeper.config.authenticate_resource_owner)`

With this proposed change, it will memoize `nil` doorkeeper_token
values too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ User-visible changes worth mentioning.
 ## master
 
 - [#PR ID] Add your PR description here.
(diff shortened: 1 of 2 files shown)


nbulaj (approved):
LGTM
:+1:

Issue #1466, opened 2021-01-01
Possibility to customize the response of oauth/token/info


energenious-eu:
I am trying to let an Oauth client authorize a user based on his/her email address registered on the AuthServer.

I am not owner of the client's code and have figured out that the email of the user is read within the response of the GET request to the /oauth/token/info endpoint.

To solve this issue i have patched to TokenInfoController this way:

```ruby
module Doorkeeper
    class TokenInfoController < Doorkeeper::ApplicationMetalController
      def show
        if doorkeeper_token&.accessible?
          email = User.find(doorkeeper_token.resource_owner_id).email
          custom_response = doorkeeper_token.serializable_hash.merge "email" => email
          render json: custom_response, status: :ok
        else
          error = OAuth::InvalidTokenResponse.new
          response.headers.merge!(error.headers)
          render json: error.body, status: error.status
        end
      end
    end
  end
```

but of course this was a quick fix. Is there a proper way to do this in the doorkeeper.rb file?

By the way: Is this actually a use case for OpenID Connect instead of Oauth ?

nbulaj:
Hi @energenious-eu . Hi about to create a custom token info controller and put required change there?

```ruby
Rails.application.routes.draw do
  use_doorkeeper scope: 'endpoint' do
    # Customize controllers
    controllers token_info: 'my_token_info'
  end
end
```

https://github.com/doorkeeper-gem/doorkeeper/wiki/Customizing-routes#version--05

PR #1482, opened 2021-02-06
Simplify TokenInfoController to be overridable (extract response rendering)

nbulaj:
Now you can override response(s) for `Doorkeeper::TokenInfoController` with more elegant way:

```ruby
class TokenInfoController < Doorkeeper::TokenInfoController
  protected

  def doorkeeper_token_to_json
    {}
  end

  def error_to_json(error)
    {}
  end
end

# config/routes.rb
Rails.application.routes.draw do
  use_doorkeeper do
    controllers token_info: 'token_info_controller'
  end
end
```

Relates to #1466

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ User-visible changes worth mentioning.
 ## master
 
 - [#PR ID] Add your PR description here.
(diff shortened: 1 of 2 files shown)


Chasemycharlie (commented):
Duplicate of #

Chasemycharlie (commented):
@doorkeeper-bot

PR #1487, opened 2021-03-05
[*] Replaces find_matching_token implementation with more performant version

PhilipSkinner:
### Summary

Performance of find_matching_token a major issue - can take systems offline.

This patch rewrites it to only fetch 1 valid access token from the database.

### Other Information

I used the existing tests as coverage over this particular modification - all tests are still passing.

I did notice that in order to get the tests to pass you must accept that an application may have no scopes, and if it has no scopes then no verification of the incoming scopes is carried out, nor does the system verify that the token being returned contains only claims available to the application.

Let me know if this is an oversight of the current system and I will correct this behavior - right now it works as it did before.

### DIFF
--- a/lib/doorkeeper/models/access_token_mixin.rb
+++ b/lib/doorkeeper/models/access_token_mixin.rb
@@ -91,22 +91,8 @@ module Doorkeeper
         find_matching_token(tokens, application, scopes)
       end
 
-      # Interface to enumerate access token records in batches in order not
-      # to bloat the memory. Could be overloaded in any ORM extension.
-      #
-      def find_access_token_in_batches(relation, **args, &block)
-        relation.find_in_batches(**args, &block)
-      end
-
-      # Enumerates AccessToken records in batches to find a matching token.
-      # Batching is required in order not to pollute the memory if Application
-      # has huge amount of associated records.
-      #
-      # ActiveRecord 5.x - 6.x ignores custom ordering so we can't perform a
-      # database sort by created_at, so we need to load all the matching records,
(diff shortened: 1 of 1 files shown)


PhilipSkinner:
Apologies, I wrote this with a feature that isn't readily available on older versions of active record, I've corrected it to work with older versions.

PhilipSkinner:
Apologies for all the amends, ruby is a new language for me.

Let me know if there are any questions or modifications that I need to make.

lib/doorkeeper/models/access_token_mixin.rb
```diff
@@ -120,44 +106,113 @@ def find_access_token_in_batches(relation, **args, &block)
...
+      #   set of scopes
+      #
+      # @return [ActiveRecord::Relation] filtered relation
+      #   query
+      #
+      def filter_by_application_scopes(relation, application, scopes)
+        application.scopes.each do |scope_val|
+          next unless scopes.exists? scope_val
+
+          relation = relation.where([
```
nbulaj, line 158: I don't sure actually that this implementation matches previous one with `Doorkeeper::OAuth::Helpers::ScopeChecker
` :thinking: Even with passing specs it looks scary.. I have to to a deep dive into the changes to understand if everything is OK

PhilipSkinner: Ok, I can see the differences related to the grant types, is that what you are worried about here? If so, I'll see about getting this updated.

I am worried about the fact that the existing tests enforce that if a client has no scopes, that we allow any scopes through, is that on purpose?

PR #1488, opened 2021-03-05
Fix ROPC client authentication validation when it's optional

toupeira:
### Summary

When `skip_client_authentication_for_password_grant` is enabled and invalid client credentials are present in a Basic auth header, an access token is still granted.

We do validate if `client_id` is present as a query-string parameter, and should do the same if an authorization header is present.

### Other Information

See previous PRs https://github.com/doorkeeper-gem/doorkeeper/pull/1420 and https://github.com/doorkeeper-gem/doorkeeper/pull/1458.

I guess this option should really be called `require_client_authentication_for_password_grant` to make the intent clearer, but I guess it's too late for that now :wink: 

Still, this change should be useful for correctness and troubleshooting purposes. I don't think there's really a security aspect to this since clients can also just not send any credentials at all.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ User-visible changes worth mentioning.
 ## main
 
 - [#PR ID] Add your PR description here.
+- [#1488] Verify client authentication for Resource Owner Password Grant when
+  `config.skip_client_authentication_for_password_grant` is set and the client credentials
+  are sent in a HTTP Basic auth header.
 
 ## 5.5.0
 
--- a/lib/doorkeeper/oauth/password_access_token_request.rb
+++ b/lib/doorkeeper/oauth/password_access_token_request.rb
@@ -10,12 +10,13 @@ module Doorkeeper
       validate :resource_owner, error: :invalid_grant
       validate :scopes, error: :invalid_scope
 
-      attr_reader :client, :resource_owner, :parameters, :access_token
+      attr_reader :client, :credentials, :resource_owner, :parameters, :access_token
 
-      def initialize(server, client, resource_owner, parameters = {})
+      def initialize(server, client, credentials, resource_owner, parameters = {})
         @server          = server
         @resource_owner  = resource_owner
         @client          = client
+        @credentials     = credentials
         @parameters      = parameters
         @original_scopes = parameters[:scope]
         @grant_type      = Doorkeeper::OAuth::PASSWORD
@@ -60,7 +61,7 @@ module Doorkeeper
       #
       def validate_client
         if Doorkeeper.config.skip_client_authentication_for_password_grant
-          !parameters[:client_id] || client.present?
+          client.present? || (!parameters[:client_id] && credentials.blank?)
         else
           client.present?
         end
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -278,6 +278,10 @@ module Doorkeeper
     # Rationale: https://github.com/doorkeeper-gem/doorkeeper/issues/1189
     option :token_reuse_limit,              default: 100
 
+    # Don't require client authentication for password grants. If client credentials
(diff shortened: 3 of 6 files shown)


lib/doorkeeper/oauth/password_access_token_request.rb
```diff
@@ -10,12 +10,13 @@ class PasswordAccessTokenRequest < BaseRequest
       validate :resource_owner, error: :invalid_grant
       validate :scopes, error: :invalid_scope
 
-      attr_reader :client, :resource_owner, :parameters, :access_token
+      attr_reader :client, :credentials, :resource_owner, :parameters, :access_token
 
-      def initialize(server, client, resource_owner, parameters = {})
+      def initialize(server, client, credentials, resource_owner, parameters = {})
```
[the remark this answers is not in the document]
toupeira: We don't have an initializer in `BaseRequest`, so this can probably be ignored?

toupeira:
@nbulaj could you please give this a review if you have time? :grinning: 

Unfortunately we have to enable `skip_client_authentication_for_password_grant` too in GitLab, which is why I stumbled over this: https://gitlab.com/gitlab-org/gitlab/-/merge_requests/55873

But as mentioned I don't think it's a security concern as such, just a correctness fix.

Not sure what to do about the Codeclimate complaint. :thinking:

nbulaj:
Morning @toupeira 

I see a lot of failed jobs, looks like specs should be fixed to use new number of args in the initializer.

> I guess this option should really be called **require_client_authentication_for_password_grant** to make the intent clearer

Actually Doorkeeper now requires authentication by default for this grant, so such option will be confusing :thinking: Or you found that we can still use blank credentials without the option? :scream:

toupeira:
> I see a lot of failed jobs, looks like specs should be fixed to use new number of args in the initializer.

@nbulaj oh thanks, totally missed the Travis failures! :see_no_evil: Specs should be fixed now.

> Actually Doorkeeper now requires authentication by default for this grant, so such option will be confusing :thinking:

Yeah what I meant is that we're not really "skipping" client authentication, we were still requiring valid client credentials when they were present in the query-string, but not when sent in a HTTP Basic auth header (which this PR fixes). So calling the setting `require_` (with a default of `true`) would have been slightly clearer, but I'm definitely not advocating for changing it now since it would just add more confusion :wink: 

To clarify this more, I also added a spec now for the query-string case at https://github.com/doorkeeper-gem/doorkeeper/pull/1488/files#diff-5acf666b403221a90d0741e6500d7d44598331dd4f0d2a332496e564883ded5bR226 (which still passed with the previous code in `validate_client`), and a comment at https://github.com/doorkeeper-gem/doorkeeper/pull/1488/files#diff-d4ed41e5cd854dad8edf2e617b038a62d12547817c447a6864ed0addd33a622dR281 to explain what the setting does.

> Or you found that we can still use blank credentials without the option? :scream:

No, this only applies if `skip_client_authentication_for_password_grant` is enabled.

And the goal is just to make the behaviour consistent, there's not really a security implication because clients can also just not send any credentials at all to skip the client authentication :grinning:

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ User-visible changes worth mentioning.
 
 - [#PR ID] Add your PR description here.
 - [#1477] Respect `ActiveRecord::Base.pluralize_table_names` for Doorkeeper table names.
+- [#1488] Verify client authentication for Resource Owner Password Grant when
```
nbulaj, line 12: Look like outdated copy :thinking: 

We already released 5.5 https://github.com/doorkeeper-gem/doorkeeper/blob/main/CHANGELOG.md#main

toupeira: @nbulaj whoops indeed, I had an old checkout and missed the rename from `master` to `main`! :facepalm: 

Rebased now, and moved the changelog entry up.

nbulaj:
Thanks @toupeira :handshake:

PR #1495, opened 2021-03-19
Make respond_to available for API-Only applications

### LINKED ISSUE #1494: ApplicationsController not working in Api only Mode
### Steps to reproduce
I recently found out that the applications endpoint is now exposed in API only mode and I wanted to try it out. However, every time I call the `/oauth/applications` endpoint I get the following error:

```
NoMethodError at /v1/oauth/applications
=========================================

undefined method `respond_to' for #<Doorkeeper::ApplicationsController:0x00000000064208>
Did you mean?  respond_to?

```

I did some research and found out that the controller extends from the `ActionController::API`:

https://github.com/doorkeeper-gem/doorkeeper/blob/e048bedb7813bb273e05554a66cc1c0411638d44/lib/doorkeeper/config.rb#L382

This controller does not include the `respond_to` unless you include `MimeResponds` (https://stackoverflow.com/a/35985101/10127698). 

I am not sure if it is needed, but I did have the responders gem installed when I tested this.

Matthijsy:
### Summary
This PR fixes that the `respond_to` was not found in the API only mode. I also found a little bug that when removing an authorized application in API only mode the response code was 200 with an empty payload instead of a 204. 

Resolved #1494

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ User-visible changes worth mentioning.
 ## main
 
 - [#PR ID] Add your PR description here.
+- [#1495] Fix `respond_to` undefined in API-only mode 
 - [#1488] Verify client authentication for Resource Owner Password Grant when
   `config.skip_client_authentication_for_password_grant` is set and the client credentials
   are sent in a HTTP Basic auth header.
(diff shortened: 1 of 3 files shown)


nbulaj (approved):
LGTM :+1: 

Let's finish instructions from the bot (ignore about tests) and it's ready to merge

Matthijsy:
Thanks @nbulaj ! I have updated the changelog and squashed everything into one commit. I cannot find any tests related to the api-only mode, so I did not include/update any tests. Should I still do so or is it fine like this?

PR #1501, opened 2021-04-09
Ensure AccessToken#expired? always returns a boolean

jhass:
I noticed compared to `revoked?` `expired?` may return `nil` :)


nbulaj (approved):

ThisIsMissEm:
@nbulaj I left a comment on the commit, but this method can and does still return `nil` making the token introspection stuff fail, if `access_token_expires_in nil` in the configuration

PR #1511, opened 2021-05-18
Fix that authorization code is returned by fragment if response_mode is fragament

nhosoya:
### Summary

There is a problem that the authorization code is returned in the query even if "fragment" is specified for response_mode in the authorization request.

### Other Information

Whether the authorization code is returned as a query or as a fragment is determined by the following in `CodeResponse`.
https://github.com/doorkeeper-gem/doorkeeper/blob/a56b1a5379e97fed5cf838f834702724e070db38/lib/doorkeeper/oauth/code_response.rb#L43-L47

In order for this to be determined correctly, `response_on_fragment` must be passed when initializing `CodeResponse`.
However, since `response_on_fragment` is not specified when `CodeReponse` is initialized in `CodeRequest`.
Therefore, the authorization code is always returned in query.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ User-visible changes worth mentioning.
 - [#1502] Drop support for Ruby 2.4 because of EOL.
 - [#1504] Updated the url fragment in the comment.
 - [#1512] Fix form behavior when response mode is form_post.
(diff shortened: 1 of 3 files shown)


nbulaj (approved):
LGTM :+1:

PR #1512, opened 2021-05-18
Fix form behavior when response mode is form_post

nhosoya:
### Summary

There are three problems with the behavior when `response_mode=form_post` is specified

- Form is not automatically posted.
- When no state is specified, an empty state is posted.
- Unnecessary authenticity_token is posted.
- response_mode is not posted when approving or denying.

### Other Information

N/A


nhosoya:
In order to add meaningful tests, I think we need to implement an E2E testing mechanism in this gem, since we need to test javascript.
Is it really necessary to go that far?

nbulaj (approved):
LGTM :+1: 

About the specs - it will be great, but I don't have time now to write and check them :) So I think we can avoid for now

Issue #1129, opened 2018-07-31
Provide the ability to use JWT's `jti` instead of the whole token for db storage ( or any alternative )


coalwater:
I want to create a JWT with a unique UUID as a `jti` claim [(defined in section 3 point 7)](https://tools.ietf.org/html/rfc7523#section-3), and would like to use the UUID to identify the token in the database instead of the whole token.

What currently happens is that the whole encoded JWT is saved in the db, which could increase the db size too much too fast.

## Steps to reproduce
Create a basic set up of `doorkeeper` + `doorkeeper-jwt`
Do a `POST /oauth/token` to create a new token
A record will be inserted with the whole encoded token into `oauth_access_tokens.token`

It seems that it's not possible to do that with the currenct setup, I've also checked a previous ticket #983

## Doorkeeper setup

```ruby
Doorkeeper.configure do
  orm :active_record
  grant_flows %w[password]
  access_token_expires_in 2.hours
  access_token_generator 'Doorkeeper::JWT'
  use_refresh_token
  resource_owner_authenticator do
    User.first # for testing
  end
  resource_owner_from_credentials do |routes|
    User.first # for testing
  end
end
```

## JWT setup 

```ruby
Doorkeeper::JWT.configure do
  token_payload do |opts|
    user = User.find(opts[:resource_owner_id])
    created_at = opts[:created_at].to_i

    {
      iss: 'https://exmaple.com/oauth',
      sub: user.user_name,
      aud: 'https://example.com/oauth',
      user_id: user.id,
      username: user.username,
      iat: created_at,
      exp: created_at + opts[:expires_in],
      jti: SecureRandom.uuid # The value that should be inserted in the db
    }
  end
  secret_key 'MY-SECRET'
  encryption_method :hs512
end
```

nbulaj:
Hi @coalwater . I think you need to take a look at [Doorkeeper-JWT gem](https://github.com/chriswarren/doorkeeper-jwt/blob/master/lib/doorkeeper-jwt.rb#L8) as it generates a token and returns the whole token for saving to the database. 

Also I could produce a PR if you have a clear vision on what we need to do to support such case. Thanks!

coalwater:
@nbulaj Thanks for your reply, I just needed someone to guide me, I will try to look into this and I'll get back to you if i have any questions.
Thanks again

coalwater:
I think we can close this issue and I'll move this to the other gem.

Issue #983, opened 2017-08-09
Token are always persisted in database (JWT)


mdebo:
Hi
Thank you for your usefull work!
I'm using Dookeeper and JWT. So i use your gem combine with doorkeeper-jwt.
As I use JWT, I don't want to persist token in database. In fact, JWT gives the advantage of not needing to check the token in a db every time since you can just use cryptography to verify that the token is legitimate. I dont want to lose this benefit...
So, there is an option to tell doorkeeper to not store token in database? I can't find it in the wiki..
Thank you in advance

nbulaj:
Hi @mdebo. Currently, that is impossible as far as I know. Every token is saved to the database. It is required to change some core architecture of the doorkeeper gem to implement such a functionality. You are very welcome with the PR if you want.

nbulaj:
As it requires to change core Doorkeeper methods, feel free to send a PR @mdebo . Maybe it can be solved with some "NULL-orm" that will ignore DB and store / find only some structs based on decoded JWT.

pandwoter:
Hi @nbulaj, would it be ok if I instead will try to provide a PR with this feature? As we want to have it in the doorkeeper codebase rather than maintaining fork...

PR #1515, opened 2021-06-29
Add an option for token customization before persisting it in the database

pandwoter:
### Summary

Before there was no possibility to customize the token before actual persistence, so in the case of using JWT as an access token the gem users were forced to store the whole token payload in the database (and update the default `VARCHAR(255)` token column type to `TEXT` as JWT tokens are quite big). This PR provides the possibility for customization like this so with these changes you can easily migrate to JWT tokens and store token JTI as a token value (as there is no sense in the actual JWT token persistence as a whole). 

In terms of `doorkeeper.rb` this option can be used like this:
```ruby
  # https://github.com/doorkeeper-gem/doorkeeper#custom-access-token-generator
  if jwt.enabled?
    access_token_generator '::Doorkeeper::JWT'

    customized_token do |token|
      keyset = JWT::JWK.new(OpenSSL::PKey::RSA.new(Base64.decode64(ENV.fetch('JWT_PRIVATE_KEY'))))

      decoded_token = JWT.decode(
        token, nil, true, { algorithms: ['RS256'], jwks: ->(_) { { keys: [keyset.export] } }}
      )

      decoded_token.first['jti']
    end
  end
```

This block will be used for token persistence ~~and for token verification~~.

**NOTE:** It's up to the end-users to use this customization as a building block in their overwritten controllers. A raw token that was used for customized token generation can be accessed through the getter method `plaintext_token_not_restored`.

### Related issues which can benefit from this feature

https://github.com/doorkeeper-gem/doorkeeper/issues/983
https://github.com/doorkeeper-gem/doorkeeper/issues/1129

### DIFF
--- a/spec/models/doorkeeper/access_token_spec.rb
+++ b/spec/models/doorkeeper/access_token_spec.rb
@@ -19,8 +19,6 @@ RSpec.describe Doorkeeper::AccessToken do
 
   describe "#generate_token" do
     it "generates a token using the default method" do
-      FactoryBot.create :access_token
-
       token = FactoryBot.create :access_token
       expect(token.token).to be_a(String)
     end
@@ -100,6 +98,20 @@ RSpec.describe Doorkeeper::AccessToken do
           end
         end
       end
+    end
+
+    it "customize the token before saving" do
+      Doorkeeper.configure do
+        orm DOORKEEPER_ORM
+        customized_token { |token| "#{token}_custom_part" }
+      end
+
+      owner = FactoryBot.create :resource_owner
+      token = FactoryBot.create :access_token,
+                                resource_owner_id: owner.id,
+                                resource_owner_type: owner.class.name
+
+      expect(token.token).to include("_custom_part")
     end
 
     it "generates a token using a custom object" do
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ User-visible changes worth mentioning.
 ## main
 
 - [#PR ID] Add your PR description here.
+- [#1515] Add an option for token customization before persisting it in the database.
 
 ## 5.5.2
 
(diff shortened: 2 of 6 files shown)


spec/models/doorkeeper/access_token_spec.rb
```diff
@@ -19,8 +19,6 @@ def self.generate; end
 
   describe "#generate_token" do
     it "generates a token using the default method" do
-      FactoryBot.create :access_token
```
pandwoter, line 22: I think this is placed here by mistake, feel free to comment if it's done on purpose.

pandwoter:
@nbulaj, what do you think of these? If there is a need I can in my spare time provide an example application which will heavily use this `customized_token` option to not persist the whole JWT.

pandwoter:
Removed customization in `from_request`, cause if needed it can be adjusted in the `AccessToken` model by the end-users.

nbulaj:
Hi @pandwoter . Thanks for the PR! I just wondering if it will be better to move this logic somehow to doorkeeper-JWT gem? :thinking:  I see it more benefit there then in the core. WDYT? Also does it works correctly when hashing/encryption for the tokens enabled?

pandwoter:
Hi 👋 , unfortunately, as far as I know, we can't expose the original token from the token generator class (that's why I placed this PR here). Regarding the hashing/encryption - let me check, I don't know for sure.

nbulaj:
Well nothing stops you from adding just one more option (`option :customized_token, default: ->(access_token) { access_token }`) to `doorkeeper-jwt` itself and something like:

```ruby
ActiveSupport.on_load(:active_record) do
  Doorkeeper.config.access_token_model.prepend(MyCustomizedMixin) # MyCustomizedMixin will handle the changes you proposed
end
```

I'm not the fan of prepending modules like this, but.. from the other side I don't want to introduce a ton configuration options and prefer to use extensions over all-in-one :disappointed:

pandwoter:
Ok then, I'll close this for now. Mb in the future I will come up with separate strategies classes for handling JWT auth (which will skip the persistence-related logic). (eg in `Doorkeeper::Oauth::AuthrorizationCodeRequest` and pretty much the same logic is in all classes which represent different OAuth flows).
https://github.com/doorkeeper-gem/doorkeeper/blob/252dea036bcee834f0f1acbf1fca352bcb553de9/lib/doorkeeper/oauth/authorization_code_request.rb#L27-L43

PR #1516, opened 2021-06-30
Form errors using Bootstrap's invalid-class

itay-grudev:
### Summary

Changed `doorkeeper_errors_for` to use Bootstrap's `invalid-feedback` class, which produces red highlighted text next to the corresponding field. Below is a screenshot of how it looks like:

image


nbulaj (approved):
Nice one :+1:

PR #1518, opened 2021-08-05
remove redirect uri from auth request

pauloancheta:
Hello everyone!

I'm currently stumped on this one. According to the company I'm trying to integrate with, `redirect_uri` shouldn't be required for requesting an access token. Which makes sense to me but I'm not 100% in what is supposed to happen. I tried making request_uri optional from my DB but it looks like `request_uri` is required from the `AuthorizationCodeRequest`.

```bash
# account linking
account_linking=$(
    curl -X GET ${BASE_URL}/oauth/authorize \
    -F response_type=code \
    -F scopes=${SCOPES} \
    -F redirect_uri=${REDIRECT_URI} \
    -F client_id=${CLIENT_ID} \
    -F client_secret=${CLIENT_SECRET}
)

# request access token
# redirect_uri shoulnd't be required here
code="tOgZlTLHWtfFpqxsHSDRCOID-ky0VU2Yvpc9Mw1Dh0g"
access_token=$(
    curl -vvv -X POST ${BASE_URL}/oauth/token \
    -F grant_type=authorization_code \
    -F code=${code} \
    -F client_id=${CLIENT_ID} \
    -F client_secret=${CLIENT_SECRET}
)

# echo $account_linking
echo $access_token
```

This PR removes this `request_uri` requirement.

### DIFF
--- a/lib/doorkeeper/oauth/authorization_code_request.rb
+++ b/lib/doorkeeper/oauth/authorization_code_request.rb
@@ -7,7 +7,7 @@ module Doorkeeper
       validate :client,       error: :invalid_client
       validate :grant,        error: :invalid_grant
       # @see https://tools.ietf.org/html/rfc6749#section-5.2
-      validate :redirect_uri, error: :invalid_grant
+      validate :redirect_uri, error: :invalid_grant, if: -> { !Doorkeeper.config.allow_blank_redirect_uri }
       validate :code_verifier, error: :invalid_grant
 
       attr_reader :grant, :client, :redirect_uri, :access_token, :code_verifier,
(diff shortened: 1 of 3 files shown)


nbulaj:
Hi @pauloancheta . Thanks for reporting this!

Indeed redirect_uri should be optional per OAuth docs:

pauloancheta:
@nbulaj Thank you for a speedy reply! After reading up on the classes, I do see that I can't just remove the validation. That would break existing behaviour. What do you think about using the config `Doorkeeper.configuration.allow_blank_redirect_uri`.

Might be because it's Friday but it's confusing that `false` disables the flag. To me, having false states that "do not allow blank redirect_uri` which is the opposite of what the documentation says.

nbulaj:
This will be the correct way to solve the issue @pauloancheta - https://github.com/doorkeeper-gem/doorkeeper/pull/1519
Thanks once again for raising this!

PR #1521, opened 2021-08-16
Be able to reuse_access_token

pauloancheta:
I was trying to refresh tokens on my application and it was creating new tokens instead of reusing the access tokens as the flag should do.


nbulaj:
Hi @pauloancheta . I have to check what OAuth2 specs say about the refresh tokens (I mean if they could be reused). Thanks for the PR btw!

dshoemaker:
Hi all, 

First, many thanks for working on this project. I want to chime in here and provide a use case for reusing refresh tokens. Certain client applications may need a token that acts as a service token, these tokens will be used long-term and the resource owner will not be available to re-authorize when the refresh token is expired. 

Per the RFC that @nbulaj linked above https://datatracker.ietf.org/doc/html/rfc6749#section-1.5:
>(H)  The authorization server authenticates the client and validates
        the refresh token, and if valid, issues a new access token (and,
        optionally, a new refresh token).
 
This section here indicates that new refresh tokens are optional. Doorkeeper should allow refresh token reuse because it more closely matches the RFC. Happy to work on this if you all believe it's a reasonable thing to add.

PR #1522, opened 2021-08-29
all link that was redirected is now changed to the true url

mal-wtag:
### Summary

Whenever I was going through the doorkeeper's documentation and clicking the links with `http://tools.ietf.org/html/draft-ietf-oauth-v2-22#section-4.1` it was sometimes failing to redirect. Also as all the link was from drafts some working links didn't have the proper information. That's why I tried to fix all the URLs in the documentation so that no one faces this anymore.

### Other Information

Also if you search the website for topics like `Oauth Authorization` you will see they are also notifying that their links have changed for example `RFC 6749 (was draft-ietf-oauth-v2)`

Screen Shot 2021-08-29 at 5 59 04 PM

So I thought maybe it's best if the doorkeeper has all the new links.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,18 +14,18 @@ functionality to your Ruby on Rails or Grape application.
 
 Supported features:
 
-- [The OAuth 2.0 Authorization Framework](https://tools.ietf.org/html/rfc6749)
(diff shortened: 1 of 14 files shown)


nbulaj (approved):
I like it, thanks!
LGTM :+1:

PR #1525, opened 2021-09-10
Change translation source for forbidden token error message.

NeilSlater-Clixifix:
### Summary

Changes error message source text for forbidden tokens due to incorrect scope when accessing a resource. 

The previous re-use of the scope descriptions leads to odd error messages, e.g. `{"error": "Access your public data"}`, because the text is intended for a different purpose.

The fix is to use a new dedicated source for the text of fobidden token errors.

The new locale key is `doorkeeper.errors.messages.forbidden_token.missing_scope`. The new default English error response this produces is `{"error": 'Access to this resource requires scope "public".'}`

### DIFF
--- a/lib/doorkeeper/oauth/forbidden_token_response.rb
+++ b/lib/doorkeeper/oauth/forbidden_token_response.rb
@@ -23,7 +23,8 @@ module Doorkeeper
       end
 
       def description
-        @description ||= @scopes.map { |s| I18n.t(s, scope: %i[doorkeeper scopes]) }.join("\n")
+        @description ||= I18n.t("doorkeeper.errors.messages.forbidden_token.missing_scope",
+                                oauth_scopes: @scopes.map(&:to_s).join(" "),)
       end
 
       protected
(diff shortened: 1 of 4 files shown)


lib/doorkeeper/oauth/forbidden_token_response.rb
```diff
@@ -3,6 +3,8 @@
 module Doorkeeper
   module OAuth
     class ForbiddenTokenResponse < ErrorResponse
+      MISSING_SCOPE_I18N = "doorkeeper.errors.messages.forbidden_token.missing_scope"
```
nbulaj, line 6: I don't sure we have to allocate a constant for this, let's move it below

nbulaj (approved):
LGTM :+1: 
Just a minor change

lib/doorkeeper/oauth/forbidden_token_response.rb
NeilSlater-Clixifix: OK, I will make that change, also will resolve conflict with CHANGELOG.

NeilSlater-Clixifix: @nbulaj I have adjusted the code to not use a constant, and rebased over latest `main` to resolve the conflict in CHANGELOG

Issue #1527, opened 2021-09-15
Authorization request doesn't strictly validate redirect_uri query part


pluff:
### Summary

redirect_uri at authorization request stage is not validated to match exactly one of redirect_uri in registered application. [Oauth spec states it should be exact match](https://www.oauth.com/oauth2-servers/redirect-uris/redirect-uri-validation/) but doorkeeper allows custom query params in redirect_uri.

This can be a problem for security reasons if target of `redirect_uri` is vulnerable to Reflected XSS or other attacks.

### Steps to reproduce

1. Register an app with redirect_uri = "https://example.com/redir.php"
2. Make a call to 
`/oauth/authorize?client_id=CLIENT_ID&redirect_uri=https://example.com/redir.php?myxssparam=X`
3. The app is authorized and redirected to malicious redirect uri.

### Expected behavior

According to [Oauth spec redirect_uri in request must be exact match](https://www.oauth.com/oauth2-servers/redirect-uris/redirect-uri-validation/)

### Actual behavior

Query part is skipped here:
https://github.com/doorkeeper-gem/doorkeeper/blob/main/lib/doorkeeper/oauth/pre_authorization.rb#L99-L111
https://github.com/doorkeeper-gem/doorkeeper/blob/main/lib/doorkeeper/oauth/helpers/uri_checker.rb#L47-L49
https://github.com/doorkeeper-gem/doorkeeper/blob/47037da4def738e4cfd930bd433f35629a5869f6/lib/doorkeeper/oauth/helpers/uri_checker.rb#L18-L39

PR #1528, opened 2021-09-16
Fix redirect_uri validation with query params

pluff:
### Summary

According to OAuth 2.0 protocol redirect_uri
shouldn't contain any extra query params.
We should consider exact query params list match
when comparing redirect_uris

More on topic: https://tools.ietf.org/id/draft-ietf-oauth-security-topics-18.html#name-protecting-redirect-based-f

Fixes https://github.com/doorkeeper-gem/doorkeeper/issues/1527

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ User-visible changes worth mentioning.
 ## main
 
 - [#1519] Allow to skip redirect_uri parameter for authorization request as stated in RFC 6749 section 4.1.1.
+- [#1528] Don't allow extra query params in redirect_uri
 - [#PR ID] Add your PR description here.
 
 ## 5.5.2
--- a/lib/doorkeeper/oauth/helpers/uri_checker.rb
+++ b/lib/doorkeeper/oauth/helpers/uri_checker.rb
@@ -19,12 +19,13 @@ module Doorkeeper
           url = as_uri(url)
           client_url = as_uri(client_url)
 
-          unless client_url.query.nil?
+          unless client_url.query.nil? && url.query.nil?
             return false unless query_matches?(url.query, client_url.query)
 
             # Clear out queries so rest of URI can be tested. This allows query
             # params to be in the request but order not mattering.
             client_url.query = nil
+            url.query = nil
           end
 
           # RFC8252, Paragraph 7.3
@@ -34,7 +35,6 @@ module Doorkeeper
             client_url.port = nil
           end
 
-          url.query = nil
           url == client_url
         end
 
--- a/spec/lib/oauth/authorization_code_request_spec.rb
+++ b/spec/lib/oauth/authorization_code_request_spec.rb
@@ -130,11 +130,11 @@ RSpec.describe Doorkeeper::OAuth::AuthorizationCodeRequest do
   end
 
   context "when redirect_uri contains some query params" do
-    let(:redirect_uri) { client.redirect_uri + "?query=q" }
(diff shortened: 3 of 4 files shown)


nbulaj (approved):
Hey @pluff , thanks for the PR :handshake:  !
I checked the document and think that you're right so :green_circle:  light

rmm5t:
Concern. This should have never been incorporated into a patch-level release (5.5.2 -> 5.5.3). This is a breaking change, even if it is more of a direct interpretation of the spec.

rmm5t:
@pluff @nbulaj  omniauth-oauth2 sends the full callback URL (with query params) as the `redirect_uri` during the authorization code verification. I think this PR makes sense for the initial `PreAuthorization#validate_redirect_uri` check, but it doesn't make sense for the `AuthorizationCodeRequest#validate_redirect_uri` check of the `redirect_uri`.

Thoughts? 

In other words, this PR breaks integration with omniauth-oauth2.

rmm5t:
To be fair, It may very well be the case that this is a long standing issue with omniauth-oauth2:
https://github.com/omniauth/omniauth-oauth2/pull/70#issuecomment-567729265

menisy:
Is this still allowing the `state` query param to be allowed?
https://www.oauth.com/oauth2-servers/redirect-uris/redirect-uri-registration/#per-request

nbulaj:
Above one sounds reasonable :thinking: :point_up:

menisy:
@nbulaj doesn't seem to work, I produced a failing spec. I can work on it.

nbulaj:
Yeah I checked original RFC and found a lot of spec for the `state` parameter:

* https://datatracker.ietf.org/doc/html/rfc6749#section-3.1.2.2
* https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.1
* ... and more ...

menisy:
@nbulaj let me know if [this](https://github.com/doorkeeper-gem/doorkeeper/pull/1534) looks okay, please and thank you!

menisy:
Seems like I misinterpreted the spec about the `state` param being part of the `redirect_uri`.

However, this implementation is also not very accurate as per this excerpt from the spec:
https://datatracker.ietf.org/doc/html/rfc6749#section-3.1.2.2

 The authorization server SHOULD require the client to provide the
   complete redirection URI (the client MAY use the "state" request
   parameter to achieve per-request customization).  **If requiring the
   registration of the complete redirection URI is not possible, the
   authorization server SHOULD require the registration of the URI
   scheme, authority, and path (allowing the client to dynamically vary
   only the query component of the redirection URI when requesting
   authorization)**.

With this change, this is no longer possible!

nbulaj:
Released 5.5.4 with reverted changes

menisy:
@nbulaj thanks a lot for the prompt action, let's look into this more closely once you (we) have the time.

nbulaj:
@pluff had to revert the PR because of different rules for redirect URI comparison across the docs. We should look at the original RFC more closely and attentively. Sorry for that :disappointed:

menisy:
Seems to be more of an issue on the client side:
https://github.com/omniauth/omniauth-oauth2/issues/141

It feels a bit like egg/chicken situation, the more I read the spec the more I'm getting confused TBH.

nbulaj:
Yeah, the same thoughts. One of the ideas is to check OAuth implementation in other frameworks/languages to check main trend-offs for redirect URI validation. General RFC allows even pattern matching, but we still don't support it. So.. we have to "fight" with it :)

PR #1529, opened 2021-09-20
redirect_uri is optional in token request if it was not present in authorization request

danielsbird:
### Summary

This PR implements section 4.1.3 of the OAuth 2.0 specification with regards to `redirect_uri`:

> 4.1.3.  Access Token Request
>
> The client makes a request to the token endpoint by sending the following parameters using the "application/x-www-form-urlencoded" format per Appendix B with a character encoding of UTF-8 in the HTTP request entity-body:
>
> redirect_uri
>     REQUIRED, if the "redirect_uri" parameter was included in the authorization request as described in Section 4.1.1, and their values MUST be identical.

Right now Doorkeeper requires that the `redirect_uri` parameter is included in the access token request even if it was not included in the authorization request. This PR makes `redirect_uri` optional if it was not included in the authorization request.

This PR isn't ready to be merged because I haven't yet figured out what code I need to write so that this change can be easily integrated into an existing application that uses Doorkeeper. I'm hoping someone can give me some direction on how to do that. The problem I need to overcome is that in order for the Ruby code in this PR to work it requires that a `NOT NULL` constraint be dropped from one of the database tables generated by Doorkeeper:

https://github.com/doorkeeper-gem/doorkeeper/blob/d9d51f091335bf23356bd6a96b5ade2dacbbe49f/lib/generators/doorkeeper/templates/migration.rb.erb#L26

If an application that uses Doorkeeper were to try using the code I have written in this PR without first making the `redirect_uri` column nullable then that application would encounter an error when trying to insert a row into the `oauth_access_grants` table with a null `redirect_uri`.

I think what I need to do is create a generator that can be run by developers before they try to integrate the changes in this PR into their application. The generator will create an Active Record database migration that removes the `NOT NULL` constraint from the `redirect_uri` column of the `oauth_access_grants` table. I imagine the generator will be similar to the one that enables Doorkeeper to implement PKCE: https://github.com/doorkeeper-gem/doorkeeper/blob/2833eeed725c7c5dcfa53946dc424c108588d3d8/lib/generators/doorkeeper/pkce_generator.rb

Is that the right thing to do?


lib/doorkeeper/oauth/authorization_code_request.rb
```diff
@@ -75,9 +75,18 @@ def validate_grant
...
+        #     authorization request as described in Section 4.1.1, and their
+        #     values MUST be identical.
+        #
+        # @see https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3
+        return true if redirect_uri.blank? && grant.redirect_uri.blank?
+
         Helpers::URIChecker.valid_for_authorization?(
           redirect_uri,
-          grant.redirect_uri,
+          grant.redirect_uri.presence || client.redirect_uri,
```
danielsbird, line 89: The purpose of this line of code is to tell Doorkeeper what to do in a situation where the authorization request did not include a `redirect_uri` param but the access token request does include a `redirect_uri` param. As far as I know the OAuth 2.0 specification doesn't say what should happen in that situation.

> redirect_uri
>        #     REQUIRED, if the "redirect_uri" parameter was included in the
>        #     authorization request as described in Section 4.1.1, and their
>        #     values MUST be identical.
https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3

My options were:
* make the token request fail because the `redirect_uri` param is present
* ignore the presence of the `redirect_uri` param
* validate the `redirect_uri` against the application's pre-registered `redirect_uri` (the same pre-registered `redirect_uri` which Doorkeeper redirected the resource owner to during the authentication request)

I chose the third option.

lib/doorkeeper/oauth/pre_authorization.rb
```diff
@@ -97,7 +97,14 @@ def validate_resource_owner_authorize_for_client
...
 
       def validate_redirect_uri
-        return false if redirect_uri.blank?
+        # 4.1.1.  Authorization Request
+        #   redirect_uri
+        #     OPTIONAL.  As described in Section 3.1.2.
+        #
+        # @see https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.1
+        return true if redirect_uri.blank? &&
+                       client.redirect_uri.present? &&
```
danielsbird, line 106: I'm checking for the presence of `client.redirect_uri` because I think it's possible for a Doorkeeper application to not have a pre-registered redirect URI. My understanding is that if an application allows the authorization code flow and/or implicit flow and that application doesn't have a pre-registered redirect URI then that is a misconfiguration but, since I think it is possible, I decided to handle that situation by rejecting authorization requests that don't include a `redirect_uri` param if the target application does not have a pre-registered redirect URI.

spec/lib/oauth/authorization_code_request_spec.rb
```diff
@@ -169,6 +156,41 @@
     end
   end
 
+  context "when the grant's redirect_uri is present" do
```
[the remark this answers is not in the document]
danielsbird: I _think_ this is fixed. When I run `rubocop spec/lib/oauth/authorization_code_request_spec.rb` in my terminal I see `1 file inspected, no offenses detected`.

Am I supposed to resolve this comment? I believed Hound would resolve it once I fixed the problem.

danielsbird:
@nbulaj This is ready for review. 🙂

PR #1534, opened 2021-10-05
Allow state param query in redirect_uri

menisy:
### Summary

This PR allows the use of `state` query param in the `redirect_uri`.
According to the specs here:
https://datatracker.ietf.org/doc/html/rfc6749#section-3.1.2.2
here
https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.1
and according to the guide [here](https://www.oauth.com/oauth2-servers/redirect-uris/redirect-uri-registration/#per-request) as well, it is recommended that the client can use the `state` query param to customize the redirect URI according to its needs.

### Other Information

PS: The spec also mention an optional query param `scope`. It's not being covered by this PR though. If you think it should be, please let me know and I will modify the implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ User-visible changes worth mentioning.
 
 ## main
 
-- [#PR ID] Add your PR description here.
+- [#1534] Allow the use of `state` query param in the `redirect_uri` conforming with the specs.
 
 ## 5.5.3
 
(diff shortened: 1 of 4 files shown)


menisy:
Closed, it seems that I misinterpreted the spec, the `state` is a param passed along the `redirect_uri` in the authorization request and not part of the `redirect_uri` itself.

PR #1535, opened 2021-10-05
Revert disallowing query params in redirect_uri

menisy:
### Summary

This PR reverts the change introduced in #1528 for disallowing query params in `redirect_uri`. As per the specs, it SHOULD (even though discouraged) be allowed for the client to set query params in the redirect URI if it needs to. In that case, the authorization server should only match the scheme, authority, and path.

Ref: https://datatracker.ietf.org/doc/html/rfc6749#section-3.1.2.2

> The authorization server SHOULD require the client to provide the
complete redirection URI (the client MAY use the "state" request
parameter to achieve per-request customization). **If requiring the
registration of the complete redirection URI is not possible, the
authorization server SHOULD require the registration of the URI
scheme, authority, and path (allowing the client to dynamically vary
only the query component of the redirection URI when requesting
authorization).**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ User-visible changes worth mentioning.
 
 ## main
 
-- [#PR ID] Add your PR description here.
+- [#1535] Revert changes introduced in #1528 to allow query params in `redirect_uri` as per the spec.
 
 ## 5.5.3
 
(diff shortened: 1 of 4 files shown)


nbulaj:
Well, the important thing is the first part of the sentence:

> If requiring the registration of the complete redirection URI is not possible

I don't see the reasons why it's impossible for Doorkeeper :thinking: 

Also I have to find more time to go through the docs more carefully with the redirect URI validation rules

nbulaj:
Anyway, since this change break a lot of apps I consider to merge and release it. After we(I) can think about more detailed check of the specs and releasing minor version with the fix

nbulaj (approved):

PR #1549, opened 2022-01-19
Add support for rails ~> 7.0.0

joeyparis:
### Summary

Expands dependency range to include Rails ~> 7.0.0

### Other Information

Just expands the Rails dependency range to include the first versions of Rails 7. I needed to add `sprockets-rails` to the Gemfile for the tests to succeed, but after adding that to the Gemfile all the tests still pass.


nbulaj (approved):
LGTM :+1:

PR #1552, opened 2022-02-02
Fix #1541: Remove duplicated IDs to improve accessibility

### LINKED ISSUE #1541: Accessibility - duplicate IDs on authorization page
### Context
We're working to make our application accessible such that it meets the [WCAG - Web Content Accessibility Guidelines](https://www.w3.org/TR/WCAG20-TECHS/). 

### Steps to reproduce
If we access the web page rendered by the [authorization/new.html.erb file](https://github.com/doorkeeper-gem/doorkeeper/blob/main/app/views/doorkeeper/authorizations/new.html.erb) and we inspect the submit buttons we find duplicate IDs that are causing [this guideline failure](https://www.w3.org/TR/WCAG20-TECHS/F77.html). To replicate this you can paste the HTML from the previously mentioned file into [this HTML checker tool](https://validator.w3.org/nu/#textarea).

The reason why this happens is that the page has the same ID on the fields generated by the **hidden_field_tag** for the 2 buttons (authorize and deny)
(I am attaching a screenshot here of the 2 HTML forms that have fields with duplicated IDs)
Selecție_098

### Expected behavior
The HTML page should not have duplicated IDs.

### Actual behavior
The HTML page raises an accessibility error due to duplicated IDs.
Selecție_097

### Suggested solution
A solution would be to change the **id** attribute form the HTML generated to **class**
In order to do that I would suggest instead of using **hidden_field_tag** - you could use something like: 
```ruby
<input type="hidden" name="client_id" class="client_id" value="<%= @pre_auth.client.uid %>"> 
```

This will change the HTML into this:
Selecție_099

Which would make this page meet the accessibility guidelines.

nbulaj:
Fixes #1541

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ User-visible changes worth mentioning.
 
 ## main
 
+- [#1552] Remove duplicate IDs on Auth form to improve accessibility.
 - [#1542] Improve performance of `Doorkeeper::AccessToken#matching_token_for` using database specific SQL time math.
 
   **[IMPORTANT]**: API of the `Doorkeeper::AccessToken#matching_token_for` method has changed and now it returns
--- a/app/views/doorkeeper/authorizations/new.html.erb
+++ b/app/views/doorkeeper/authorizations/new.html.erb
@@ -21,25 +21,25 @@
 
   <div class="actions">
     <%= form_tag oauth_authorization_path, method: :post do %>
-      <%= hidden_field_tag :client_id, @pre_auth.client.uid %>
-      <%= hidden_field_tag :redirect_uri, @pre_auth.redirect_uri %>
-      <%= hidden_field_tag :state, @pre_auth.state %>
-      <%= hidden_field_tag :response_type, @pre_auth.response_type %>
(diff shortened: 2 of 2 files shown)


nbulaj:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Warning">
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">There're library changes, but not tests. That's OK as long as you're refactoring existing code.</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please include a changelog entry. <br />
You can find it at <a href="https://github.com/doorkeeper-gem/doorkeeper/blob/main/CHANGELOG.md">CHANGELOG.md</a>.You can skip this warning only if you made some typo fix or other small changes that didn't affect the API.</td>
    </tr>
  </tbody>
</table>

Here's an example of a CHANGELOG.md entry:
```markdown
- [#1552] Fix #1541: remove duplicate ids for accessibility.
```


<p align="right" data-meta="generated_by_danger">
  Generated by :no_entry_sign: <a href="https://danger.systems/">Danger</a>
</p>

Issue #1557, opened 2022-02-21
Able to obtain a token with default scopes even if they are not present in the application scopes using client credentials request


enrico:
When grant type is "client credentials" and application scopes are present:

If the the scope parameter is NOT present in the `/oauth/token` request then this api is returning a token with default scopes (for exp. 'public' scope) even if 'public' scope is not present in the application scopes.

This is similar to issue #1002

nbulaj:
I think we can close this #1558 being merged

PR #1558, opened 2022-02-21
Bug: able to obtain a token with default scopes even if they are not present in the application scopes when using client credentials

enrico:
see description of issue #1557


nbulaj (approved):
LGTM

nbulaj:
Thanks @enrico

enrico:
@nbulaj you're welcome!
I'm assuming you are planning to incorporate this in 5.6.0 , correct? If so, do you have an ETA for the release?

PR #1570, opened 2022-05-16
Per realm configuration

moiristo:
### Summary

Our case is as follows:

* We have an API that can be accessed by registered and authorized 'apps'. 
  * These apps are OAuth2 applications (i.e. `application_class  'AppStore::Application')`. 
  * Scopes look like `category|read`, `user|read`, `message|write`
  * Resource owners are `Organization` records
* A new requirement in our system is to support OpenId Connect. This use case is very different:
  * Applications are the default `Doorkeeper::Application` records
  * Scopes are `openid`, `email`, `profile`
  * Resource owners are `User` records

To support this, I've modified doorkeeper to support multiple realms. You can configure a realm using `Doorkeeper.configure(realm)`. Not specifying a realm is basically the same as how DK currently works. So, for our case I have two configure blocks: `Doorkeeper.configure {}` and `Doorkeeper.configure('api.example.com') {}`. 

Switching between realms happens as follows:

* On the model level, you can specify `self.doorkeeper_realm` to set the realm of that particular model
* On controller level, you can specify `Doorkeeper::Current.realm`, for example in a `prepend_before_action`
* I have custom controllers for the API controllers, therefore I've wrapped the doorkeeper routes in a module to support this:
  ```ruby
  scope module: :app_store do
    use_doorkeeper(realm: 'api.example.com') { skip_controllers :applications, :authorized_applications }
  end
  ```
* In these controllers, I inherit from the base controllers to customize:
  ```ruby
  class AppStore::Doorkeeper::AuthorizationsController < ::Doorkeeper::AuthorizationsController
  end
  ```

This PR is not finished as the new functionality is not properly tested yet. The current tests pass though. Things that still need to be done:

* Add tests
* Move the `base_controller` settings somewhere. Setting it per realm makes no sense.
* Version bump

The big question though: would you be interested in a finished PR anyway? As I couldn't really find anyone else needing this, I would understand if you don't see any benefit in adding support for multiple realms.

### DIFF
--- a/lib/doorkeeper.rb
+++ b/lib/doorkeeper.rb
@@ -19,6 +19,7 @@ module Doorkeeper
   autoload :AccessGrantMixin, "doorkeeper/models/access_grant_mixin"
   autoload :AccessTokenMixin, "doorkeeper/models/access_token_mixin"
   autoload :ApplicationMixin, "doorkeeper/models/application_mixin"
+  autoload :Current, "doorkeeper/models/current"
 
   module Helpers
     autoload :Controller, "doorkeeper/helpers/controller"
@@ -86,6 +87,7 @@ module Doorkeeper
   end
 
   module Models
+    autoload :Configurable, "doorkeeper/models/concerns/configurable"
     autoload :Accessible, "doorkeeper/models/concerns/accessible"
     autoload :Expirable, "doorkeeper/models/concerns/expirable"
     autoload :ExpirationTimeSqlMath, "doorkeeper/models/concerns/expiration_time_sql_math"
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -21,16 +21,18 @@ module Doorkeeper
   class MissingConfigurationBuilderClass < StandardError; end
 
   class << self
-    attr_reader :orm_adapter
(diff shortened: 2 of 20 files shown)


moiristo:
Thanks for the feedback! 👍

nbulaj:
Hey @moiristo . Sorry, didn't have time for proper look into this PR.

Unfortunately this looks for me like some specific case which currently brings a lot of changes into the gem (and could break some existing setups which we are trying to avoid as much as possible). I don't know how many projects / devs / etc will need such behavior. My vision for the gem is to have the core OAuth functionality in it and implemented in a such way where you can simply override/configure some functionality. That's why I prefer extensions over core changes (openid_connect, resource indicators and other gems). 

We can leave this PR opened and see if anybody else will give it a thumb up :+1:  so I can see such case is commonly needed. Thanks for understanding!

moiristo:
Thanks, I understand! As this was part of an issue in which we could opt for Oauth and SAML, we decided in the end to use SAML instead, so this code is currently not in use anyway. Overriding functionality to support this was something I did try, but the code became very messy.

I think I'm indeed the only one atm as this PR has been open for months and did not receive any thumbs up during that time :) So it's fine with me to leave it closed, I think people will still be able to find it anyway.

PR #1572, opened 2022-06-01
Add support for configuring a custom application secret generator

dbussink:
### Summary

Right now the application secret generator can't be configured and the only way to ensure that application secrets have an identifiable prefix is by overwriting `renew_secret`.

Generating identifiable tokens and secrets is important and relevant for things like the GitHub Secret Scanning program and we use this already for other things like regular OAuth tokens where we overwrite them with a prefix so they are clearly identifiable for secret scanning.

The changes here also allow such an override for the application secret.

### Other Information

This adds a new reference to the documentation, but that documentation would also need to be created. If this change is considered a good idea, I'm happy to also send a pull request for the documentation then to add it there as well.

### DIFF
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -387,11 +387,20 @@ module Doorkeeper
     option :access_token_generator,
            default: "Doorkeeper::OAuth::Helpers::UniqueToken"
 
+    # Use a custom class for generating the application secret.
+    # https://doorkeeper.gitbook.io/guides/configuration/other-configurations#custom-application-secret-generator
+    #
+    # @param application_secret_generator [String]
+    #   the name of the application secret generator class
+    #
+    option :application_secret_generator,
+           default: "Doorkeeper::OAuth::Helpers::UniqueToken"
+
     # Default access token generator is a SecureRandom class from Ruby stdlib.
     # This option defines which method will be used to generate a unique token value.
     #
-    # @param access_token_generator [String]
-    #   the name of the access token generator class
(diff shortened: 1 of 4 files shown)


dbussink:
The code climate suggestion here is something I'm not sure how the Doorkeeper project would like that to be fixed. A new helper class? Is the duplication ok here? Suggestions welcome on that one.

nbulaj:
> Another question I have that can also be a separate PR, is whether it's worth it / useful to also be able to configure the OAuth Application UID generation? Right now that's also using a base64 identifier, but I think it can be a nicer user experience for end users if that could also be a different format?

Agree, but don't sure how useful and popular will be such option. So.. don't know what to say :) My vision is not to pollute config with *eveything everybody needs* but allow to easily customize behavior in a custom setup classes like we have for the models (without any need to play with `prepend` and so on)

dbussink:
> Agree, but don't sure how useful and popular will be such option. So.. don't know what to say :) My vision is not to pollute config with _eveything everybody needs_ but allow to easily customize behavior in a custom setup classes like we have for the models (without any need to play with `prepend` and so on)

Yeah, so we are doing this now from a usability perspective. We didn't want to have the application uid as case sensitive and do things like setting up a case sensitive collation on the unique database index. 

Making it case insensitive was nicer, since it also avoids issues for the user as many other "username" like identifiers are normally case insensitive. 

As for what the best say to customize is, right now we override `generate_uid` which works but of course isn't a supported API. If there's a preference for allowing it in some way, happy to also open a PR for that.

PR #1575, opened 2022-06-06
pass application into AccessToken#create method

### LINKED ISSUE #1574: Doorkeeper::AccessToken.find_or_create_for unnecessarily refetches application object
When creating a new token for an application, despite passing an application object into the method, it's refetched from the database when the token is generated.

### Steps to reproduce
Create a token for an existing application:
```ruby
app = Doorkeeper::Application.find_or_create_by(name: "API")
Doorkeeper::AccessToken.find_or_create_for(
      application: app,
      resource_owner: id,
      scopes: Doorkeeper::OAuth::Scopes.from_array(%w(public))
)
```

### Expected behavior
When passing an object into the `find_or_create_for` method, it should use that object reference rather than refetching the application object from the DB.

### Actual behavior
Here is the `create_for(...)` method
```ruby
      def create_for(application:, resource_owner:, scopes:, **token_attributes)
        token_attributes[:application_id] = application&.id
        token_attributes[:scopes] = scopes.to_s

        if Doorkeeper.config.polymorphic_resource_owner?
          token_attributes[:resource_owner] = resource_owner
        else
          token_attributes[:resource_owner_id] = resource_owner_id_for(resource_owner)
        end
        create!(token_attributes)
      end
```

Because `:application` is used as a keyword argument, it's removed from the `token_attributes` hash which is passed further down into the `create!` method. Anytime a new `AccessToken` is created, `generate_token` is called which in turn calls `attributes_for_token_generator`, which passes a full application reference to the token generator.

```ruby
    # Set of attributes that would be passed to token generator to
    # generate unique token based on them.
    #
    #  @return [Hash] set of attributes
    #
    def attributes_for_token_generator
      {
        resource_owner_id: resource_owner_id,
        scopes: scopes,
        application: application, # see here, this causes a refetch
        expires_in: expires_in,
        created_at: created_at,
      }.tap do |attributes|
        if Doorkeeper.config.polymorphic_resource_owner?
          attributes[:resource_owner] = resource_owner
        end
      end
    end
```

Because the `create!` method is only getting passed the `application_id` and not the full application, when `application` is called from inside `attributes_for_token_generator` it causes the application to be fetched from the database again.  

This could be avoided by passing the application into `token_attributes` rather than passing in the `id` only.

### System configuration
Doorkeeper 5.4.0

stereoscott:
This avoids refetching the application object from the database when it's referenced inside of `attributes_for_token_generator`. Closes #1574

### DIFF
--- a/lib/doorkeeper/models/access_token_mixin.rb
+++ b/lib/doorkeeper/models/access_token_mixin.rb
@@ -212,7 +212,7 @@ module Doorkeeper
       # @return [Doorkeeper::AccessToken] new access token
       #
       def create_for(application:, resource_owner:, scopes:, **token_attributes)
-        token_attributes[:application_id] = application&.id
+        token_attributes[:application] = application
         token_attributes[:scopes] = scopes.to_s
 
         if Doorkeeper.config.polymorphic_resource_owner?
--- a/spec/lib/oauth/base_request_spec.rb
+++ b/spec/lib/oauth/base_request_spec.rb
@@ -18,7 +18,7 @@ RSpec.describe Doorkeeper::OAuth::BaseRequest do
            created_at: 0
   end
 
-  let(:client) { double :client, id: "1" }
+  let(:client) { Doorkeeper::Application.new(id: "1") }
 
   let(:scopes_array) { %w[public write] }
 


nbulaj:
I thought it should be cached if already called :thinking: 
Also I;m wondering if we could break any existing ORM extension with such changes. Have to think

stereoscott:
> I thought it should be cached if already called 🤔
> Also I;m wondering if we could break any existing ORM extension with such changes. Have to think

In my case the application passed into the method was initially fetched by name, but within the`AccessToken` model it fetches by ID so it missed the query cache.

It's likely this hasn't come up before because for the built-in implementation where tokens are being generated via `Doorkeeper::ApplicationsController` the application is fetched by the `application_id` provided in the params.  Not sure of other use cases where the application object may not be fetched by ID and if those warrant this change.

nbulaj:
OK, analyzed some source code and I think it should be fine. Thanks!

nbulaj (approved):
LGTM

Could you please add an entry to the CHANGELOG ?

nbulaj:
Aha, tests shows some issues

PR #1593, opened 2022-10-22
Add support for Trilogy MySQL adapter

zavan:
This adds support for GitHub's [Trilogy](https://github.blog/2022-08-25-introducing-trilogy-a-new-database-adapter-for-ruby-on-rails/)  MySQL adapter to `ExpirationTimeSqlMath`. It uses the same class as `mysql` and `mysql2`.


nbulaj (approved):
LGTM :+1:

PR #1597, opened 2022-11-03
Add optional support to use the url path for the native authorization code flow. Ports forward [#1143] from 4.4.3

starsolutions:
### Summary

Provide the same option to opt out of the breaking api change introduced in https://github.com/doorkeeper-gem/doorkeeper/pull/1003 that was provided in the 4.x branch via this pr(https://github.com/doorkeeper-gem/doorkeeper/pull/1144).   

Added comments regarding this issue to the original issue https://github.com/doorkeeper-gem/doorkeeper/issues/1143

### Other Information

The reason for pulling this option forward is that there are users who are stuck on the 4.x branch because they have clients that rely on this legacy behavior.

### DIFF
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -159,6 +159,15 @@ module Doorkeeper
         @config.instance_variable_set(:@reuse_access_token, true)
       end
 
+      # Choose to use the url path for native autorization codes 
+      # Enabling this flag sets the authorization code response route for
+      # native redirect uris to oauth/authorize/<code>. The default is
+      # oauth/authorize/native?code=<code>.
+      # Rationale: https://github.com/doorkeeper-gem/doorkeeper/issues/1143
+      def use_url_path_for_native_authorization
+        @config.instance_variable_set(:@use_url_path_for_native_authorization, true)
+      end
+
       # TODO: maybe make it more generic for other flows too?
       # Only allow one valid access token obtained via client credentials
       # per client. If a new access token is obtained before the old one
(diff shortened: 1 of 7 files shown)


jmeckley (approved):

lib/doorkeeper/config.rb
```diff
@@ -623,6 +632,11 @@ def token_grant_types
     def deprecated_token_grant_types_resolver
       @deprecated_token_grant_types ||= calculate_token_grant_types
     end
+    
+    def native_authorization_code_route
+      @opt_out_native_route_change ||= false
```
nbulaj, line 637: Let's use `unless defined?(@opt_out_native_route_change)` to be safe

lib/doorkeeper/config.rb
```diff
@@ -159,6 +159,15 @@ def reuse_access_token
         @config.instance_variable_set(:@reuse_access_token, true)
       end
 
+      # Opt out of breaking api change to the native authorization code flow.
+      # Opting out sets the authorization code response route for native
+      # redirect uris to oauth/authorize/<code>. The default is
+      # oauth/authorize/native?code=<code>.
+      # Rationale: https://github.com/doorkeeper-gem/doorkeeper/issues/1143
+      def opt_out_native_route_change
```
nbulaj, line 167: Can we rename the option to be something like `use_url_path_for_native_authorization`? To be more explicit and meaningful

nbulaj (commented):
Some small comments. Otherwise I'm OK with the proposed changes

lib/doorkeeper/config.rb
starsolutions: makes sense, updated!

lib/doorkeeper/config.rb
starsolutions: Makes sense, that's more clear, updated!

starsolutions:
Thanks for the feedback! I updated based on that, and updated a few of the comments to match the new wording.

nbulaj (approved):
LGTM :+1: 

Just one thing I forgot - could you please also add a changelog entry? :pray:

jmeckley (approved):

PR #1599, opened 2022-11-06
fix: remove unnecessarily refetches application object

saiqulhaq:
### Summary

Fix https://github.com/doorkeeper-gem/doorkeeper/issues/1574
passed the specs too

image

### DIFF
--- a/lib/doorkeeper/oauth/base_request.rb
+++ b/lib/doorkeeper/oauth/base_request.rb
@@ -28,8 +28,9 @@ module Doorkeeper
 
       def find_or_create_access_token(client, resource_owner, scopes, server)
         context = Authorization::Token.build_context(client, grant_type, scopes, resource_owner)
+        application = client.is_a?(Doorkeeper::Application) ? client : client.application if client
         @access_token = server_config.access_token_model.find_or_create_for(
-          application: client,
+          application: application,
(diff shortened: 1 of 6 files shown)


nbulaj (approved):
I believe we should use some universal interface to access the application instance :thinking: Because for now it looks tricky

Can we add a changelog entry as well? :pray:

nbulaj:
OK I wanna to release a minor update and include this one into it, so will merge and add myself

lib/doorkeeper/oauth/base_request.rb
```diff
@@ -28,8 +28,9 @@ def scopes
 
       def find_or_create_access_token(client, resource_owner, scopes, server)
         context = Authorization::Token.build_context(client, grant_type, scopes, resource_owner)
+        application = client.is_a?(Doorkeeper::Application) ? client : client.application if client
```
sealabcore, line 31: Doing this breaks apps that are using Custom application classes as defined in doorkeeper.rb

`application_class "MyCustomApplicationClass"`

And defined like so
```
# frozen_string_literal: true

class MyCustomApplicationClass < ApplicationRecord
  include ::Doorkeeper::Orm::ActiveRecord::Mixins::Application
end
```

PR #1602, opened 2022-11-25
Allow custom data to be sent inside access tokens

JeremyC-za:
### Summary

I've added a new config option - `custom_access_token_fields`. By default, this will be an empty array and have no effect, and when in use, it will be an array of symbols. This array of symbols will then be permitted parameters keys when authorising and granting access to an application. A developer can then configure their OAuth consent screen to allow custom data to be submitted during authorisation. This custom data will then be saved to the Access Grant, and will be included in any subsequently generated access tokens.

### Example Use Case

You have a multi-tenanted platform, where our example user has access to `tenant_1`, `tenant_2`, and `tenant_3`. The example user would like to grant an OAuth application access to their data, but only for `tenant_2`, and not the others. Using this new config option, the developer of the platform can specify that a `tenant_id` will be submitted when granting access - `custom_access_token_fields [:tenant_id]`. The developer can then add a `tenant_id` column to the `oauth_access_grants` and `oauth_access_tokens` database tables, and then add a tenant select dropdown to their OAuth consent screen, allowing the `tenant_id` to be selected and submitted and saved. This tenant_id is then included in access tokens. When receiving a request using one of these access tokens, then developer can add measures to check that the requested resource belongs to the tenant for which the access token was created.

### DIFF
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -321,6 +321,15 @@ module Doorkeeper
     option :access_token_generator,
            default: "Doorkeeper::OAuth::Helpers::UniqueToken"
 
+    # Allows additional data to be received when granting access to an Application, and for this
+    # additional data to be sent with subsequently generated access tokens. The access grant and
+    # access token models will both need to respond to the specified field names.
+    #
+    # @param fields [Array] The array of custom field names to be saved
+    #
+    option :custom_access_token_fields,
+           default: []
+
     # Use a custom class for generating the application secret.
     # https://doorkeeper.gitbook.io/guides/configuration/other-configurations#custom-application-secret-generator
     #
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -323,11 +323,11 @@ module Doorkeeper
 
     # Allows additional data to be received when granting access to an Application, and for this
     # additional data to be sent with subsequently generated access tokens. The access grant and
-    # access token models will both need to respond to the specified field names.
+    # access token models will both need to respond to the specified attribute names.
     #
-    # @param fields [Array] The array of custom field names to be saved
+    # @param attributes [Array] The array of custom attribute names to be saved
     #
-    option :custom_access_token_fields,
+    option :custom_access_token_attributes,
            default: []
 
     # Use a custom class for generating the application secret.
--- a/lib/doorkeeper/oauth/base_request.rb
+++ b/lib/doorkeeper/oauth/base_request.rb
@@ -26,12 +26,12 @@ module Doorkeeper
         @scopes ||= build_scopes
       end
 
-      def find_or_create_access_token(client, resource_owner, scopes, custom_fields, server)
+      def find_or_create_access_token(client, resource_owner, scopes, custom_attributes, server)
         context = Authorization::Token.build_context(client, grant_type, scopes, resource_owner)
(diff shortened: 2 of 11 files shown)


JeremyC-za:
@nbulaj let me know what you think about these changes. I'm not sure about the codeclimate issue - seems like the code will be worse if I try to refactor that 🤷‍♂️

nbulaj:
Thanks @JeremyC-za , will check a little bit later, don't have enough time. I like the idea, but I have to think if it could be done with current implementation or we still need some changes to make it possible

jonathansimmons:
Beyond just adding a thumbs up, I'll say this would be a huge feature addition allowing us to prioritize the user experience in multi-tenant setups. 

Happy to discuss or contribute, but I'm not sure I understand the doorkeeper DSL and flows enough yet.

ashkulz:
@nbulaj can you elaborate on what exactly would break for other extensions?

nbulaj:
Actually all the ORM extensions @ashkulz . Maybe more, have to be tested with doorkeeper-openid_connect as well (as it's also patches some Doorkeeper internals)

nbulaj:
@JeremyC-za could you please rebase you branch on top of current `main` one ? Thanks :bow:

lib/doorkeeper/oauth/base_request.rb
```diff
@@ -26,16 +26,20 @@ def scopes
...
       end
 
-      def find_or_create_access_token(client, resource_owner, scopes, server)
+      def find_or_create_access_token(client, resource_owner, scopes, custom_fields, server)
         context = Authorization::Token.build_context(client, grant_type, scopes, resource_owner)
+        token_model = server_config.access_token_model
         application = client.is_a?(Doorkeeper::Application) ? client : client.application if client
-        @access_token = server_config.access_token_model.find_or_create_for(
+
+        token_params = {
```
nbulaj, line 34: Let's name it as `token_attributes` and rename `custom_fields` into `custom_attributes`

lib/doorkeeper/config.rb
```diff
@@ -396,6 +396,15 @@ def configure_secrets_for(type, using:, fallback:)
     option :access_token_generator,
            default: "Doorkeeper::OAuth::Helpers::UniqueToken"
 
+    # Allows additional data to be received when granting access to an Application, and for this
+    # additional data to be sent with subsequently generated access tokens. The access grant and
+    # access token models will both need to respond to the specified field names.
+    #
+    # @param fields [Array] The array of custom field names to be saved
+    #
+    option :custom_access_token_fields,
```
nbulaj, line 405: Let's rename it into `custom_access_token_attributes`

JeremyC-za:
@nbulaj I've rebased and actioned your feedback. Let me know if there's anything else I can do.

nbulaj (approved):
LGTM :+1: 
Thanks!

ashkulz:
Thanks for merging this and doing a followup PR (#1634), @nbulaj!

ashkulz:
@nbulaj can you publish a release to RubyGems?

nbulaj:
Released with 5.6.5

raivil:
@JeremyC-za @nbulaj 
I've tried adding this new feature and found one issue with rails migrations. Added to https://github.com/doorkeeper-gem/doorkeeper/issues/1642

PR #1608, opened 2022-12-02
Simplify PreAuthorization class

nbulaj:
Simplify `PreAuthorization` class (no need to pass Doorkeeper config it, it's accessible in the class itself)


nbulaj:
Hm, it brakes Doorkeeper OpenID-connect : https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/master/lib/doorkeeper/openid_connect/oauth/pre_authorization.rb#L9

PR #1615, opened 2022-12-14
Remove doorkeeper config error

### LINKED ISSUE #1577: Missing configuration should be a log warning, not an error
Currently, if you add `gem "doorkeeper"` to your Rails project **without** an initializer, the project will fail to boot with this error:

```
/gems/doorkeeper/lib/doorkeeper/config.rb:33:in `configuration':
Configuration for doorkeeper missing.
Do you have doorkeeper initializer? (Doorkeeper::MissingConfiguration)
```

Instead, this should be simply be a warning log line, which is what all other gems do nearly universally. I've tested on the 200+ gems in my project and Doorkeeper is the only one that causes such a failure.

tonybruess:
This error should be removed. No other gem behaves like this. It prevents `tapioca` from generating static types.

Closes #1577

### DIFF
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -5,14 +5,6 @@ require "doorkeeper/config/option"
 require "doorkeeper/config/validations"
 
 module Doorkeeper
-  # Defines a MissingConfiguration error for a missing Doorkeeper configuration
-  #
-  class MissingConfiguration < StandardError
-    def initialize
-      super("Configuration for doorkeeper missing. Do you have doorkeeper initializer?")
-    end
-  end
-
   # Doorkeeper option DSL could be reused in extensions to build their own
   # configurations. To use the Option DSL gems need to define `builder_class` method
   # that returns configuration Builder class. This exception raises when they don't
@@ -30,7 +22,7 @@ module Doorkeeper
     # @return [Doorkeeper::Config] configuration instance
     #
     def configuration
-      @config || (raise MissingConfiguration)
+      @config ||= Config::Builder.new(&block).build
     end
 
     alias config configuration
@@ -159,7 +151,7 @@ module Doorkeeper
         @config.instance_variable_set(:@reuse_access_token, true)
       end
 
-      # Choose to use the url path for native autorization codes 
+      # Choose to use the url path for native autorization codes
       # Enabling this flag sets the authorization code response route for
       # native redirect uris to oauth/authorize/<code>. The default is
       # oauth/authorize/native?code=<code>.
@@ -632,7 +624,7 @@ module Doorkeeper
     def deprecated_token_grant_types_resolver
       @deprecated_token_grant_types ||= calculate_token_grant_types
     end
-    
+
     def native_authorization_code_route
       @use_url_path_for_native_authorization = false unless defined?(@use_url_path_for_native_authorization)
       @use_url_path_for_native_authorization ? '/:code' : '/native'


nbulaj (commented):
Hi @tonybruess 

Well, config is builded in the Rails initializers using the `Doorkeeper.configure { ... }` block. And it's done when Rails loads them. ORM setup and so on invoked in engine `#to_prepare` block. So if you see the error - seems like something is calling Doorkeeper files/models/etc before configuration block is invoked (monkey-patching? `prepend`s?).

Your changes just generates default config if `.configure` wasn't called before and I'm pretty sure it leads to a lot of bugs

I suggest to check your initializers and Doorkeeper setup.

tonybruess:
Hi @nbulaj,

This change is required for `sorbet/sorbet` to generate types for the doorkeeper gem. It requires the gem to generate types which means the gem must not throw an error if it is not configured. Of the hundreds of gems in my app, doorkeeper is the only one that behaves like this.

nbulaj:
Yeah, agree, it's sad, but GitLab [has 250+ gems](https://github.com/gitlabhq/gitlabhq/blob/master/Gemfile.lock) and no patches needed. I don't sure what Sorbet exactly does with the app, we need some investigation here to understand who is the caller of `Doorkeeper.config` invoked **before** app initializers loaded.

tonybruess:
@nbulaj Gitlab doesn't use tapioca/sorbet which is why they are not seeing this issue. Any rails app that uses sorbet/tapioca with doorkeeper will have this issue.

The `tapoica gem` command loads up a gem and tries to require as much of it as possible. I don't know how it specifically works but you can look at it here: https://github.com/Shopify/tapioca/blob/main/lib/tapioca/gem/pipeline.rb#L17

leemeichin:
just gonna weigh on in this and say that the current implementation is also bugged:

https://github.com/doorkeeper-gem/doorkeeper/issues/1617
https://github.com/doorkeeper-gem/doorkeeper/issues/1577

So even if this PR is done for the sake of Sorbet, there is still a bug that prevents you from setting up the gem.

nbulaj:
The problem with the configuration is that it's called somewhere **before** the initializer invoked (which we don't expect to be). Why it's called before - I don't know. Looks like Sorbet tries to load model files (which are tied to Doorkeeper config via associations, name configurations and other options) _before_ the initializer. And in other projects somerthing else loads files before the initializer invoked as well.

To fix this we have to reimplement configuration behavior and models implementation for the Doorkeeper. Or ask developers to check their Zeitwerk / Ruby / Rails / etc setp (debug why models or smth else are loaded before initializers executed).

nbulaj:
@tonybruess could you please try to set Doorkeeper in your Gemfile from this PR? https://github.com/doorkeeper-gem/doorkeeper/pull/1627 and check locally?

leemeichin:
I tried it on a clean setup on my repo and still got the same error (basically deleted the existing doorkeeper config and tried to generate a fresh one).

nbulaj:
@leemeichin could you please give one more try?

```ruby
# Gemfile

gem "doorkeeper", github: "doorkeeper-gem/doorkeeper", branch: "fixes/lazy-load-associations"
```

And report if issue still present?

leemeichin:
@nbulaj no dice, unfortunately. Still missing config error.

nbulaj:
Any backtrace you can provide @leemeichin ? I need to understand which exactly line causes an error

nbulaj:
Should be resolved via https://github.com/doorkeeper-gem/doorkeeper/pull/1627 and https://github.com/doorkeeper-gem/doorkeeper/pull/1633

PR #1620, opened 2023-01-11
Migrate danger workflow to use ruby/setup-ruby@1

sato11:
### Summary
Thanks for merging #1618 🚀 However I also find the danger workflow is not working and therefore propose that we do a follow-up. 

This is only to replace actions/setup-ruby@1 with ruby/setup-ruby@1, where what I did is basically followed the official readme of the latter: https://github.com/ruby/setup-ruby

This is going to resolve not only the runtime error but also (although hopefully) nodejs v12 deprecation warning, which to me seems to come incidentally with actions/setup-ruby@1.
https://github.com/doorkeeper-gem/doorkeeper/actions/runs/3882097161

### DIFF
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -25,4 +25,4 @@ jobs:
         danger_file: 'Dangerfile'
         danger_id: 'danger-pr'
       env:
-        DANGER_GITHUB_API_TOKEN: ${{ secrets.DANGER_GITHUB_TOKEN }}
+        DANGER_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -14,15 +14,10 @@ jobs:
       (github.event_name  == 'pull_request')
     steps:
     - uses: actions/checkout@v3
-    - uses: actions/setup-ruby@v1
+    - uses: ruby/setup-ruby@v1
       with:
-        ruby-version: '2.7'
-    - uses: actions/cache@v3
-      with:
-        path: vendor/bundle
-        key: ${{ runner.os }}-gems-${{ hashFiles('Gemfile') }}
-        restore-keys: |
-          ${{ runner.os }}-gems-
+        ruby-version: '3.2'
+        bundler-cache: true
     - uses: MeilCli/danger-action@v5
       with:
         plugins_file: 'Gemfile'


sato11:
Oops now I see MeilCli/danger-action@5 failing. Let me do some experiments I'll squash in the end 💭

sato11:
So with this change the job start to run and now it's moaning that it cannot find the commit 200c6cd.

```diff
-DANGER_GITHUB_API_TOKEN: ${{ secrets.DANGER_GITHUB_TOKEN }}
+DANGER_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

I'd like to retry the job but since it seems I'm not authorized to see retry button in the GitHub Actions UI I'm adding an empty commit for now.

sato11:
It turns out that mysterious missing commit is not a temporal glitch. I assume this is because my commit belongs to sato11/doorkeeper (forked) and the job belongs to doorkeeper-gem/doorkeeper (origin)?

I curiously find that MeilCli/danger-action has left cautious statement that it has a restriction regarding using forks: https://github.com/MeilCli/danger-action#attention-for-repository-oss-or-using-dependabot

hashie/hashie and ruby-grape/grape seem to have dropped dependency to MeilCli/danger-action to run danger and migrated to maintain danger workflow by themselves. The following PRs are interesting to learn from:

- https://github.com/hashie/hashie/pull/561
- https://github.com/ruby-grape/grape/pull/2153

But they come with an apparent cost of managing maintainer's PAT as well as responsibility of rotating them periodically. See for example:

- https://github.com/ruby-grape/grape/pull/2291

I believe this is a moment of decision, where in order to fix CI we need to either

1. use someone's PAT just like hashie and ruby-grape teams do
1. come up with a better idea so we can run danger just like it used to run
1. forget about danger for now and cancel the workflow definition

I feel more comfortable to see the green CI as soon as possible rather than we stop here and it remains red. I'd Iike to propose that we choose 3. unless danger is more vital job than I imagine. I'll be creating another PR later today.

nbulaj:
Hey @sato11 . Unfortunately the third option is the option I wanna to use some time ago as currently Danger makes more issues rather than helps with code review. I think we have to wait when some stable solution will be ready and not to use something custom. Or even move to another code review tooling. So I;m for disabling Danger for now.

sato11:
Hi @nbulaj 👋 Thanks a ton for consideration!

I'm closing this as this has become irrelevant to the option we're proceeding with, and will submit the next one 👍

PR #1622, opened 2023-01-11
Drop support for Rubies 2.5 and 2.6

sato11:
### Summary
After #1618 I recognized some remaining stuff:

- Ruby 2.5 has been untested since #1502 (it says 2.4 but 2.5 was excluded from travis at this point)
- Ruby 2.6 is going to be untested from now on

So I propose that we exclude these versions. How does that sound?


nbulaj (approved):
LGTM :+1:

PR #1623, opened 2023-01-11
Add RuboCop's GHA workflow

ydah:
### Summary

This PR is that proposes to add RuboCop's GHA workflow.

However, RuboCop pointed out that there seemed to be a lot of offenses in the latest main branch.

```
244 files inspected, 342 offenses detected, 126 offenses autocorrectable
```

Therefore, additional steps were taken as follows

- Run `rubocop --auto-gen-config` again
- Modify cop that can be safely autocorrected.
- Disable cop that is considered unnecessary
- Add GHA
- Fix RuboCop failing to run on GHA


nbulaj:
Hey @ydah , thanks for the PR! We currently have Hound configured to check & add PR comments using rubocop, not sure we'll benefit from migrating to GA :thinking:

ydah:
@nbulaj 

>We currently have Hound configured to check & add PR comments using rubocop

Aah! I missed that one😅.
You're right, that might not be a rational reason to make it a GHA. Thanks for letting me know!
But I'm a little concerned that there were a few RuboCop offenses left 🤔.

ydah:
Would it be better to remove the GHA commit and change the resolution of the RuboCop violation to be of interest for this pull request?

nbulaj:
Hey :wave: Could you please explain this point? 
> and change the resolution of the RuboCop violation to be of interest for this pull request

I'm not sure I've get your idea correctly :smiley_cat: 

We have Hound which uses[ some specific version of the Rubocop ](https://github.com/doorkeeper-gem/doorkeeper/blob/main/.hound.yml). And we have our own Rububocop rules which we're applying to Doorkeeper codebase

PR #1624, opened 2023-01-12
Drop dependency to danger for now to fix ci

sato11:
### Summary
Danger workflow has been failing since last month. Investigation took place in https://github.com/doorkeeper-gem/doorkeeper/pull/1620#issuecomment-1378171407 where the consensus was reached to drop it until a better solution will appear.

This I hope will render CI stable again.

/cc @nbulaj


nbulaj (approved):
One day we will return it, I promise!

Issue #1617, opened 2023-01-06
`rails generate doorkeeper:install` fails with `Doorkeeper::MissingConfiguration` error


leemeichin:
### Steps to reproduce

1. Add doorkeeper to a Rails app
2. Run `rails generate doorkeeper:install` as suggested by docs

### Expected behavior

It creates the doorkeeper initializer, updates routes, adds i18n, etc.

### Actual behavior

```
bundle exec rails generate doorkeeper:install

[..]/doorkeeper-5.6.2/lib/doorkeeper/config.rb:33:in `configuration': Configuration for doorkeeper missing. Do you have doorkeeper initializer? (Doorkeeper::MissingConfiguration)
```

No files are generated.

### System configuration

**Rails version**: 7.0.4

**Ruby version**: `3.2.0`

**Gemfile.lock**:

<details>
  <summary>Gemfile.lock content</summary>
  
```
...
    doorkeeper (5.6.2)
      railties (>= 5)
...
```
</details>

leemeichin:
never mind, looks like I've duplicated https://github.com/doorkeeper-gem/doorkeeper/issues/1577#issue-1284626166 :)

nbulaj:
Hey @leemeichin 

could you please try with 

```ruby
# Gemfile

gem "doorkeeper", github: "doorkeeper-gem/doorkeeper", branch: "fixes/lazy-load-associations"
```

and report any issues you found? (and backtrace please if present, I have to see which Doorkeeper lined caused an error)

PR #1627, opened 2023-01-12
Lazy evaluate Doorkeeper config when loading files and executing initializers

nbulaj:
Root of the issue is the fact that Rails or some specific gems (like Sorbet) could load model files **before** initializers have been run. Since some models have constructions like:

```ruby
class Some
  if Doorkeper.config.blabla
    # do smth
  end
end
```

in their class body - it will trigger an error because `@config` still doesn't exist (`Doorkeeper.configure` block wasn't called).

The same is applied to Doorkeeper railtie (engine): it adds some custom parameters filters based on the OAuth configuration, but it does it even for the new Rails apps which have not yet generated an initializer (just added a gem and run `rails g doorkeeper:install`).

What should be done:

* Avoid checking config values in railtie initializers (config could be nil for Rails apps which just installed the gem. see above issues)
* Avoid checking config in class definitions (move adding Polymorphic resource owner into an ORM hook)
* 
---

Addresses:

- https://github.com/doorkeeper-gem/doorkeeper/pull/1615
- https://github.com/doorkeeper-gem/doorkeeper/issues/1617
- https://github.com/doorkeeper-gem/doorkeeper/issues/1577


lib/doorkeeper/engine.rb
```diff
@@ -2,10 +2,12 @@
 
 module Doorkeeper
   class Engine < Rails::Engine
-    initializer "doorkeeper.params.filter" do |app|
-      parameters = %w[client_secret authentication_token access_token refresh_token]
-      parameters << "code" if Doorkeeper.config.grant_flows.include?("authorization_code")
-      app.config.filter_parameters << /^(#{Regexp.union(parameters)})$/
+    initializer "doorkeeper.params.filter", after: :load_config_initializers  do |app|
```
johnnyshields, line 5: double whitespace before `do` here

lib/doorkeeper/config.rb
```diff
@@ -33,6 +35,10 @@ def configuration
       @config || (raise MissingConfiguration)
```
johnnyshields, line 35: This line still raises an error.

zavan:
This broke my initializer in development (I haven't tested in production):

```
Jan 30 12:08:52 ubuntu-jammy rails[186539]: /usr/local/lib/ruby/gems/3.2.0/gems/activesupport-7.0.4.2/lib/active_support/inflector/methods.rb:280:in `constantize': uninitialized constant DoorkeeperAccessToken (NameError)
Jan 30 12:08:52 ubuntu-jammy rails[186539]:       Object.const_get(camel_cased_word)
Jan 30 12:08:52 ubuntu-jammy rails[186539]:             ^^^^^^^^^^
Jan 30 12:08:52 ubuntu-jammy rails[186539]:         from /usr/local/lib/ruby/gems/3.2.0/gems/activesupport-7.0.4.2/lib/active_support/core_ext/string/inflections.rb:74:in `constantize'
Jan 30 12:08:52 ubuntu-jammy rails[186539]:         from /usr/local/lib/ruby/gems/3.2.0/gems/doorkeeper-5.6.3/lib/doorkeeper/config.rb:425:in `access_token_model'
```

I have in my initializer/config:
```ruby
Doorkeeper.configure do
  # ...
  access_token_class 'DoorkeeperAccessToken'
  # ...
end
```

Rails v7.0.4.2

nbulaj:
Yeah, can I see your `DoorkeeperAccessToken` implementation? Where it is located? How it looks like @zavan ?

zavan:
> Yeah, can I see your `DoorkeeperAccessToken` implementation? Where it is located? How it looks like @zavan ?

It's a regular rails model in `app/models/doorkeeper_access_token.rb`:

```ruby
class DoorkeeperAccessToken < ApplicationRecord
  include ::Doorkeeper::Orm::ActiveRecord::Mixins::AccessToken

  after_save :sync_legacy_token

  private

  def sync_legacy_token
    #
  end
end
```

Downgrading to v5.6.2 works.

nbulaj:
@zavan could you please try with `fix-custom-models` branch ?

zavan:
> @zavan could you please try with `fix-custom-models` branch ?

It works 😄

PR #1639, opened 2023-02-22
Add grant type vlaidation to avoid Internal Server Error for DELETE /oauth/authorize endpoint

### LINKED ISSUE #1638: Internal Server Error for DELETE /oauth/authorize endpoint with invalid response type
### Steps to reproduce

Call `DELETE /oauth/authorize` endpoint with `response_type=test` or something other response type which is not supported by Doorkeeper.

I created minimal reproductive repository with API mode enabled: https://github.com/kyori19/rails-doorkeeper-repro
Detailed steps to reproduce is written there.

### Expected behavior

Returns 400 Bad Request with error code `unsupported_response_type` (as `GET /oauth/authorize` does).

### Actual behavior

Returns 500 Internal Server Error.
Ruby stacktrace is written on the repository above.

### System configuration

See the repository above.

nbulaj:
Fixes #1638

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ User-visible changes worth mentioning.
 - [#ID] Add your PR description here.
 - [#1602] Allow custom data to be stored inside access grants/tokens.
 - [#1634] Code refactoring for custom token attributes.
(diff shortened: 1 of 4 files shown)


kyori19:
thanks for fixing! 👍

PR #1644, opened 2023-03-13
Update HTTP headers

marjanovic93:
Pragma is a deprecated HTTP header. It changes from Cache-Control: no-store to Cache-Control: no-store, no-cache to Doorkeeper.

### Summary

Those headers are set by Doorkeeper. We have:

```
Pragma: no-cache
Cache-Control: no-store
```

which are set in https://github.com/doorkeeper-gem/doorkeeper/search?q=no-cache

Pragma is a deprecated HTTP header and Atlassian's suggestion is what is favored now so we should contribute the simple change from `Cache-Control: no-store` to `Cache-Control: no-store, no-cache` to Doorkeeper and pick that up in a gem update.

### Other Information

If there's anything else that's important and relevant to your pull
request, mention that information here. This could include
benchmarks, or other information.

If you are updating CHANGELOG.md file or are asked to update it by reviewers,
please add the changelog entry at the top of the file.

Thanks for contributing to Doorkeeper project!

### DIFF
--- a/spec/requests/endpoints/token_spec.rb
+++ b/spec/requests/endpoints/token_spec.rb
@@ -15,8 +15,7 @@ RSpec.describe "Token endpoint" do
   it "respond with correct headers" do
     post token_endpoint_url(code: @authorization.token, client: @client)
 
-    expect(headers["Pragma"]).to eq("no-cache")
-    expect(headers["Cache-Control"]).to be_in(["no-store", "private, no-store"])
+    expect(headers["Cache-Control"]).to be_in(["no-store", "no-cache, no-store", "private, no-store"])
     expect(headers["Content-Type"]).to eq("application/json; charset=utf-8")
   end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ User-visible changes worth mentioning.
 ## main
 
 - [#ID] Add your PR description here.
+- [#1644] Update HTTP headers.
 
 # 5.6.5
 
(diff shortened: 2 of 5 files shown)


nbulaj (approved):
LGTM :+1: 

For the record: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Pragma

nbulaj:
@marjanovic93 could you please add a changelog entry?

marjanovic93:
@nbulaj sure, updated! Back to you 🏓

nbulaj:
I see some tests are failling, could you please check? And also squash all the commints into a single one later :pray: Thanks!
https://github.com/doorkeeper-gem/doorkeeper/actions/runs/4414282695/jobs/7770367021

marjanovic93:
Sure @nbulaj 🚀 

It should be fixed, your turn! 🏓

nbulaj (approved):
All is good now, thanks :bow:

spec/requests/endpoints/token_spec.rb
```diff
@@ -15,8 +15,7 @@
   it "respond with correct headers" do
     post token_endpoint_url(code: @authorization.token, client: @client)
 
-    expect(headers["Pragma"]).to eq("no-cache")
-    expect(headers["Cache-Control"]).to be_in(["no-store", "private, no-store"])
+    expect(headers["Cache-Control"]).to be_in(["no-store", "no-cache, no-store", "private, no-store"])
```
zubin, line 18: Bit late here but, given that `Pragma: no-cache` is always present here, would the equivalent not be:

```ruby
    expect(headers["Cache-Control"]).to be_in(["no-store, no-cache", "private, no-store, no-cache"])
```

Issue #1589, opened 2022-10-08
Non-confidential apps should always re-prompt for user consent


nbudin:
(Mentioning @rgammans, since he told me about this issue and wanted to be notified about replies.)

### Steps to reproduce

1. Log into the server application (in whatever way is necessary to pass its `resource_owner_authenticator` check)
2. Create a non-confidential application in Doorkeeper admin
3. Initiate authentication against it using a flow that only includes client_id, such as the `authorization_code` flow with PKCE

### Expected behavior
According to [RFC 8252 section 8.6](https://www.rfc-editor.org/rfc/rfc8252#section-8.6), the authentication server should re-prompt for user consent, since the client's identity cannot be assured simply from the client_id parameter.  (In Doorkeeper, I think this means that it should redirect to `AuthorizationsController#new`.)

### Actual behavior
The authentication server immediately redirects with a code parameter.

### System configuration
You can help us to understand your problem if you will share some very
useful information about your project environment (don't forget to
remove any confidential data if it exists).

**Doorkeeper initializer**:

```ruby
# config/initializers/doorkeeper.rb
class NullResourceOwner
  def id
    nil
  end
end

Doorkeeper.configure do
  Devise::Doorkeeper.configure_doorkeeper(self)

  # Change the ORM that doorkeeper will use (needs plugins)
  orm :active_record

  # This block will be called to check whether the resource owner is authenticated or not.
  resource_owner_authenticator do
    user_signed_in? ? current_user : NullResourceOwner.new
  end

  # If you didn't skip applications controller from Doorkeeper routes in your application routes.rb
  # file then you need to declare this block in order to restrict access to the web interface for
  # adding oauth authorized applications. In other case it will return 403 Forbidden response
  # every time somebody will try to access the admin web interface.
  #
  admin_authenticator { authorize Doorkeeper::Application.new, :manage? }

  # If you are planning to use Doorkeeper in Rails 5 API-only application, then you might
  # want to use API mode that will skip all the views management and change the way how
  # Doorkeeper responds to a requests.
  #
  # api_only

  # Enforce token request content type to application/x-www-form-urlencoded.
  # It is not enabled by default to not break prior versions of the gem.
  #
  # enforce_content_type

  # Authorization Code expiration time (default 10 minutes).
  #
  # authorization_code_expires_in 10.minutes

  # Access token expiration time (default 2 hours).
  # If you want to disable expiration, set this to nil.
  #
  # access_token_expires_in 2.hours

[...]

PR #1646, opened 2023-03-17
Block public clients automatic authorization skip

adam-h:
Non-confidential applications should not be able to skip the authorization stop, even if they have an existing matching_token.

From the [issue](https://github.com/doorkeeper-gem/doorkeeper/issues/1589):
> According to RFC 8252 section 8.6, the authentication server should re-prompt for user consent, since the client's identity cannot be assured simply from the client_id parameter

Fixes https://github.com/doorkeeper-gem/doorkeeper/issues/1589


nbulaj (approved):
LGTM :+1: 
Thanks!

PR #1652, opened 2023-04-08
Add custom attributes support to token generator

raphaelcosta:
### Summary

Adds support to the custom attributes added by the PR https://github.com/doorkeeper-gem/doorkeeper/pull/1602 in the custom access_token_generator. It's now possible to use the custom attributes in the JWT payload for example.

### Other Information

Related PR https://github.com/doorkeeper-gem/doorkeeper/pull/1602

### DIFF
--- a/lib/doorkeeper/models/access_token_mixin.rb
+++ b/lib/doorkeeper/models/access_token_mixin.rb
@@ -435,6 +435,10 @@ module Doorkeeper
         if Doorkeeper.config.polymorphic_resource_owner?
           attributes[:resource_owner] = resource_owner
         end
+
+        Doorkeeper.config.custom_access_token_attributes.each do |attribute_name|
+          attributes[attribute_name] = public_send(attribute_name)
+        end
       end
     end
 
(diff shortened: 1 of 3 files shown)


raphaelcosta:
@JeremyC-za I would like to have your opinion here since it's based on your PR. Thanks!

lib/doorkeeper/models/access_token_mixin.rb
```diff
@@ -435,6 +435,10 @@ def attributes_for_token_generator
         if Doorkeeper.config.polymorphic_resource_owner?
           attributes[:resource_owner] = resource_owner
         end
+
+        Doorkeeper.config.custom_access_token_attributes.each do |attribute_name|
+          attributes[attribute_name] = send(attribute_name)
```
nbulaj, line 440: can we at least use `public_send` as bot suggested?

raphaelcosta: Yes, done!

nbulaj (approved):
LGTM :+1:

raphaelcosta:
> Could you please also add a changelog entry? And squash the commits then 🙇 Thanks in advance

Done, thanks!

nbulaj:
Thanks @raphaelcosta !

PR #1660, opened 2023-05-19
Stop auto-creation of AccessGrant when a matching token is found if custom access token attributes are used

JeremyC-za:
### Summary

There are two changes in this PR, discussed in more detail below:
1. Prevent new access grants being created automatically when a matching token is found if `custom_access_token_attributes` are used.
2. Added another config option `revoke_previous_authorization_code_token` to mirror `revoke_previous_client_credentials_token`.

#### Change 1:
What happens currently is that when you authorize an application, and you already have a valid access token (not expired or revoked) for that application, then a new access grant is created immediately. This happens in `AuthorizationsController#render_success`, which checks for `matching_token?` and then calls `authorize_response`, which creates the new grant. 

This is a problem if `custom_access_token_attributes` are used, since those attributes from the existing token aren't copied into the new grant. In my case, I need to display the custom attributes (tenant ids) from the existing token so I can allow the user to edit them if necessary before the new access grant is created. I did this by loading the custom attributes from the existing token into the `@pre_auth` object, which is the accessible in the view, and can be used as needed.

I also needed the existing token to be revoked once a token is generated from the new access grant, which leads to the next change.

#### Change 2:
We already have a config option `revoke_previous_client_credentials_token`. That's only available for the `client_credentials` grant type though. I need the same behaviour for the `authorization_code` grant type, so I've added another config option for it. 

There is a TODO to make `revoke_previous_client_credentials_token` more generic for the other flows, but I don't really understand the other ones and I just need this grant type, so I opted add a new config option rather than refactor the existing one.

### DIFF
--- a/app/controllers/doorkeeper/authorizations_controller.rb
+++ b/app/controllers/doorkeeper/authorizations_controller.rb
@@ -31,7 +31,7 @@ module Doorkeeper
     private
 
     def render_success
-      if skip_authorization? || (matching_token? && pre_auth.client.application.confidential?)
+      if skip_authorization? || can_authorize_response?
         redirect_or_render(authorize_response)
       elsif Doorkeeper.configuration.api_only
         render json: pre_auth
@@ -52,10 +52,16 @@ module Doorkeeper
       end
     end
 
+    def can_authorize_response?
+      Doorkeeper.config.custom_access_token_attributes.empty? && pre_auth.client.application.confidential? && matching_token?
+    end
+
     # Active access token issued for the same client and resource owner with
     # the same set of the scopes exists?
     def matching_token?
-      Doorkeeper.config.access_token_model.matching_token_for(
+      # We don't match tokens on the custom attributes here - we're in the pre-auth here,
+      # so they haven't been supplied yet (there are no custom attributes to match on yet)
+      @matching_token ||= Doorkeeper.config.access_token_model.matching_token_for(
         pre_auth.client,
         current_resource_owner,
         pre_auth.scopes,
--- a/lib/doorkeeper/models/access_token_mixin.rb
+++ b/lib/doorkeeper/models/access_token_mixin.rb
@@ -83,14 +83,18 @@ module Doorkeeper
       #   Resource Owner model instance or it's ID
       # @param scopes [String, Doorkeeper::OAuth::Scopes]
       #   set of scopes
+      # @param custom_attributes [Nilable Hash]
+      #   A nil value, or hash with keys corresponding to the custom attributes
+      #   configured with the `custom_access_token_attributes` config option.
+      #   A nil value will ignore custom attributes.
       #
       # @return [Doorkeeper::AccessToken, nil] Access Token instance or
       #   nil if matching record was not found
       #
-      def matching_token_for(application, resource_owner, scopes, include_expired: true)
+      def matching_token_for(application, resource_owner, scopes, custom_attributes: nil, include_expired: true)
         tokens = authorized_tokens_for(application&.id, resource_owner)
         tokens = tokens.not_expired unless include_expired
-        find_matching_token(tokens, application, scopes)
+        find_matching_token(tokens, application, custom_attributes, scopes)
       end
 
       # Interface to enumerate access token records in batches in order not
@@ -114,11 +118,15 @@ module Doorkeeper
       #   Application instance
       # @param scopes [String, Doorkeeper::OAuth::Scopes]
       #   set of scopes
+      # @param custom_attributes [Nilable Hash]
+      #   A nil value, or hash with keys corresponding to the custom attributes
+      #   configured with the `custom_access_token_attributes` config option.
+      #   A nil value will ignore custom attributes.
       #
       # @return [Doorkeeper::AccessToken, nil] Access Token instance or
       #   nil if matching record was not found
       #
-      def find_matching_token(relation, application, scopes)
+      def find_matching_token(relation, application, custom_attributes, scopes)
         return nil unless relation
 
         matching_tokens = []
@@ -126,7 +134,8 @@ module Doorkeeper
 
         find_access_token_in_batches(relation, batch_size: batch_size) do |batch|
           tokens = batch.select do |token|
-            scopes_match?(token.scopes, scopes, application&.scopes)
+            scopes_match?(token.scopes, scopes, application&.scopes) &&
+              custom_attributes_match?(token, custom_attributes)
           end
 
           matching_tokens.concat(tokens)
@@ -160,6 +169,31 @@ module Doorkeeper
           )
       end
 
+      # Checks whether the token custom attribute values match the custom
+      # attributes from the parameters.
+      #
+      # @param token [Doorkeeper::AccessToken]
+      #   The access token whose custom attributes are being compared
+      #   to the custom_attributes.
+      #
+      # @param custom_attributes [Hash]
+      #   A hash of the attributes for which we want to determine whether
+      #   the token's custom attributes match.
+      #
+      # @return [Boolean] true if the token's custom attribute values
+      #   match those in the custom_attributes, or if both are empty/blank.
+      #   False otherwise.
+      def custom_attributes_match?(token, custom_attributes)
+        return true if custom_attributes.nil?
+
+        token_attribs = token.custom_attributes
+        return true if token_attribs.blank? && custom_attributes.blank?
+
+        Doorkeeper.config.custom_access_token_attributes.all? do |attribute|
+          token_attribs[attribute] == custom_attributes[attribute]
+        end
+      end
+
       # Looking for not expired AccessToken record with a matching set of
       # scopes that belongs to specific Application and Resource Owner.
       # If it doesn't exists - then creates it.
@@ -181,7 +215,9 @@ module Doorkeeper
       #
       def find_or_create_for(application:, resource_owner:, scopes:, **token_attributes)
         if Doorkeeper.config.reuse_access_token
-          access_token = matching_token_for(application, resource_owner, scopes, include_expired: false)
+          custom_attributes = extract_custom_attributes(token_attributes).presence
+          access_token = matching_token_for(
+            application, resource_owner, scopes, custom_attributes: custom_attributes, include_expired: false)
 
           return access_token if access_token&.reusable?
         end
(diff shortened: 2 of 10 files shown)


JeremyC-za:
@nbulaj if you could review when you have time please 🙏

app/controllers/doorkeeper/authorizations_controller.rb
```diff
@@ -31,8 +31,14 @@ def destroy
     private
 
     def render_success
-      if skip_authorization? || (matching_token? && pre_auth.client.application.confidential?)
-        redirect_or_render(authorize_response)
+      if skip_authorization? || ((token = matching_token?) && pre_auth.client.application.confidential?)
```
nbulaj, line 34: I think we have to do some refactoring with method extractions here, now it starts to look complex. WDYT?

Maybe unify the condition and move attributes setup into some method. It will also allow to override them if required in any application (rather than overriding `render_success`)

JeremyC-za:
@nbulaj I've done some refactoring. Let me know what you think.

app/controllers/doorkeeper/authorizations_controller.rb
```diff
@@ -49,10 +50,21 @@ def render_error
       end
     end
 
+    def can_authorize_response?
+      matching_token? && pre_auth.client.application.confidential? && Doorkeeper.config.custom_access_token_attributes.empty?
+    end
```
PhilippeChab, line 55: Could we change the order of this check? 

```suggestion
    def can_authorize_response?
      Doorkeeper.config.custom_access_token_attributes.empty? && pre_auth.client.application.confidential? && matching_token?
    end
```

Like this we don't have to do the whole `matching_token?` processing if one of the two boolean checks fails.

app/controllers/doorkeeper/authorizations_controller.rb
```diff
@@ -49,10 +50,21 @@ def render_error
...
+
+    def prepare_custom_attributes
+      return if Doorkeeper.config.custom_access_token_attributes.empty?
+      return if (token = matching_token?).blank?
+      custom_attributes = token.attributes.symbolize_keys.slice(*Doorkeeper.config.custom_access_token_attributes)
+      pre_auth.custom_access_token_attributes = custom_attributes
+    end
```
PhilippeChab, line 62: Is there a reason we are not setting this in the `OAuth::PreAuthorization` directly? It seems like logic that should not be in the controller.

app/controllers/doorkeeper/authorizations_controller.rb
```diff
@@ -49,10 +50,21 @@ def render_error
...
+
     # Active access token issued for the same client and resource owner with
     # the same set of the scopes exists?
     def matching_token?
-      Doorkeeper.config.access_token_model.matching_token_for(
+      @matching_token ||= Doorkeeper.config.access_token_model.matching_token_for(
```
PhilippeChab, line 67: As said [here](https://github.com/doorkeeper-gem/doorkeeper/issues/1665), `matching_token_for` should consider custom attributes if they are provided.

JeremyC-za: I've set it up to ignore matching on the custom attributes here - like my comment in the code says, since we're in the pre-auth here the custom attributes haven't been provided yet, so there's nothing to match on. This could find a matching token then, but because `#can_authorize_response?` always returns false if custom attributes are used, it will just copy that tokens custom attributes into the pre-auth, instead of copying them and generating a new grant/token.

JeremyC-za (commented):
@PhilippeChab I've pushed some changes which I think should handle the scenario you mentioned. Let me know what you think.

cc @nbulaj

app/controllers/doorkeeper/authorizations_controller.rb
PhilippeChab: I suppose someone could send `/oauth/authorize?tenant_id=TENANT_ID&...`, but I'm not sure it's part of the standard.

I don't mind not handling it.

lib/doorkeeper/models/access_token_mixin.rb
```diff
@@ -160,6 +169,30 @@ def scopes_match?(token_scopes, param_scopes, app_scopes)
...
+      #   to the param_attributes.
+      #
+      # @param param_attributes [Hash]
+      #   A hash of the attributes for which we want to determine whether
+      #   the token's custom attributes match.
+      #
+      # @return [Boolean] true if the token's custom attribute values
+      #   match those in the param_attributes, or if both are empty/blank.
+      #   False otherwise.
+      def custom_attributes_match?(token, param_attributes)
```
PhilippeChab, line 186: Can we rename `param_attributes` to `custom_attributes`? These do not necessarily come from the request, but from the authorization code record as well, and it's a bit confusing.

lib/doorkeeper/models/access_token_mixin.rb
```diff
@@ -160,6 +169,30 @@ def scopes_match?(token_scopes, param_scopes, app_scopes)
...
+        return true if param_attributes.nil?
```
PhilippeChab, line 187: This is fine for now, but it will be an issue if we handle `/oauth/authorize?tenant_id=TENANT_ID&...`.

It would be a good idea to leave a comment I believe.

nbulaj:
Hey @JeremyC-za . What is the state of this PR? Do we need to do more work on it?

JeremyC-za:
Hey @nbulaj, it's almost ready. There are a couple things I want to look into (which may not end up needing code changes), I just haven't been able to find the time lately. I intend on getting to it by the end of next week though.

JeremyC-za (commented):
@PhilippeChab I've responded to your comments, let me know what you think. 

@nbulaj In my opinion this is ready to be reviewed and hopefully merged. Thanks.

app/controllers/doorkeeper/authorizations_controller.rb
```diff
@@ -31,11 +31,12 @@ def destroy
     private
 
     def render_success
-      if skip_authorization? || (matching_token? && pre_auth.client.application.confidential?)
+      if skip_authorization? || can_authorize_response?
         redirect_or_render(authorize_response)
       elsif Doorkeeper.configuration.api_only
         render json: pre_auth
       else
+        pre_auth.load_custom_attributes_from_token(matching_token?) if matching_token?
```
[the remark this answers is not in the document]
PhilippeChab: I've given some thoughts to this and I think it should not be part of the gem's default behaviour. I don't believe this kind of thing is part of the oauth standard and I feel like doorkeeper users do not expect this kind of behaviour by default.

I think this is something you should handle on your end by overriding the controller.

lib/doorkeeper/oauth/pre_authorization.rb
```diff
@@ -65,6 +67,13 @@ def form_post_response?
         response_mode == "form_post"
       end
 
+      def load_custom_attributes_from_token(matching_token)
```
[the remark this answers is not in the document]
PhilippeChab: I think the link is not pointing to the correct reference, but as I said in the other thread, I think this is something that you should handle on your end and not in doorkeeper directly.

ashkulz:
@nbulaj would it be possible for you to review this?

app/controllers/doorkeeper/authorizations_controller.rb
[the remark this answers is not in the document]
nbulaj: I agree with @PhilippeChab , this looks more like customization which should be done on developers end for any purposes they need. Can we make it customizable or at least configurable?

lib/doorkeeper/models/access_token_mixin.rb
```diff
@@ -114,19 +118,24 @@ def find_access_token_in_batches(relation, **args, &block)
...
+      # @param custom_attributes [Nilable Hash]
+      #   A nil value, or hash with keys corresponding to the custom attributes
+      #   configured with the `custom_access_token_attributes` config option.
+      #   A nil value will ignore custom attributes.
       #
       # @return [Doorkeeper::AccessToken, nil] Access Token instance or
       #   nil if matching record was not found
       #
-      def find_matching_token(relation, application, scopes)
+      def find_matching_token(relation, application, custom_attributes, scopes)
```
nbulaj, line 129: Do we need to pass `custom_attributes` here since we have it defined on a model level? Sorry it's hard to me to understand if we're passing some new attributes or the same as defined in the model (so can be find there already)

PhilippeChab: I believe it's the custom attributes from the access grant retrieved [here](https://github.com/doorkeeper-gem/doorkeeper/blob/0aa94c5a82035ec4840785156760a2930e5de27a/lib/doorkeeper/oauth/authorization_code_request.rb#L105).

nbulaj: Ah, OK, makes sense

app/controllers/doorkeeper/authorizations_controller.rb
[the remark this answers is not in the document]
PhilippeChab: @JeremyC-za could you make the last changes so that we can go forward with the PR. It has been stale for quite some time now.

Thank you!

JeremyC-za:
@nbulaj @PhilippeChab Apologies for the radio silence. I've pushed a change so the custom attributes are no longer copied into the access grant automatically when a matching token is found (as per [this conversation](https://github.com/doorkeeper-gem/doorkeeper/pull/1660#discussion_r1318822094)). I think that was all that was todo, but let me know if I've missed something.

A recap of what this PR includes at this point in time:
- `AccessTokenMixin#find_matching_token` now considers custom attributes, which should resolve #1665.
- When using `custom_access_token_attributes`, an AccessGrant and code will no longer be automatically created when a matching token is found (due to the changes added in `AuthorizationsController#render_success`).
- Introduced `revoke_previous_client_credentials_token` config option (unchanged from **Change 2** in PR description).

nbulaj (approved):
LGTM :+1: 

We can ignore truffle-ruby-head for Rails 7.1 I think. Have to make it experimental BTW

nbulaj:
@JeremyC-za can I ask you please to squash the commits and add a changelog entry? :pray:  Thanks!

JeremyC-za:
@nbulaj done 👍🏻

nbulaj:
Thanks a lot @JeremyC-za @PhilippeChab

PR #1668, opened 2023-08-31
Provide an option to skip orm hooks

cwjenkins:
### Summary

Currently trying to dockerize an application that contains doorkeeper and upon running assets precompilation it wants ~~a database connection~~ to evaluate database.yml config. 
This provides a way to skip the orm hooks.

### DIFF
--- a/lib/doorkeeper.rb
+++ b/lib/doorkeeper.rb
@@ -158,6 +158,8 @@ module Doorkeeper
     end
 
     def run_orm_hooks
+      return if config.skip_orm_hooks
+
       config.clear_cache!
 
       if @orm_adapter.respond_to?(:run_hooks)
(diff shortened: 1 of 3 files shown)


nbulaj:
Hey @cwjenkins , thanks for your contribution! It looks a little bit odd for me to introduce such a config option for such case.. What log do you see while compiling the assets in docker ?

nbulaj:
Will be great to understand at what step it tried to connect to the DB and why it requires that step @cwjenkins . In such case we can think how to make it lazy evaluatable (btw it should be already :thinking: )

nbulaj:
Yeah, but I believe it shouldn't be a config option. We have to change autoloading to be fully lazy-evaluatable so it won't affect anything in Docker or in any other env

nbulaj:
Hello there. We merged one more lazy AR initialization change in #1804 and most likely it can solve existing issue here as well. Is it possible now to validate a `main` branch maybe? :eyes: I know it's been a long time, but still... Thanks!

nbulaj:
We did more work on lazy load. So I'm gonna close this one, feel free to report if the issue is still there. Thanks!

PR #1673, opened 2023-10-06
Honor custom_access_token_attributes in client credentials grant flow

mlr:
### Summary

This change supports the `custom_access_token_attributes` configuration for Client Credentials grant_flow type.

Without this change, the custom configured attributes are not passed along during the token creation for client credentials requests, which leads to an error similar to:

> Mysql2::Error - Field 'tenant_id' doesn't have a default value

If a default of null is provided for the field instead, the result will be the column is not populated with the tenant_id provided during the client credentials request. 

### Other Information

Steps to reproduce:

1. Install doorkeeper following standard instructions for Rails application
2. Configure the initializer with:
    `grant_flows %w[client_credentials]` and
    `custom_access_token_attributes [:tenant_id]`
4. Modify the `oauth_access_tokens` and `oauth_access_grants` tables with the `tenant_id` field as described by the initializer comment for custom_access_token_attributes.
5. Create a Doorkeeper::Application and obtain the client id and client secret.
6. Make a credentials request to obtain a token, i.e.
   ```
   curl -H 'Content-Type: application/json' -XPOST 'http://localhost:3000/oauth/token' \
      -d '{ "client_id": "[client id]", "client_secret": "[client secret]", "grant_type": "client_credentials", "tenant_id": "1" }'
7. Observe the database exception above which happens because the `tenant_id` field is not passed along during token creation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ User-visible changes worth mentioning.
 - [#1662] Specify uri_redirect validation class explicitly.
 - [#1652] Add custom attributes support to token generator.
 - [#1667] Pass `client` instead of `grant.application` to `find_or_create_access_token`.
+- [#1673] Honor `custom_access_token_attributes` in client credentials grant flow.
 
 ## 5.6.6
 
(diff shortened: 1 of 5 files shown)


nbulaj (approved):
LGTM :+1:
 
Thanks!

PR #1677, opened 2023-11-14
Fix URIHelper.valid_for_authorization? breaking for non url URIs

Zeko369:
Adds an additional except clause for validating `.loopback_uri?` since it was failing for non URL uris i.e. "vscode://file/home/user/.vimrc"

Also added tests


nbulaj (approved):
LGTM :+1: 

Can you also please add a changelog entry? Thanks!

nbulaj:
@Zeko369 just pinging if you lost it :) Thanks!

nbulaj:
Will do myself, no worries

PR #1692, opened 2024-02-17
Recommend explicit expires_in in initializer

### LINKED ISSUE #1670: Do not recommend setting access_token_expires_in to nil
The doorkeeper initializer's inline comments mention

```ruby
  # Access token expiration time (default: 2 hours).
  # If you want to disable expiration, set this to `nil`.
  #
  # access_token_expires_in 2.hours
```
https://github.com/doorkeeper-gem/doorkeeper/blob/main/lib/generators/doorkeeper/templates/initializer.rb#L94

But that text is misleading. Yes, if you set it to nil, _Doorkeeper_ will not expire the token. And when a caller requests a token they'll get back something like:

```json
{"access_token":"...","token_type":"Bearer","created_at":1694623393}
```

a response which omits expires_in. But there is no guarantee that callers will interpret the absence of `expires_in` to mean "never expires".

See for example https://github.com/spring-projects/spring-security/issues/8701 in which they discuss workarounds for handling OAuth responses from Gitlab. At the time of the post, Gitlab was using Doorkeeper with nil expires_in. When the caller uses Spring Security's OAuth code to interpret the response, they interpret a missing `expires_in` to mean expire it in 1 second.

One might say Spring is doing it wrong, they should interpret it the same way Doorkeeper intends. But as is mentioned in that thread, the official spec for `expires_in` does not give any particular meaning:

https://www.rfc-editor.org/rfc/rfc6749#section-5.1
> expires_in
> RECOMMENDED. The lifetime in seconds of the access token. For
> example, the value "3600" denotes that the access token will
> expire in one hour from the time the response was generated.
> If omitted, the authorization server SHOULD provide the
> expiration time via other means or document the default value.

The default value when expires_in is omitted is not defined. (Hence the suggestion to provide it via other means or document it).

The doorkeeper generator should warn users against relying on `access_token_expires_in nil`, as it does not have a well established meaning. Prefer `access_token_expires_in 100.years` or similar, which would be functionally equivalent and avoid the risk of unexpected behavior by callers. (You set expires_in nil expecting the caller to reuse the token. But if their code interprets that as "expire in 1 second", they end up calling you for a new token on every request. Not what you expected...)

I'm ignoring the question of whether a token that never expires is a good idea. The trouble is that "nil = never expire" is not part of the spec. Nil simply means one has omitted to specify the lifetime of the access token, leaving it open to differing interpretations as to how to handle it.

p.s. Gitlab no longer uses nil for expires_in:
https://gitlab.com/gitlab-org/gitlab/-/issues/21745

Paultdx:
### Summary

Resolves #1670

Wording is the way i saw it, but if anyone wants to add / change something i will be grateful. 
Also uses some wording by @benissimo as an example.

### DIFF
--- a/lib/generators/doorkeeper/templates/initializer.rb
+++ b/lib/generators/doorkeeper/templates/initializer.rb
@@ -91,7 +91,10 @@ Doorkeeper.configure do
   # authorization_code_expires_in 10.minutes
 
   # Access token expiration time (default: 2 hours).
-  # If you want to disable expiration, set this to `nil`.
+  # If you set this to `nil` Doorkeeper will not expire the token and omit expires_in in response.
+  # It is RECOMMENDED to set expiration time explicitly.
+  # Prefer access_token_expires_in 100.years or similar,
+  # which would be functionally equivalent and avoid the risk of unexpected behavior by callers.
   #
   # access_token_expires_in 2.hours
 


nbulaj (approved):
LGTM

Issue #1678, opened 2023-11-22
Always requiring `redirect_uri` is not compliant to RFC 6749


westnordost:
It looks like doorkeeper always requires the `redirect_uri` parameter to be set in authorization requests. 

https://github.com/doorkeeper-gem/doorkeeper/blob/9fc81d5009aef533ca8116285148cb6e37549ff2/lib/doorkeeper/oauth/pre_authorization.rb#L100-L107

This does not seem to be compliant to [RFC 6749 -  The OAuth 2.0 Authorization Framework](https://datatracker.ietf.org/doc/html/rfc6749):

- **[4.1.1](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.1). Authorization Request** lists `redirect_uri` as OPTIONAL

- **[4.1.3](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3). Access Token Request** lists `redirect_uri` only as REQUIRED if it was supplied in the authorization request

- **[3.1.2.3](https://datatracker.ietf.org/doc/html/rfc6749#section-3.1.2.3). Dynamic Configuration** clarifies that `redirect_uri` is REQUIRED _only_ in case no redirection URI has been specified during client registration:
  > If multiple redirection URIs have been registered, if only part of the redirection URI has been registered, or if no redirection URI has been registered, the client MUST include a redirection URI with the authorization request using the "redirect_uri" request parameter.
  >
  > When a redirection URI is included in an authorization request, the authorization server MUST compare and match the value received against at least one of the registered redirection URIs (or URI components) as defined in [[RFC3986] Section 6](https://datatracker.ietf.org/doc/html/rfc3986#section-6), if any redirection URIs were registered.  If the client registration included the full redirection URI, the authorization server MUST compare the two URIs using simple string comparison as defined in [[RFC3986] Section 6.2.1](https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.1).

I understand that by default, doorkeeper also does not allow leaving the redirect url blank during client registration ([but it can be changed in the configuration](https://github.com/doorkeeper-gem/doorkeeper/wiki/Allow-blank-redirect-URI-for-Applications)).

### Expected behavior

- if client registered a redirection URI, the `redirect_uri` parameter is optional in the authorization request
- the `redirect_uri` parameter is only required in the access token request if it was specified in the authorization request. In this case, it must match

### Actual behavior

- `redirect_uri` parameter is always required in the authorization request and access token request

[...]

PR #1701, opened 2024-03-14
Make redirect_uri optional for Authorization request

### LINKED ISSUE #1682: Doorkeeper::Errors::InvalidRedirectUri Raised When No Redirect URI Set
### Steps to reproduce

Create an application with no `redirect_uri` set, i.e., `redirect_uri=nil`.

Set `allow_blank_redirect_uri = true` in the Doorkeeper config.

Hit `GET /oauth/authorize` with `redirect_uri=''` or `redirect_uri=nil` or omit the `redirect_uri` parameter.

Observe an `Doorkeeper::Errors::InvalidRedirectUri` exception with the message `The requested redirect uri is malformed or doesn't match client redirect URI.`

### Expected behavior

In versions 5.6.6 and before an authorization code was returned.

### Actual behavior

An `Doorkeeper::Errors::InvalidRedirectUri` exception with the message `The requested redirect uri is malformed or doesn't match client redirect URI.`

### System configuration

Set `allow_blank_redirect_uri = true` in the Doorkeeper config.

nbulaj:
Fixes #1682 and #1678

* Grant also should be allowed to have blank redirect_uri? :thinking: 
* Review the full flow
* Add tests

### DIFF
--- a/lib/doorkeeper/oauth/authorization_code_request.rb
+++ b/lib/doorkeeper/oauth/authorization_code_request.rb
@@ -59,7 +59,7 @@ module Doorkeeper
         @missing_param =
           if grant&.uses_pkce? && code_verifier.blank?
             :code_verifier
-          elsif redirect_uri.blank?
+          elsif redirect_uri.blank? && !allow_blank_redirect_uri?
             :redirect_uri
           end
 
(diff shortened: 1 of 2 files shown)


lib/doorkeeper/oauth/authorization_code_request.rb
```diff
@@ -109,6 +117,12 @@ def custom_token_attributes_with_data
           .slice(*Doorkeeper.config.custom_access_token_attributes)
           .symbolize_keys
       end
+
+      def allow_blank_redirect_uri?
```
ThisIsMissEm, line 121: It may be better to phrase this in the way that the spec reads, so: `require_redirect_uri?` and then check `present?` instead of `blank?`

ThisIsMissEm (commented):
This looks good, but we'll need tests to confirm it works, only comment is about making the `allow_blank_redirect_uri?` into the same language as the spec: `require_redirect_uri?`

PR #1702, opened 2024-03-22
Fix bugs for error response in the form_post and error view

lurz:
**Bug**
When the `form_post.html.erb` view attempts to render an error, it 
incorrectly posts an empty form rather than displaying the appropriate 
error and error_description.

**Root cause**
The form_post view used the body of `@authorize_response`. This 
issue arose when the `redirect_or_render` method was invoked with 
either `authorization.deny` or `pre_auth.error_response`, resulting in 
`@authorize_response` containing only an empty body, failing to 
display the error details as intended.

**Fix**
Instead of using `@authorize_response`, we now introduce a local 
variable `auth`, which represents the authorization object passed to 
the `redirect_or_render` function. This ensures that the correct error 
information is provided to the view. Additionally, I have backfilled 
tests in the authorizations controller spec to verify the fix.

**Additional Findings During Testing**
During the test phase, I discovered `error.html.erb` always renders an
incorrect error description. It turns out that 
`respond_to(:error_response)` always return false in the view. I changed 
it to use `local_assigns` as the correct condition.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ User-visible changes worth mentioning.
 
 - [#1696] Add missing `#issued_token` method to `OAuth::TokenResponse`
 - [#1697] Allow a TokenResponse body to be customized.
+- [#1702] Fix bugs for error response in the form_post and error view
 
 ## 5.6.9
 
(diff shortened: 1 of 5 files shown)


nbulaj (approved):
LGTM, thanks! :bow:

Issue #1654, opened 2023-04-15
Option to enforce that clients use PKCE


hickford:
OAuth best practice is to enforce that clients use PKCE.   Draft OAuth 2.1 insists authorization servers enforce the use of PKCE by public clients, and recommends enforcing it for all clients  https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-08.html#name-countermeasures-2

> using code_challenge and code_verifier is REQUIRED for clients, and **authorization servers MUST enforce their use**, unless ... The client is a confidential client. ... In this case, using and enforcing code_challenge and code_verifier is still RECOMMENDED.

Thus it would be useful to have an 'enforce client use of PKCE' option with choices: none, public, all.

nbulaj:
Nice idea @hickford , thanks! Would you mind to create a PR for it?

hickford:
@nbulaj I don't have the expertise with Ruby

westnordost:
Perhaps it would be more straightforward to have *one* option to enforce OAuth 2.1 compliance. (In case there are more requirements for OAuth 2.1 in the final version)

bhuone-garbu:
We definitely need to implement the PKCE flow for non-public clients. I'm pleased to see that the PR has been merged. Can we confirm if there are any plans to release a new patch for Doorkeeper?

nbulaj:
Which patch you're talking about @bhuone-garbu ? #1705 was released with 5.7.0

bhuone-garbu:
I mean #1705 was only merged a month ago but the latest release `5.7.0` was 2 months ago

nbulaj:
LOL yeah, changelog entry was added to the wrong place
https://github.com/doorkeeper-gem/doorkeeper/pull/1705/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR15

bhuone-garbu:
> LOL yeah, changelog entry was added to the wrong place https://github.com/doorkeeper-gem/doorkeeper/pull/1705/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR15
> 
> Or not 🤔

either way, I can confirm the `5.7.0` definitely doesn't contain what the `main` branch has

https://github.com/doorkeeper-gem/doorkeeper/blob/7ffdeec302b747d21e11334d02b2ede0959fc633/lib/doorkeeper/config.rb#L116-L120

[...]

PR #1705, opened 2024-05-01
Add force_pkce option

mattmanning:
The force_pkce option is disabled by default. When enabled, it requires non-confidential clients to use PKCE when requesting an access_token using an authorization code.

### Summary

PKCE should be used for all non-confidential clients (those who can't protect their client_secret). However, since the OAuth 2.0 spec doesn't require it, it should be disabled by default and only used if desired by the server. https://www.rfc-editor.org/rfc/rfc7636#section-5

### Other Information

This implementation is slightly different than what was discussed in https://github.com/doorkeeper-gem/doorkeeper/issues/1654. A more stringent option that also requires PKCE for confidential clients could later be added to a bundled OAuth 2.1 option.

### DIFF
--- a/doorkeeper.gemspec
+++ b/doorkeeper.gemspec
@@ -52,7 +52,6 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency "factory_bot", "~> 6.0"
   gem.add_development_dependency "generator_spec", "~> 0.10.0"
   gem.add_development_dependency "grape"
-  gem.add_development_dependency "pry-byebug"
   gem.add_development_dependency "rake", ">= 11.3.0"
   gem.add_development_dependency "rspec-rails"
   gem.add_development_dependency "timecop"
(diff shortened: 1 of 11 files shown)


anolson (commented):
👍

mattmanning:
@nbulaj Any feedback on this? We've now implemented it as monkeypatches in my company's app, but it would be cool to get it into the gem.

I'm happy to make changes to the design or the implementation if you have any issues with the current code.

Thanks!

doorkeeper.gemspec
```diff
@@ -52,6 +52,7 @@ Gem::Specification.new do |gem|
   gem.add_development_dependency "factory_bot", "~> 6.0"
   gem.add_development_dependency "generator_spec", "~> 0.10.0"
   gem.add_development_dependency "grape"
+  gem.add_development_dependency "pry-byebug"
```
nbulaj, line 55: I don't find pry / byebug useful, can we avoid adding it there please?

nbulaj (commented):
Just a minor comment, otherwise LGTM :+1:

mattmanning:
@nbulaj Thank you for looking!

I have removed pry-byebug. Please let me know if you would like any other changes.

nbulaj (approved):
LGTM :bowing_man:

PR #1708, opened 2024-05-12
Do not instantiate ActiveRecord model early

### LINKED ISSUE #1703: Doorkeeper is loading ActiveRecord too early
### Steps to reproduce
We found a weird behavior as we were upgrading Rails (7.0 => 7.1). When upgrading Rails, it generates a file called `config/initializers/new_framework_defaults.rb` where it sets a bunch of `Rails.application.config.x.y = :something` in preparation for changing Rails defaults to the latest version.

We noticed that some of these configs, especially the ones in `config.active_record.X` did not take hold. Meaning, they did not change the behavior of ActiveRecord. We dug into this further and found out that this happens when gem loads ActiveRecord too early. Here's a Rails issue describing the problem: https://github.com/rails/rails/issues/50133. There's even work (in Rails) towards warning when this happens: https://github.com/rails/rails/pull/46047

For Doorkeeper, specifically, we traced it down to here:
https://github.com/doorkeeper-gem/doorkeeper/blob/1cc7145ba195cd48a703abfb904b11fbdeff0aa1/lib/doorkeeper/config.rb#L440-L442

Calling `access_token_model` will constantize the configured class and essentially loads an autoloaded model that lives inside `app/models/..` within the application. The problem is that this method is invoked very early in the setup phase of the doorkeeper gem:

=> https://github.com/doorkeeper-gem/doorkeeper/blob/main/lib/doorkeeper/orm/active_record.rb#L41
from https://github.com/doorkeeper-gem/doorkeeper/blob/main/lib/doorkeeper.rb#L160-L172
from https://github.com/doorkeeper-gem/doorkeeper/blob/main/lib/doorkeeper/engine.rb#L23-L25

I've created a repo to help illustrate the problem. The README has the information.
https://github.com/ngan/doorkeeper-activerecord-load-issue

tachyons:
### Summary 
Do not instantiate active record models before loading active record, as it can cause several side effects.

Fixes #1703

### DIFF
--- a/lib/doorkeeper/orm/active_record.rb
+++ b/lib/doorkeeper/orm/active_record.rb
@@ -33,12 +33,14 @@ module Doorkeeper
       end
 
       def self.initialize_configured_associations
-        if Doorkeeper.config.enable_application_owner?
-          Doorkeeper.config.application_model.include ::Doorkeeper::Models::Ownership
-        end
+        ActiveSupport.on_load(:active_record) do
+          if Doorkeeper.config.enable_application_owner?
+            Doorkeeper.config.application_model.include ::Doorkeeper::Models::Ownership
+          end
 
-        Doorkeeper.config.access_grant_model.include ::Doorkeeper::Models::PolymorphicResourceOwner::ForAccessGrant
-        Doorkeeper.config.access_token_model.include ::Doorkeeper::Models::PolymorphicResourceOwner::ForAccessToken
+          Doorkeeper.config.access_grant_model.include ::Doorkeeper::Models::PolymorphicResourceOwner::ForAccessGrant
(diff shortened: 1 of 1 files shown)


nbulaj:
I remember we already had such hook and it also lead to a lot of issues, unfortunately. I believe we can find them in the issues section

nbulaj (approved):
Well, let's try to use AS hooks once again :crossed_fingers:

nbulaj:
Can you please add a changelog entry?

tachyons:
I checked the old issues and found that active support hooks caused issues
for STI loading. So I wanted to add tests for that before
proceeding further. But I haven't got time for that yet.

nbulaj:
We merged  #1804 so I believe we can close this one. Thanks!

tachyons:
Thanks @nbulaj

PR #1712, opened 2024-06-28
Add Pragma: no-cache to token response

ransombriggs:
### Summary

According to the [spec](https://datatracker.ietf.org/doc/html/rfc6749#section-5.1) we should be returning a `Pragma` header in the token response.

> The authorization server MUST include the HTTP "Cache-Control"
>   response header field [[RFC2616](https://datatracker.ietf.org/doc/html/rfc2616)] with a value of "no-store" in any
>   response containing tokens, credentials, or other sensitive
>   information, as well as the "Pragma" response header field [[RFC2616](https://datatracker.ietf.org/doc/html/rfc2616)]
>   with a value of "no-cache".

When I was doing a review of our implementation I found that the `Pragma` header is missing so adding it to be compliant.

### DIFF
--- a/lib/doorkeeper/oauth/token_response.rb
+++ b/lib/doorkeeper/oauth/token_response.rb
@@ -30,6 +30,7 @@ module Doorkeeper
         {
           "Cache-Control" => "no-store, no-cache",
           "Content-Type" => "application/json; charset=utf-8",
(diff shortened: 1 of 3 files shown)


nbulaj (approved):
LGTM, thanks!

PR #1715, opened 2024-07-11
Fix token introspection invalid request reason

ransombriggs:
### Summary

I was validating the client authentication in the token introspection endpoint and got a "The request is missing a required parameter, includes an unsupported parameter value, or is otherwise malformed." error which was kind of confusing and so I took a look at the code and this appears to be not intentional.  The `invalid_request_reason` is a private method so when the `try(:invalid_request_reason)` runs it returns `nil` and so it falls back to the default.  I went with a quick fix but making the `attr_reader` public so that the error_description is set properly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ User-visible changes worth mentioning.
 
 Add your entry here.
 
+- [#1715] Fix token introspection invalid request reason
 - [#1714] Fix `Doorkeeper::AccessToken.find_or_create_for` with empty scopes which raises NoMethodError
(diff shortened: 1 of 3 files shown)


nbulaj (approved):
Makes sense, thanks! LGTM

Can you please also add a changelog entry and squash the commits? :pray:  Thanks!

ransombriggs:
I rebased from `main`, added to the changelog and force-pushed [d86fc47](https://github.com/doorkeeper-gem/doorkeeper/pull/1715/commits/d86fc47abac1219b70b271be603708f4cc3920fd)

PR #1723, opened 2024-07-27
Add Ruby 3.3 to GitHub Actions testing matrix

ThisIsMissEm:
### Summary

Ruby 3.3.4 is the current latest stable version of Ruby, however, we are currently not testing it in GitHub Actions (we were only going up to ruby 3.2)

This adds support for 3.3 to the testing matrix.

I do note that according to the Gemspec, we support `>= 2.7`, so we should probably also include Ruby 2.7 in the testing matrix at least.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,11 +19,12 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        os: [ ubuntu-latest ]
+        os: [ubuntu-latest]
         ruby:
-          - '3.0'
-          - '3.1'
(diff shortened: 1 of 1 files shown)


nbulaj (approved):
Thanks!

> I do note that according to the Gemspec, we support >= 2.7, so we should probably also include Ruby 2.7 in the testing matrix at least.

While we're trying to support older Rubies as well (I mean we're trying to allow legacy projects to still use newer versions of the gem) I don't think there is much sense to add EOL ruby to the test matrix. Inconvenient, yes, but that's how I personally see it.

PR #1732, opened 2024-09-23
Fix Gemfile & Dockerfile to get passing tests

ransombriggs:
### Summary

This tweaks the Dockerfile and Gemfile to get a passing docker build.  This could probably be tweaked back more towards its original style, but it was much easier to debug when I could run the bundle process as the doorkeeper user.

PR #1735, opened 2024-10-02
Add pkce_code_challenge_methods option

### LINKED ISSUE #1717: Option to specify supported PKCE code_challenge_methods supported
In [OAuth 2.0 Security Best Current Practices](https://www.ietf.org/archive/id/draft-ietf-oauth-security-topics-27.html#name-authorization-code-grant), the PKCE code_challenge_method of `plain` MUST NOT be used, as it leaks the code_verifier in the authorization request:

> When using PKCE, clients SHOULD use PKCE code challenge methods that do not expose the PKCE verifier in the authorization request. Otherwise, attackers that can read the authorization request (cf. Attacker A4 in [Section 3](https://www.ietf.org/archive/id/draft-ietf-oauth-security-topics-27.html#secmodel)) can break the security provided by PKCE. **Currently, S256 is the only such method.**

Doorkeeper's PKCE support currently supports both `plain` and `S256` and there is no easy way to disable 'plain' to be inline with security best current practices.

## Potential Resolution

Add a `code_challenge_methods_supported` option that is an array of methods, defaulting to `['plain', 'S256']` which is then used when [validating the code_challenge_method](https://github.com/doorkeeper-gem/doorkeeper/blob/main/lib/doorkeeper/oauth/pre_authorization.rb#L153-L158).

Additionally this should be exposed via `Doorkeeper.configuration.code_challenge_methods_supported`, such that servers implementing OAuth Authorization Server Metadata (also now recommended, issue #1587) can return `code_challenge_methods_supported` property.

ThisIsMissEm:
### Summary

This fixes #1717, allowing a user to prevent using the `plain` method for PKCE code challenges, since using `plain` is deemed insecure (as noted in the PKCE RFC)

This is based off of #1732, I also have some local changes that pull in irb and debug gems for interactive debugging, which I found helpful when testing in the console via:

```sh
$ docker run -it --rm doorkeeper:test /srv/bin/console
```

### Other Information

In `validate_pkce_code_challenge_methods` I did want to log an error if the database hadn't been configured for PKCE, however, due to how this code is instantiated, I wasn't able to get that working due to the model class not yet being loaded by `bin/console` (so I'm assuming issues may happen in actual usage too).

The [error message does list](https://github.com/doorkeeper-gem/doorkeeper/blob/de0fb1fc0cbe8b9ac2a3ba49c21a0730ea21dd0b/config/locales/en.yml#L103) the supported methods always as `plain or S256` — to fix that, we'd need to parameterize the localisation message to receive the supported code challenge methods. I wasn't sure if I should make such a change here.

This also limits us to only ever supporting plain or S256, but to [my knowledge no other PKCE code challenge methods exist](https://www.iana.org/assignments/oauth-parameters/oauth-parameters.xhtml#pkce-code-challenge-method), and doorkeeper wouldn't have support for them anyway due to how it [hardcodes which ones it supports](https://github.com/doorkeeper-gem/doorkeeper/blob/5bf9d406dd710209cb2bc0600a20aa52178ff32b/lib/doorkeeper/oauth/authorization_code_request.rb#L98).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,21 +28,23 @@ ENV LANG=en_US.UTF-8
 ENV LANGUAGE=en_US:en
 ENV LC_ALL=en_US.UTF-8
 
-ENV BUNDLER_VERSION=2.5.11
-RUN gem install bundler -v ${BUNDLER_VERSION} -i /usr/local/lib/ruby/gems/$(ls /usr/local/lib/ruby/gems) --force
-
 WORKDIR /srv
-
-COPY Gemfile doorkeeper.gemspec /srv/
-COPY lib/doorkeeper/version.rb /srv/lib/doorkeeper/version.rb
-
-RUN bundle install
 
 COPY . /srv/
 
-RUN chown -R doorkeeper:doorkeeper /srv/coverage
+RUN chown -R doorkeeper:doorkeeper /srv
 
 # Set the running user for resulting container
 USER doorkeeper
 
-CMD ["rake"]
+RUN mkdir -p /srv/.local/gem/share
+ENV GEM_HOME=/srv/.local/gem/share
+
+ENV BUNDLER_VERSION=2.5.11
+RUN gem install bundler -v ${BUNDLER_VERSION}
+
+# This is a fix for sqlite alpine issues
+RUN bundle config force_ruby_platform true
+RUN bundle install
+
+CMD ["bundle", "exec", "rake"]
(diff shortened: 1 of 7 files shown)


nbulaj (approved):
I like it :bow:  Thanks a lot!

ransombriggs:
I pushed up https://github.com/doorkeeper-gem/doorkeeper/pull/1747 to try and fix the error message when `plain` is passed.

PR #1736, opened 2024-10-02
Improve developer experience

ThisIsMissEm:
### Summary

Whilst working on #1735, I noticed it was hard to debug the code, this change adds a fairly opinionated way to debug the code when developing the doorkeeper gem. Additionally I've fixed the warning in the tests that I mentioned in #1732.

Now when developing in either the code or the specs, you can use the `debugger` keyword to jump into an interactive debugging session.

### Other Information

Why include irb in the Gemfile? That's been recommended by the irb authors, due to irb bundled with ruby frequently being out of date with the state of irb development.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -27,3 +27,8 @@ gem "sqlite3", "~> 1.4", platform: [:ruby, :mswin, :mingw, :x64_mingw]
 
 gem "tzinfo-data", platforms: %i[mingw mswin x64_mingw]
 gem "timecop"
+
+gem 'irb', '~> 1.8'
+
+# Interactive Debugging tools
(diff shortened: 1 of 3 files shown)


nbulaj (approved):
LGTM :+1:

nbulaj:
Hm interesting that 1 test started to fail :thinking:

ThisIsMissEm:
> Hm interesting that 1 test started to fail 🤔

I think that test case may be flakey, have seen failures for it in other branches too

stanhu:
This change broke tests for Rails < 7.1: https://github.com/doorkeeper-gem/doorkeeper/pull/1742. I don't think this was flaky.

Issue #1587, opened 2022-09-11
Add support for OAuth 2.0 Authorization Server Metadata


emersion:
See https://www.rfc-editor.org/rfc/rfc8414

This allows clients to auto-discover OAuth 2.0 parameters instead of having to hardcode them.

emersion:
Note, it seems like [doorkeeper-openid_connect](https://github.com/doorkeeper-gem/doorkeeper-openid_connect) supports this: https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/152

It would still be nice to support the RFC without an additional plugin.

emersion:
Ping

nbulaj:
Hey @emersion , thanks for the feedback. I think we can add it into Doorkeeper itself, yes, and then reuse in doorkeeper openid_ connect extending default data. Would you mind prepare a PR?

emersion:
Sounds good. I don't have time for this at the moment, but will keep in my TODO list.

ThisIsMissEm:
@nbulaj I think I can maybe secure some funding for this work, would you like me to implement it if I can secure that funding? (I'm wanting to implement this in Mastodon)

ThisIsMissEm:
@emersion we implemented this directly in Mastodon 4.3 via https://github.com/mastodon/mastodon/pull/29191, since I had the funding to implement and didn't receive a response here.

ThisIsMissEm:
> Hey @emersion , thanks for the feedback. I think we can add it into Doorkeeper itself, yes, and then reuse in doorkeeper openid_ connect extending default data. Would you mind prepare a PR?

@nbulaj keep in mind that `.well-known/oauth-authorization-server` and `.well-known/openid-configuration` are different specifications; They're very similar but different.

Issue #1720, opened 2024-07-26
Support for RFC 9207 - OAuth 2.0 Authorization Server Issuer Identification


ThisIsMissEm:
> This document specifies a new parameter called iss. This parameter is used to explicitly include the issuer identifier of the authorization server in the authorization response of an OAuth authorization flow. The iss parameter serves as an effective countermeasure to "mix-up attacks".

[RFC 9207](https://www.rfc-editor.org/rfc/rfc9207.html)

I'd advocate for supporting this, since it prevents [mix-up attacks](https://www.ietf.org/archive/id/draft-ietf-oauth-security-topics-27.html#name-mix-up-attacks) as noted in OAuth 2.0 Security Best Current Practices.

This would require additional configuration of an `issuer` which defaults to the `root_url` for the Rails App. It would also require #1587.

I would be interested in implementing both in Doorkeeper if there is interest in receiving pull requests for these.

PR #1737, opened 2024-10-02
Feat: Implement oauth authorization server metadata

ThisIsMissEm:
### Summary

This is early work on supporting #1587, it's currently based off an older commit from #1732, #1735, and #1736 since those allow me to work semi-productively on the doorkeeper codebase.

I'm unsure if we'd really need a custom controller for discovery, as I think the initializer for configuration should probably handle the "adding extra stuff to the `.well-known/oauth-authorization-server` endpoint, instead of actually overriding the controller.

I don't think it makes sense for the `scope` option with doorkeeper to affect the `.well-known/oauth-authorization-server` endpoint, however, nesting that within a scope can work (allowing multiple oauth providers on the one domain, similar to how keycloak works)

I'm still to figure out all the necessary stuff for actually building out the data for response. I think I'm probably inclined to create a `Doorkeeper::OAuth::DiscoveryResponse` class, and a `custom_discovery_response` configuration block option, then have the controller just apply that class.

There's also the question of "should this be rename-able by other plugins?", e.g., OIDC that uses `.well-known/openid-configuration` instead of `.well-known/oauth-authorization-server` — I'd probably guess maybe no, and instead the OIDC plugin could just remove the `.well-known/oauth-authorization-server` route and add its own which is calling a subclass of `Doorkeeper::OAuth::DiscoveryResponse`?

### DIFF
--- a/app/controllers/doorkeeper/discovery_controller.rb
+++ b/app/controllers/doorkeeper/discovery_controller.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+module Doorkeeper
+  class DiscoveryController < Doorkeeper::ApplicationMetalController
+    def show
+      render json: {}, status: 200
+    end
+  end
+end
--- a/app/controllers/doorkeeper/discovery_controller.rb
+++ b/app/controllers/doorkeeper/discovery_controller.rb
@@ -3,7 +3,18 @@
 module Doorkeeper
   class DiscoveryController < Doorkeeper::ApplicationMetalController
     def show
-      render json: {}, status: 200
+      headers.merge!(discovery_response.headers)
+      render json: discovery_response.body,
+             status: discovery_response.status
+    end
+
+    private
+
+    def discovery_response
+      @discovery_response ||= Doorkeeper::OAuth::DiscoveryResponse.new(
+        root_url,
+        -> (**args) { url_for(**args) }
+      )
     end
   end
 end
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -321,6 +321,9 @@ module Doorkeeper
     #
     option :realm,                          default: "Doorkeeper"
 
+    # Issuer URL for the OAuth Authorization Server
+    option :issuer,                         default: nil
+
     # Forces the usage of the HTTPS protocol in non-native redirect uris
     # (enabled by default in non-development environments). OAuth2
     # delegates security in communication to the HTTPS protocol so it is
@@ -396,6 +399,10 @@ module Doorkeeper
 
     option :application_class,
            default: "Doorkeeper::Application"
+
+    # Allows setting a hash of custom data on the OAuth 2.0 Authorization
+    # Server Metadata discovery response.
+    option :custom_discovery_data, default: {}
 
     # Allows to set blank redirect URIs for Applications in case
     # server configured to use URI-less grant flows.
--- a/lib/doorkeeper/oauth/discovery_response.rb
+++ b/lib/doorkeeper/oauth/discovery_response.rb
@@ -1,0 +1,115 @@
+# frozen_string_literal: true
+
+module Doorkeeper
+  module OAuth
+    class DiscoveryResponse < BaseResponse
+      def initialize(root_url, url_builder)
+        @root_url = root_url
+        @url_builder = url_builder
+      end
+
+      def body
+        @body ||= {
+          issuer: issuer || @root_url,
+          authorization_endpoint: authorization_endpoint,
+          token_endpoint: token_endpoint,
+          revocation_endpoint: revocation_endpoint,
+          userinfo_endpoint: userinfo_endpoint,
+          scopes_supported: scopes_supported,
+          response_types_supported: response_types_supported,
+          response_modes_supported: response_modes_supported,
+          grant_types_supported: grant_types_supported,
+          token_endpoint_auth_methods_supported: token_endpoint_auth_methods_supported,
+          code_challenge_methods_supported: code_challenge_methods_supported,
+        }.merge(custom_discovery_data)
+      end
+
+      def status
+        :ok
+      end
+
+      def headers
+        {
+          "Cache-Control" => "public",
+          "Content-Type" => "application/json; charset=utf-8",
+        }
+      end
+
+      private
+
+      def config
+        @config ||= Doorkeeper.configuration
+      end
+
+      def url_for(**args)
+        @url_builder.call(**args)
+      end
+
+      def custom_discovery_data
+        config.custom_discovery_data.symbolize_keys
+      end
+
+      def issuer
+        config.issuer
+      end
+
+      def authorization_endpoint
+        mapping = Doorkeeper::Rails::Routes.mapping[:authorizations] || {}
+
+        url_for(
+          controller: mapping[:controllers] || "doorkeeper/authorizations",
+          action: 'new'
+        )
+      end
+
+      def token_endpoint
+        mapping = Doorkeeper::Rails::Routes.mapping[:tokens] || {}
+        
+        url_for(
+          controller: mapping[:controllers] || "doorkeeper/tokens",
+          action: 'create'
+        )
+      end
+
+      def userinfo_endpoint
+        nil
+      end
+
+      def revocation_endpoint
+        mapping = Doorkeeper::Rails::Routes.mapping[:tokens] || {}
+        
+        url_for(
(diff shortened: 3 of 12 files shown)


ThisIsMissEm:
@nbulaj @ransombriggs would love some feedback on my implementation proposal in the PR description — I think that's the right way forwards, but haven't yet begun to implement.

ransombriggs:
I would give feedback but I am not familiar with https://www.rfc-editor.org/rfc/rfc8414 and really just a bystander on this project.  I got it to do what I needed and then was moved onto other things.

nbulaj:
I have some feeling that the work should be done inside https://github.com/doorkeeper-gem/doorkeeper-openid_connect :thinking: 

I remember it handles all these stuff around Discovery and related. Have to recheck deeply

ThisIsMissEm:
> I have some feeling that the work should be done inside https://github.com/doorkeeper-gem/doorkeeper-openid_connect 🤔
> 
> I remember it handles all these stuff around Discovery and related. Have to recheck deeply

RFC 8414 is part of OAuth and is separate from OIDC's Discovery, but is compatible in some ways.

RFC 8414 uses `/.well-known/oauth-authorization-server` and OIDC Discovery uses `/.well-known/openid_configuration` (iirc), but the contents of the documents is largely the same. These differences are noted in RFC 8414

app/controllers/doorkeeper/discovery_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+      render json: discovery_response.body,
+             status: discovery_response.status
+    end
+
+    private
+
+    def discovery_response
+      @discovery_response ||= Doorkeeper::OAuth::DiscoveryResponse.new(
+        root_url,
+        -> (**args) { url_for(**args) }
```
ThisIsMissEm, line 16: I'm not particularly keen on this, but including UrlFor or similar in the `Doorkeeper::OAuth::DiscoveryResponse` class didn't work as I'd expect it to, and I need the url builder in order to generate the absolute URLs for the authorization server metadata.

lib/doorkeeper/config.rb
```diff
@@ -321,6 +321,9 @@ def configure_secrets_for(type, using:, fallback:)
     #
     option :realm,                          default: "Doorkeeper"
 
+    # Issuer URL for the OAuth Authorization Server
+    option :issuer,                         default: nil
```
ThisIsMissEm, line 325: Technically this is maybe OpenID more, but OAuth 2.0 does have a spec for it, see: #1720

lib/doorkeeper/config.rb
```diff
@@ -397,6 +400,10 @@ def configure_secrets_for(type, using:, fallback:)
     option :application_class,
            default: "Doorkeeper::Application"
 
+    # Allows setting a hash of custom data on the OAuth 2.0 Authorization
+    # Server Metadata discovery response.
+    option :custom_discovery_data, default: {}
```
ThisIsMissEm, line 405: This is intended for applications to use, e.g., Mastodon where we want to set `service_documentation` and `app_registration_url` properties in the metadata.

lib/doorkeeper/oauth/discovery_response.rb
```diff
@@ -0,0 +1,115 @@
...
+        config.authorization_response_types
+      end
+
+      def response_modes_supported
+        config.authorization_response_flows.flat_map(&:response_mode_matches).uniq
+      end
+
+      def grant_types_supported
+        grant_types_supported = config.grant_flows.dup
+        grant_types_supported << 'refresh_token' if !!config.refresh_token_enabled?
```
ThisIsMissEm, line 101: This is technically broken, because `use_refresh_token` configuration can be a block/proc, which accepts `server` and `context` which, well, I don't think we have here, nor would it be possible to have?

spec/requests/endpoints/discovery_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+    expect(json_response).to have_key("code_challenge_methods_supported")
+
+    expect(json_response["issuer"]).to be_a(String)
+
+    expect(json_response["scopes_supported"]).to be_a(Array)
+    expect(json_response["response_types_supported"]).to be_a(Array)
+    expect(json_response["response_modes_supported"]).to be_a(Array)
+    expect(json_response["grant_types_supported"]).to be_a(Array)
+    expect(json_response["token_endpoint_auth_methods_supported"]).to be_a(Array)
+    expect(json_response["code_challenge_methods_supported"]).to be_a(Array)
```
ThisIsMissEm, line 35: I'm not sure if I should really test logic of the discovery response here, so instead I'm just asserting properties and types for things that must be arrays.

ThisIsMissEm:
@nbulaj @ransombriggs I think this is ready for a look, I think it works okay for now? Not sure if we need to do something else or not, there's a few things I've noted in the comments above.

spec/requests/endpoints/discovery_spec.rb
kmayer: Very similar to what we did for a one-off in our application:

```
      get oauth_discovery_path

      expect(response).to have_http_status(200)

      expect(response.content_type).to eq("application/json; charset=utf-8")

      object = JSON.parse(response.body)

      expect(object.dig("authorization_endpoint")).to eq("http://www.example.com/oauth/authorize")
      expect(object.dig("token_endpoint")).to eq("http://www.example.com/oauth/token")
      expect(object.dig("revocation_endpoint")).to eq("http://www.example.com/oauth/revoke")
      ...
```

ThisIsMissEm: Yeah, similar to the Mastodon one as well: https://github.com/mastodon/mastodon/blob/main/spec/requests/well_known/oauth_metadata_spec.rb

I'm just not sure I want to be validating the values or not here, since the paths can be configured with doorkeeper.

spec/requests/endpoints/discovery_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+RSpec.describe "Discovery endpoint" do
+  before do
+    default_scopes_exist :read
+    optional_scopes_exist :write, :publish
+  end
+
+  it "returns json" do
+    get "/.well-known/oauth-authorization-server"
+
+    response_status_should_be(200)
```
ThisIsMissEm, line 14: We should add a content-type assertion here as well.

lib/doorkeeper/oauth/discovery_response.rb
```diff
@@ -0,0 +1,115 @@
...
+        mapping = Doorkeeper::Rails::Routes.mapping[:tokens] || {}
+        
+        url_for(
+          controller: mapping[:controllers] || "doorkeeper/tokens",
+          action: 'revoke'
+        )
+      end
+
+      def scopes_supported
+        config.scopes.to_a
```
ThisIsMissEm, line 88: Might be better as:
```suggestion
        config.scopes.map(&:to_s)
```

Might depend on how `Scopes` implements `to_a`

lib/doorkeeper/oauth/discovery_response.rb
ThisIsMissEm: Thinking about this more, if we did drastically change how refresh tokens work, as suggested in #1771, then we could have options like:

```
Doorkeeper.configure do
  orm :active_record

  # Enable using them in general
  use_refresh_tokens true

  # Set the expiry for refresh tokens:
  refresh_token_expires_in 30.days

  # Allow refresh tokens to only be used by specific clients:
  allow_refresh_tokens_for do |client, access_grant|
    access_grant.scopes.exists?('offline_access')
  end
end
```

Or something.

lib/doorkeeper/oauth/discovery_response.rb
```diff
@@ -0,0 +1,115 @@
...
+
+      def grant_types_supported
+        grant_types_supported = config.grant_flows.dup
+        grant_types_supported << 'refresh_token' if !!config.refresh_token_enabled?
+        grant_types_supported
+      end
+
+      # FIXME: https://github.com/doorkeeper-gem/doorkeeper/pull/1770
+      def token_endpoint_auth_methods_supported
+        %w(none client_secret_basic client_secret_post)
```
ThisIsMissEm, line 107: This will be fixed by #1772

alex-ross:
Is this kind of forgotten or is it someway to enable a `.well-known/oauth-authorization-server` trough this gem?

ThisIsMissEm:
@alex-ross this isn't forgotten, work is just progressing more slowly because of time constraints

PR #1738, opened 2024-10-07
skip_forgery_protection always

chris-iotum:
see also #TBD


chris-iotum:
That was accidental; I apologize.

Issue #431, opened 2014-07-14
Letting user choose from dynamic scopes


chendo:
I'd like to implement the ability to allow the user to choose which resources an OAuth2 client can access when they're presented with the authorization page. Is this possible currently? If not, would you accept a pull request with the feature if I manage to implement it myself?

grzesiek:
This would be incredibly useful +1

PR #1739, opened 2024-10-09
Add support for dynamic scopes

stanhu:
This commit adds support for dynamic scopes, which are disabled by default.

As discussed in https://github.com/keycloak/keycloak/discussions/8486, a dynamic scope notation is in the form:

```
<static-part>:<variable-part>
```

The objective of this feature is to have a static part of the scope that represents an entity and a variable part that identifies the entity.

For example, a scope of `user:1` could be interpreted as allowing access to perform actions of user 1. A wildcard (`*`) is allowed in the variable part, such as `user:*`. This scope allows the request to perform actions as any users.

Dynamic scopes can be enabled via:

```ruby
Doorkeeper.configure do
  enable_dynamic_scopes
end
```

A custom delimiter can also be configured:

```ruby
Doorkeeper.configure do
  enable_dynamic_scopes(delimiter: '-')
end
```

Relates to https://github.com/doorkeeper-gem/doorkeeper/issues/431

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,3 @@
+exclude_patterns:
+- "lib/doorkeeper/config.rb"
+- "spec/"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ User-visible changes worth mentioning.
 
 Add your entry here.
 
+- [#1739] Add support for dynamic scopes
 - [#1715] Fix token introspection invalid request reason
 - [#1714] Fix `Doorkeeper::AccessToken.find_or_create_for` with empty scopes which raises NoMethodError
 - [#1712] Add `Pragma: no-cache` to token response
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -31,6 +31,16 @@ module Doorkeeper
         @config.instance_variable_set(:@confirm_application_owner, true)
       end
 
+      # Provide support for dynamic scopes (e.g. user:*) (disabled by default)
+      # Optional parameter delimiter (default ":") if you want to customize
+      # the delimiter separating the scope name and matching value.
+      #
+      # @param opts [Hash] the options to configure dynamic scopes
+      def enable_dynamic_scopes(opts = {})
+        @config.instance_variable_set(:@enable_dynamic_scopes, true)
+        @config.instance_variable_set(:@dynamic_scopes_delimiter, opts[:delimiter] || ':')
+      end
+
       # Define default access token scopes for your provider
       #
       # @param scopes [Array] Default set of access (OAuth::Scopes.new)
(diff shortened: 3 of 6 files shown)


stanhu:
@nbulaj What do you think of this pull request? Also I'm not sure how best to fix the Code Climate errors. I tried to disable the `method-count` check for just this file, but it seems that Code Climate can't do that.

grzesiek:
+1

samg:
We're long time users of Doorkeeper and having the ability to use dynamic scopes in Doorkeeper would be very helpful to the projects we have planned. 😄  Are there any plans to merge this and include it in the gem?

nbulaj:
hey @stanhu :wave: Sounds interesting, but I wasn't able to find some RFC. So it looks like a custom addition to the spec. Do we have any docs to reference? Except keycloack :smile:

stanhu:
@nbulaj I don't think there is a spec governing how scopes are handled; I think that is an application-specific decision. I will keep looking.

stanhu:
As https://datatracker.ietf.org/doc/html/rfc6749#page-23 mentions:

> The strings are defined by the authorization server. 

RabbitMQ, for example, supports wildcard OAuth2 scopes: https://www.rabbitmq.com/docs/oauth2#scope-translation

nbulaj:
Anyway I like it :+1:

nbulaj:
I see tests failing but seems to be related to something else :thinking:  Looks like something change in Rails

stanhu:
The tests are failing due to https://github.com/doorkeeper-gem/doorkeeper//commit/b4bd6803147934d41d0d19642ef8e97d48984090. Rails is showing the exception because `ActionDispatch::Request.show_exception?` is returning true since older Rails versions expect that `false` is used: https://github.com/rails/rails/blob/v7.0.8.4/actionpack/lib/action_dispatch/http/request.rb#L186

https://github.com/rails/rails/pull/45867 deprecated `false` in favor of `:all`, `:none`,  and `:rescuable` for Rails 7.1.

stanhu:
https://github.com/doorkeeper-gem/doorkeeper/pull/1742 fixes the broken Rails tests.

nbulaj:
Can we rebase now @stanhu ? Above one is merged now :bow:

stanhu:
@nbulaj Done, thanks!

stanhu:
@nbulaj Friendly ping: could you take a look at this pull request?

nbulaj (approved):
Just checked, awesome work, LGTM :+1:

nbulaj:
Can we please add a changelog entry? :pray:

stanhu:
@nbulaj Thank you for the merge! Would you mind tagging a new release?

nbulaj:
Will review rest of the opened MRs tomorrow and yes, will prepare a release. Maybe more can be included as well

PR #1741, opened 2024-10-16
load the session if the session is not preloaded already

laomatt:
### Summary

When I upgraded an app to rails 7, the token controller was not pre loading the session which is necessary for the creation of the token.  This is just to make sure the session is created if its not pre loaded.


ThisIsMissEm:
Why's the session needed for the creation of the token? The token creation only relies on the authorization code? You might want to use the `before_successful_authorization` configuration hook instead.

PR #1742, opened 2024-10-24
Fix broken tests for older Rails versions

stanhu:
### Summary

https://github.com/doorkeeper-gem/doorkeeper/pull/1736 broke tests for Rails < 7.1 because these versions compared
`config.action_dispatch.show_exceptions == false`. Conditionally set this value in the test set up to fix the broken tests.

### Other Information

If there's anything else that's important and relevant to your pull
request, mention that information here. This could include
benchmarks, or other information.

If you are updating CHANGELOG.md file or are asked to update it by reviewers,
please add the changelog entry at the top of the file.

Thanks for contributing to Doorkeeper project!


nbulaj (approved):
Thanks :bow:

Issue #1743, opened 2024-10-30
Refresh tokens that are paired with expired access tokens cannot be revoked


ransombriggs:
### Steps to reproduce

> What we need to do to see your problem or bug?

The token revocation endpoint accepts either access or refresh tokens, but will not revoke refresh tokens if the access token it is paired with has expired.  This is not the right behavior because the refresh tokens have no expiration and we should be able to revoke them at anytime.

I am pushing up a PR soon with specs for the behavior I would expect as well as an initial awkward fix.

### Expected behavior

> Tell us what should happen

I should be able to revoke a refresh token after the `expired?` method returns true.

### Actual behavior

> Tell us what happens instead

It silently fails to revoke.

nbulaj:
And this one can be closed as well? https://github.com/doorkeeper-gem/doorkeeper/pull/1744

ransombriggs:
yup, closing

PR #1744, opened 2024-10-30
Allow for expired refresh tokens to be revoked

ransombriggs:
### Summary

This alters the revocation endpoint so that when a refresh token is revoked, we check `revoked?` rather than `accessible?` since the extra `expired?` check in `accessible?` does not apply to refresh tokens since they never expire.

https://github.com/doorkeeper-gem/doorkeeper/issues/1743

### DIFF
--- a/app/controllers/doorkeeper/tokens_controller.rb
+++ b/app/controllers/doorkeeper/tokens_controller.rb
@@ -113,17 +113,36 @@ module Doorkeeper
       # The authorization server responds with HTTP status code 200 if the token
       # has been revoked successfully or if the client submitted an invalid
       # token
-      token.revoke if token&.accessible?
+      revocable_token.revoke if revocable_token.revocable?
     end
 
     def token
-      @token ||=
+      revocable_token&.token
+    end
+
+    def revocable_token
+      return @revocable_token if defined? @revocable_token
+
+      @revocable_token =
         if params[:token_type_hint] == "refresh_token"
-          Doorkeeper.config.access_token_model.by_refresh_token(params["token"])
+          refresh_token
         else
-          Doorkeeper.config.access_token_model.by_token(params["token"]) ||
-            Doorkeeper.config.access_token_model.by_refresh_token(params["token"])
+          access_token || refresh_token
         end
+    end
+
+    def refresh_token
+      token = Doorkeeper.config.access_token_model.by_refresh_token(params["token"])
+      return unless token
+
+      RevocableTokens::RevocableRefreshToken.new(token)
+    end
+
+    def access_token
+      token = Doorkeeper.config.access_token_model.by_token(params["token"])
+      return unless token
+
+      RevocableTokens::RevocableAccessToken.new(token)
     end
 
     def strategy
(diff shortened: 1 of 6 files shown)


app/controllers/doorkeeper/tokens_controller.rb
```diff
@@ -113,19 +113,53 @@ def revoke_token
...
     def token
-      @token ||=
+      token_and_type&.[](:token)
+    end
+
+    def token_type
+      token_and_type&.[](:type)
+    end
+
+    def token_and_type
```
ransombriggs, line 132: I admit that this implementation is pretty awkward, I went with the simplest / quickest just to show the issue.  I think I would rather create an object here that is the pair instead, but not sure if y'all would think that is overkill or would match your patterns, just let me know and I can refactor this however you want it to look.

app/controllers/doorkeeper/tokens_controller.rb
```diff
@@ -113,19 +113,53 @@ def revoke_token
       # The authorization server responds with HTTP status code 200 if the token
       # has been revoked successfully or if the client submitted an invalid
       # token
-      token.revoke if token&.accessible?
```
ransombriggs, line 116: We have already checked `token.blank?` by the time we call this so the `&.` felt unnecessary, but I can bring it back.

spec/controllers/tokens_controller_spec.rb
```diff
@@ -183,8 +191,89 @@
         expect(response.status).to eq 200
       end
 
-      it "revokes the access token" do
-        post :revoke, params: { client_id: client.uid, token: access_token.token }
+      it "does not revoke the access token when token_type_hint == refresh_token" do
```
ransombriggs, line 194: The `token` method was not fully tested so added some specs before I made my changes.

app/controllers/doorkeeper/tokens_controller.rb
nbulaj: Agree, maybe better to have some clear DTO object. Can we refactor it please? :pray:  We have some (well not good actually) examples like OAuth::Client

ransombriggs: I went a slightly different direction and refactored in [5dd6454](https://github.com/ransombriggs/doorkeeper/commit/5dd6454dc940aa0d50b505e4308e39d37e8ec8ba) into classes that respond to `revocable?` and `revoke`.  I think it is cleaner than this the original code, but can keep tweaking.

nbulaj (approved):
LGTM, thanks! :bow:

nbulaj:
@ransombriggs can we please squash all the commits into a single one for the clear history? :pray:  Thanks!

app/controllers/doorkeeper/tokens_controller.rb
```diff
@@ -113,19 +113,38 @@ def revoke_token
...
+      revocable_token.revoke if revocable_token.revocable?
     end
 
     def token
-      @token ||=
+      revocable_token&.token
+    end
+
+    def revocable_token
+      return @revocable_token if defined? @revocable_token
```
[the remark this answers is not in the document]
ransombriggs: I went with the `defined?` pattern so that `nil` is memozied properly in the case where the token cannot be found.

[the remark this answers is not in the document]
kmayer: ah! my bad. thank you.

nbulaj:
Thanks a lot @ransombriggs ! :bow:

Issue #1746, opened 2024-10-31
The error message for invalid pkce method is not affected by pkce_code_challenge_methods configuration


ransombriggs:
### Steps to reproduce

I am upgrading to get the functionality in https://github.com/doorkeeper-gem/doorkeeper/pull/1735 because we do not accept the `plain` method but noticed that the error message still says that plain is an acceptable method.
 
> What we need to do to see your problem or bug?

```
  Doorkeeper.configure
    pkce_code_challenge_methods ["S256"]
  end
```

```
        # from a request spec that we have for our hack that I am trying to remove
        get oauth_authorization_path,
            params: {
              client_id: application.uid,
              redirect_uri: application.redirect_uri,
              response_type: "code",
              scope: scope,
              code_challenge: code_challenge,
              code_challenge_method: "plain",
            },
            headers: {
              "cookie" => session_cookie,
            }
```

### Expected behavior

 The code challenge method must be S256.

### Actual behavior

 The code challenge method must be plain or S256.

nbulaj:
I believe we can close this one via https://github.com/doorkeeper-gem/doorkeeper/pull/1747 , right?

ransombriggs:
yup, closing

PR #1747, opened 2024-10-31
Fix unknown pkce method error when configured

ransombriggs:
### Summary

I am upgrading to get the configuration update from https://github.com/doorkeeper-gem/doorkeeper/pull/1735 but noticed that the error message is not correct when it is configured in that it makes it sound like we accept `plain`.

https://github.com/doorkeeper-gem/doorkeeper/issues/1746

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -101,9 +101,9 @@ en:
         access_denied: 'The resource owner or authorization server denied the request.'
         invalid_scope: 'The requested scope is invalid, unknown, or malformed.'
         invalid_code_challenge_method:
-          zero: 'The authorization server does not support PKCE as there are no accepted code_challenge methods'
-          one: 'The code challenge method must be %{challenge_methods}.'
-          other: 'The code challenge method must be one of %{challenge_methods}.'
+          zero: 'The authorization server does not support PKCE as there is no accepted code_challenge_method'
+          one: 'The code_challenge_method must be %{challenge_methods}.'
+          other: 'The code_challenge_method must be one of %{challenge_methods}.'
         server_error: 'The authorization server encountered an unexpected condition which prevented it from fulfilling the request.'
         temporarily_unavailable: 'The authorization server is currently unable to handle the request due to a temporary overloading or maintenance of the server.'
 
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -100,7 +100,10 @@ en:
         unauthorized_client: 'The client is not authorized to perform this request using this method.'
         access_denied: 'The resource owner or authorization server denied the request.'
         invalid_scope: 'The requested scope is invalid, unknown, or malformed.'
-        invalid_code_challenge_method: 'The code challenge method must be one of %{challenge_methods}.'
+        invalid_code_challenge_method:
+          zero: 'There are no acceptable code_challenge methods'
+          one: 'The code challenge method must be %{challenge_methods}.'
+          other: 'The code challenge method must be one of %{challenge_methods}.'
         server_error: 'The authorization server encountered an unexpected condition which prevented it from fulfilling the request.'
         temporarily_unavailable: 'The authorization server is currently unable to handle the request due to a temporary overloading or maintenance of the server.'
 
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -100,7 +100,7 @@ en:
         unauthorized_client: 'The client is not authorized to perform this request using this method.'
         access_denied: 'The resource owner or authorization server denied the request.'
         invalid_scope: 'The requested scope is invalid, unknown, or malformed.'
-        invalid_code_challenge_method: 'The code challenge method must be plain or S256.'
+        invalid_code_challenge_method: 'The code challenge method must be one of %{challenge_methods}.'
         server_error: 'The authorization server encountered an unexpected condition which prevented it from fulfilling the request.'
         temporarily_unavailable: 'The authorization server is currently unable to handle the request due to a temporary overloading or maintenance of the server.'
 
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -101,7 +101,7 @@ en:
         access_denied: 'The resource owner or authorization server denied the request.'
         invalid_scope: 'The requested scope is invalid, unknown, or malformed.'
         invalid_code_challenge_method:
-          zero: 'The authorization server does not support PKCE as there is no accepted code_challenge_method'
+          zero: 'The authorization server does not support PKCE as there are no accepted code_challenge_method values.'
           one: 'The code_challenge_method must be %{challenge_methods}.'
           other: 'The code_challenge_method must be one of %{challenge_methods}.'
         server_error: 'The authorization server encountered an unexpected condition which prevented it from fulfilling the request.'
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -101,7 +101,7 @@ en:
         access_denied: 'The resource owner or authorization server denied the request.'
         invalid_scope: 'The requested scope is invalid, unknown, or malformed.'
         invalid_code_challenge_method:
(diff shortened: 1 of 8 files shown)


config/locales/en.yml
```diff
@@ -100,7 +100,7 @@ en:
         unauthorized_client: 'The client is not authorized to perform this request using this method.'
         access_denied: 'The resource owner or authorization server denied the request.'
         invalid_scope: 'The requested scope is invalid, unknown, or malformed.'
-        invalid_code_challenge_method: 'The code challenge method must be plain or S256.'
+        invalid_code_challenge_method: 'The code challenge method must be one of %{challenge_methods}.'
```
ransombriggs, line 103: This could probably use some tweaking, just let me know how you think this should read.

lib/doorkeeper/oauth/error_response.rb
```diff
@@ -12,6 +12,7 @@ def self.from_request(request, attributes = {})
           attributes.merge(
             name: error_name_for(request.error),
             exception_class: exception_class_for(request.error),
+            translate_options: request.error.try(:translate_options),
```
ransombriggs, line 15: I was not completely confident what classes `request.error` are expected here.  I can remove the `try` if we always expect to receive a `DoorkeeperError` here.

config/locales/en.yml
ThisIsMissEm: You'd probably want to use the one / other syntax here.. 
```suggestion
        invalid_code_challenge_method:
          one: 'The code challenge method must be %{challenge_methods}.'
          other: 'The code challenge method must be one of %{challenge_methods}.'
```

Which would need a `count` on `challenge_methods`, iirc.

lib/doorkeeper/errors.rb
```diff
@@ -45,6 +49,14 @@ def self.name_for_response
       end
     end
 
+    class InvalidCodeChallengeMethod < BaseResponseError
+      def self.translate_options
+        {
+          challenge_methods: Doorkeeper.config.pkce_code_challenge_methods_supported.join(", ")
```
ThisIsMissEm, line 55: Re: pluralisation in https://github.com/doorkeeper-gem/doorkeeper/pull/1747/files#r1825345954
```suggestion
          challenge_methods: Doorkeeper.config.pkce_code_challenge_methods_supported.join(", "),
          count: Doorkeeper.config.pkce_code_challenge_methods_supported.length
```

lib/doorkeeper/oauth/error_response.rb
ThisIsMissEm: Maybe there's a default that can be set? grab :translate_options or fallback to {}

spec/lib/oauth/error_spec.rb
```diff
@@ -17,5 +18,21 @@
...
+          {
+            challenge_methods: "foo, bar"
+          }
+        )
+      end
+
+      it "is translated from translation messages with variables" do
+        expect(error.description).to eq("The code challenge method must be one of foo, bar.")
+      end
+    end
```
ThisIsMissEm, line 36: We should have a test case for the "one" option, i.e., just "foo" not "foo, bar"

(this likely more applies to pre_authorization_spec.

config/locales/en.yml
ransombriggs: Awesome suggestions, I went ahead and updated in ccc97732f3977e57d3a7c541689c30f418de6697

lib/doorkeeper/errors.rb
ransombriggs: Thanks, used this in ccc97732f3977e57d3a7c541689c30f418de6697

spec/lib/oauth/error_spec.rb
ransombriggs: I added a test for `%w(foo)` and `[]` in ccc97732f3977e57d3a7c541689c30f418de6697

lib/doorkeeper/oauth/error_response.rb
ransombriggs: I default it [here](https://github.com/doorkeeper-gem/doorkeeper/pull/1747/files#diff-990b33ab855d22431465363199d973eb7dd1c5eb2dbfbfa0151801ada778bdf5R7) in `error.rb` because that caught all the usages of `Error` in one place.

config/locales/en.yml
```diff
@@ -100,7 +100,10 @@ en:
         unauthorized_client: 'The client is not authorized to perform this request using this method.'
         access_denied: 'The resource owner or authorization server denied the request.'
         invalid_scope: 'The requested scope is invalid, unknown, or malformed.'
-        invalid_code_challenge_method: 'The code challenge method must be plain or S256.'
+        invalid_code_challenge_method:
+          zero: 'There are no acceptable code_challenge methods'
```
ThisIsMissEm, line 104: this should never be reachable, but if it is, say
```suggestion
          zero: 'This authorization server does not support PKCE as there a no accepted code_challenge methods'
```

Would be better?

lib/doorkeeper/oauth/error_response.rb
ThisIsMissEm: More thinking of if the Error somehow doesn't derive from Doorkeeper::OAuth::Error, e.g., it comes from activerecord or something

spec/lib/oauth/pre_authorization_spec.rb
```diff
@@ -331,6 +331,22 @@
...
+            pkce_code_challenge_methods []
+          end
+        end
+
+        it "rejects plain as a code_challenge_method" do
+          attributes[:code_challenge] = "a45a9fea-0676-477e-95b1-a40f72ac3cfb"
+          attributes[:code_challenge_method] = "plain"
+
+          expect(pre_auth).to_not be_authorizable
+          expect(pre_auth.error_response.description).to eq("There are no acceptable code_challenge methods")
```
ThisIsMissEm, line 346: Also, we should use `code_challenge` or `code challenge` consistently.

config/locales/en.yml
ransombriggs: That is better, tweaked it just a little bit to align with other messages and updated in [f042917](https://github.com/doorkeeper-gem/doorkeeper/pull/1747/commits/f04291758eae274b9fc92a141e71b7d4aeae8c46)

spec/lib/oauth/pre_authorization_spec.rb
ransombriggs: Good catch, this comment made me realize that we should probably be calling it `code_challenge_method` since that is the actual param, so I made that change in [d244e69](https://github.com/doorkeeper-gem/doorkeeper/pull/1747/commits/d244e6937dbe3f0e6194ae15438d6cbafd30092d)

config/locales/en.yml
```diff
@@ -100,7 +100,10 @@ en:
         unauthorized_client: 'The client is not authorized to perform this request using this method.'
         access_denied: 'The resource owner or authorization server denied the request.'
         invalid_scope: 'The requested scope is invalid, unknown, or malformed.'
-        invalid_code_challenge_method: 'The code challenge method must be plain or S256.'
+        invalid_code_challenge_method:
+          zero: 'The authorization server does not support PKCE as there is no accepted code_challenge_method'
```
ThisIsMissEm, line 104:
```suggestion
          zero: 'The authorization server does not support PKCE as there are no accepted code_challenge_method'
```

ThisIsMissEm (approved):
Tiny comment, but LGTMO

config/locales/en.yml
ransombriggs: I updated this message in 7e2bf4b

ThisIsMissEm (approved):
Looks good!

nbulaj (approved):
LGTM :+1: 

Thanks!

PR #1752, opened 2024-11-29
Bump the range of supported Ruby and Rails versions

naitoh:
Ruby 3.0 has reached the EOL:
https://www.ruby-lang.org/en/downloads/branches/

Also since Rails 6 has been dropped support for and Rails 7.2, 8.0 are out
corresponding gemfiles are now dispensable to ci.
https://rubyonrails.org/maintenance

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,15 +21,17 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         ruby:
-          - "3.0"
           - "3.1"
           - "3.2"
           - "3.3"
         gemfile:
-          - gemfiles/rails_6_0.gemfile
-          - gemfiles/rails_6_1.gemfile
           - gemfiles/rails_7_0.gemfile
           - gemfiles/rails_7_1.gemfile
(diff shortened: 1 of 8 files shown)


doorkeeper.gemspec
```diff
@@ -31,8 +31,8 @@ Gem::Specification.new do |gem|
     "funding_uri" => "https://opencollective.com/doorkeeper-gem",
   }
 
-  gem.add_dependency "railties", ">= 5"
```
nbulaj, line 34: This can break a lot of applications. I believe we should set such requirements only if we're using something version-specific. So I'm for ignoring EOL rubies/rails in our CI, but against strict dependencies versions when we really don't need them

naitoh: OK, I see.
I will cancel this  gem.add_dependency  change.

naitoh:
@nbulaj 
I have removed the changes to the doorkeeper.gemspec from this PR.
Also fixed CI errors.

Thanks.

nbulaj (approved):
LGTM

PR #1754, opened 2024-12-04
Fix refresh tokens with dynamic scopes

stanhu:
### Summary

Suppose your OAuth application supports the scopes `public user:*`. Previously a refresh token with the `public user:1` scope returned a scope of `public` by default, unless `user:1` were explicitly requested. This happened because `Doorkeeper::Oauth::BaseRequest` does a set intersection:

```
"public user:*" & "public user:1" = "public"
```

With dynamic scopes, this doesn't work because the wildcard `user:*` doesn't match `user:1`.

To fix this, introduce `Scopes#allowed` and deprecate the use of `&`.

### DIFF
--- a/lib/doorkeeper/oauth/scopes.rb
+++ b/lib/doorkeeper/oauth/scopes.rb
@@ -70,8 +70,26 @@ module Doorkeeper
         end
       end
 
+      # DEPRECATED: With dynamic scopes, #allowed should be called because
+      # A & B doesn't really make sense with dynamic scopes.
+      #
+      # For example, if A = user:* and B is user:1, A & B = [].
+      # If we modified this method to take dynamic scopes into an account, then order
+      # becomes important, and this would violate the principle that A & B = B & A.
       def &(other)
+        return allowed(other) if dynamic_scopes_enabled?
+
         self.class.from_array(all & to_array(other))
+      end
+
+      # Returns a set of scopes that are allowed, taking dynamic
+      # scopes into account. This instance's scopes is taken as the allowed set,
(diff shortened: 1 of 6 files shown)


stanhu:
@nbulaj Just one small fix we ran into while trying to use dynamic scopes. Could you review?

lib/doorkeeper/oauth/scopes.rb
```diff
@@ -70,10 +70,26 @@ def <=>(other)
...
       end
 
+      # DEPRECATED: With dynamic scopes, #allowed should be called because
+      # A & B doesn't really make sense with dynamic scopes.
+      #
+      # For example, if A = user:* and B is user:1, A & B = [].
+      # If we modified this method to take dynamic scopes into an account, then order
+      # becomes important, and this would violate the principle that A & B = B & A.
       def &(other)
         self.class.from_array(all & to_array(other))
```
nbulaj, line 80: Can we somehow check here if dynamic scopes used? And run `allowed` behavior? Wanna to avoid braking changes in case somebody uses scores directly

stanhu: @nbulaj Ok, updated!

nbulaj (approved):
LGTM, thanks! :bow:

stanhu:
@nbulaj Thanks so much again! Sorry to bother you again, but would you be so kind to tag a new release? 

Thanks again for your work on this project.

nbulaj:
Hey @stanhu , already released :point_up:

PR #1755, opened 2024-12-19
Fix the error message for force_pkce

lurz:
#### Bug
When the `force_pkce` option is enabled for a public client, and the authorization code request doesn't include a code_challenge, the error message returned is

"The authorization server encountered an unexpected condition which prevented it from fulfilling the request."

#### Root cause
[#1705] didn't add the error message in the locales.

#### Fix
Base on RFC https://datatracker.ietf.org/doc/html/rfc7636#section-4.4.1, we should return an "invalid_request" error with a proper description. I added a new entry for the error message and included a test to verify the proper message is returned.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ User-visible changes worth mentioning.
 ## main
 
 Add your entry here.
+- [#1755] Fix the error message for force_pkce
 
 ## 5.8.1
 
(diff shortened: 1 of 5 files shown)


nbulaj:
@mattmanning can you please take a look? :pray:

mattmanning (approved):
LGTM

Thanks!

nbulaj (approved):
+1 from my side as well, thanks!

PR #1760, opened 2025-01-16
Bump the range of supported Ruby versions

naitoh:
### 1. add Bundled gems for Ruby 3.4

https://bugs.ruby-lang.org/issues/20187

It has already been backported to the 7-0-stable branch, but is not included in 7.0.8.7.
- https://github.com/rails/rails/blob/7-0-stable/activesupport/activesupport.gemspec#L41-L43
- https://github.com/rails/rails/blob/v7.0.8.7/activesupport/activesupport.gemspec

### 2. Specify concurrent-ruby 1.3.4 for Rails 7.0

concurrent-ruby 1.3.5 don't work on Rails 7.0

> NameError:
>  uninitialized constant ActiveSupport::LoggerThreadSafeLevel::Logger
    
See: https://github.com/ruby-concurrency/concurrent-ruby/pull/1062

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,7 @@ jobs:
           - "3.1"
           - "3.2"
           - "3.3"
+          - "3.4"
         gemfile:
           - gemfiles/rails_7_0.gemfile
           - gemfiles/rails_7_1.gemfile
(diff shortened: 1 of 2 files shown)


nbulaj (approved):
LGTM

Can we please squash the commits into a single one? :pray:

naitoh:
> Can we please squash the commits into a single one? 🙏

I squashed these commits, thanks.

PR #1761, opened 2025-01-17
Memoize authentication failure

c960657:
`Doorkeeper::Rails::Helpers#doorkeeper_token` uses memoization, but only when authentication is successful (positive hits).

Let it memoize authentication failures as well.


nbulaj (approved):
Makes sense to me, thanks :bow: 

Can we please add a changelog entry and squash the commits? Thanks!

PR #1762, opened 2025-01-27
Allow missing client to trigger invalid client error when force_pkce is enabled

gkemmey:
Currently, the force_pkce handling is treating a missing client as a non-confidential client and forcibly validating the code_verifier param. This updates the conditional to only force_pkce if the client is present.


nbulaj (approved):
Makes sense to me, thanks!

PR #1767, opened 2025-03-31
Make sure error handling happens on a controller level opposed to action level to account for the controller being extended

filipesperandio:
### Summary

When extending the `TokensController` for any purpose, if any helper method is called and ends up raising an error, that error wasn't being handled by the controller since the rescue was happening at the action level.
Turning an action level `rescue` into a controller `rescue_from` will ensure controllers extending `TokensController` will have the same error handling while enabling it to keep calling the helper methods without worrying about error handling.


nbulaj (approved):
I like it! Thanks!

Can we add a changelog entry please as well?

nbulaj:
OK I can do it my self actually, wanna to do a release soon. So thanks anyway!

Issue #984, opened 2017-08-22
Is there Digest support for Doorkeeper Client Credentials Authentication


jannelukkari:
Is there any plans for digest support to https://github.com/doorkeeper-gem/doorkeeper/wiki/Client-Credentials-flow

nbulaj:
@jaluke take a look at [this answer](https://security.stackexchange.com/a/49544):

> Since you use SSL, don't do HTTP digest; use HTTP basic instead. SSL offers a much better protection than what HTTP digest offers (when compared to HTTP basic); and HTTP digest is not compatible with proper password server-side storage mechanisms (bcrypt...).

Do you really need http digest authentication ?

jaluke:
I would prefer solution which do not transfer secrets. This why SSL + digest (ish) would be my goal. SSL usually terminates before application server so its a security enhancement to current client credential as well.

nbulaj:
~~I think this must be implemented in order to conform with _4.4.3.1.  Threat: Accidental Exposure of Passwords at Client Site_ of [RFC6819](https://tools.ietf.org/html/rfc6819)~~

**UPD**: It's about Resource Owner Password Credentials

> **4.4.3.1.  Threat: Accidental Exposure of Passwords at Client Site**
> 
>    If the client does not provide enough protection, an attacker or
>    disgruntled employee could retrieve the passwords for a user.
> 
>    Countermeasures:
> 
>    o  Use other flows that do not rely on the client's cooperation for
>        secure resource owner credential handling.
> 
>    o  Use digest authentication instead of plaintext credential
>        processing.

Would you like to submit a PR @jaluke ?

ThisIsMissEm:
That is all a long way to say: if you'd want Digest support for client authentication, then you'd need to [write an internet draft](https://authors.ietf.org/) to propose it & register it with that IANA registry linked above.

Also keep in mind that RFC 6819 was updated by the recently published RFC 9700 (which forms the basis of OAuth 2.1)

PR #1770, opened 2025-04-16
Implement helper method to retrieve token_endpoint_auth_methods

ThisIsMissEm:
### Summary

Whilst getting back to Mastodon OAuth work, I remembered that we didn't have a method to retrieve the configured `token_endpoint_auth_methods` for OAuth Authorization Server Metadata ([RFC 8414](https://www.rfc-editor.org/rfc/rfc8414.html)), and we needed [to hard-code this](https://github.com/mastodon/mastodon/blob/main/app/presenters/oauth_metadata_presenter.rb#L63) in mastodon.

Having a helper to determine which client authentication methods are supported is helpful for implementing RFC 8414 (there's WIP for this in #1587 / PR #1737 )

Essentially the `client_credentials` option maps to `token_endpoint_auth_methods` in the OAuth Authorization Server Metadata  per the [IANA registry](https://www.iana.org/assignments/oauth-parameters/oauth-parameters.xhtml#token-endpoint-auth-method) and the Dynamic Client Registration RFC (linked below)

I do note that this current implementation doesn't really support extensibility, even though additional methods like `private_key_jwt` have been registered.

(The dynamic client registration spec is the thing that actually defined `none`, `client_secret_basic`, and `client_secret_post` as methods: https://www.rfc-editor.org/rfc/rfc7591.html#section-2 )

### Other Information

I'm not sure if this documentation page is correct, which implies `client_credentials` option can be a proc/method: https://github.com/doorkeeper-gem/doorkeeper/wiki/Changing-how-clients-are-authenticated

If that is the case, we may need an alternative mechanism for this.

### DIFF
--- a/spec/lib/config_spec.rb
+++ b/spec/lib/config_spec.rb
@@ -287,11 +287,56 @@ RSpec.describe Doorkeeper::Config do
     it "can change the value" do
       Doorkeeper.configure do
         orm DOORKEEPER_ORM
-        client_credentials :from_digest, :from_params
+        client_credentials :from_basic
       end
 
       expect(config.client_credentials_methods)
-        .to eq(%i[from_digest from_params])
+        .to eq(%i[from_basic])
+    end
+  end
+
+  # Returns token endpoint auth methods based on client_credentials per
+  # https://www.iana.org/assignments/oauth-parameters/oauth-parameters.xhtml#token-endpoint-auth-method
+  describe 'token_endpoint_auth_methods' do
+    it 'returns methods according to defaults' do
+      expect(config.client_credentials_methods).to eq(%i[from_basic from_params])
+      expect(config.token_endpoint_auth_methods).to contain_exactly('none', 'client_secret_post', 'client_secret_basic')
+    end
+
+    it "returns none even if no methods are configured" do
+      Doorkeeper.configure do
+        orm DOORKEEPER_ORM
+        client_credentials
+      end
+
+      expect(config.client_credentials_methods)
+        .to eq([])
+
+      expect(config.token_endpoint_auth_methods).to contain_exactly('none')
+    end
+
+    it 'returns client_secret_post if configured' do
+      Doorkeeper.configure do
+        orm DOORKEEPER_ORM
+        client_credentials :from_params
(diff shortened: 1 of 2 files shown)


spec/lib/config_spec.rb
```diff
@@ -287,11 +287,56 @@
     it "can change the value" do
       Doorkeeper.configure do
         orm DOORKEEPER_ORM
-        client_credentials :from_digest, :from_params
+        client_credentials :from_basic
```
ThisIsMissEm, line 290: `:from_digest` was never a "supported" client authentication method, from what I can tell, so instead I've switched this to an actually supported client authentication method. (I looked through the git history for https://github.com/doorkeeper-gem/doorkeeper/blob/main/lib/doorkeeper/oauth/client/credentials.rb )

nbulaj: True. We have some longstanding issue for digest auth

https://github.com/doorkeeper-gem/doorkeeper/issues/984

nbulaj:
>  I'm not sure if this documentation page is correct, which implies client_credentials option can be a proc/method: https://github.com/doorkeeper-gem/doorkeeper/wiki/Changing-how-clients-are-authenticated

> If that is the case, we may need an alternative mechanism for this.

This seems to be very old page, I checked `config.rb` and other files and don't see any place where we yield a block. So I believe it currently supports only collection, yes

ThisIsMissEm:
@nbulaj what do you think of this idea for being able to register different client authentication mechanisms?

That'd give an upgrade path for people using the block syntax — just write & register a custom client authentication mechanism.

Also, from what I can tell 'none' for public clients (non-confidential ones) is always an option? (This is currently handled by supporting `blank?` on credentials by just checking `uid`)

nbulaj:
Yeah this sounds great to me @ThisIsMissEm :+1:

ThisIsMissEm:
> Yeah this sounds great to me @ThisIsMissEm 👍

@nbulaj good! Because it's already mostly implemented in #1772

ThisIsMissEm:
Closing this as it's better implemented by #1772

PR #1772, opened 2025-04-20
Add Client Authentication Methods Registry

ThisIsMissEm:
### Summary

In #1770, I realised that really we do need the ability for extensions to be able to register additional client authentication mechanisms. This implements a strategy similar to that of the grant flows registry.

Essentially we should be able to support all of the registered methods, either in core or via extensions: https://www.iana.org/assignments/oauth-parameters/oauth-parameters.xhtml#token-endpoint-auth-method

### Implementation Notes

This allows for another gem that extends doorkeeper to do something like:

```ruby
  Doorkeeper::ClientAuthentication.register(
    :private_key_jwt,
    PrivateKeyJwt
  )
```

And provided `PrivateKeyJwt` implements what it needs to, then that client authentication mechanism would then be supported by Doorkeeper!

For the user, they'd need to do:
```ruby
Doorkeeper.configure do
  # ...
  client_authentication [ :none, :client_secret_basic, :client_secret_post, :private_key_jwt ]
  # ...
end
```

### Other Information

This would be a breaking change, unless we managed to convert `:from_params` and `:from_basic` across to the new values — however, this does use `request.request_parameter` instead of `request.parameters` so only looks at the request body, not the query params as well.

So like, if you had:

```ruby
Doorkeeper.configure do
  # ...

  # disabling `:from_params` for client_secret_post / none
  client_credentials :from_basic
  # ...
end
```

You'd now need:

```ruby
Doorkeeper.configure do
  # ...
  client_authentication [ :client_secret_basic ]
  # ...
end
```

I'm not yet 100% happy with this code, but it's a start.

I haven't yet integrated this in fully, nor added tests for it all.

---
This work was in part funded by my wonderful supporters: https://support.thisismissem.social


lib/doorkeeper/client_authentication.rb
```diff
@@ -0,0 +1,26 @@
...
+require "doorkeeper/client_authentication/registry"
+
+module Doorkeeper
+  module ClientAuthentication
+    extend Registry
+
+    register(
+      :none,
+      mechanism: Doorkeeper::ClientAuthentication::Mechanisms::None,
+      authenticates_client: false
```
ThisIsMissEm, line 13: I'm not sure if I really need this, since we have `blank?` on `Doorkeeper::ClientAuthentication::Credentials`

lib/doorkeeper/request.rb
```diff
@@ -3,6 +3,16 @@
...
   module Request
     class << self
+      def client_authentication_mechanism(request)
+        client_authentication_mechanisms = client_authentication_mechanisms.detect do |mechanism|
+          mechanism.matches_request?(request)
+        end
+
+        if client_authentication_mechanisms
+          client_authentication_mechanisms.mechanism
+        end
```
ThisIsMissEm, line 13: I'm not sure what the "fallback" would be here? I think it'd just be a flow that returned empty `Credentials`?

lib/doorkeeper/server.rb
```diff
@@ -37,8 +42,7 @@ def resource_owner
     end
 
     def credentials
```
ThisIsMissEm, line 44: This is used by the `client` method above, however, we probably need to change that method to:

```
def client
  @client ||= OAuth::Client.find(credentials.uid)
end
```
and have:
```
def client_authenticated?
  return unless client
  return client if credentials.blank? && !client.confidential?
  return client if client.secret_matches?(credentials.secret)
end
```

i.e., we're separating the "getting of the client" from the "authenticating of a client", such that for things like token revocation, we can just get the client, authenticate it if necessary, but then assert that the access token or refresh token was issued to that client.

I'm not 100% sure on the exact specifics here though.

spec/support/helpers/url_helper.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 module UrlHelper
+  # FIXME
```
ThisIsMissEm, line 4: @nbulaj so thanks to doing this, I've just realized all our tests are technically wrong. We're passing everything through query parameters, when the `client_secret` should never go via query parameters, as far as I know?

> Including the client credentials in the request-body using the two parameters is NOT RECOMMENDED and SHOULD be limited to clients unable to directly utilize the HTTP Basic authentication scheme (or other password-based HTTP authentication schemes).  The parameters can only be transmitted in the request-body and MUST NOT be included in the request URI.

That last sentence in particular.

Here in our tests we're passing the parameters via the request URI, not the request body.
- `/oauth/token` is a POST request has client authentication
- `/oauth/revoke` is a POST request has client authentication
- `/oauth/authorize` is a GET request (doesn't include `client_secret`)

So the URLs that we are generating here are largely incorrect, and that's going to require some pretty serious work to fix. We likely need to change most of these to be like a `build_token_request` or `build_refresh_token_request` method which returns the url, body, and headers.

ThisIsMissEm: I have a fix in https://github.com/doorkeeper-gem/doorkeeper/pull/1772/commits/93b2d41e77dc1a53608ef8eb7047e073535c6022

spec/requests/endpoints/token_spec.rb
```diff
@@ -22,10 +22,10 @@
 
   it "accepts client credentials with basic auth header" do
     post token_endpoint_url,
-         params: {
+         params: token_endpoint_params(
```
ThisIsMissEm, line 25: This was failing because of missing `grant_type` parameter. I'm not sure how it ever passed.

ThisIsMissEm: (it's clearly using `authorization_code` though, due to the presence of the `code` parameter

lib/doorkeeper/client_authentication/fallback_mechanism.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Doorkeeper
+  module ClientAuthentication
+    class FallbackMechanism
+      def authenticate(request)
+        Doorkeeper::ClientAuthentication::Credentials.new(nil, nil)
```
ThisIsMissEm, line 7: I think this should actually just be returning nil, not an instance of Credentials, which is why the `lib/doorkeeper/oauth/token_introspection.rb` condition failed.

ThisIsMissEm:
So this actually works now, and all the tests pass.

All we need now is:
- tests for the classes in `doorkeeper/oauth/client_authentication/`
- any tests we want for the client authentication registry stuff (I haven't looked at GrantFlow to determine what we might need here)

lib/doorkeeper/server.rb
ThisIsMissEm: Now that I think about it, maybe as it is is better? You only get a `client` when it's authenticated, even if there were credentials on the request. Maybe it's important to know _how_ the client is authenticated, I'm not sure?

spec/lib/oauth/client_authentication/none_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      binding.break
+      expect(described_class.matches_request?(request)).to_not be true
+    end
+
+    private
+
+    # I'm not sure if there's a better way to get a mock rack request for
+    # testing. Here we don't need a full request spec, but we do need enough to
+    # check that the logic of these classes works.
+    def mock_request(params, credentials = nil)
```
ThisIsMissEm, line 38: There's gotta be a better way to get a fake request that acts in a given way in ruby, surely?

spec/lib/oauth/client_authentication/none_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      expect(described_class.matches_request?(request)).to be true
+    end
+
+    it "doesn't match if the request has client_secret" do
+      request = mock_request({
+        client_id: '1234',
+        client_secret: "5678"
+      })
+
+      expect(described_class.matches_request?(request)).to_not be true
```
ThisIsMissEm, line 21: `matches_request?` actually returns `true` or `nil`, not true or false, just due to a quirk of boolean logic in ruby.

lib/doorkeeper/config/option.rb
```diff
@@ -53,7 +53,7 @@ def option(name, options = {})
             value = if attribute_builder
                       attribute_builder.new(&block).build
                     else
-                      block || args.first
+                      block || (options[:array] == true ? args : args.first)
```
ThisIsMissEm, line 56: We previously didn't have a way to use the `option` DSL to define options that take multiple arguments and convert them to a list/array. This seemed like a reasonable thing to add?

lib/doorkeeper/config.rb
```diff
@@ -253,9 +243,25 @@ def configure_secrets_for(type, using:, fallback:)
     option :orm,                            default: :active_record
     option :native_redirect_uri,            default: "urn:ietf:wg:oauth:2.0:oob", deprecated: true
     option :grant_flows,                    default: %w[authorization_code client_credentials]
+    option :client_authentication,          default: %w[client_secret_basic client_secret_post none]
```
ThisIsMissEm, line 246: This doesn't actually work as I documented in the PR description, instead it's:

```ruby
Doorkeeper.configure do
  client_authentication %w[client_secret_basic client_secret_post none]
end
```

Instead of
```ruby
Doorkeeper.configure do
  client_authentication :client_secret_basic, :client_secret_post, :none
end
```

With the new `array` flag for `option` we could make it what I thought it was.

lib/doorkeeper/config/option.rb
ThisIsMissEm: If we choose to add this, I don't mind if it's as `options[:array]` or `options[:list]` or something.

lib/doorkeeper/config.rb
ThisIsMissEm: We probably also need validation that this isn't empty, since that'd cause most of OAuth to fail, I think.

lib/doorkeeper/oauth/client_authentication/client_secret_basic.rb
```diff
@@ -0,0 +1,22 @@
...
+      class ClientSecretBasic
+        def self.matches_request?(request)
+          request.authorization.present? && request.authorization.downcase.start_with?('basic')
+        end
+
+        def authenticate(request)
+          value = request.authorization.to_s.split(" ", 2).second
+          client_id, client_secret = Base64.decode64(value).split(':', 2)
+
+          return unless client_id.present? && client_secret.present?
```
ThisIsMissEm, line 15: I'm not 100% sure if you can do client authentication with just the username, e.g.,

```
Authorization: Basic Base64Encode(pPqlv1hayMY3NXSMnikiqMLw3G2tyibuRex2HHo3aVE:)
```

My reading of the spec is that without the password this mechanism wouldn't match? Though, previously the code would have accepted that as a value authentication, and it'd return credentials without the `client_secret`, so maybe just the `client_id` (username) is necessary?

ThisIsMissEm: I've asked about this here: https://mailarchive.ietf.org/arch/msg/oauth/QGJfkCMqN2kVMRMDwvCiF3mZ6r4/

lib/doorkeeper/config.rb
```diff
@@ -70,8 +70,26 @@ def scopes_by_grant_type(hash = {})
...
+            Kernel.warn("[DOORKEEPER] Unknown client_credentials method detected: #{method}")
+          end
+        }.reject(&:nil?)
+
+        if client_authentication.empty?
+          Kernel.warn("[DOORKEEPER] No known client_credentials method detected, ignoring option")
+        else
+          @config.instance_variable_set(:@client_credentials_methods, client_authentication.concat([:none]))
+        end
       end
```
ThisIsMissEm, line 93: This is pretty messy, but we basically have to handle all the different permutations this supported.

lib/doorkeeper/config.rb
```diff
@@ -70,8 +70,30 @@ def scopes_by_grant_type(hash = {})
       # `params` object.
       #
       # @param methods [Array] Define client credentials
+      # @deprecated
       def client_credentials(*methods)
```
ThisIsMissEm, line 74: This is pretty messy because of all the different ways `client_credentials` could be set, and we can only convert known methods across automatically.

We could move the deprecated line to the `else` at the end, and basically say: "here's what you need instead: client_authentication [...]"

lib/doorkeeper/config.rb
```diff
@@ -70,8 +70,30 @@ def scopes_by_grant_type(hash = {})
...
+            if method.respond_to? :call
+              Kernel.warn("[DOORKEEPER] Unknown client_credentials method detected, received callable block")
+            else
+              Kernel.warn("[DOORKEEPER] Unknown client_credentials method detected: #{method}")
+            end
+          end
+        }.reject(&:nil?)
+
+        if client_authentication.empty?
+          Kernel.warn("[DOORKEEPER] No known client_credentials method detected, ignoring option")
```
ThisIsMissEm, line 93:
```suggestion
          Kernel.warn("[DOORKEEPER] No known client_credentials method detected, cannot automatically convert to client_authentication option")
```

lib/doorkeeper/config/validations.rb
```diff
@@ -16,6 +17,16 @@ def validate!
 
       private
 
+      def validate_client_authentication_value
+        return if client_authentication.is_a?(Array)
+
+        ::Rails.logger.warn(
```
ThisIsMissEm, line 23: I can't figure out a way to test this, but this is where supporting a list of arguments for `option` would be a good idea.

spec/lib/oauth/client_authentication/client_secret_post_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    it "doesn't match if the request is missing client_secret" do
+      request = mock_request request_parameters: {
+        client_id: '1234'
+      }
+
+      expect(described_class.matches_request?(request)).to_not be true
+    end
+
+    it "doesn't match if the parameters are in the query parameters" do
```
ThisIsMissEm, line 24: Note: previously this would have passed, and that was incorrect.

.rspec
```diff
@@ -1 +1 @@
---colour
+--colour --format documentation
```
ThisIsMissEm, line 1: This is temporary, but was necessary to help debug the tests — I'm wondering if we can set this based on environment?

spec/lib/oauth/client_authentication/none_spec.rb
ThisIsMissEm: This comment still applies, I just changed the method signature to use keyword args.

ThisIsMissEm:
I think I'm now settled on and happy with this code and the test coverage, but let me know if it needs more.

lib/doorkeeper/server.rb
```diff
@@ -8,6 +8,10 @@ def initialize(context)
       @context = context
     end
 
+    def client_authentication_method_for_request
+      Request.client_authentication_method(context.request)
+    end
+
```
ThisIsMissEm, line 14: Arguably we don't need this, there's no tests for it because it just returns the class, it doesn't instantiate it, because there's nothing the class should need as an instance — all the methods are class methods.

lib/doorkeeper/server.rb
```diff
@@ -37,8 +41,7 @@ def resource_owner
     end
 
     def credentials
-      methods = Doorkeeper.config.client_credentials_methods
-      @credentials ||= OAuth::Client::Credentials.from_request(context.request, *methods)
+      @credentials ||= client_authentication_method_for_request.authenticate(context.request)
```
ThisIsMissEm, line 44: We could just do this instead:

```
      method = Request.client_authentication_method(context.request)
      @credentials ||= method.authenticate(context.request)
```

lib/doorkeeper/request.rb
```diff
@@ -3,6 +3,21 @@
 module Doorkeeper
   module Request
     class << self
+      def client_authentication_method(request)
+        # TODO: Should we support theoretically more than one method matching a
+        # request and then check each for authentication? Currently we only
+        # check the first that matches the request
```
ThisIsMissEm, line 9: Currently we only return the first matching method for a request, not all potentially matching methods. I'm not sure if we should support that?

lib/doorkeeper/oauth/client_authentication/client_secret_post.rb
```diff
@@ -0,0 +1,23 @@
...
+
+module Doorkeeper
+  module OAuth
+    module ClientAuthentication
+      class ClientSecretPost
+        def self.matches_request?(request)
+          request.method.upcase === "POST" && request.request_parameters[:client_id].present? && request.request_parameters[:client_secret].present?
+        end
+
+        def self.authenticate(request)
```
ThisIsMissEm, line 11: I'm not sure if `authenticate` is really the method name that's correct here.. maybe more `credentials` or `parse` or something? Maybe `call` or `handle` even?

ThisIsMissEm: Could also call it `from_request` or something? Given how it's called?

lib/doorkeeper/oauth/client_authentication/client_secret_basic.rb
ThisIsMissEm: Consensus in the mailing list seems to imply that these methods only support POST requests, and require the `client_secret` to be present.

lib/doorkeeper/config.rb
ThisIsMissEm: This also now handles the custom authenticator case, where the `client_credentials` was a lambda or class.

lib/doorkeeper/config.rb
```diff
@@ -579,8 +610,23 @@ def pkce_code_challenge_methods_supported
...
+      return @client_authentication_methods if defined?(@client_authentication_methods)
+      
+      methods = if instance_variable_defined?("@client_credentials_methods")
+        if instance_variable_defined?("@client_authentication")
+          Kernel.warn("[DOORKEEPER] Both client_credentials and client_authentication are set, using client_authentication")
+          client_authentication
+        else
+          instance_variable_get("@client_credentials_methods")
+        end
+      else
```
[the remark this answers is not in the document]
ThisIsMissEm: I think this would depend on the order in which these were registered — we really more want a hook that's like "validations to run after configuration", which may be the `Doorkeeper::Config::Validations` class.

nbulaj:
Wow so many comments and changes :smiley:  I'll need some time & :beer:  to check everything. Thanks anyway, I'll get back to it ASAP (sorry have busy days currently)

lib/doorkeeper/request.rb
ThisIsMissEm: Yeah, this is actually a bug. We need to collect an array of potential client credentials, and then check if we:
- had zero, then return an error when client authentication is required (`invalid_client`, 401)
- had more than one, return an error (`invalid_request`, 400, https://www.rfc-editor.org/rfc/rfc6749.html#section-2.3 but the error type comes from section 5.2)

I discovered this whilst implementing client authentication mechanisms in Hollo (an activitypub server)

spec/support/helpers/url_helper.rb
nbulaj: Well I believe these are legacy helpers and tests which are pass params via the url params. So if we can refactor them - will be great, thanks!

ThisIsMissEm: @nbulaj yeah, I've gone ahead and fixed how we do all the requests in https://github.com/doorkeeper-gem/doorkeeper/commit/93b2d41e77dc1a53608ef8eb7047e073535c6022

It was a fairly significant change, but we can resolve this conversation point.

ThisIsMissEm:
```suggestion
```

This FIXME is resolved.

spec/requests/endpoints/token_spec.rb
nbulaj: `token_endpoint_url` is a spec helper which has this:

```ruby
 grant_type: options[:grant_type] || "authorization_code",
```

So I believe with `token_endpoint_params` you just override the params from the URL generated by the helper

lib/doorkeeper/oauth/client.rb
```diff
@@ -17,6 +17,8 @@ def self.find(uid, method = Doorkeeper.config.application_model.method(:by_uid))
         new(application)
       end
 
+      # TODO: Figure out a way to have this just get the client but not assert
+      # authentication if not secret
```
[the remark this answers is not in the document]
ThisIsMissEm: Perhaps we could do it in a non-breaking way? I'm not sure.

For openid_connect, I'd expect it'd need to write their own client authentication methods, which they currently don't have. (i.e., they haven't been able to implement private_key_jwt or similar for client authentication — they just do stuff with ID tokens and amending responses to use JWTs

kwent:
Would love to see this land :)

nbulaj:
Also wanna to return to this great PR, there a lot of changes and open questions, but I wanna to have it merged :eyes: :rocket:

55728:
Last night, I read the PR with some 🥛 & have some suggestions 😊

As a collaborator on doorkeeper-openid_connect, I wanted to chime in here since this directly affects what we can build on the extension side.

I really like the direction of separating match & extract from authenticate (as you mentioned in your comment on `client_secret_post.rb`) — that two-phase design would be exactly what we'd need to implement `private_key_jwt` and `client_secret_jwt` in doorkeeper-openid_connect.

For doorkeeper-openid_connect, the ideal contract would be something like: a registered method can resolve and verify the client itself, rather than always going through `by_uid_and_secret`. That way asymmetric key methods (especially `private_key_jwt`) would fit naturally into the registry.

I also grepped the openid_connect side for compatibility:

**Credentials relocation**: doorkeeper-openid_connect has zero references to `Doorkeeper::OAuth::Client::Credentials`, so moving it doesn't affect us.

**The actual breaking point** is the config reader rename. doorkeeper-openid_connect relies on `Doorkeeper.config.client_credentials_methods` in the `token_endpoint_auth_methods_supported` discovery metadata ([here](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/master/lib/doorkeeper/openid_connect/token_endpoint_auth_methods_supported_mixin.rb#L12)). This PR removes that reader (renamed to `client_authentication_methods`) without an alias, so it would raise `NoMethodError` on the openid_connect side. The gemspec pins `doorkeeper < 6.0`, so we won't break silently — but we'll need to update when this lands.

**The upside**: since the registry exposes RFC 8414-compliant names directly (`client_secret_basic`, etc.), we can actually delete the internal `from_basic` → `client_secret_basic` mapping and just do `client_authentication_methods.map { |m| m.name.to_s }`. And any method an extension registers (e.g. `private_key_jwt`) would flow into discovery automatically — exactly the synergy this PR enables. 🎉

Happy to help validate the interface from the openid_connect side as this evolves. @nbulaj mentioned wanting that compatibility check, so consider this an open offer 👀✨

ThisIsMissEm:
@55728 it'd actually be really helpful to have someone take over this pull request; as I'm no longer contributing to mastodon, I'm not a user of doorkeeper, which is largely why this has stalled out.

I would also like to highlight https://github.com/doorkeeper-gem/doorkeeper/pull/1772#discussion_r2201182825 to you, as it is a pretty significant point that I'd previously missed (I only picked it up when helping a client build [oauth4swift](https://github.com/germ-network/oauth4swift/) and working on another OAuth project)

ThisIsMissEm:
@55728 oh, also of interest to you is probably #1587 / https://github.com/doorkeeper-gem/doorkeeper/pull/1737 which is effectively the same as oidc's metadata endpoint (and it's designed for you to be able to adopt iirc): https://github.com/doorkeeper-gem/doorkeeper/pull/1737

55728:
@ThisIsMissEm Happy to take this on 😊 I'll open a fresh PR from my fork (building on your commits as the base) since I can't push to your branch directly.

I've also been looking at #1737 — happy to pick that up too 👀✨

Thanks for all the groundwork here, and for the hand-off!

55728:
@ThisIsMissEm Would you be okay with me picking up #1773 (force PKCE for all clients) as well? 👀✨

55728:
> [!TIP]
> The #1772 take-over is split into three reviewable pieces — [1/3] #1839, [2/3] #1840, [3/3] #1841. #1839 is test-only and can be reviewed/merged independently.

@nbulaj Would appreciate your review when you get a chance — starting with #1839 would be easiest 👀✨

nbulaj:
Gonna close this one since we're sequentially work on if in multiple new PRs. Huge thanks to all participants :bow:

PR #1773, opened 2025-04-23
Ensure force PKCE always forces PKCE, fixes #1654

### LINKED ISSUE #1654: Option to enforce that clients use PKCE

ThisIsMissEm:
### Summary

Previously `force_pkce` option could be bypassed if the client was confidential, that's no longer the case with the [Security Practices BCP](https://www.rfc-editor.org/rfc/rfc9700), which recommends PKCE for all clients.

Fixes #1654 

**This is technically a breaking change**

### Other Information

We could perhaps simplify the tests a little, but I'm not sure that's necessary right now.

### DIFF
--- a/lib/doorkeeper/oauth/authorization_code_request.rb
+++ b/lib/doorkeeper/oauth/authorization_code_request.rb
@@ -63,7 +63,7 @@ module Doorkeeper
         @missing_param =
           if grant&.uses_pkce? && code_verifier.blank?
             :code_verifier
-          elsif client && !client.confidential && Doorkeeper.config.force_pkce? && code_verifier.blank?
+          elsif Doorkeeper.config.force_pkce? && code_verifier.blank?
             :code_verifier
           elsif redirect_uri.blank?
             :redirect_uri
--- a/lib/doorkeeper/oauth/pre_authorization.rb
+++ b/lib/doorkeeper/oauth/pre_authorization.rb
@@ -147,7 +147,6 @@ module Doorkeeper
 
       def validate_code_challenge
         return true unless Doorkeeper.config.force_pkce?
-        return true if client.confidential
         return true if code_challenge.present?
 
         @invalid_request_reason = :invalid_code_challenge
--- a/spec/lib/oauth/authorization_code_request_spec.rb
+++ b/spec/lib/oauth/authorization_code_request_spec.rb
@@ -176,10 +176,10 @@ RSpec.describe Doorkeeper::OAuth::AuthorizationCodeRequest do
       end
 
       context "when the app is confidential" do
-        it "issues a new token for the client" do
+        it "does not issue a new token" do
           expect do
             request.authorize
-          end.to change { client.reload.access_tokens.count }.by(1)
+          end.not_to change { client.reload.access_tokens.count }
         end
       end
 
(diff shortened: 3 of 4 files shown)


nbulaj (approved):
LGTM, thanks

ThisIsMissEm:
> how does this affect implementations currently forcing public apps to use pkce, but not confidential ones? unless i'm misunderstanding, this is a breaking change i'd have to pass on to third-party integrators, or disable the `force_pkce` option entirely.

@gkemmey  This would be a breaking change, in which case, I'd recommend disabling `force_pkce` and then introducing it again to your ecosystem. I'd be hesitant to add a block invocation for `force_pkce` when it should really be an "on" or "off" option. Keep in mind, that the OAuth 2.1 and OAuth Security Current Best Practices _strongly_ recommend PKCE for all authorization code requests due to the volume of security issues that PKCE fixes.

Keep in mind this would not affect client credentials grant flow, only authorization code grant flow.

molawson:
> I'd be hesitant to add a block invocation for force_pkce when it should really be an "on" or "off" option.

@ThisIsMissEm Can you say a little more about this perspective?  While I love the idea of moving more providers and clients in the direction of the most secure methods of operation available, having some in-between functionality (i.e. a "partially on" state as the code works prior to this change) seems hugely beneficial for enabling apps to move in this more secure direction incrementally.  I understand that's not ideal and makes for more code to maintain (I love the simplicity of this change!), but there seems like a reasonable amount of upside to that incremental/layered approach.

ThisIsMissEm:
If anything, we'd add a `require_pkce_for` block, but `force_pkce` should always be all clients or none, you can't partially "force" something.

nbulaj:
Hey all. Thanks for the great insights! This indeed can affect existing projects & customers so I think if we need to add some warning message? Or at least a changelog entry with a `breaking` mark? And release it as a part of 6.x version. WDYT? We had some similar changes before and indeed reports were created.

paulRbr:
Thanks both of you @ThisIsMissEm and @gkemmey how can we help to make this change go forward? I've done a quick review and didn't see anything off (but I'm new to the gem so I'm trusting @nbulaj 's approval of April 2025 😊 )

ThisIsMissEm:
I started merging in that change, but the test coverage here doesn't make sense to me right now. see: https://github.com/ThisIsMissEm/doorkeeper/pull/1/changes#r2750112774

ThisIsMissEm:
@55728 yeah, you can take over this if you want, however, note: OAuth 2.1 requires PKCE.

Also note the comment history here and https://github.com/doorkeeper-gem/doorkeeper/pull/1773#issuecomment-3586970000

55728:
@ThisIsMissEm Thanks for the heads-up!

55728:
Closing in favor of #1842, which has been merged 🎉

#1842 is based on @ThisIsMissEm's work here — the only behavioral difference is that it preserves the `client &&` guard in `AuthorizationCodeRequest` to avoid regressing #1762 (missing app → `InvalidClient` instead of `ArgumentError`).

Thank you @ThisIsMissEm for the original work and the RFC 9700 reference — it made the follow-up straightforward 😊

PR #1775, opened 2025-05-27
Fix not null migration generator

thatsanicehat:
### Summary

The generator added in https://github.com/doorkeeper-gem/doorkeeper/pull/1727 wasn't named consistently with the other generators, and the method within the generator was misnamed, so it couldn't run. This normalizes the naming schema of the generator and allows it to function.


nbulaj:
> Should I be adding a CHANGELOG entry for this?

Yep, let's add it to be clear. And squash the commits. Thanks!

PR #1778, opened 2025-08-18
fix: ensure that token revocation is idempotent

hannahyeates:
### Summary

We were debugging an issue related to how we lock the access tokens table when we refresh an access token, and in this process, we found that revoking an access token is not idempotent. In the existing implementation, the `revoked_at` timestamp will continue to be updated each time `revoke!` is called on the same access token. Since a token can only be revoked once, this should be enforced in code.


nbulaj (approved):
LGTM :white_check_mark:

PR #1779, opened 2025-08-18
fix: only lock previous access token model if revoke_previous_refresh_token_on_use is false

hannahyeates:
This PR resolves lock contention issues on the OAuth access tokens table when creating new access tokens from refresh tokens.

## Problem
We were experiencing database lock contention on our `access_tokens` table during token refresh operations because it was trying to acquire a lock while there was another process that had also acquired a lock on the table. While resolving this, we noticed that was that the default behaviour was to acquire a lock on the access token record, even when the revoke_previous_access_token_on_use configuration was set to true.

## Solution
When revoke_previous_access_token_on_use is enabled, we don't need to revoke the previous access token, which means we can safely allow multiple access tokens to exist simultaneously. This eliminates the need for table locking during token creation since there's no race condition to prevent.

This change formalizes the behavior where multiple access tokens can coexist when the revocation setting is disabled, removing unnecessary locking that was causing performance bottlenecks.

## Changes
Remove table locking during access token creation when revoke_previous_access_token_on_use is true
Allow concurrent token refresh operations without acquiring a lock
Maintain existing behavior for configurations where token revocation is required

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ User-visible changes worth mentioning.
 
 Add your entry here.
 - [#1775] Fix Applications Secret Not Null Constraint generator
-- [#1779] only lock previous access token model when creating a new token from its refresh token if revoke_previous_refresh_token_on_use is false
+- [#1779] Only lock previous access token model when creating a new token from its refresh token if revoke_previous_refresh_token_on_use is false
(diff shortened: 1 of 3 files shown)


hannahyeates:
@nbulaj Thanks for your review! Can you please merge this for me? I don't seem to have permissions to do so.

PR #1786, opened 2025-12-08
Remove obsolete post_install_message in gemspec

francois-ferrandis:
Hello there!

First of all, thank you for maintaining this great gem, it has been very helpful to out team.

I am opening this PR because I believe this post install message has been out there for enough time now for devs to notice it. This how it looks from my end: my team adopted this gem last year and this warning about a new default behavior introduced years ago does not provide any valuable information.

So maybe the time has come to remove that warning, and maybe release 6.0.0 to invite devs to checkout the changelog.

I could be missing some info, of course, so this need confirmation from a maintainer.

Thanks! :slightly_smiling_face:

### DIFF
--- a/doorkeeper.gemspec
+++ b/doorkeeper.gemspec
@@ -34,18 +34,6 @@ Gem::Specification.new do |gem|
   gem.add_dependency "railties", ">= 5"
   gem.required_ruby_version = ">= 2.7"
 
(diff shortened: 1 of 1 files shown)


nbulaj (approved):
Agree, time passed :slightly_smiling_face: We can remove it now, thanks!

PR #1788, opened 2026-01-09
fix: changes regex for basic auth to be case-insensitive and adds tests for mixed case prefixes

caiubi:
### Summary

The [RFC](https://www.rfc-editor.org/rfc/rfc7617.html) for `basic` authentication specifies that the "scheme" should be matched case-insensitively, so I changed the regex and added some specs.

### Other Information

N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ User-visible changes worth mentioning.
 ## main
 
 Add your entry here.
+- [#1788] Fix regex for basic auth to be case-insensitive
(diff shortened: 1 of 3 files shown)


nbulaj (approved):
I checked [OAuth RFC](https://datatracker.ietf.org/doc/html/rfc6749) which refers [HTTP Authentication RFC](https://datatracker.ietf.org/doc/html/rfc2617) and it tells about case-insensitive auth token as well. So LGTM :+1:

Issue #1688, opened 2024-01-25
Cleanup job removes not expired tokens


jensljungblad:
This is a follow-up to https://github.com/doorkeeper-gem/doorkeeper/issues/1612 which was closed by https://github.com/doorkeeper-gem/doorkeeper/pull/1625.

The PR fixed half of the problem. The issue is that the rake task `doorkeeper:db:cleanup:expired_tokens` will still delete tokens which haven't yet expired. The reason is that the `expires_in` column is not considered.

Compare the logic in 
```ruby
class StaleRecordsCleaner
  # This method compares against `created_at`
  def clean_expired(ttl)
    table = @base_scope.arel_table
  
    @base_scope
      .where.not(expires_in: nil)
      .where(table[:created_at].lt(Time.current - ttl))
      .in_batches(&:delete_all)
  end
end
```
With
```ruby
module Expirable
  # This method compares against `expires_at`
  def expired?
    !!(expires_in && Time.now.utc > expires_at)
  end
end
```

fredplante:
Hello @jensljungblad 

I don't think there is an issue with this, could you provide a failing test that proves your point?

In `Expirable`, `expires_at` is computed as `created_at + expires_in`.
In `StaleRecordsCleaner#clean_expired`, `ttl` is basically the same as `expires_in`.

so we have:

`created_at < (Time.current - expires_in)`
VS
`Time.now.utc > (created_at + expires_in)`

And these 2 statements are equivalent. (I only have a doubt about `Time.current` vs `Time.now.utc`, I guess it's fine as long as you have UTC configured as a time zone)

Floppy:
I've just run into this; the thing is that `ActiveRecord::StaleRecordsCleaner` uses a single TTL argument to work out if a key should be removed.

https://github.com/doorkeeper-gem/doorkeeper/blob/main/lib/doorkeeper/orm/active_record/stale_records_cleaner.rb#L25

Line 30 in this method simply compares the creation date plus that TTL to now, and removed everything that falls before that window.

https://github.com/doorkeeper-gem/doorkeeper/blob/main/lib/doorkeeper/orm/active_record/stale_records_cleaner.rb#L30

In the rake task, this TTL is set to the default expiry: 

https://github.com/doorkeeper-gem/doorkeeper/blob/e4a7b6d66faeb2c61f0a7158887f0f8210716537/lib/doorkeeper/rake/db.rake#L24

[...]

PR #1792, opened 2026-02-11
Fixes/check for expires in when cleaning

Paultdx:
Presumably fixes https://github.com/doorkeeper-gem/doorkeeper/issues/1688.
Consider expires_in when comparing with Time.current.

Original pr with conversation: https://github.com/doorkeeper-gem/doorkeeper/pull/1690


nbulaj (approved):
Thanks!

Issue #1583, opened 2022-08-07
uninitialized constant Doorkeeper::Application while booting


masterkain:
I have a rails 6.1 app, with factory_bot, doorkeeper and a factory such as

```
FactoryBot.define do
  factory :application, class: 'Doorkeeper::Application' do
    name { 'oauth client app' }
    redirect_uri { 'https://oauthclient.dev/users/auth/account/callback' }
  end
end
```

my doorkeeper config has `enable_application_owner confirmation: true`, I already tried to put `require 'doorkeeper/orm/active_record/application'` atop the file with the same result.

when I load the test environment through guard I get, even if the specified factory class is a string:

```
rake aborted!
NameError: uninitialized constant Doorkeeper::Application
Did you mean?  ApplicationJob
/bundle/vendor/ruby/3.0.0/gems/activesupport-6.1.6.1/lib/active_support/inflector/methods.rb:288:in `const_get'
/bundle/vendor/ruby/3.0.0/gems/activesupport-6.1.6.1/lib/active_support/inflector/methods.rb:288:in `block in constantize'
/bundle/vendor/ruby/3.0.0/gems/activesupport-6.1.6.1/lib/active_support/inflector/methods.rb:284:in `each'
/bundle/vendor/ruby/3.0.0/gems/activesupport-6.1.6.1/lib/active_support/inflector/methods.rb:284:in `inject'
/bundle/vendor/ruby/3.0.0/gems/activesupport-6.1.6.1/lib/active_support/inflector/methods.rb:284:in `constantize'
/bundle/vendor/ruby/3.0.0/gems/activesupport-6.1.6.1/lib/active_support/core_ext/string/inflections.rb:74:in `constantize'
/bundle/vendor/ruby/3.0.0/gems/doorkeeper-5.5.4/lib/doorkeeper/config.rb:465:in `application_model'
/bundle/vendor/ruby/3.0.0/gems/doorkeeper-5.5.4/lib/doorkeeper/orm/active_record.rb:35:in `block in initialize_application_owner!'
/bundle/vendor/ruby/3.0.0/gems/activesupport-6.1.6.1/lib/active_support/lazy_load_hooks.rb:71:in `class_eval'
/bundle/vendor/ruby/3.0.0/gems/activesupport-6.1.6.1/lib/active_support/lazy_load_hooks.rb:71:in `block in execute_hook'
[...]

Issue #1319, opened 2019-10-17
NameError: uninitialized constant Doorkeeper::AccessToken on rake doorkeeper:db:cleanup


j-arnaiz:
### Steps to reproduce
What we need to do to see your problem or bug?
On RakeFile:
```
# frozen_string_literal: true

# Add your own tasks in files placed in lib/tasks ending in .rake,
# for example lib/tasks/capistrano.rake, and they will automatically be available to Rake.

require_relative 'config/application'

Rails.application.load_tasks
Doorkeeper::Rake.load_tasks
```
Doorkeeper installed and configured with Rails 5.2.3

```
### Actual behavior
rake doorkeeper:db:cleanup fails with message
rake aborted!
NameError: uninitialized constant Doorkeeper::AccessToken
/usr/local/bundle/gems/bootsnap-1.4.4/lib/bootsnap/load_path_cache/core_ext/active_support.rb:79:in `block in load_missing_constant'
/usr/local/bundle/gems/bootsnap-1.4.4/lib/bootsnap/load_path_cache/core_ext/active_support.rb:8:in `without_bootsnap_cache'
/usr/local/bundle/gems/bootsnap-1.4.4/lib/bootsnap/load_path_cache/core_ext/active_support.rb:79:in `rescue in load_missing_constant'
/usr/local/bundle/gems/bootsnap-1.4.4/lib/bootsnap/load_path_cache/core_ext/active_support.rb:58:in `load_missing_constant'
/usr/local/bundle/gems/doorkeeper-5.1.0/lib/doorkeeper/rake/db.rake:16:in `block (4 levels) in <main>'
/usr/local/bundle/gems/rake-12.3.2/exe/rake:27:in `<top (required)>'
```

### System configuration
Ruby 2.6.3 with mysql and ActiveRecord

**Doorkeeper initializer**:
Doorkeeper is working fine in endpoints and even on rails console. But rake cleanup task is not working.

**Ruby version**: ``
2.6.3

**Gemfile.lock**:

<details>
  <summary>Gemfile.lock content</summary>
  
```
GEM
  remote: https://rubygems.org/
  specs:
    actioncable (5.2.3)
      actionpack (= 5.2.3)
      nio4r (~> 2.0)
      websocket-driver (>= 0.6.1)
    actionmailer (5.2.3)
      actionpack (= 5.2.3)
      actionview (= 5.2.3)
      activejob (= 5.2.3)
[...]

Issue #1281, opened 2019-06-27
Doorkeeper models not available in rake "environment"


thegcat:
### Steps to reproduce

We use a rake task to create Doorkeeper Application objects, however it seems the availability of Doorkeeper models in rake tasks is broken for us.

The rake task https://stackoverflow.com/questions/tagged/doorkeeper
Relevant section in the README of our app: https://github.com/kif-ev/oskiosk-server#create-a-new-oauth-token

This used to work, I however haven't been able to find if the issue was with Doorkeeper or Rails by downgrading minor versions of each.

I have also tried adding `Doorkeeper::Rake.load_tasks` to the `Rakefile` and running the Doorkeeper rake tasks, it also fails to load the Doorkeeper models.

### Expected behavior

Doorkeeper models should be available in Rake tasks when the rake task depends on the `environment` task.

### Actual behavior

They aren't.

### System configuration

Mostly see https://github.com/kif-ev/oskiosk-server

nbulaj:
Hi @thegcat . Looks like your rake task doesn't load ActiveRecord by some reason and Doorkeeper models isn't loaded too as they use `ActiveSupport#on_load` lazy loading hooks. Try to call `ActiveRecord::Base` somewhere in rake task before calling Doorkeeper models and if it started top work - that's the case. I don't sure why is such happening, there we need some debugging to check why Rails doesn't load ActiveRecord for you case

thegcat:
Thank you for the suggestion, this works indeed!

As to why this doesn't happen automatically: I can't remember exactly but I might have created this project way back when as a RAILS API project, do you know if this might have an incidence in this case?

nbulaj:
Also need to check it out, don't sure about it.

nbulaj:
You can add `require 'active_record/base'` on top of your Rake task to call AR hooks that will invoke Doorkeeper model initialization. I just don't sure why it wasn't called if you use `task: :environment` :thinking: , maybe autoloading in development env or something like that.

thegcat:
Thanks, that does the trick. As this is probably not directly related to doorkeeper and this fixes our problem I'll just close this here.

PR #1807, opened 2026-03-27
Add reproduction tests for on_load deferral (issue #1703)

### LINKED ISSUE #913: Loading ActiveRecord (models) in intializer leads to settings conflicts with Rails
See this issue: https://github.com/rails/rails/issues/23589

Using doorkeeper will cause ActiveRecord to be loaded before Rails' own `new_framework_defaults.rb`. It seems that it's not 'safe' to load AR models in an initializer and that such things should be deferred until later using `ActiveSupport.on_load(:active_record)`.

55728:
## Summary

Follow-up to #1804. Adds regression tests that explicitly verify the deferral behavior introduced in 511a7806.

- **`does not call config model accessors at registration time`** -- Ensures `application_model` / `access_token_model` / `access_grant_model` are NOT called when `initialize_configured_associations` is invoked (i.e., `constantize` does not trigger early AR loading).
- **`calls config model accessors only when the on_load hook fires`** -- Captures the deferred block and verifies model accessors are called only when the block executes (simulating AR finishing initialization).

## Background: History of `on_load` in Doorkeeper

| Date | Commit | Author | What happened |
|---|---|---|---|
| 2018-02 | `9170e5b4` | Nikita Bulaj | **Introduced** `on_load(:active_record)` to defer model `require` calls (Fix #913) |
| 2021-03 | `6db479cb` | Nikita Bulaj | **Removed** `on_load` entirely, replaced with Ruby `autoload` |
| 2023-01 | `53d70a89` | Nikita Bulaj | Added `initialize_configured_associations` with eager `constantize` -- **reintroduced early loading problem** |
| 2026-03 | `511a7806` | Kenta Ishizaki | **Re-introduced** `on_load`, but only for module `include` calls (not model definitions) |

The key difference from the 2018 approach: model classes remain immediately available via `autoload`. Only the `include` of `Ownership` and `PolymorphicResourceOwner` modules is deferred. This avoids the past issues (#1319, #1281, #1117, #1583, #1043) while fixing the early AR loading reported in #1703.

Reproduction app by @ngan: https://github.com/ngan/doorkeeper-activerecord-load-issue

## Test plan

- [x] `bundle exec rspec spec/lib/doorkeeper/orm/active_record_spec.rb` passes (7 examples, 0 failures)
- [x] Confirm tests fail if the `on_load` wrapper is removed from `initialize_configured_associations`
- [x] Full CI green

### DIFF
--- a/spec/lib/doorkeeper/orm/active_record_spec.rb
+++ b/spec/lib/doorkeeper/orm/active_record_spec.rb
@@ -11,6 +11,49 @@ if DOORKEEPER_ORM == :active_record
       end
     end
 
+    # Reproduction test for https://github.com/doorkeeper-gem/doorkeeper/issues/1703
+    #
+    # Without the on_load wrapper, calling initialize_configured_associations
+    # eagerly triggers constantize on model class names (via access_token_model,
+    # access_grant_model, etc.), which forces ActiveRecord to load before
+    # config.active_record.* settings have been applied. This test verifies
+    # that the on_load block does NOT execute eagerly at registration time.
+    #
+    # See also: https://github.com/ngan/doorkeeper-activerecord-load-issue
+    describe "deferral of model loading (issue #1703)" do
+      it "does not call config model accessors at registration time" do
+        # Stub on_load to capture the block WITHOUT executing it,
+        # simulating the state before ActiveRecord is fully initialized.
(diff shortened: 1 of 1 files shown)


nbulaj (approved):
Nice! Let's include it, thanks 🙇

PR #1809, opened 2026-04-01
Remove hardcoded colons from views, move punctuation to i18n

### LINKED ISSUE #1784: Typo: double colon on default authorization page
Heya

We noticed a small textual hiccup in the default template for granting an authorization.

In the file `app/views/doorkeeper/authorizations/new.html.erb` there is a translated string

```erb      
<p><%= t('.able_to') %>:</p>
```

There is a colon after the text at the end of the `<p>` tag.

The translation itself also includes a colon. To me this makes sense, since not all languages handle colons the same. Notably in French there should be a whitespace before the colon. (Which is currently missing in the french translation for this case.)

```yml
---
de:
  doorkeeper:
    authorizations:
      new:
        able_to: 'Diese Anwendung erhält folgende Berechtigungen:'
```

This results in a double colon.

I suggest removing the colon in the view template (and adding it to the translations in all locales).

55728:
### Summary

Remove all hardcoded `:` from view templates and move punctuation into translation strings, as discussed in #1784.

### Problem

The view templates had hardcoded colons after `t()` calls, while some translations (`de`, `ko`, `ru`) already included a colon in their strings. This resulted in a double colon (`::`) being displayed for those locales. Hardcoded colons also prevent languages with different punctuation rules (e.g. French ` :`) from following their own conventions.

### Changes

- `authorizations/new.html.erb` — remove hardcoded `:` after `t(.able_to)`
- `applications/show.html.erb` — remove hardcoded `:` after `application_id`, `secret`, `scopes`, `confidential`, `callback_urls`
- `authorizations/show.html.erb` — remove hardcoded `:` after `title`
- Add colons to the default `en.yml` translations so English output remains unchanged
- Add `[IMPORTANT]` upgrade note to `CHANGELOG.md`

### Companion PR

- **doorkeeper-i18n**: doorkeeper-gem/doorkeeper-i18n#73 — adds colons to all affected keys across all 25 locales

Fixes #1784

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ User-visible changes worth mentioning.
 
 ## main
 
+- [#1784] Remove hardcoded colon from authorization view template, move punctuation to translation files
 - [#1781] Honor `handle_auth_errors :raise` in `AuthorizationsController#authorize_response`
 - [#1795] Fix: detailed error 'insufficient_scope' in protected resources 403s
 - [#1797] Fix `doorkeeper:db:cleanup` rake task failure on PostgreSQL
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,6 @@ User-visible changes worth mentioning.
 
 ## main
 
-- [#1784] Remove hardcoded colon from authorization view template, move punctuation to translation files
 - [#1781] Honor `handle_auth_errors :raise` in `AuthorizationsController#authorize_response`
 - [#1795] Fix: detailed error 'insufficient_scope' in protected resources 403s
 - [#1797] Fix `doorkeeper:db:cleanup` rake task failure on PostgreSQL
@@ -18,6 +17,13 @@ User-visible changes worth mentioning.
 - [#1806] Fix token revocation bypass for public clients (RFC 7009)
 - [#1815] Expose `current_resource_owner` as a view helper in `Doorkeeper::ApplicationController`.
 - [#1818] Fix token introspection returning `exp: 0` for non-expiring tokens.
+- [#1784] Remove hardcoded colons from view templates, move punctuation to i18n translation strings.
+
+  **[IMPORTANT]**: if you have customized Doorkeeper views (`authorizations/new`, `authorizations/show`,
+  `applications/show`) or overridden the default `en.yml` translations, you may need to update them.
+  Colons are no longer hardcoded in the views — they are now part of the translation strings.
+  Update the [doorkeeper-i18n](https://github.com/doorkeeper-gem/doorkeeper-i18n) gem to get the
+  updated translations for all locales.
 
 ## 5.9.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ User-visible changes worth mentioning.
 
 ## main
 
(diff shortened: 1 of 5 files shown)


nbulaj:
Hello. Should we also cover more cases?

image

nbulaj (commented):
Can we please include all the cases with `:` ?

55728:
Thanks @nbulaj! Updated to cover all hardcoded colon cases:

- `applications/show.html.erb` — `application_id`, `secret`, `scopes`, `confidential`, `callback_urls`
- `authorizations/show.html.erb` — `title`

The companion doorkeeper-i18n PR (doorkeeper-gem/doorkeeper-i18n#73) has also been updated with colons for all 6 keys across all 25 locales.

nbulaj:
Hm, I'm currently think what will be if somebody wants to reuse some translation in a case when `:` is not required :thinking: LIke `t('.scopes')` .. Introduce their own?

55728:
Good point! Label-style translations like `scopes` or `confidential` could definitely be reused in contexts where a colon isn't appropriate.

Reverted the label keys and kept the colon only in the sentence-style translations (`able_to` and `authorizations.show.title`) where reuse is unlikely. The hardcoded `:` stays in the view for the label-style keys.

55728:
Hi @nbulaj, just rebased to resolve the CHANGELOG conflict 👍
To summarize the current approach based on your feedback: I kept the colon only in sentence-style translations (able_to and authorizations.show.title) where reuse in other contexts is unlikely, and left label-style keys (scopes, confidential, etc.) as pure labels with the colon hardcoded in the view. This way the label keys stay reusable.
Let me know if you'd prefer a different direction — happy to iterate! 😊

nbulaj:
TBH I'm currently not sure that we wanna to do such changes. Project exists >10 years and nobody reported about such issues. Any developer can override views or i18n values in case they need to change smth. But adding PR changes currently can break existing apps UI when apps already included `:` somewhere and now they will have `::` .

55728:
Totally fair to pause — appreciate you weighing it carefully 😊

Either direction works for me. doorkeeper is load-bearing in my stack, and I'm here for the long run as a contributor — happy to keep showing up for the ecosystem alongside you!

nbulaj:
OK or maybe still fix default behavior and make a note in `Upgrading` section (and changelog as well) where we can mention changes in the views & i18n translations :eyes: This should help upgrade smoothly

55728:
Thanks @nbulaj — sounds good! I'll re-apply the full fix (all hardcoded colons → i18n) and add notes to the *Upgrading* section and CHANGELOG documenting the view/translation changes for downstream apps. The companion doorkeeper-i18n PR will be updated accordingly.

nbulaj (approved):
OK let's finish with current implementation and see how it goes 🙂  BTW nobody reported about it for a long so I believe it's not a big deal for developers to override views/i18n. But I wanna to have correct default values as a baseline so we're good to go

I'm gonna to include it in 5.9.1 as I don't see this as a breaking API change. It's easy to manage/adjust i18n translations

55728:
Just a heads-up — the companion i18n change is ready at doorkeeper-gem/doorkeeper-i18n#73 whenever you're ready to review. Would be great to land them together for 5.9.1 👀✨

PR #1815, opened 2026-04-15
Expose current_resource_owner as a view helper

### LINKED ISSUE #1674: Enhancement: expose `current_resource_owner` to views
Just as Devise makes `current_user` [accessible to both the controllers and views](https://github.com/heartcombo/devise/blob/e2242a95f3bb2e68ec0e9a064238ff7af6429545/lib/devise/controllers/helpers.rb#L138) via `helper_method`, it would make sense to make `current_resource_owner` accessible to views as well. 

Currently, when custom views need to access the resource owner, it can only be done via the hacky and undocumented `@current_resource_owner` [memoized variable](https://github.com/doorkeeper-gem/doorkeeper/blob/0aa94c5a82035ec4840785156760a2930e5de27a/lib/doorkeeper/helpers/controller.rb#L18C1-L18C1).

55728:
## Summary

Exposes `current_resource_owner` as a view helper, following the pattern used by Devise for `current_user`.

Previously, custom views needing access to the resource owner had to rely on the undocumented `@current_resource_owner` memoized instance variable. This PR makes the helper officially available in views.

Closes #1674.

## Changes

- Register `current_resource_owner` via `helper_method` in `Doorkeeper::Helpers::Controller`'s `included` hook, guarded by `respond_to?(:helper_method)` so `ApplicationMetalController` is unaffected
- Add specs covering helper registration and end-to-end view rendering
- Add a regression test ensuring the metal controller stays view-helper-free
- Update CHANGELOG

## Notes

Picking this up since the original reporter hasn't had a chance to open a PR.

cc @nbulaj — following your suggestion in the issue thread 🙌

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ User-visible changes worth mentioning.
 - [#1802] Fix `filter_parameters` not applied when `Doorkeeper.configure` is called inside to_prepare.
 - [#1804] Use `ActiveSupport.on_load(:active_record)` in ORM hooks to prevent loading ActiveRecord models too early
 - [#1806] Fix token revocation bypass for public clients (RFC 7009)
+- [#1815] Expose `current_resource_owner` as a view helper in `Doorkeeper::ApplicationController`.
 
 ## 5.9.0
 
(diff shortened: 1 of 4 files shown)


nbulaj (approved):
Thanks, LGTM 👍

PR #1818, opened 2026-04-22
Fix token introspection returning `exp: 0` for non-expiring tokens

### LINKED ISSUE #1817: Token introspection returns `exp: 0` for non-expiring tokens (RFC 7662 violation)
### Summary

When introspecting a token with `expires_in: nil` (non-expiring token), the response includes `"exp": 0`, which represents the Unix epoch (1970-01-01 UTC). Clients interpreting this value will consider the token expired over 50 years ago.

Per [RFC 7662 §2.2](https://datatracker.ietf.org/doc/html/rfc7662#section-2.2), the `exp` field is OPTIONAL. For non-expiring tokens, it should be omitted entirely.

This is not an edge case — non-expiring tokens are created whenever `access_token_expires_in` is set to `nil` in the Doorkeeper configuration, which is a supported and documented option.

### Steps to reproduce

1. Create a non-expiring access token:

```ruby
token = Doorkeeper::AccessToken.create!(
  application: app,
  resource_owner_id: user.id,
  expires_in: nil
)
```

2. Introspect the token:

```
POST /oauth/introspect
Authorization: Bearer <valid_token>
token=<never_expiring_token>
```

3. Response includes:

```json
{
  "active": true,
  "scope": "public",
  "client_id": "abc123",
  "token_type": "Bearer",
  "exp": 0,
  "iat": 1713800000
}
```

### Expected behavior

The `exp` field should be omitted for non-expiring tokens:

```json
{
  "active": true,
  "scope": "public",
  "client_id": "abc123",
  "token_type": "Bearer",
  "iat": 1713800000
}
```

### Root cause

In `lib/doorkeeper/oauth/token_introspection.rb`, `success_response` unconditionally calls `@token.expires_at.to_i`:

```ruby
exp: @token.expires_at.to_i,
```

`Models::Expirable#expires_at` returns `nil` when `expires_in` is `nil`, and `nil.to_i` returns `0`.

### Proposed fix

Conditionally include `exp` only when the token has an expiration:

```ruby
def success_response
  response = {
    active: true,
    scope: @token.scopes_string,
    client_id: @token.try(:application).try(:uid),
    token_type: @token.token_type,
    iat: @token.created_at.to_i,
  }
  response[:exp] = @token.expires_at.to_i if @token.expires_at
  customize_response(response)
end
```

### Environment

- Ruby 4.0.2
- Rails 8.0.5
- Doorkeeper main

55728:
## Summary

When introspecting a token with `expires_in: nil` (non-expiring), the response includes `"exp": 0`, which represents the Unix epoch (1970-01-01 UTC). Clients interpreting this value will consider the token expired over 50 years ago.

Per [RFC 7662 §2.2](https://datatracker.ietf.org/doc/html/rfc7662#section-2.2), the `exp` field is OPTIONAL. For non-expiring tokens, it should be omitted entirely.

## Changes

- **`lib/doorkeeper/oauth/token_introspection.rb`**: Conditionally include `exp` only when `@token.expires_at` is present.
- **`spec/controllers/tokens_controller_spec.rb`**: Add test for introspecting a non-expiring token — verifies `exp` is absent and `iat` is present.

## Before

```json
{
  "active": true,
  "scope": "public",
  "client_id": "abc123",
  "token_type": "Bearer",
  "exp": 0,
  "iat": 1713800000
}
```

## After

```json
{
  "active": true,
  "scope": "public",
  "client_id": "abc123",
  "token_type": "Bearer",
  "iat": 1713800000
}
```

Closes #1817

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ User-visible changes worth mentioning.
 - [#1804] Use `ActiveSupport.on_load(:active_record)` in ORM hooks to prevent loading ActiveRecord models too early
 - [#1806] Fix token revocation bypass for public clients (RFC 7009)
 - [#1815] Expose `current_resource_owner` as a view helper in `Doorkeeper::ApplicationController`.
+- [#1818] Fix token introspection returning `exp: 0` for non-expiring tokens.
 
 ## 5.9.0
 
--- a/lib/doorkeeper/oauth/token_introspection.rb
+++ b/lib/doorkeeper/oauth/token_introspection.rb
@@ -102,14 +102,18 @@ module Doorkeeper
 
       # 2.2. Introspection Response
       def success_response
-        customize_response(
+        response = {
           active: true,
           scope: @token.scopes_string,
           client_id: @token.try(:application).try(:uid),
           token_type: @token.token_type,
-          exp: @token.expires_at.to_i,
           iat: @token.created_at.to_i,
-        )
+        }
+        # `exp` is OPTIONAL per RFC 7662 §2.2; omit it for non-expiring tokens
+        # so clients don't interpret `0` as "expired at 1970-01-01".
+        response[:exp] = @token.expires_at.to_i if @token.expires_at
+
+        customize_response(response)
       end
 
       # If the introspection call is properly authorized but the token is not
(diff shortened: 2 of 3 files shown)


nbulaj (approved):
Makes sense to me, thanks!

BTW for everyone who will check this PR: setting `nil` as an expiration time is allowed, but not recommended in RFC 6749 and RFC 6819 🙂

PR #1829, opened 2026-05-25
Fix re-entrant `on_load(:active_record)` NameError on `rails db:seed` (preserves #1703 fix)

### LINKED ISSUE #1828: doorkeeper 5.9.1 start failing on `rails db:seed` phase
it raises `uninitialized constant ApplicationRecord (NameError)` probably because of https://github.com/doorkeeper-gem/doorkeeper/pull/1804

55728:
Fixes #1828.

## Summary

`Doorkeeper::Orm::ActiveRecord.run_hooks` (introduced in #1804) registers an `ActiveSupport.on_load(:active_record)` callback that calls `constantize` on the configured Doorkeeper model classes. If `ActiveRecord::Base` has not yet been loaded — typical for `rails db:seed` and any flow with `config.eager_load = false` — that callback queues, and then fires later when something *else* triggers the `AR::Base` autoload.

The "something else" is normally the host app referencing its own model:

```
seeds.rb: User.create!(...)
  ↳ autoload User
  ↳ class User < ApplicationRecord
    ↳ autoload ApplicationRecord
    ↳ class ApplicationRecord < ActiveRecord::Base   # ← Ruby resolves superclass FIRST
      ↳ autoload ActiveRecord::Base
      ↳ require "active_record/base"
      ↳ ActiveSupport.run_load_hooks(:active_record, ActiveRecord::Base)   # ← Doorkeeper's queued hook fires HERE
      ↳ Doorkeeper.config.access_grant_model        # constantize "FooGrant"
        ↳ autoload FooGrant
        ↳ class FooGrant < ApplicationRecord        # ApplicationRecord not yet defined (mid-load)
        ↳ NameError: uninitialized constant ApplicationRecord
```

Ruby evaluates the superclass expression before assigning the class constant, so `ApplicationRecord` is still undefined when the AR load hooks fire from inside its own definition.

## Why the obvious fix (force-load AR in `to_prepare`) doesn't work

The first attempt on this branch — touching `::ActiveRecord::Base` at the end of `run_hooks` to flush the queued callback — passes the standalone re-entrancy spec but **re-introduces #1703**.

Confirmed empirically against ngan's reproduction app (https://github.com/ngan/doorkeeper-activerecord-load-issue):

| Variant | #1703 (`configs` in AR `after_initialize`) | #1828 (re-entry) |
| --- | --- | --- |
| pre-#1804 | ❌ `Hash` (regression) | ✓ no bug |
| #1804 alone | ✓ `OrderedOptions` | ❌ re-entry introduced |
| `to_prepare` + AR force-load | ❌ `Hash` (regresses #1804) | ✓ fixed |
| **This PR** | **✓ `OrderedOptions`** | **✓ fixed** |

Root cause: in Rails 7 the finisher order is `:run_prepare_callbacks` → `:finisher_hook`, so `config.to_prepare` runs **before** `config.after_initialize`. Loading `::ActiveRecord::Base` inside `to_prepare` fires AR's own `:set_configs` `on_load(:active_record)` block (which reassigns its local `configs` to a `Hash`) before its `after_initialize` block has applied framework defaults — silently dropping `config.active_record.*` settings from `new_framework_defaults_*.rb`. The two requirements are genuinely opposed at this phase.

## Fix

Keep #1804's `on_load(:active_record)` wrapper inside `run_hooks` unchanged, and add a new engine initializer that drains the queue from `config.after_initialize`, ordered after `active_record.set_configs`:

```ruby
initializer "doorkeeper.orm.flush_active_record_hooks", after: "active_record.set_configs" do
  config.after_initialize do
    ::ActiveRecord::Base if defined?(::ActiveRecord)
  end
end
```

By the time this block fires:

- AR's own `:set_configs` `after_initialize` has already applied framework defaults to AR (#1703 preserved).
- AR's `:set_configs` `on_load` callback has not yet been triggered, because nothing has loaded `AR::Base` yet.
- Touching `::ActiveRecord::Base` autoloads AR in this clean context, running the queued `on_load` callbacks — including Doorkeeper's — before any host-app code (db:seed, controllers, ...) can autoload models and trigger them re-entrantly (#1828 fixed).

The `if defined?(::ActiveRecord)` guard keeps the initializer harmless for Mongoid / Sequel users; the `after:` constraint is silently ignored if `active_record.set_configs` is not registered.

## Tests

- Subprocess regression spec for #1828 in `spec/lib/doorkeeper/orm/active_record_spec.rb` (added in df5291fa), updated to simulate the engine `after_initialize` flush in addition to `run_hooks`. The dummy app pre-loads `ActiveRecord::Base`, so this scenario can only be reproduced in a fresh Ruby subprocess that replays the exact autoload chain (User → ApplicationRecord → ActiveRecord::Base → Doorkeeper hook → FooToken/FooGrant) end-to-end.
- ngan's #1703 reproduction app verified manually: with this PR, `configs` stays `OrderedOptions` in AR's `after_initialize`, matching the no-doorkeeper baseline.
- Full suite: `bundle exec rspec` → 1296 examples, 0 failures.

## Compatibility

- No public API change.
- #1703 fix from #1804 preserved (verified against the original repro).
- All existing STI / `on_load` tests in `spec/lib/doorkeeper/orm/active_record_spec.rb` continue to pass.
- New engine initializer is a no-op for non-AR ORMs.

### DIFF
--- a/lib/doorkeeper/orm/active_record.rb
+++ b/lib/doorkeeper/orm/active_record.rb
@@ -30,6 +30,19 @@ module Doorkeeper
 
       def self.run_hooks
         initialize_configured_associations
+        # Force ActiveRecord::Base to load now so any pending
+        # on_load(:active_record) callbacks (including the one just registered
+        # above) fire in this safe context — typically `config.to_prepare`,
+        # well after `:load_config_initializers`.
+        #
+        # Without this, a queued callback can fire re-entrantly during a host
+        # app autoload chain (e.g. `rails db:seed` evaluating
+        # `class ApplicationRecord < ActiveRecord::Base`) and then constantize
+        # a user-configured Doorkeeper model that inherits from
+        # `ApplicationRecord` while `ApplicationRecord` itself is still
+        # mid-definition — raising `NameError: uninitialized constant
+        # ApplicationRecord` (issue #1828).
+        ::ActiveRecord::Base
       end
 
       def self.initialize_configured_associations
--- a/lib/doorkeeper/orm/active_record.rb
+++ b/lib/doorkeeper/orm/active_record.rb
@@ -30,19 +30,6 @@ module Doorkeeper
 
       def self.run_hooks
         initialize_configured_associations
-        # Force ActiveRecord::Base to load now so any pending
-        # on_load(:active_record) callbacks (including the one just registered
-        # above) fire in this safe context — typically `config.to_prepare`,
-        # well after `:load_config_initializers`.
-        #
-        # Without this, a queued callback can fire re-entrantly during a host
-        # app autoload chain (e.g. `rails db:seed` evaluating
-        # `class ApplicationRecord < ActiveRecord::Base`) and then constantize
-        # a user-configured Doorkeeper model that inherits from
-        # `ApplicationRecord` while `ApplicationRecord` itself is still
-        # mid-definition — raising `NameError: uninitialized constant
-        # ApplicationRecord` (issue #1828).
-        ::ActiveRecord::Base
       end
 
       def self.initialize_configured_associations
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ User-visible changes worth mentioning.
 
 ## main
 
-- Add here
+- [#1829] Fix `NameError: uninitialized constant ApplicationRecord` on `rails db:seed` and other non-eager-loading flows caused by `on_load(:active_record)` firing re-entrantly during `ApplicationRecord` autoload.
 
 ## 5.9.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ User-visible changes worth mentioning.
 
 ## main
 
-- [#1829] Fix `NameError: uninitialized constant ApplicationRecord` on `rails db:seed` and other non-eager-loading flows caused by `on_load(:active_record)` firing re-entrantly during `ApplicationRecord` autoload.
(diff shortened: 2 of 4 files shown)


lib/doorkeeper/orm/active_record.rb
```diff
@@ -30,6 +30,19 @@ module Mixins
...
+        # well after `:load_config_initializers`.
+        #
+        # Without this, a queued callback can fire re-entrantly during a host
+        # app autoload chain (e.g. `rails db:seed` evaluating
+        # `class ApplicationRecord < ActiveRecord::Base`) and then constantize
+        # a user-configured Doorkeeper model that inherits from
+        # `ApplicationRecord` while `ApplicationRecord` itself is still
+        # mid-definition — raising `NameError: uninitialized constant
+        # ApplicationRecord` (issue #1828).
+        ::ActiveRecord::Base
```
nbulaj: We thought about such solution in some previous PRs, but it looks ... hacky? I mean calling AciveRecord manually to load it looks like we have some wrong setup somewhere 👀  Which I haven't found yet

55728: > We thought about such solution in some previous PRs, but it looks ... hacky? I mean calling ActiveRecord manually to load it looks like we have some wrong setup somewhere 👀 Which I haven't found yet

Thanks for the pointer 😊 — you were right. Ran @ngan's repro app for #1703 against this branch and confirmed **the force-load was silently re-introducing #1703** (`to_prepare` runs before `after_initialize` in Rails 7, so loading AR there fires `:set_configs`'s `on_load` before its `after_initialize` applies framework defaults).

Pushed a second commit that moves the flush into `config.after_initialize` (ordered after `active_record.set_configs`) in `engine.rb`, keeping #1804's `on_load` wrapper in `run_hooks` unchanged. Updated the PR description with the full analysis and a comparison table.

nbulaj:
Do we still need this when have https://github.com/doorkeeper-gem/doorkeeper/pull/1830 ? :eyes:

55728:
Nope — #1830 supersedes this. Closing 👋

PR #1832, opened 2026-06-08
Gate `belongs_to :owner` on `enable_application_owner?` at include time

### LINKED ISSUE #1831: Unconditional inclusion ::Doorkeeper::Models::Ownership has confusing side effect
Doorkeeper v5.9.2 is the first release to include commit ab58c37553fc399e80e8cf83a76db5323235b0f8.
The choice to include `Ownership` unconditionally in `Application`

https://github.com/doorkeeper-gem/doorkeeper/commit/ab58c37553fc399e80e8cf83a76db5323235b0f8#diff-777f43ffabf1b2b30465eb1d6f1343e7406d4e8cdaca45c5ea58787dd01dfceeR12-R18

has consequences which may confuse some users:

```ruby
> RUBY_VERSION
=> "4.0.5"
> ActiveRecord::VERSION::STRING
=> "8.1.3"
> Doorkeeper::VERSION::STRING
=> "5.9.2"

# There are no Ownership database columns
> Doorkeeper::Application.attribute_names.sort
=> ["blocked", "confidential", "created_at", "id", "mobile_app_id", "name", "redirect_uri", "scopes", "secret", "type", "uid", "updated_at"]

# But the model has the owner association
> Doorkeeper::Application.reflect_on_all_associations.map(&:name).sort
=> [:access_grants, :access_tokens, :authorized_applications, :authorized_tokens, :owner]
> Doorkeeper::Application.first!.owner
  Doorkeeper::Application Load (1.8ms)  SELECT "oauth_applications".* FROM "oauth_applications" ORDER BY "oauth_applications"."id" ASC LIMIT 1
=> nil

# But if your code actually reflects on the association
> Doorkeeper::Application.reflect_on_association(:owner).join_foreign_type
=> "owner_type"
> Doorkeeper::Application.first!.owner_type
  Doorkeeper::Application Load (1.2ms)  SELECT "oauth_applications".* FROM "oauth_applications" ORDER BY "oauth_applications"."id" ASC LIMIT 1
:in '<main>': undefined method 'owner_type' for an instance of Doorkeeper::Application (NoMethodError)
Did you mean?  owner_changed?
```

I personally can work around this, but others may be confused - and most likely not now but some time down the line when it's way harder to remember this implementation choice. So if the maintainers decide to keep it this way, this issue will serve as AI fodder for those who seek answers.

55728:
### Problem

Since #1830 (ab58c37), `Doorkeeper::Models::Ownership` is included unconditionally from `Mixins::Application`'s `included` block. This means `belongs_to :owner` is always declared — even when `enable_application_owner` is not set and the schema lacks the `owner_id`/`owner_type` columns. The `:owner` reflection is visible, but calling `owner_type` raises `NoMethodError`, which is confusing.

### Root cause

The original commit chose unconditional inclusion to avoid "freezing" the gate at first-load time. In practice, the real constraint was that the test suite exercised the enabled behavior by reconfiguring `Doorkeeper.configure { enable_application_owner }` at runtime and calling `Doorkeeper.run_orm_hooks` — which is now a no-op. Since `ActiveSupport::Concern`'s `included` block runs only once, the association could never be retro-fitted onto an already-loaded model, so the tests relied on it being always present.

Note: gating the `:owner` association on the flag also restores the pre-#1830 behavior — the old `on_load(:active_record)` path included `Ownership` conditionally (`if enable_application_owner?`), so OFF-by-default never declared the association. #1830's unconditional include was the regression; this PR is not a new user-facing behavior change.

### Fix

Gate the association and its presence validation on `enable_application_owner?` inside `Ownership`'s `included` block:

```ruby
included do
  if Doorkeeper.config.enable_application_owner?
    belongs_to :owner, polymorphic: true, optional: true
    validates :owner, presence: true, if: :validate_owner?
  end
end
```

This mirrors how `PolymorphicResourceOwner::ForAccessGrant` / `ForAccessToken` gate `belongs_to :resource_owner` — same pattern, same timing.

`enable_application_owner?` reads `option_set?` (an instance variable check), so this introduces no `constantize` call (#1703 safe) and no `ActiveSupport.on_load(:active_record)` (#1828 safe).

Also synced the now-stale comment above `include ::Doorkeeper::Models::Ownership` in `Mixins::Application`, which still argued for unconditional inclusion.

### Test changes

`enable_application_owner` is now formally a **load-time switch**: its effect is evaluated once when the model class is defined. Tests that exercise the enabled path can no longer reconfigure an already-loaded singleton.

Introduced `build_application_model` helper (`spec/support/helpers/application_model_helper.rb`) that builds a fresh `ActiveRecord::Base` subclass with `Mixins::Application` included **after** the feature is configured. This keeps examples order-independent and avoids mutating the global `Doorkeeper::Application`.

The two #1830 regression specs that asserted "`:owner` association is present even when the feature is off" are inverted to assert the opposite — the intended behavioral change of this PR.

Also fixed a pre-existing label typo: duplicate `"without confirmation"` → `"with confirmation"`.

Closes #1831

### DIFF
--- a/lib/doorkeeper/orm/active_record/mixins/application.rb
+++ b/lib/doorkeeper/orm/active_record/mixins/application.rb
@@ -9,13 +9,12 @@ module Doorkeeper::Orm::ActiveRecord::Mixins
       self.strict_loading_by_default = false if respond_to?(:strict_loading_by_default)
 
       include ::Doorkeeper::ApplicationMixin
-      # Included unconditionally: this block runs once at parent-class
-      # autoload time, so gating on `enable_application_owner?` would
-      # freeze behavior at first-load time. The actual owner validation
-      # is still gated dynamically via `validate_owner?` →
-      # `confirm_application_owner?`, and `belongs_to :owner` is lazy on
-      # schemas that lack the columns.
-      include ::Doorkeeper::Models::Ownership
+      # `enable_application_owner?` is read once, at parent-class autoload
+      # time (#1831): with the feature off the model exposes no `:owner`
+      # association — avoiding a misleading reflection on schemas that lack
+      # the owner columns. The flag is therefore a load-time switch; turning
+      # it on later requires defining a fresh model class.
+      include ::Doorkeeper::Models::Ownership if Doorkeeper.config.enable_application_owner?
 
       has_many :access_grants,
                foreign_key: :application_id,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ User-visible changes worth mentioning.
 ## main
 
 - Please add here
+- [#1832] Fix confusing `belongs_to :owner` side effect: `Doorkeeper::Models::Ownership` is now included only when `enable_application_owner?` is set (read at include time), so models no longer expose a misleading `owner` association/reflection when the application owner feature is disabled and the schema lacks the owner columns.
 
 ## 5.9.2
 
--- a/spec/lib/config_spec.rb
+++ b/spec/lib/config_spec.rb
@@ -394,23 +394,19 @@ RSpec.describe Doorkeeper::Config do
 
     if DOORKEEPER_ORM == :active_record
       context "when enabled without confirmation", active_record: true do
-        class ApplicationWithOwner < ActiveRecord::Base
-          include Doorkeeper::Orm::ActiveRecord::Mixins::Application
-        end
+        # The owner association is wired at include time (#1831), so the model
+        # is built after `enable_application_owner` has been configured.
+        let(:application_with_owner) { build_application_model }
 
         before do
           Doorkeeper.configure do
             orm DOORKEEPER_ORM
             enable_application_owner
-
-            application_class "ApplicationWithOwner"
           end
-
-          Doorkeeper.run_orm_hooks
         end
 
         it "adds support for application owner" do
-          instance = ApplicationWithOwner.new(FactoryBot.attributes_for(:application))
+          instance = application_with_owner.new(FactoryBot.attributes_for(:application))
 
           expect(instance).to respond_to :owner
           expect(instance).to be_valid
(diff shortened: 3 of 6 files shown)


lib/doorkeeper/orm/active_record/mixins/application.rb
```diff
@@ -9,12 +9,13 @@ module Application
...
 
       include ::Doorkeeper::ApplicationMixin
-      # Included unconditionally: this block runs once at parent-class
-      # autoload time, so gating on `enable_application_owner?` would
-      # freeze behavior at first-load time. The actual owner validation
-      # is still gated dynamically via `validate_owner?` →
-      # `confirm_application_owner?`, and `belongs_to :owner` is lazy on
-      # schemas that lack the columns.
+      # The module is mixed in unconditionally, but its `included` block
+      # gates `belongs_to :owner` (and the owner presence validation) on
```
nbulaj: Since we already in the `included` block here which is lazy as well - can we move ` if Doorkeeper.config.enable_application_owner?` here and include conditionally ? 👀  Or I missed smth?

55728: Good call 👀✨ — the include site is just as lazy (both run in the same `included` block at parent-class autoload), so I moved the gate up:

```ruby
include ::Doorkeeper::Models::Ownership if Doorkeeper.config.enable_application_owner?
```

`Doorkeeper::Models::Ownership` is now untouched and the whole change lives at the include site. One observable difference: with the feature off, the model no longer carries the `Ownership` module in its ancestors nor defines `validate_owner?` — I updated the specs that asserted those.

nbulaj (approved):

leoarnold:
@55728 Looks good to me (see also [here](https://github.com/doorkeeper-gem/doorkeeper/issues/1831#issuecomment-4680409816))! :clap:

55728:
Both Copilot's points are valid — updated in the next force-push:
1. CHANGELOG: dropped the `owner_type` / `NoMethodError` claim; the fix only removes the misleading `:owner` reflection.
2. Helper comment: reworded to "As of #1832 (fixes #1831)" since the behavior change is introduced by this PR, not #1831.

55728:
Hound skipped his walk today 🐕💤

55728:
@leoarnold Thank you for verifying against your app 😊 — that's exactly the behavior this PR restores 🎉

PR #1837, opened 2026-06-19
Drop Ruby 3.1 from CI matrix and bump required_ruby_version

55728:
Ruby 3.1 reached EOL on 2025-03-25. Since i18n 1.15.0 (released 2026-06-17) uses `Fiber[:i18n_config]` (Ruby 3.2+ Fiber storage API), the Ruby 3.1 CI jobs now fail with `NoMethodError` on every run.

>[!NOTE]
> #### This is the same root cause addressed in doorkeeper-openid_connect:
> - https://github.com/doorkeeper-gem/doorkeeper-openid_connect/issues/314

Changes:
- Remove Ruby 3.1 from CI matrix (and its `exclude` entry)
- Set `fail-fast: false` to prevent one job's failure from cancelling siblings
- Bump `required_ruby_version` from `>= 2.7` to `>= 3.2`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,11 +17,10 @@ jobs:
       !(   contains(github.event.pull_request.title,  '[ci skip]')
         || contains(github.event.pull_request.title,  '[skip ci]'))
     strategy:
-      fail-fast: true
+      fail-fast: false
       matrix:
         os: [ubuntu-latest]
         ruby:
-          - "3.1"
           - "3.2"
           - "3.3"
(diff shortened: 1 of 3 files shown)


55728:
Closing this — the upstream cause has been resolved in

- ruby-i18n/i18n#739

which restores Ruby 3.1 support by falling back to thread-local variables on Rubies without `Fiber[]`.

Once i18n 1.15.2 is released, the Ruby 3.1 CI jobs should pass again without changes on the doorkeeper side.

PR #1840, opened 2026-06-19
[2/3] Add a client authentication methods registry (RFC 6749 §2.3)

55728:
## Summary

This continues @ThisIsMissEm's work in #1772, which she generously offered to hand off. It **supersedes #1772**. Depends on #1839.

> [!NOTE]
> Original work by @ThisIsMissEm (#1772). This PR is a clean rewrite of her registry design, with the indifferent-access fix and split into a reviewable series (#1839 → this → #1841).

Replaces the hardcoded `client_credentials` extraction with a pluggable client authentication methods registry, mirroring the existing grant-flow registry. Extension gems (e.g. doorkeeper-openid_connect) can register additional methods from the [IANA registry](https://www.iana.org/assignments/oauth-parameters/oauth-parameters.xhtml#token-endpoint-auth-method) such as `private_key_jwt`:

```ruby
Doorkeeper::ClientAuthentication.register(
  :private_key_jwt,
  PrivateKeyJwt,
)
```

## New API

- `Doorkeeper::ClientAuthentication` registry (`register` / `get`) with a `Method` wrapper (exposing `#name` / `#strategy`) and a `FallbackMethod` (matches everything, authenticates to nothing — preserving the previous "no credentials" behaviour).
- Built-in strategies under `Doorkeeper::OAuth::ClientAuthentication`: `ClientSecretBasic`, `ClientSecretPost`, `None`. The post/none strategies read from the request **body** only (not the query string), per §2.3.1.
- `Doorkeeper::ClientAuthentication::Credentials` (moved from `Doorkeeper::OAuth::Client::Credentials`; the old constant remains as a deprecated alias — see Deprecations).

## Configuration

New `client_authentication` option, default `%i[client_secret_basic client_secret_post none]`. It accepts an array or varargs, and the listed methods are used **verbatim** and in order — unlike the legacy option, nothing (in particular `:none`) is ever appended, so a restrictive configuration is never silently broadened:

```ruby
Doorkeeper.configure do
  client_authentication %i[client_secret_basic client_secret_post none]
  # or: client_authentication :client_secret_basic, :client_secret_post
end
```

The configuration is validated at boot: unregistered method names (a typo, or an extension that failed to load) are warned about and ignored, a configuration that resolves to zero usable methods logs an error (every authentication attempt would fail), and setting both `client_credentials` and `client_authentication` warns that the latter takes precedence.

## Implementation note

The original prototype accessed `request.request_parameters[:client_id]` with symbol keys. `request_parameters` is string-keyed at the Rack layer, so whether a bare symbol lookup works depends on the Rails version: on newer Rails it returns an `ActiveSupport::HashWithIndifferentAccess` (symbol access happens to work), but on the older Rails versions Doorkeeper still supports it returns a plain string-keyed `Hash`, where `params[:client_id]` returns `nil` — so `client_secret_post` authentication would have silently failed there. This rewrite normalises with `with_indifferent_access` so the strategies behave identically across all supported Rails versions.

The strategy unit specs now build their mock request with string-keyed parameters (matching what Rack/ActionDispatch actually hand back), so they genuinely exercise the indifferent-access handling — dropping `with_indifferent_access` now fails these specs rather than passing on the symbol keys the specs were written with. The live flow specs cover the end-to-end path.

## Breaking / behavioural changes

- **Body-only credentials for `client_secret_post` and `none`** — these strategies read client credentials from the request body only (`request.request_parameters`), per RFC 6749 §2.3.1. Clients that sent `client_id`/`client_secret` in the query string of the token endpoint will no longer authenticate and must send them in the request body instead.
- **A Basic authentication attempt never falls through** — any request whose `Authorization` header decodes to a non-blank Basic `client_id` is claimed by `client_secret_basic`, so invalid credentials fail with `invalid_client` instead of silently falling through to another configured method or the fallback (which would have downgraded a failed authentication attempt to "no authentication provided").
- **`none` requires a blank `Authorization` header** — a request carrying a non-blank one (Basic, Bearer, or anything else) is attempting header-based authentication and is no longer treated as an unauthenticated public client. This is narrower than the legacy `from_params` extractor, which read the body `client_id` regardless of the header.

### Known deviation kept for compatibility

RFC 6749 §2.3.1 requires `client_id`/`client_secret` to be form-urlencoded before being placed into the Basic header. Doorkeeper has never URL-decoded them (like much of the ecosystem), and `ClientSecretBasic` deliberately keeps that behaviour — adding the decoding now would break every existing client whose credentials contain URL-encodable characters.

## Deprecations

All three keep working for the deprecation window, with warnings:

- **`client_credentials` config option** — deprecated in favour of `client_authentication`. `:from_basic` / `:from_params` are automatically converted to `:client_secret_basic` / `:client_secret_post`; `:none` (public client support) is appended **only when `:from_params` was configured**, since that is the only legacy method that accepted a bare `client_id` without a secret — a Basic-only configuration is not silently broadened. Callable extractors are wrapped in a `LegacyCallable` adapter so they keep working too.
- **`Doorkeeper.config.client_credentials_methods`** — renamed to `client_authentication_methods` (which returns the resolved registry `Method` objects). The old name remains as a deprecated alias that keeps returning the **legacy symbol names** (`:from_basic` / `:from_params`) existing callers were written against; names without a legacy equivalent (e.g. `:none`, or a registered `:private_key_jwt`) pass through unchanged.
- **`Doorkeeper::OAuth::Client::Credentials`** — moved to `Doorkeeper::ClientAuthentication::Credentials`. The old constant remains as an alias marked with `deprecate_constant`, so referencing it warns but keeps working for one release.

## Notes for doorkeeper-openid_connect

doorkeeper-openid_connect reads `Doorkeeper.config.client_credentials_methods` in its discovery metadata (`token_endpoint_auth_methods_supported_mixin`). Because the deprecated alias keeps returning the legacy `:from_basic` / `:from_params` symbols, the extension's existing mapping continues to work completely unchanged. Once it migrates to `client_authentication_methods`, it can drop its internal `from_basic → client_secret_basic` mapping and any registered method (e.g. `private_key_jwt`) flows into discovery automatically, since the registry exposes RFC 8414-compliant names directly.

## Follow-ups (out of scope)

- Reject requests that use more than one client authentication method (RFC 6749 §2.3) — #1841.
- A richer `Method` contract that lets a method resolve & verify the client itself (rather than always going through `by_uid_and_secret`), which is what asymmetric-key methods like `private_key_jwt` need. The registry in this PR is the groundwork.


55728:
> [!TIP]
> **Reading guide**: This PR is structured as two commits for easier review:
> 1. **1st commit** — Registry infrastructure + strategies + config option (purely additive, no behavior change)
> 2. **2nd commit** — Wire it up, deprecation path, breaking change
>
> Reviewing commit-by-commit is recommended. If separate PRs would be easier, I can split them — just let me know.

lib/doorkeeper/oauth/client_authentication/client_secret_post.rb
```diff
@@ -0,0 +1,29 @@
...
+      # RFC 6749 §2.3.1 "client_secret_post": client credentials are sent in
+      # the request body. The query string is intentionally ignored so that
+      # credentials must be supplied in the body as the spec requires.
+      class ClientSecretPost
+        def self.matches_request?(request)
+          params = request.request_parameters.with_indifferent_access
+
+          request.post? &&
+            params[:client_id].present? &&
+            params[:client_secret].present?
```
55728, line 15: >[!NOTE]
> **Body-only**: We read from `request.request_parameters` (the parsed POST body) instead of `request.parameters` (which merges body + query string). This is per RFC 6749 §2.3.1: *"Including the client credentials in the request-body [...] is NOT RECOMMENDED and SHOULD be limited to clients unable to directly utilize the HTTP Basic authentication scheme."* — credentials in the query string are not a recognized authentication method. The same applies to the `None` strategy. This is the source of the breaking change documented in the CHANGELOG.

lib/doorkeeper/client_authentication/fallback_method.rb
```diff
@@ -0,0 +1,19 @@
...
+    # no credentials.
+    class FallbackMethod
+      def self.matches_request?(_request)
+        true
+      end
+
+      def self.authenticate(_request)
+        nil
+      end
+    end
```
55728, line 17: >[!NOTE]
> The FallbackMethod preserves backward compatibility: when no registered method matches the request, `Server#credentials` receives `nil` — the same result the old `Credentials.from_request` returned when no extraction method succeeded. Without this, unrecognized requests would raise instead of proceeding unauthenticated (which Doorkeeper intentionally allows for some grant flows).

lib/doorkeeper/client_authentication.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    register(
+      :client_secret_post,
+      Doorkeeper::OAuth::ClientAuthentication::ClientSecretPost,
+    )
+
+    register(
+      :client_secret_basic,
+      Doorkeeper::OAuth::ClientAuthentication::ClientSecretBasic,
+    )
```
55728, line 51: >[!NOTE]
> Registration order here doesn't affect runtime behavior — it only populates the registry hash. The *evaluation order* at request time is determined by the `client_authentication` config array (default: `[:client_secret_basic, :client_secret_post, :none]`), which is resolved in `config.client_authentication_methods`.

lib/doorkeeper/client_authentication/registry.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module Doorkeeper
+  module ClientAuthentication
+    # Holds the registered client authentication methods and provides the DSL
+    # to register and look them up by name.
+    module Registry
+      mattr_accessor :methods
+      self.methods = {}
```
55728: >[!NOTE]
> `mattr_accessor` rather than a plain `@methods` ivar so that the registry is a proper module-level singleton shared across all `extend`ers. The name `methods` shadows `Module#methods` but is only accessed through `Registry.methods` or `self.methods`, not through instances, so the collision is harmless.

lib/doorkeeper/config.rb
```diff
@@ -596,8 +634,43 @@ def pkce_code_challenge_methods_supported
...
+            @client_credentials_methods
+          end
+        else
+          client_authentication
+        end
+
+      @client_authentication_methods = methods.filter_map do |name|
+        Doorkeeper::ClientAuthentication.get(name)
+      end
+    end
```
55728, line 673: >[!NOTE]
> Priority logic for the migration period:
> 1. `client_authentication` set + `client_credentials` set → `client_authentication` wins (with a warning)
> 2. `client_credentials` set alone → honored (it was already auto-converted from `:from_basic`/`:from_params` to the new names by the deprecated `client_credentials` writer)
> 3. Neither set (default) → `client_authentication` default value
>
> The result is memoized in `@client_authentication_methods` so the resolution (and any warnings) happen only once.

lib/doorkeeper/server.rb
```diff
@@ -37,8 +41,7 @@ def resource_owner
     end
 
     def credentials
-      methods = Doorkeeper.config.client_credentials_methods
-      @credentials ||= OAuth::Client::Credentials.from_request(context.request, *methods)
+      @credentials ||= client_authentication_method_for_request.authenticate(context.request)
     end
```
55728, line 45: >[!NOTE]
> This is the core behavior change. Previously: iterate an array of extraction lambdas (`from_basic`, `from_params`) and return the first non-blank `Credentials`. Now: `Request.client_authentication_method` matches the *first strategy whose `matches_request?` returns true*, then calls `.authenticate` on that strategy alone. The semantic difference: the old approach tried every method regardless of whether the request "looked like" that method; the new approach picks exactly one method based on request shape (Basic header present? → `client_secret_basic`; body has both `client_id` + `client_secret`? → `client_secret_post`; etc.). If nothing matches, `FallbackMethod` returns `nil`.

lib/doorkeeper/config.rb
```diff
@@ -596,8 +634,43 @@ def pkce_code_challenge_methods_supported
...
+    #   and will be removed afterwards.
     def client_credentials_methods
-      @client_credentials_methods ||= %i[from_basic from_params]
+      Kernel.warn(
+        "[DOORKEEPER] Doorkeeper.config.client_credentials_methods has been renamed to " \
+        "client_authentication_methods and will be removed in a future version.",
+      )
+
+      client_authentication_methods
     end
```
55728, line 698: >[!NOTE]
> doorkeeper-openid_connect reads `Doorkeeper.config.client_credentials_methods` in `token_endpoint_auth_methods_supported_mixin` to populate the OIDC discovery document. This alias keeps that working for one release cycle. On the upside, the new registry returns RFC 8414-compliant method names directly (`:client_secret_basic` etc.), so doorkeeper-openid_connect can drop its internal `from_basic → client_secret_basic` translation once it migrates.

lib/doorkeeper/oauth/client_authentication/client_secret_post.rb
```diff
@@ -0,0 +1,29 @@
...
+
+module Doorkeeper
+  module OAuth
+    module ClientAuthentication
+      # RFC 6749 §2.3.1 "client_secret_post": client credentials are sent in
+      # the request body. The query string is intentionally ignored so that
+      # credentials must be supplied in the body as the spec requires.
+      class ClientSecretPost
+        def self.matches_request?(request)
+          params = request.request_parameters.with_indifferent_access
```
55728, line 11: >[!NOTE]
> `with_indifferent_access` is intentional: Rails `request_parameters` are **string-keyed** in real requests, so accessing `:client_id` with a symbol key would silently return `nil`. The original implementation in #1772 used bare symbol access, which passed in unit tests (where params are set with symbol keys) but would have failed in production. The **live flow specs** (which exercise real string-keyed request parameters) verify that this works correctly end-to-end.

lib/doorkeeper/config.rb
```diff
@@ -65,13 +65,43 @@ def scopes_by_grant_type(hash = {})
...
+
+        if client_authentication.empty?
+          Kernel.warn(
+            "[DOORKEEPER] No known client_credentials method detected, " \
+            "cannot automatically convert to client_authentication option",
+          )
+        else
+          @config.instance_variable_set(:@client_credentials_methods, client_authentication.push(:none))
+        end
       end
```
55728, line 96: >[!NOTE]
> The auto-conversion appends `:none` because the old `from_params` extracted *any* `client_id` from `request.parameters` — including public clients without a secret. Under the new registry, public clients are handled by the explicit `None` strategy, so `:none` is needed to preserve that behavior for `:from_params` users. Note: for `client_credentials :from_basic` only configurations, this adds `None` authentication that wasn't previously available — a conscious trade-off favoring safe defaults over exact equivalence.

lib/doorkeeper/oauth/client_authentication/client_secret_post.rb
nbulaj: While it's a breaking change - in a real world apps which support OAuth have to correctly send the payload using the standard, right, which says to do it via the request body? It's only the case of the old implementation which allowed to use any payload source including query string. So maybe we actually _don't break anything_ 👼

nbulaj (commented):
This is the only breaking change, right @55728 ?

ThisIsMissEm:
> This is the only breaking change, right @55728 ?

This will be breaking, but was never spec behaviour, so should have never been used. But I do know of certain clients to mastodon that this will break, and to be honest I'm okay with that. This is an identified security issue with a CWE attached.

55728:
Yes — this is the only **runtime / client-facing** breaking change.

And I agree completely. Per [RFC 6749 §2.3.1](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1), the body parameters "can only be transmitted in the request-body and **MUST NOT be included in the request URI**." So a spec-compliant client was never permitted to put `client_id`/`client_secret` in the query string in the first place — the old implementation was accepting a transport the spec explicitly forbids. Well-behaved clients should see no change. 👼

One thing to flag so we're precise about "only": there's also an **API-level** breaking change — `Doorkeeper::OAuth::Client::Credentials` has moved to `Doorkeeper::ClientAuthentication::Credentials` with no backward-compat alias (noted in the CHANGELOG upgrade note), so downstream code referencing the old constant must update. The renamed config readers (`client_credentials_methods` → `client_authentication_methods`) *do* keep a delegating alias, so those aren't breaking.

nbulaj:
Thanks @ThisIsMissEm and @55728 :bow: . I believe these changes gonna go into the following 6.x release and we can introduce such breaking changes. Anyway it will be great to note them in the changelog and in the[ upgrade guide](https://github.com/doorkeeper-gem/doorkeeper/wiki/Migration-from-old-versions) to make update of the gem smoother

55728:
I've expanded the CHANGELOG entry and drafted an upgrade guide section for the wiki ([Migration from old versions](https://github.com/doorkeeper-gem/doorkeeper/wiki/Migration-from-old-versions)). Posting it here first for review before editing the wiki directly:

---

## From 5.x to 6.0

### Configuration changes:

* **The `client_credentials` option is deprecated in favour of `client_authentication`.** If your initializer contains `client_credentials :from_basic, :from_params`, replace it with:

  ```ruby
  client_authentication %i[client_secret_basic client_secret_post none]
  ```

  For backwards compatibility the old option still works (with a deprecation warning): `:from_basic` is mapped to `:client_secret_basic`, `:from_params` to `:client_secret_post`, and `:none` is appended automatically so public clients keep working. If both options are set, `client_authentication` takes precedence. The default when neither is set is `%i[client_secret_basic client_secret_post none]`.

* You can now register custom client authentication strategies (RFC 6749 §2.3) via `Doorkeeper::ClientAuthentication.register(:name, Strategy)`. See [Changing how clients are authenticated](https://github.com/doorkeeper-gem/doorkeeper/wiki/Changing-how-clients-are-authenticated).

### API changes:

* **Client authentication now enforces RFC 6749 §2.3.1 transport rules.** The `client_secret_post` and `none` strategies now only accept credentials from the POST request body. Credentials sent via query string are no longer accepted. Per [RFC 6749 §2.3.1](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1), "The parameters can only be transmitted in the request-body and MUST NOT be included in the request URI." Spec-compliant clients should see no change, but any client relying on query-string credentials must switch to sending them in the request body.

* **`Doorkeeper::OAuth::Client::Credentials` has moved to `Doorkeeper::ClientAuthentication::Credentials`.** There is no backward-compatible alias. Downstream code (including extension gems such as `doorkeeper-openid_connect`) referencing the old constant must update.

* The configuration reader `client_credentials_methods` has been renamed to `client_authentication_methods`. A delegating alias is kept for one release — calling the old name still works but emits a deprecation warning, and it will be removed in a future version.

---

Let me know if anything needs adjusting!

ThisIsMissEm:
We could reference the names as being from the IANA OAuth Registry: https://www.iana.org/assignments/oauth-parameters/oauth-parameters.xhtml#token-endpoint-auth-method

explicitly "none" doesn't mean "don't send client_id" it means that only a client_id is needed in the request body or sometimes query params (can't recall if that's definitely to spec or not)

55728:
> We could reference the names as being from the IANA OAuth Registry: https://www.iana.org/assignments/oauth-parameters/oauth-parameters.xhtml#token-endpoint-auth-method
> 
> explicitly "none" doesn't mean "don't send client_id" it means that only a client_id is needed in the request body or sometimes query params (can't recall if that's definitely to spec or not)

Agreed on the definition — and I think the current `None` strategy already matches it. Per the IANA registry (defined in RFC 7591): `none` means the client is a public client and does not have a client secret, identifying itself via the `client_id` request parameter (RFC 6749 §3.2.1).

The implementation requires `client_id` to be present and `client_secret` to be blank, reading from the request body only, and returns `Credentials(client_id, nil)`. The truly "no credentials" case (no `client_id` at all) doesn't match `None` and falls through to `FallbackMethod` — so `none` ("only a client_id in the body") and "send nothing" stay distinct.

One improvement I'll make: the code comment currently cites RFC 6749 §2.3; I'll update it to reference the IANA Token Endpoint Auth Methods registry / RFC 7591 for the `none` definition as you suggested.

55728:
All six addressed in the latest push — thanks Copilot 🤖

1. `start_with?("basic ")` — require space delimiter
2. `request.authorization.blank?` — defensive consistency
3. `validate_client_authentication_registered` — warn on unresolved method names at boot (pulled into its own validation rather than inlining in `filter_map`)
4. `methods` → `registered_methods`
5. Spec teardown updated to match
6. `Kernel.warn` → `Rails.logger.warn`

Specs added for 1–3.

55728:
Addressed Copilot's feedback 👀✨

55728:
Pushed a follow-up addressing the review feedback (Copilot + a self-review pass), rebased on `main` and split into topic-scoped commits for easier review 👀✨

Why these changes:

- **Legacy `client_credentials` conversion** — only append `:none` for `:from_params` (the one legacy method that accepted a public client), so a Basic-only config isn't silently broadened; and wrap callable extractors in a `LegacyCallable` adapter instead of dropping them.
- **`client_credentials_methods` alias** — now returns the legacy `:from_basic` / `:from_params` symbols rather than internal `Method` objects, so doorkeeper-openid_connect's `token_endpoint_auth_methods_supported` keeps working unchanged.
- **`client_authentication` option** — honoured verbatim in both array and varargs forms (no silent reset to the default that would re-enable `:none`).
- **`Doorkeeper::OAuth::Client::Credentials`** — kept as a deprecating alias so referencing code doesn't hard-raise `NameError`.
- **Validation** — warn on conflicting options at boot, and log an error when the config resolves to zero usable methods.

CHANGELOG updated.

55728:
Updated the upgrade-guide draft:

---

## From 5.x to 6.0

### Configuration changes:

* **The `client_credentials` option is deprecated in favour of `client_authentication`.** If your initializer contains `client_credentials :from_basic, :from_params`, replace it with:

  ```ruby
  client_authentication %i[client_secret_basic client_secret_post none]
  ```

  For backwards compatibility the old option still works (with a deprecation warning): `:from_basic` is mapped to `:client_secret_basic`, `:from_params` to `:client_secret_post`, and `:none` is appended **only when `:from_params` was configured** — that is the sole legacy method that accepted a bare `client_id` without a secret, so a Basic-only (`client_credentials :from_basic`) configuration is not silently broadened to accept public clients. A callable extractor (e.g. `client_credentials ->(request) { [client_id, client_secret] }`) is wrapped in a legacy adapter and keeps working during the deprecation window. If both options are set, `client_authentication` takes precedence (warned about at boot). The default when neither is set is `%i[client_secret_basic client_secret_post none]`.

* You can now register custom client authentication strategies (RFC 6749 §2.3) via `Doorkeeper::ClientAuthentication.register(:name, Strategy)`. See [Changing how clients are authenticated](https://github.com/doorkeeper-gem/doorkeeper/wiki/Changing-how-clients-are-authenticated).

### API changes:

* **Client authentication now enforces RFC 6749 §2.3.1 transport rules.** The `client_secret_post` and `none` strategies now only accept credentials from the POST request body. Credentials sent via query string are no longer accepted. Per [RFC 6749 §2.3.1](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1), "The parameters can only be transmitted in the request-body and MUST NOT be included in the request URI." Spec-compliant clients should see no change, but any client relying on query-string credentials must switch to sending them in the request body.

* **`Doorkeeper::OAuth::Client::Credentials` has moved to `Doorkeeper::ClientAuthentication::Credentials`.** The old constant is kept as a deprecating alias for one release, so referencing code does not raise `NameError` — but its legacy `.from_request` / `.from_basic` / `.from_params` class methods are gone (client credential extraction now goes through the registry). Downstream code (including extension gems such as `doorkeeper-openid_connect`) referencing the old constant should update.

* The configuration reader `client_credentials_methods` has been renamed to `client_authentication_methods`. A delegating alias is kept for one release — calling the old name still works but emits a deprecation warning, and it will be removed in a future version. The alias returns the legacy `:from_basic` / `:from_params` symbols (not the internal `Method` objects), so consumers that map those symbols — such as `doorkeeper-openid_connect`'s `token_endpoint_auth_methods_supported` — keep working unchanged for that release.

---

Let me know if anything needs adjusting 😊

55728:
Addressed Copilot's feedback 👀✨

lib/doorkeeper/oauth/client_authentication/client_secret_basic.rb
```diff
@@ -0,0 +1,53 @@
...
+        def self.credentials_from(request)
+          authorization = request.authorization.to_s
+          return unless authorization.downcase.start_with?("basic ")
+
+          value = authorization.split(" ", 2).last
+          client_id, client_secret = Base64.decode64(value.to_s).split(":", 2)
+          return if client_id.blank?
+
+          [client_id, client_secret]
+        end
```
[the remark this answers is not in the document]
55728: Thanks — I considered this, but I'd rather keep the strategies stateless. Detection runs once per request (Server#credentials is memoized), so the header is decoded at most twice, and that's a Base64.decode64 + split on a short string — noise next to the DB lookup and secure secret comparison the token request already performs. Caching on request.env would introduce a shared-state contract into an otherwise pure matches_request?/authenticate API, which doesn't seem worth it without a measurable win. (ClientSecretPost has the same double-read shape via request_parameters, which Rack already caches.)

55728:
Thanks Mr. Copilot 😆✨

55728:
@nbulaj I'd appreciate your review whenever you have a moment. No rush at all! Thanks 😊

CHANGELOG.md
```diff
@@ -9,6 +9,25 @@ User-visible changes worth mentioning.
...
+- [#1840] Introduce a pluggable client authentication registry (RFC 6749 §2.3). Client authentication is now resolved through named strategies instead of the hard-coded basic/params lookup, making the supported methods explicit, configurable, and extensible.
+  - **New `client_authentication` option**: declare which methods Doorkeeper accepts and the order in which they are tried, e.g. `client_authentication %i[client_secret_basic client_secret_post none]` (the default). The three built-in strategies are `client_secret_basic` (HTTP Basic, RFC 6749 §2.3.1), `client_secret_post` (credentials in the request body) and `none` (public clients with no secret).
+  - **Custom methods**: register your own strategy with `Doorkeeper::ClientAuthentication.register(:my_method, MyStrategy)`. A strategy responds to `matches_request?` and `authenticate`, returning a `Doorkeeper::ClientAuthentication::Credentials`. See the [wiki](https://github.com/doorkeeper-gem/doorkeeper/wiki/Changing-how-clients-are-authenticated) for the contract.
+
+  This ships with the following **breaking changes**:
```
nbulaj: Looks like we have a lot for the changelog  entry 🥲  Maybe we can make it shorter and move detailed descriptions to the upgrade guides?

nbulaj:
Nice work BTW, appreciated :muscle:

CHANGELOG.md
55728: I've unintentionally written an OAuth fairy tale 😅

Shortened the entry to a one-liner per item and prepared an Upgrade Guide for the wiki (will publish it when this ships).

While at it I noticed the query-string restriction also affects revocation and introspection (they go through the same `Server#credentials` path), so I:
- spelled that out in the CHANGELOG (`… token, revocation and introspection`)
- added request specs covering body-credential success + query-string rejection for all three endpoints
- updated the Upgrade Guide accordingly

Please take another look when you get a chance!

55728: Here's the draft Upgrade Guide I've prepared for the wiki — planning to publish it under [Migration from old versions](https://github.com/doorkeeper-gem/doorkeeper/wiki/Migration-from-old-versions) when this ships. Let me know if anything needs adjusting!

<details>
<summary>Upgrade Guide: Client Authentication Registry (#1840)</summary>

# Client Authentication Registry (#1840)

Doorkeeper resolves client authentication at the token/revocation/introspection endpoints through a registry of named strategies (RFC 6749 §2.3) instead of the previous hard-coded basic/params lookup. This page collects the breaking changes, deprecations and migration steps for this change.

## What changed

The new `client_authentication` config option declares which methods Doorkeeper accepts and the order in which they are tried:

```ruby
Doorkeeper.configure do
  # This is the default
  client_authentication %i[client_secret_basic client_secret_post none]
end
```

The three built-in strategies are:

| Strategy | Description |
|---|---|
| `client_secret_basic` | Credentials in the HTTP `Authorization: Basic` header (RFC 6749 §2.3.1) |
| `client_secret_post` | Credentials (`client_id`/`client_secret`) in the request body |
| `none` | Public clients: a bare `client_id` in the request body, no secret |

You can register your own strategy with:

```ruby
Doorkeeper::ClientAuthentication.register(:my_method, MyStrategy)
```

A strategy responds to `matches_request?` and `authenticate`, returning a `Doorkeeper::ClientAuthentication::Credentials`. See [Changing how clients are authenticated](https://github.com/doorkeeper-gem/doorkeeper/wiki/Changing-how-clients-are-authenticated) for the full contract.

## Breaking changes

### Client credentials are no longer read from the query string

The `client_secret_post` and `none` strategies now look at the request body only (`request.request_parameters`), per RFC 6749 §2.3.1 — sending secrets in a URI query is explicitly discouraged by the spec (secrets end up in server logs, proxies and browser history). This applies to **every endpoint that authenticates clients: token, revocation (`/oauth/revoke`) and introspection (`/oauth/introspect`)** — they all resolve client authentication through the same registry. Clients that passed `client_id`/`client_secret` in the query string of any of these endpoints will stop authenticating and must move the credentials into the request body (or use HTTP Basic).

### The `none` strategy requires a blank `Authorization` header

A request that carries a non-blank `Authorization` header (e.g. a `Bearer` token, or a malformed `Basic` header) is attempting header-based authentication and is no longer silently treated as an unauthenticated public client the way the legacy `from_params` extractor did. Public clients must send their `client_id` in the request body without an `Authorization` header.

### `Doorkeeper::OAuth::Client::Credentials` has moved

The class now lives at `Doorkeeper::ClientAuthentication::Credentials`. The old constant remains as a deprecated alias for one release (marked with `deprecate_constant`, so Ruby warns on access when deprecation warnings are enabled) so referencing code does not raise `NameError` — but its legacy `.from_request` / `.from_basic` / `.from_params` class methods are gone (extraction now goes through the registry). Update references to the new constant.

### `client_authentication` is honoured exactly as written

Both the array form (`client_authentication %i[client_secret_basic client_secret_post none]`) and the varargs form (`client_authentication :client_secret_basic, :client_secret_post`) are accepted, and nothing (in particular `:none`) is appended — a restrictive configuration is never silently broadened. A configuration that resolves to zero registered methods is logged as an error at boot instead of silently failing every client authentication.

## Deprecations

### `client_credentials` option → `client_authentication`

The `client_credentials` config option is deprecated in favour of `client_authentication`. During the deprecation window the legacy values are mapped automatically:

| Legacy value | Mapped to |
|---|---|
| `:from_basic` | `:client_secret_basic` |
| `:from_params` | `:client_secret_post` (+ `:none`, see below) |

`:none` (public client support) is appended only when `:from_params` was configured — that is the only legacy method that accepted a bare `client_id` without a secret. A Basic-only configuration (`client_credentials :from_basic`) is therefore **not** silently broadened to accept public clients.

If both `client_credentials` and `client_authentication` are set, `client_authentication` wins and Doorkeeper warns about the conflict at boot.

### Callables passed to `client_credentials`

A callable passed to `client_credentials` (e.g. `client_credentials ->(request) { [client_id, client_secret] }`) is wrapped in a legacy adapter (`Doorkeeper::ClientAuthentication::LegacyCallable`) so it keeps working during the deprecation window. Migrate it to a registered strategy via `Doorkeeper::ClientAuthentication.register` before the `client_credentials` option is removed.

### `client_credentials_methods` → `client_authentication_methods`

`Doorkeeper.config.client_credentials_methods` has been renamed to `client_authentication_methods`. The old name remains as a delegating alias that emits a one-shot deprecation warning. The alias returns the legacy method-name symbols (e.g. `:from_basic`, `:from_params`) rather than the internal `Method` objects, so external consumers such as `doorkeeper-openid_connect` (which map those symbols in `token_endpoint_auth_methods_supported`) keep working unchanged for one release.

## Upgrade notes

1. **Audit your clients**: if any of them send `client_id`/`client_secret` in the token endpoint query string, switch them to the request body or HTTP Basic *before* upgrading.
2. **Update your initializer**: replace `client_credentials :from_basic, :from_params` with `client_authentication %i[client_secret_basic client_secret_post none]` (drop `none` if you don't serve public clients).
3. **Migrate custom callables**: if you passed a lambda/proc to `client_credentials`, rewrite it as a strategy class and register it with `Doorkeeper::ClientAuthentication.register`, then reference it by name in `client_authentication`.
4. **Update constant references**: `Doorkeeper::OAuth::Client::Credentials` → `Doorkeeper::ClientAuthentication::Credentials`.
5. **Update config readers**: `Doorkeeper.config.client_credentials_methods` → `Doorkeeper.config.client_authentication_methods`.
6. **Check public clients**: make sure they send a bare `client_id` in the request body with no `Authorization` header — requests with a stray non-blank header are now rejected.

</details>

nbulaj (approved):
LGTM ✅

55728:
Thank you for the review 😊

PR #1842, opened 2026-06-20
Force PKCE for all clients, including confidential ones (RFC 9700)

### LINKED ISSUE #1654: Option to enforce that clients use PKCE

55728:
## Summary

This continues @ThisIsMissEm's work in #1773. It **supersedes #1773**. Closes #1654.

> [!NOTE]
> Based on @ThisIsMissEm's change in #1773, with a fix to preserve the #1762 error-handling behaviour for missing clients.

When `force_pkce` is enabled, confidential clients were previously exempt from the PKCE requirement. This PR removes that exemption so that **all** clients — public and confidential — must use PKCE, in line with the [OAuth 2.0 Security BCP (RFC 9700)](https://www.rfc-editor.org/rfc/rfc9700) and the upcoming OAuth 2.1 specification.

## What changed

- `pre_authorization.rb` — removed `return true if client.confidential` from `validate_code_challenge`, so the authorize endpoint requires `code_challenge` from all clients.
- `authorization_code_request.rb` — removed the `!client.confidential` condition from the `force_pkce` check at the token endpoint.

## Difference from #1773

The original PR removed the `client &&` guard along with `!client.confidential`, which caused a regression: when the client is missing (`nil`), the request would return `InvalidRequest` (missing `code_verifier`) instead of the correct `InvalidClient`. This was fixed by #1762 and must not be regressed.

This PR keeps `client &&` intact — only `!client.confidential` is removed:

```ruby
# Before
elsif client && !client.confidential && Doorkeeper.config.force_pkce? && code_verifier.blank?

# This PR
elsif client && Doorkeeper.config.force_pkce? && code_verifier.blank?
```

The existing "missing app → InvalidClient" test passes without modification, confirming no regression.

## Breaking change

If you enable `force_pkce` and have confidential clients that do not yet send `code_challenge`/`code_verifier`, their authorization requests will start to be rejected. This aligns with RFC 9700's recommendation that PKCE should be required for all clients.

### DIFF
--- a/lib/doorkeeper/oauth/authorization_code_request.rb
+++ b/lib/doorkeeper/oauth/authorization_code_request.rb
@@ -64,7 +64,7 @@ module Doorkeeper
         @missing_param =
           if grant&.uses_pkce? && code_verifier.blank?
             :code_verifier
-          elsif client && !client.confidential && Doorkeeper.config.force_pkce? && code_verifier.blank?
+          elsif client && Doorkeeper.config.force_pkce? && code_verifier.blank?
             :code_verifier
           elsif redirect_uri.blank?
             :redirect_uri
--- a/lib/doorkeeper/oauth/pre_authorization.rb
+++ b/lib/doorkeeper/oauth/pre_authorization.rb
@@ -147,7 +147,6 @@ module Doorkeeper
 
       def validate_code_challenge
         return true unless Doorkeeper.config.force_pkce?
-        return true if client.confidential
         return true if code_challenge.present?
 
         @invalid_request_reason = :invalid_code_challenge
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ User-visible changes worth mentioning.
 ## main
 
 - [#1839] Send client credentials in the request body (not the query string) in the token endpoint specs, per RFC 6749 §2.3.1. Test-only change: the `*_endpoint_url` helpers now return a path plus a matching `*_endpoint_params` builder so flow specs post credentials through the body.
+- [#1842] **[BREAKING]** `force_pkce` now requires PKCE for all clients, including confidential ones, in line with the OAuth 2.0 Security BCP (RFC 9700) and OAuth 2.1. Previously confidential clients were exempt. If you enable `force_pkce` and have confidential clients that do not yet send a `code_challenge`/`code_verifier`, their authorization requests will start to be rejected.
 - Please add here
 
 ## 5.9.3
--- a/lib/generators/doorkeeper/templates/initializer.rb
+++ b/lib/generators/doorkeeper/templates/initializer.rb
@@ -188,8 +188,8 @@ Doorkeeper.configure do
   #
   # revoke_previous_authorization_code_token
 
-  # Require non-confidential clients to use PKCE when using an authorization code
-  # to obtain an access_token (disabled by default)
+  # Require all clients (including confidential ones) to use PKCE when using an
+  # authorization code to obtain an access_token (disabled by default)
   #
   # force_pkce
 
--- a/spec/lib/oauth/authorization_code_request_spec.rb
+++ b/spec/lib/oauth/authorization_code_request_spec.rb
@@ -176,10 +176,15 @@ RSpec.describe Doorkeeper::OAuth::AuthorizationCodeRequest do
       end
 
       context "when the app is confidential" do
-        it "issues a new token for the client" do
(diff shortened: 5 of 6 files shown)


nbulaj:
Any chance @ThisIsMissEm can review as a original author? :pray:

lib/doorkeeper/oauth/pre_authorization.rb
```diff
@@ -147,7 +147,6 @@ def validate_scopes
 
       def validate_code_challenge
         return true unless Doorkeeper.config.force_pkce?
```
ThisIsMissEm, line 149: is this the right way round? if `force_pkce?` is true, then we'd want to return true no?

ThisIsMissEm (commented):
This looks good, did we decide to not do the "pkce_for" type thing which applies based on a block/proc?

lib/doorkeeper/oauth/pre_authorization.rb
55728: Good question — but `unless` is correct here. This is a validator: the return value means "does this pass validation", not "is PKCE forced". So:

- `force_pkce?` false → no PKCE requirement → the check passes → `return true`.
- `force_pkce?` true → we must not short-circuit; we fall through and require `code_challenge` to be present, failing otherwise.

If we returned `true` when `force_pkce?` were true, PKCE would never actually be enforced — the opposite of the intent. I'll add a short comment to make the guard's meaning obvious.

55728:
> This looks good, did we decide to not do the "pkce_for" type thing which applies based on a block/proc?

No — we deliberately didn't add the block/proc (`pkce_for`) approach in this PR. The scope here is intentionally minimal: keep `force_pkce` a boolean and just change its meaning from "public clients only" (the old confidential exemption) to "all clients", per RFC 9700.

One thing worth deciding explicitly though: by removing the confidential exemption, the previous de facto "public-only" enforcement is no longer reachable. #1654 originally proposed a none / public / all choice. So our options are:

1. This PR as-is — boolean `force_pkce` = all clients (simplest, matches RFC 9700/2.1).
2. A none/public/all enum.
3. A proc/predicate like `allow_grant_flow_for_client` (there's precedent at config.rb:325).

My preference is (1) for this PR and to track the proc/enum as a documented follow-up — but happy to fold it in if you'd rather land the richer API now.

ThisIsMissEm:
Yeah, I'd go 1, but follow-up with 3 pre-release

55728:
Sounds good — I'll keep this PR as (1) and open a follow-up issue for the proc/predicate approach (3) so it's tracked for pre-release 👀✨

lib/doorkeeper/oauth/authorization_code_request.rb
```diff
@@ -64,7 +64,7 @@ def validate_params
         @missing_param =
           if grant&.uses_pkce? && code_verifier.blank?
             :code_verifier
-          elsif client && !client.confidential && Doorkeeper.config.force_pkce? && code_verifier.blank?
+          elsif client && Doorkeeper.config.force_pkce? && code_verifier.blank?
```
ThisIsMissEm, line 67: Would we have a client for token_endpoint_auth_method none?

ThisIsMissEm (approved):

lib/doorkeeper/oauth/authorization_code_request.rb
55728: Yes — the `None` strategy reads `client_id` from the request body and resolves it via `by_uid_and_secret(uid, nil)`, so public clients (`token_endpoint_auth_method: none`) do have a `client` object here. The `force_pkce` check applies to them as well.

55728:
Thank you for the review @ThisIsMissEm 😊

nbulaj (approved):

nbulaj:
Thanks all :bow:

55728:

PR #1845, opened 2026-06-30
Define missing `MissingConfigurationBuilderClass` error

55728:
## Summary

`Doorkeeper::Config::Option.extended` guards against a class adopting the options DSL without defining `self.builder_class` by raising `Doorkeeper::MissingConfigurationBuilderClass`. That constant was **never defined**, so instead of the intended descriptive message, callers got a `NameError` (`uninitialized constant Doorkeeper::MissingConfigurationBuilderClass`).

This PR defines the error and fixes the reference.

## What changed

- Define the error as `Doorkeeper::Errors::MissingConfigurationBuilderClass` (a `DoorkeeperError`), matching the namespace of every other Doorkeeper error (e.g. `NoOrmCleaner`).
- Reference it correctly from the option DSL guard in `lib/doorkeeper/config/option.rb`.
- Add a regression spec (`spec/lib/errors_spec.rb`) covering the raised error.

## Before / After

**Before** — extending the DSL without a `builder_class`:

```
NameError: uninitialized constant Doorkeeper::MissingConfigurationBuilderClass
```

**After**:

```
Doorkeeper::Errors::MissingConfigurationBuilderClass:
  Define `self.builder_class` method for <Class> that returns your custom Builder class to use options DSL!
```


nbulaj (approved):
✅

nbulaj:
Thanks!

PR #1846, opened 2026-07-06
Pin ignored scope parameter on authorization_code token requests (RFC 6749 §4.1.3)

### LINKED ISSUE #1765: Token Endpoint appears to accept a scope parameter for authorization_code grant types
When using the token endpoint to exchange an authorization code for an access token, it appears that the endpoint accepts a `scope` parameter, when I can't find any reference to this in the OAuth specs.

The `scope` parameter is only used in [client_credentials](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4.2) grant type or other insecure grant types ([implicit](https://datatracker.ietf.org/doc/html/rfc6749#section-4.2.1), [resource owner password](https://datatracker.ietf.org/doc/html/rfc6749#section-4.3.2)). It is also used in the [`refresh_token` flow](https://datatracker.ietf.org/doc/html/rfc6749#section-6)

However, it is absolutely not present in the `authorization_code` grant flow: https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3

I think accepting `scope` here may be a bug, or be having unintended side-effects.

The Mastodon documentation appears to get this wrong: https://docs.joinmastodon.org/client/authorized/#token

55728:
## Summary of changes

Resolves #1765, which asked whether the token endpoint accepting a `scope` parameter for the `authorization_code` grant "may be a bug, or be having unintended side-effects".

**Investigation result: Doorkeeper already behaves correctly — this is a test-only change pinning that behavior.**

- `AuthorizationCodeRequest` never reads `parameters[:scope]`; the access token is always created with `grant.scopes`, so a `scope` sent at code-exchange time (whether a valid client scope or an unknown value) has no effect on the issued token. Silently ignoring parameters not defined for the request is the standard RFC 6749 server behavior.
- Only the `client_credentials`, `password` and `refresh_token` strategies read `parameters[:scope]` — exactly the grant types where RFC 6749 defines the parameter (§4.4.2, §4.3.2, §6). The `refresh_token` strategy additionally rejects scopes not contained in the original token, per §6.
- The token response always reports the actual granted `scope`, satisfying §5.1 (scope is REQUIRED in the response when it differs from the one the client requested).

### What changed

- Regression specs in `spec/requests/flows/authorization_code_spec.rb`: exchanging a code with `scope=write` (a valid client scope broader than the grant) or `scope=admin` (unknown) issues a token with the grant's scopes only, and the response advertises the actual scope. Any future change that starts honoring the parameter will trip these tests.
- A comment in `AuthorizationCodeRequest` documenting that not reading `scope` is deliberate.
- CHANGELOG entry (no behavior change).

Rejecting the parameter outright was considered and not pursued: RFC 6749 does not require it, and it would break existing clients (e.g. those following the Mastodon documentation referenced in the issue, which sends `scope` at the token endpoint).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ User-visible changes worth mentioning.
   - See the [Upgrade Guide](https://github.com/doorkeeper-gem/doorkeeper/wiki/Migration-from-old-versions) for detailed breaking changes, deprecations and migration steps.
 - [#1842] **[BREAKING]** `force_pkce` now requires PKCE for all clients, including confidential ones, in line with the OAuth 2.0 Security BCP (RFC 9700) and OAuth 2.1. Previously confidential clients were exempt. If you enable `force_pkce` and have confidential clients that do not yet send a `code_challenge`/`code_verifier`, their authorization requests will start to be rejected.
 - [#1845] Fix `NameError` when the config option DSL (`Doorkeeper::Config::Option`) is extended into a class that does not define `self.builder_class`. The guard raised `Doorkeeper::MissingConfigurationBuilderClass`, a constant that was never defined, so callers saw `uninitialized constant` instead of the intended message. The error is now defined as `Doorkeeper::Errors::MissingConfigurationBuilderClass` (a `DoorkeeperError`) and referenced correctly.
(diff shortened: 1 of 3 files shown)


nbulaj (approved):
LGTM ✅

PR #1848, opened 2026-07-09
Derive token_endpoint_auth_methods_supported from client_authentication config

55728:
Resolves the FIXME left in #1838 (now that #1840 has landed).

The RFC 8414 metadata endpoint previously hardcoded `token_endpoint_auth_methods_supported` as `["none", "client_secret_basic", "client_secret_post"]`. With the client authentication registry from #1840, this is now derived from `Doorkeeper.config.client_authentication`, so:

- The default behaviour is unchanged (the three built-in methods are advertised).
- Servers that explicitly configure `client_authentication` see their metadata match what's actually enabled.
- Extension-registered methods (e.g. `private_key_jwt`) automatically appear when added to the config.


nbulaj (approved):
👍

nbulaj (approved):
Thanks!

PR #1849, opened 2026-07-09
Support RFC 9207 Authorization Server Issuer Identification

### LINKED ISSUE #1720: Support for RFC 9207 - OAuth 2.0 Authorization Server Issuer Identification

55728:
Closes #1720.

## Summary

Adds support for [RFC 9207 (OAuth 2.0 Authorization Server Issuer Identification)](https://www.rfc-editor.org/rfc/rfc9207.html). The authorization server now includes an `iss` parameter (its issuer identifier) in authorization responses, which lets clients that talk to more than one authorization server confirm which one produced a given response — the countermeasure RFC 9207 defines against **mix-up attacks**.

This builds directly on the metadata work from #1838: it reuses the existing `issuer` configuration and needs **no new options**.

## Opt-in via `issuer`

The feature is gated on the existing `issuer` config being set. When `issuer` is `nil` (the default), no `iss` parameter is emitted and existing authorization and token responses are unchanged; the RFC 8414 metadata simply advertises the capability as `authorization_response_iss_parameter_supported: false`. Set an `issuer` and RFC 9207 turns on:

```ruby
Doorkeeper.configure do
  issuer "https://auth.example.com"
end
```

Authorization and token responses stay byte-for-byte identical by default, which matches Doorkeeper's explicit opt-in style.

## What carries `iss`

RFC 9207 §2 scopes the `iss` parameter to the authorization responses returned **to the client**, so the parameter is added exactly there and left off everywhere else:

| Response | `iss`? |
| --- | --- |
| Successful authorization response redirected to the client (code / implicit, query or fragment, incl. `form_post` and `api_only`) | ✅ |
| Error response redirected to the client's `redirect_uri` (e.g. `access_denied`, `invalid_scope`) | ✅ |
| Non-redirectable authorization errors (`invalid_client`, `invalid_redirect_uri`, …) rendered to the user | ❌ |
| Out-of-band flows that render the code/token on a server page instead of redirecting | ❌ |
| Token, introspection and protected-resource responses (they share `ErrorResponse` but are not authorization responses) | ❌ |

Error responses gate `iss` on `redirectable?`, which is precisely "will be redirected to the client", so the split above falls out of a single condition. The OOB path already omits `state` for the same reason.

## Metadata

When an issuer is configured, the RFC 8414 metadata advertises `authorization_response_iss_parameter_supported: true` (and an explicit `false` otherwise), matching the exact condition under which `iss` is emitted.

## Issuer value validation (boot-time warnings)

RFC 8414 (the metadata `issuer`) and RFC 9207 (the `iss` parameter) both expect an `https` URL with a host and no query or fragment. Rather than silently emitting a non-conformant value, Doorkeeper now logs a warning at configuration time (a warning, **not** a hard failure, so local setups on e.g. `http://localhost` keep working) when the configured `issuer`:

- is **not RFC-compliant** — not `https`, host-less (`https:foo`, `https://`), or contains a query/fragment; or
- is **not discoverable** — has a path component (`https://host/tenant`). Doorkeeper serves its metadata only at the root `/.well-known/oauth-authorization-server`, so an RFC 8414 client would derive a different (unreachable) well-known URL, and the advertised issuer would not match its lookup. Serving path-based (multi-tenant) well-known routes is left as a possible follow-up on the #1838 metadata endpoint.

Any userinfo in the value is redacted before it is logged, so a misconfigured `user:pass@host` does not leak credentials.

`custom_metadata` is intentionally still able to override computed metadata fields (including `authorization_response_iss_parameter_supported`) — it is a deliberate escape hatch, and keeping it consistent with actual behaviour is the operator's responsibility. This is documented in the generated initializer.

## Commits

1. **Support RFC 9207 Authorization Server Issuer Identification** — the `iss` parameter, its scoping, the metadata flag, and specs across the code / error / metadata / controller / flow layers (query, fragment, `form_post`, `api_only`, OOB branches).
2. **Warn on non-compliant or non-discoverable issuer configuration** — the boot-time issuer warnings and their specs.


lib/doorkeeper/oauth/error_response.rb
```diff
@@ -38,13 +38,22 @@ def initialize(attributes = {})
         @exception_class = attributes[:exception_class]
         @redirect_uri = attributes[:redirect_uri]
         @response_on_fragment = attributes[:response_on_fragment]
+        @issuer = attributes[:issuer]
```
nbulaj, line 41: Should we actually pass this value? It's resolved via `Doorkeeper.config.issuer` (aka singleton) , is there a case when we will pass another value here?

55728: Good question 👀✨ — the explicit pass is the scoping mechanism. `ErrorResponse` is shared across all endpoints (token requests via `BaseRequest`, exception rescue in `Helpers::Controller`), but RFC 9207 `iss` is for authorization responses only. If the class read the singleton directly, a configured issuer would leak into token endpoint error bodies. Only `PreAuthorization#error_response` supplies it, keeping `iss` scoped to the authorization endpoint. There's a spec pinning this behavior in `error_response_spec.rb`.

nbulaj: Ah, OK, got it now :+1:  thanks!

nbulaj (approved):
🪨 🚀 

Image

55728:
Thanks for the rock-solid review! 🪨🚀

nbulaj:
That's all to Copilot, I'm left with crumbs to review :bread: :slightly_smiling_face: .

55728:
No-no your `ErrorResponse` question caught a real hidden flavor! 🍞😆✨

PR #1851, opened 2026-07-10
Fix duplicate query parameter in the authorization callback

55728:
### Problem

`Doorkeeper::OAuth::Authorization::URIBuilder.uri_with_query` builds the authorization callback by merging the redirect_uri's existing query with the response parameters (`code`, `state`, `error`, …):

```ruby
original_query = Rack::Utils.parse_query(uri.query)   # => string keys
uri.query = build_query(original_query.merge(parameters))  # parameters => symbol keys
```

`Rack::Utils.parse_query` returns **string** keys, while `parameters` uses **symbol** keys. Because the key types differ, a collision cannot be deduped by `merge`. If a client's registered `redirect_uri` already contains a parameter that also appears in the response — most notably `state` — the callback URL ends up with that parameter **twice**:

```
https://app/cb?state=fixed&code=abc&state=user-state
```

Registered redirect URIs with query parameters are an explicitly supported configuration (see `URIChecker#query_matches?`), so this is reachable in normal use.

### Fix

Normalize the response parameter keys to strings before merging, so the response value overrides any same-named parameter baked into the registered redirect_uri and appears exactly once:

```ruby
uri.query = build_query(original_query.merge(parameters.transform_keys(&:to_s)))
```

Non-colliding registered query parameters continue to be preserved (existing behavior, still covered by its spec). This is a localized correctness fix — a malformed/duplicated callback parameter — with no config or API change.


nbulaj (approved):
Makes sense

55728:
Thanks for the review! 👀✨

PR #1852, opened 2026-07-10
Anchor `pkce_code_challenge_methods` validation with `\A..\z`

55728:
### Problem

`Config::Validations#validate_pkce_code_challenge_methods` validates each configured method against a regex that uses **line** anchors:

```ruby
pkce_code_challenge_methods.all? { |method| method =~ /^plain$|^S256$/ }
```

In Ruby `^` and `$` match at line boundaries, not string boundaries. So a multi-line value like `"plain\ngarbage"` matches (the first line is `plain`) and passes validation. The invalid value is then retained in `pkce_code_challenge_methods` — a challenge method that can never match a real request — instead of being rejected and reset to the default `["plain", "S256"]`.

### Fix

Use string anchors so the whole value must be exactly `plain` or `S256`:

```ruby
pkce_code_challenge_methods.all? { |method| method =~ /\A(?:plain|S256)\z/ }
```

Low impact (the value is developer-controlled config), but the validator is meant to catch exactly this kind of bad input and currently lets it through.


nbulaj (approved):
LGTM 👍

Issue #1693, opened 2024-02-20
Different access grants return the same access token with `reuse_access_token` enabled


pyromaniac:
Hello folks,

First of app thanks for a great gem and all your efforts!

### Steps to reproduce
We have stumbled upon a problem in out system and not sure how to handle this correctly.

We are now on Doorkeeper 5.6.8 and have `reuse_access_token` enabled.

My issue is that if I receive 2 different access grants, and try to receive access tokens with them - I get the same access token.

Imagine that you have 2 developers that are working on the same OAuth client at the same time, on their local machines they get an access grant using the consent screen and the authorize button, they they each try to get an access token but they get the same one, so their flows now interdependent. If one refreshes their token, the second one will end up with a revoked token at hands.

```
https://www.example.com/oauth/authorize?client_id=CLIENT_ID&redirect_uri=urn:ietf:wg:oauth:2.0:oob&response_type=code

=> 9a953a61513ff907a72f11cb0a5199fc67c77fe244c81069f10439d1b44b3862

curl -X POST -d "client_id=CLIENT_ID&client_secret=CLIENT_SECRET&code=9a953a61513ff907a72f11cb0a5199fc67c77fe244c81069f10439d1b44b3862&grant_type=authorization_code&redirect_uri=urn:ietf:wg:oauth:2.0:oob" "https://www.example.com/oauth/token"

=> {"access_token":"e04a3044f4e80272580dc1a56b76fa60fa2e32dc4b3e6a39a58fbb3465e77980","token_type":"Bearer","expires_in":7200,"refresh_token":"b0bd59025a4a3ea1c908e49912e786d39d3902d889915ba2e523981feac50d7f","scope":"public","created_at":1708356984}

# the same query
curl -X POST -d "client_id=CLIENT_ID&client_secret=CLIENT_SECRET&code=9a953a61513ff907a72f11cb0a5199fc67c77fe244c81069f10439d1b44b3862&grant_type=authorization_code&redirect_uri=urn:ietf:wg:oauth:2.0:oob" "https://www.example.com/oauth/token"

=> {"access_token":"e04a3044f4e80272580dc1a56b76fa60fa2e32dc4b3e6a39a58fbb3465e77980","token_type":"Bearer","expires_in":7189,"refresh_token":"b0bd59025a4a3ea1c908e49912e786d39d3902d889915ba2e523981feac50d7f","scope":"public","created_at":1708356984}



[...]

PR #1853, opened 2026-07-13
Fix token reuse ignoring custom attributes when the request has none

55728:
## Summary

With `reuse_access_token` enabled and `custom_access_token_attributes` configured, a token request that doesn't specify any custom attributes could be handed back an existing token that was created *with* custom attribute values — a partition mix-up within the same application and resource owner.

```ruby
Doorkeeper.configure do
  reuse_access_token
  custom_access_token_attributes [:tenant_name]
end

AccessToken.find_or_create_for(application: app, resource_owner: owner, scopes: scopes, tenant_name: "acme")
# => token A (tenant_name: "acme")

AccessToken.find_or_create_for(application: app, resource_owner: owner, scopes: scopes)
# before: token A is reused, tenant_name "acme" leaks into an attribute-less request
# after:  a new token without custom attributes is created
```

## Cause

`AccessTokenMixin.find_or_create_for` and the client credentials `Creator` both collapse the extracted custom attributes with `.presence`, so `{}` (attributes configured, but not supplied in this request) becomes `nil`. `custom_attributes_match?` treats `nil` as "ignore custom attributes entirely" (`return true if custom_attributes.nil?`) and matches any token regardless of its stored values.

The bug is one-directional: requests that *do* supply custom attributes were already compared correctly and are unaffected.

## Fix

Keep the empty hash intact instead of collapsing it to `nil`. `custom_attributes_match?` already handles `{}` correctly — each configured attribute is compared, so a token holding values no longer matches a request that supplies none, while two attribute-less tokens still match each other. The `nil` = "ignore custom attributes" contract is preserved for callers that rely on it (e.g. `AuthorizationsController#matching_token?` during pre-auth).

The change is deliberately confined to the two lookup call sites (plus docs/comments) so it can be backported to 5.x cleanly.

Note this is distinct from #1693, which is about `reuse_access_token` returning the same token for different access grants — grants aren't part of the matching there, whereas this PR is about configured custom attributes being ignored during matching.


nbulaj:
**_Note_**: backport into https://github.com/doorkeeper-gem/doorkeeper/tree/v5.9-stable when merged

nbulaj (approved):
LGTM ✅

PR #1850, opened 2026-07-10
Match access token reuse to the request's refresh token requirement

55728:
Fixes #1731

### Problem

When `reuse_access_token` is enabled, `AccessToken.find_or_create_for` returns any matching non-expired token for the application / resource owner / scopes. It never checked whether that token carries a refresh token.

So when a request expects a refresh token (an `authorization_code` or `password` grant while `use_refresh_token` is enabled) but the matching reusable token was issued **without** one, Doorkeeper reused it and the token response silently omitted the refresh token — even though the client is entitled to one.

This matches the report in #1731: requesting a single scope returned a refresh token, but requesting two scopes returned only an access token, because the reusable token that happened to match the wider scope set had no refresh token.

The reverse mismatch exists too: a request that does **not** ask for a refresh token could reuse a matching token that *carries* one and return it — an unrequested refresh token in the response. Since `token_response` includes `refresh_token` whenever the token has one, this is reachable when `refresh_token_enabled` is a per-request callable that varies by grant/context.

### Fix

`find_or_create_for` now reuses a matching token only when its refresh token presence matches what the **current** request asks for, in both directions:

- If the request asks for a refresh token (`use_refresh_token: true`), the candidate must already carry one; otherwise a fresh token is created (the #1731 case).
- If the request does not ask for a refresh token, the candidate must not carry one either; otherwise a fresh token is created, so the response never returns a refresh token the request didn't ask for.

```ruby
def refresh_token_matches?(access_token, token_attributes)
  access_token.refresh_token.present? == !!token_attributes[:use_refresh_token]
end
```

The decision keys off what the current request asks for (`use_refresh_token` in `token_attributes`, not global config), so grants that legitimately issue no refresh token are unaffected in steady state — a fresh token is only minted when a matching candidate's refresh token presence disagrees with the request.

PR #1855, opened 2026-07-15
Pre-release hardening: redirect params, token reuse matching, replica-safe secret upgrades

55728:
## Summary

Three small fixes to unreleased changes (#1851, #1850, #1801) plus CHANGELOG label corrections for #1840, ahead of the major release. Each fix is independent; happy to split this PR if preferred.

### 1. Retain registered `redirect_uri` params when the response param is blank

The key normalization from #1851 (which dedupes colliding query parameters) let a *blank* response parameter clobber a same-named parameter baked into the registered `redirect_uri`. Registered URI `https://client.example/cb?state=static`, authorization request without `state`:

- v5.9.3: redirects to `cb?state=static&code=...`
- before this fix: redirects to `cb?code=...` — the registered param is silently dropped

RFC 6749 §3.1.2 requires the registered query component to be retained when adding parameters, so blank response parameters are now dropped *before* the merge instead of after it. Non-blank response parameters still override the registered ones (the #1851 behavior).

### 2. Apply the refresh token requirement while matching reusable tokens

#1850 checks the refresh token presence of the reuse candidate, but only against the single newest matching token. When that token carries the wrong presence (issued through a grant with a different `use_refresh_token`), reuse failed even though an older matching token satisfied the request — e.g. a user alternating implicit and authorization_code grants with identical scopes created a fresh token on every request, forever.

The requirement now participates in the matching itself (as a filter passed to `matching_token_for` / `find_matching_token`, whose signatures stay backward compatible — the block is optional), so the newest token that fully satisfies the request is reused and a fresh token is created only when none does.

### 3. Route fallback secret upgrades through the primary database role

`upgrade_fallback_value` (used by both the token fallback lookup and, since #1801, `Application#secret_matches?`) writes during what is otherwise a read path. With `enable_multiple_database_roles` and automatic role switching, a lookup routed to a read replica made the upgrade `update` raise. The write is now wrapped in `with_primary_role`, mirroring how token creation on GET requests is handled, and `ApplicationMixin` includes the `WriteToPrimary` concern so the application secret fallback benefits alongside the token one.

### 4. CHANGELOG label corrections for #1840

`Doorkeeper::OAuth::Client::Credentials.from_request` / `.from_basic` / `.from_params` are removed, not deprecated — extensions calling them (the historical `doorkeeper-grants_assertion` pattern) raise `NoMethodError` at request time, and `deprecate_constant` emits no warning unless `Warning[:deprecated]` is enabled. The removal is now flagged **[BREAKING]** (only the constant alias remains deprecated), and the `none` strategy bullet calls out the most likely real-world casualty: a public client POSTing to `/oauth/revoke` with a body `client_id` while still sending its `Authorization: Bearer <token>` header.


lib/doorkeeper/models/access_token_mixin.rb
```diff
@@ -92,13 +92,16 @@ def revoke_all_for(application_id, resource_owner, clock = Time)
...
       #   only match tokens that have no custom attributes set.
       #
+      # @yield [token] optional additional predicate a token must satisfy to
+      #   count as a match (e.g. the refresh token requirement of the request).
+      #
       # @return [Doorkeeper::AccessToken, nil] Access Token instance or
       #   nil if matching record was not found
       #
-      def matching_token_for(application, resource_owner, scopes, custom_attributes: nil, include_expired: true)
+      def matching_token_for(application, resource_owner, scopes, custom_attributes: nil, include_expired: true, &filter)
```
nbulaj, line 101: This can be breaking for ORM extensions, maybe we have to mark API change for the ORM model(s)? 👀  It's OK since next release is a 6.0, just wanna to highlight the fact of the change

55728: Good point 👀✨ — the block is optional so existing callers are unaffected, but ORM extensions overriding `matching_token_for` or `find_matching_token` would need to accept it. I'll add a note to the CHANGELOG entry highlighting the signature change 🚀

55728: Done — added a note to the CHANGELOG entry in bd44a98 ✏️

nbulaj (approved):
Oh that missed secret storable 🙈 

Thanks!

55728:
I reviewed #1801 too so equally guilty 🤯 Glad we caught it before the release! 😌

Issue #1858, opened 2026-07-16
Introspecting a refresh token bound to an expired access token returns `active: false` even though the token is still usable


55728:
Split out from #1671 (the revocation half was fixed by #1744 in v5.8.1; this is the remaining introspection half, originally noted in https://github.com/doorkeeper-gem/doorkeeper/issues/1671#issuecomment — @syakovyn's comment).

### Steps to reproduce

1. Create an access token with `use_refresh_token: true` and let the access token expire (the refresh token has no expiry of its own and is still valid).
2. `POST /oauth/introspect` with `token=<refresh_token>` → responds `{"active": false}`.
3. `POST /oauth/token` with `grant_type=refresh_token&refresh_token=<same token>` → responds 200 and issues a new access token.

Reproduced on current `main` (5.9.x):

```ruby
# spec/requests/..._spec.rb
it "reports active: false for a still-usable refresh token" do
  access_token = FactoryBot.create(
    :access_token,
    application: client,
    resource_owner_id: resource_owner.id,
    resource_owner_type: resource_owner.class.name,
    use_refresh_token: true,
  ).tap { |t| t.update!(created_at: t.created_at - t.expires_in - 1) }

  post introspection_endpoint_url,
       params: { token: access_token.refresh_token },
       headers: { "HTTP_AUTHORIZATION" => basic_auth_header_for_client(client) }
  # => {"active":false}

  post refresh_token_endpoint_url,
       params: refresh_token_endpoint_params(client: client, refresh_token: access_token.refresh_token)
  # => 200, new access_token + refresh_token issued
end
```

### Expected behavior

Per RFC 7662 §2.2, the `active` field describes the state of the token presented in the request. A refresh token that has not been revoked (and that the token endpoint will happily accept) should introspect as `active: true`.

### Actual behavior

`active: false`. A resource server or ops tooling relying on introspection is told the token is dead while it can still mint new access tokens. (The failure direction is at least fail-closed for resource servers, but it is misleading for anything that audits or manages refresh tokens via introspection.)

### Root cause

`TokensController#introspect` resolves the presented token via `by_token || by_refresh_token`, so a refresh token resolves to its paired `AccessToken` record. `Doorkeeper::OAuth::TokenIntrospection#valid_token?` then checks `@token&.accessible?` — i.e. the *access token's* expiry — regardless of which credential was actually presented:

[...]

PR #1859, opened 2026-07-16
[test] Add regression specs for revoking refresh tokens of expired access tokens

### LINKED ISSUE #1671: It is not possible to revoke refresh token bound to the expired access token
A copy of https://github.com/doorkeeper-gem/doorkeeper/issues/1579 but with a focus on refresh tokens.

### Steps to reproduce

Request to revoke a refresh token bound to the expired access token.

### Expected behavior
The refresh token is revoked and can not be used to get a new access token.

### Actual behavior
The refresh token is not revoked and can be used to get a new access token.

### System configuration

**Doorkeeper initializer**:

```ruby
Doorkeeper.configure do
  ...
  use_refresh_token
  use_polymorphic_resource_owner
  ...
end
```

**Ruby version**: `3.1.4`

**Gemfile.lock**:

<details>
  <summary>Gemfile.lock content</summary>
  
```
...
doorkeeper (5.6.6)
...
```
</details>

### Reproduction tests

RSpec test to show the issue (simplified):
```ruby
# Simplified User that has no validations and has only an ID column. That is enough to reproduce the issue
class User < ActiveRecord::Base
end

# The spec
require 'rails_helper'

describe Doorkeeper::TokensController, type: :request do
  it 'revokes refresh token bound to expired access token' do
    mobile_app = Doorkeeper::Application.create!(name: 'Mobile App', redirect_uri: 'mobile://callback', confidential: false)
    user = User.create!
    access_token =  Dookeeper::AccessToken.create!(resource_owner: user, application: mobile_app, use_refresh_token: true, expires_in: 0) # Using expires_in = 0 to create an expired access token

    post '/oauth/revoke', params: { client_id: mobile_app.uid, token: access_token.refresh_token, token_type_hint: 'refresh_token' }

    expect(response).to have_http_status :ok
    expect(access_token.reload).to be_revoked # That means the refresh token remains valid despite the fact that the revoke endpoint returned OK
  end
end
```

55728:
## Summary

Issue #1671 reported that a refresh token bound to an already-expired access token could not be revoked: `POST /oauth/revoke` returned 200 OK, but the refresh token remained usable at the token endpoint. This was fixed in v5.8.1 by #1744, which routes refresh-token revocations through `Doorkeeper::RevocableTokens::RevocableRefreshToken`, whose `revocable?` only checks the revoked flag and no longer depends on the paired access token being unexpired.

The fix landed without a request-level regression spec covering the reporter's exact scenario, so this PR pins it end to end:

- A public client revokes the refresh token of an expired access token via `POST /oauth/revoke` with `token_type_hint=refresh_token` → 200, and the token is actually revoked.
- A subsequent `grant_type=refresh_token` request with the revoked refresh token is rejected with `invalid_grant`.

The second example is the part of the original report that had no coverage at all: `spec/controllers/tokens_controller_spec.rb` already asserts the revoked flag, but nothing verified that the revoked refresh token is rejected by the token endpoint afterwards.

Test-only change, no behavior change.

## Note

The second concern raised in the issue thread (introspecting a refresh token bound to an expired access token responds with `active: false` even though the token is still usable) is a separate, still-unfixed problem, now split out into #1858.

Closes #1671

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ User-visible changes worth mentioning.
 - [#1854] Fix the RFC 8414 metadata endpoint raising `ActionController::UrlGenerationError` (HTTP 500) when `use_doorkeeper` configures a custom controller whose namespace depth differs from `doorkeeper/metadata` (e.g. `controllers tokens: "custom_tokens"`).
 - [#1855] Perform the fallback secret upgrade-on-access write (plain → hashed token or application secret) through the primary database role, so `enable_multiple_database_roles` setups no longer attempt the write on a read replica when the lookup happens in a request routed to the reading role.
 - [#1857] Pin with regression specs that a `+` between scopes in a form-encoded token request is decoded as a space (so `scope=public+write` refreshes fine), while a percent-encoded literal `+` (`%2B`) names a single scope and is rejected when unknown, per RFC 6749 §3.3. Test-only change, closes [#1686].
+- [#1859] Pin with regression specs that a refresh token bound to an expired access token can be revoked (fixed by [#1744]) and that the revoked refresh token is rejected at the token endpoint afterwards. Test-only change, closes [#1671].
 - Please add here
 
 ## 5.9.3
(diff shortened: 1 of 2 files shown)


nbulaj (approved):
Tests again 😍

55728:
Thanks @nbulaj! 😆✨

ThisIsMissEm:
Nice one!

PR #1860, opened 2026-07-16
Fix introspection of refresh tokens bound to expired access tokens

### LINKED ISSUE #1858: Introspecting a refresh token bound to an expired access token returns `active: false` even though the token is still usable

55728:
## Summary

Fixes #1858 (split out from #1671): introspecting a refresh token whose paired access token had expired returned `{"active": false}`, while the token endpoint still accepted the same refresh token and issued new tokens. Per RFC 7662 §2.2 the `active` field describes the token presented in the request, but `TokenIntrospection#valid_token?` always checked `AccessToken#accessible?` — i.e. the *access token's* expiry — regardless of which credential was presented.

## Changes

`TokensController` already knows which credential it resolved (`by_token` vs `by_refresh_token`, honoring `token_type_hint`), so it now passes that along via a new `token_type:` keyword on `TokenIntrospection#initialize`. The keyword defaults to `:access_token`, so existing callers and subclasses keep the previous behavior.

When the presented token is a refresh token:

- `active` is based on `revoked?` alone — mirroring how the token endpoint (`RefreshTokenRequest`) and revocation (`RevocableTokens::RevocableRefreshToken`, #1744) already treat refresh tokens.
- The success response omits `token_type` and `exp`. Both are OPTIONAL per RFC 7662 §2.2 and describe the paired access token: a refresh token is not a Bearer credential and has no expiry of its own. Without this, an expired-but-active introspection would report `active: true` alongside an `exp` in the past, which is contradictory for resource servers that honor `exp`.

Behavior note: introspecting a *live* refresh token previously returned `token_type: "Bearer"` and the paired access token's `exp`; those two fields are no longer included for presented refresh tokens (mentioned in the CHANGELOG entry).

### DIFF
--- a/app/controllers/doorkeeper/tokens_controller.rb
+++ b/app/controllers/doorkeeper/tokens_controller.rb
@@ -34,7 +34,7 @@ module Doorkeeper
 
     # OAuth 2.0 Token Introspection - https://datatracker.ietf.org/doc/html/rfc7662
     def introspect
-      introspection = OAuth::TokenIntrospection.new(server, token)
+      introspection = OAuth::TokenIntrospection.new(server, token, token_type: presented_token_type)
 
       if introspection.authorized?
         render json: introspection.to_json, status: 200
@@ -136,6 +136,15 @@ module Doorkeeper
       return unless token
 
       RevocableTokens::RevocableRefreshToken.new(token)
+    end
+
+    # Revocation (RFC 7009) and introspection (RFC 7662) deliberately share
+    # the same request shape — `token` plus an optional `token_type_hint` —
+    # so both actions resolve the presented token through #revocable_token.
+    # The wrapper type doubles as the record of which credential the client
+    # actually presented.
+    def presented_token_type
+      revocable_token.is_a?(RevocableTokens::RevocableRefreshToken) ? :refresh_token : :access_token
     end
 
     def access_token
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ User-visible changes worth mentioning.
 - [#1855] Perform the fallback secret upgrade-on-access write (plain → hashed token or application secret) through the primary database role, so `enable_multiple_database_roles` setups no longer attempt the write on a read replica when the lookup happens in a request routed to the reading role.
 - [#1857] Pin with regression specs that a `+` between scopes in a form-encoded token request is decoded as a space (so `scope=public+write` refreshes fine), while a percent-encoded literal `+` (`%2B`) names a single scope and is rejected when unknown, per RFC 6749 §3.3. Test-only change, closes [#1686].
 - [#1859] Pin with regression specs that a refresh token bound to an expired access token can be revoked (fixed by [#1744]) and that the revoked refresh token is rejected at the token endpoint afterwards. Test-only change, closes [#1671].
+- [#1860] Fix introspection of refresh tokens (RFC 7662): a refresh token bound to an expired access token now introspects as `active: true`, matching the token endpoint which still accepts it. The introspection response for a presented refresh token no longer includes the paired access token's `token_type` and `exp`. Fixes [#1858].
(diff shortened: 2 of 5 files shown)


app/controllers/doorkeeper/tokens_controller.rb
```diff
@@ -34,7 +34,7 @@ def revoke
 
     # OAuth 2.0 Token Introspection - https://datatracker.ietf.org/doc/html/rfc7662
     def introspect
-      introspection = OAuth::TokenIntrospection.new(server, token)
+      introspection = OAuth::TokenIntrospection.new(server, token, presented_as: presented_token_type)
```
nbulaj: WDYT simplifying argument to just `token_type` ?

55728: Good call 👀✨ — renamed to token_type:, force-pushed. It also lines up nicely with the controller-side helper presented_token_type now.

nbulaj: Yeah I believe looks more clean , thanks a lot!

nbulaj (approved):

PR #1861, opened 2026-07-19
Add jwt-bearer grant (ID-JAG) support

manmohan-shaw-okta:
## Summary

Adds a `jwt-bearer` grant (`urn:ietf:params:oauth:grant-type:jwt-bearer`, RFC 7523) profiled by the [Identity Assertion Authorization Grant (ID-JAG) draft](https://datatracker.ietf.org/doc/draft-ietf-oauth-identity-assertion-authz-grant/), letting Doorkeeper act as the **Resource Authorization Server** side of a Cross App Access exchange: it verifies an ID-JAG assertion minted by an external Identity Provider and issues a locally-scoped access token. Minting the ID-JAG itself (the IdP side, RFC 8693 Token Exchange) is out of scope.

This mirrors equivalent PRs already up for the Node.js (`node-oauth2-server`), Python (`authlib`) and Go (`go-oauth2/oauth2`) OAuth2 libraries as part of the same cross-ecosystem Cross App Access rollout.

## Design

- **Not enabled by default** — add `:jwt_bearer` to `grant_flows` to opt in, same as every other non-default flow.
- **Client authentication is inherited "for free".** Doorkeeper resolves and authenticates the client (including `client_secret` verification) generically in `Server#client`, *before* any grant-specific request class runs — same as every other grant type. This new grant additionally checks `client.confidential` and rejects public clients by default per ID-JAG §8.1 (relaxable via `jwt_bearer_allow_public_clients`, not recommended for production).
- **Uses the `jwt` gem** (new runtime dependency) rather than hand-rolling JWS verification — it already provides `aud`/`exp`/`nbf` claim validation with configurable leeway and a `verify_jti` hook, which this grant wires up to an optional pluggable replay store.
- **Scope is narrowed, never expanded**: the assertion's own `scope` claim (if present) is a ceiling on what the request can ask for, matching RFC 6749 §3.3 narrowing semantics used elsewhere in this gem (e.g. refresh tokens).
- **Never issues a refresh token** (ID-JAG §4.4.3) — bypasses the generic `find_or_create_access_token` helper (whose `use_refresh_token` derives from the global `refresh_token_enabled?` setting) and calls the token model directly with `use_refresh_token: false`, the same approach `ClientCredentialsRequest` already uses.
- **Non-leaky errors**: every assertion-validation failure (bad `typ`, disallowed `alg`, untrusted issuer, unresolvable key, bad signature, claim mismatch, replay) funnels into the same `Result` failure value and the same `Errors::InvalidGrant`, so a response can't be used to distinguish which check failed.
- New `jwt_bearer_*` config/options live in their own file (`lib/doorkeeper/config/jwt_bearer.rb`, reopening `Doorkeeper::Config`) rather than growing the already-near-the-limit `lib/doorkeeper/config.rb` further.

## Involved parts of the project

- `lib/doorkeeper/oauth/helpers/jwt_bearer_assertion.rb` (new) — the decode/verify pipeline.
- `lib/doorkeeper/oauth/jwt_bearer_access_token_request.rb`, `lib/doorkeeper/request/jwt_bearer.rb` (new) — the grant/strategy, following the existing `BaseRequest`/`Strategy` pattern (modeled closely on `PasswordAccessTokenRequest`/`ClientCredentialsRequest`).
- `lib/doorkeeper/grant_flow.rb` — registers the new flow.
- `lib/doorkeeper/config/jwt_bearer.rb` (new), `lib/doorkeeper/config/validations.rb` — new config options + a boot-time warning if `:jwt_bearer` is enabled without `jwt_bearer_audience` configured.
- `lib/doorkeeper.rb`, `lib/doorkeeper/oauth.rb` — autoload wiring + the new `JWT_BEARER` grant-type constant.
- `doorkeeper.gemspec` — new runtime dependency on `jwt` (`>= 2.7`).

**OAuth2 workflow involved:** token endpoint only (`grant_type=urn:ietf:params:oauth:grant-type:jwt-bearer`). No changes to the authorization or revocation/introspection endpoints.

## Added tests?

Yes:
- `spec/lib/oauth/helpers/jwt_bearer_assertion_spec.rb` — unit coverage of the decode/verify helper (23 examples): RS256/ES256/PS256, `typ` confusion, `alg` allow-list (`none`/`HS256`), untrusted issuer, unresolvable key, tampered signature, audience mismatch, expired/future/clock-skew-boundary timestamps, missing claims, `client_id` mismatch, replay (with and without a configured store), PEM/JWK/multi-candidate-key resolver shapes.
- `spec/lib/oauth/jwt_bearer_access_token_request_spec.rb` — request-class unit tests (13 examples): client/public-client/assertion/resource-owner/authorize-hook validation, scope narrowing, no-refresh-token issuance.
- `spec/requests/endpoints/jwt_bearer_spec.rb` — full end-to-end `POST /oauth/token` coverage (9 examples), including proving client authentication is genuinely enforced (**wrong `client_secret` is rejected even with an otherwise-valid assertion**).

`bundle exec rake spec` (1536 examples, 0 failures) and `bundle exec rubocop` are clean on all touched files (the one remaining `Lint/MissingSuper` warning on the new request class matches the same, already-unaddressed pattern in `PasswordAccessTokenRequest`).

## OAuth2 standard

- [RFC 7523](https://www.rfc-editor.org/rfc/rfc7523) — JWT Bearer grant.
- [ID-JAG draft](https://datatracker.ietf.org/doc/draft-ietf-oauth-identity-assertion-authz-grant/) — the profile this grant implements (type header, required claims, audience-as-issuer-identifier, scope narrowing, no-refresh-token, confidential-client-only, replay protection referenced directly in code comments).
- RFC 6749 §5.2 — all failures map to existing `Errors::*` classes; no new error classes were needed.

## Reproduction

```ruby
Doorkeeper.configure do
  grant_flows %w[jwt_bearer]
  jwt_bearer_audience "https://your-resource-server.example.com"
  jwt_bearer_trusted_issuer { |issuer| MyIdpRegistry.trusted?(issuer) }
  jwt_bearer_issuer_key { |issuer, kid| MyIdpRegistry.jwks_for(issuer, kid) }
  jwt_bearer_resource_owner_from_assertion { |issuer, subject, client| User.find_by(federated_id: subject) }
end
```
```
POST /oauth/token
grant_type=urn:ietf:params:oauth:grant-type:jwt-bearer&assertion=<ID-JAG JWT>
```
See `spec/requests/endpoints/jwt_bearer_spec.rb` for a fully worked example (mints a self-signed test assertion and exchanges it end-to-end).

---

Opening as a **Draft PR** for early feedback on the design (particularly the new `jwt` dependency and the confidential-client-only default) before polishing further.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ User-visible changes worth mentioning.
 
 ## main
 
-- [#PR] Add the `jwt-bearer` grant (`urn:ietf:params:oauth:grant-type:jwt-bearer`, RFC 7523), profiled by the Identity Assertion Authorization Grant (ID-JAG) draft, letting Doorkeeper act as the Resource Authorization Server side of a Cross App Access exchange: it verifies an ID-JAG assertion minted by an external Identity Provider and issues a locally-scoped access token, without ever issuing a refresh token.
+- [#1861] Add the `jwt-bearer` grant (`urn:ietf:params:oauth:grant-type:jwt-bearer`, RFC 7523), profiled by the Identity Assertion Authorization Grant (ID-JAG) draft, letting Doorkeeper act as the Resource Authorization Server side of a Cross App Access exchange: it verifies an ID-JAG assertion minted by an external Identity Provider and issues a locally-scoped access token, without ever issuing a refresh token.
   - Not enabled by default — add `:jwt_bearer` to `grant_flows` to opt in.
   - Restricted to confidential clients by default per ID-JAG §8.1 (relax with `jwt_bearer_allow_public_clients`, not recommended for production).
   - New required config hooks: `jwt_bearer_trusted_issuer`, `jwt_bearer_issuer_key`, `jwt_bearer_resource_owner_from_assertion`. New optional hooks/options: `jwt_bearer_authorize`, `jwt_bearer_replay_store`, `jwt_bearer_audience` (required to actually issue tokens), `jwt_bearer_clock_skew`, `jwt_bearer_allowed_algorithms`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ User-visible changes worth mentioning.
 
 ## main
 
+- [#PR] Add the `jwt-bearer` grant (`urn:ietf:params:oauth:grant-type:jwt-bearer`, RFC 7523), profiled by the Identity Assertion Authorization Grant (ID-JAG) draft, letting Doorkeeper act as the Resource Authorization Server side of a Cross App Access exchange: it verifies an ID-JAG assertion minted by an external Identity Provider and issues a locally-scoped access token, without ever issuing a refresh token.
+  - Not enabled by default — add `:jwt_bearer` to `grant_flows` to opt in.
+  - Restricted to confidential clients by default per ID-JAG §8.1 (relax with `jwt_bearer_allow_public_clients`, not recommended for production).
+  - New required config hooks: `jwt_bearer_trusted_issuer`, `jwt_bearer_issuer_key`, `jwt_bearer_resource_owner_from_assertion`. New optional hooks/options: `jwt_bearer_authorize`, `jwt_bearer_replay_store`, `jwt_bearer_audience` (required to actually issue tokens), `jwt_bearer_clock_skew`, `jwt_bearer_allowed_algorithms`.
+  - Adds a new runtime dependency on the `jwt` gem for JWS decode/verify (RS256, ES256, PS256).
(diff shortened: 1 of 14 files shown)


nbulaj:
Hello. This is actually better to move to an extension , like it's done in https://github.com/exop-group/doorkeeper-device_authorization_grant/ did (using modern core options).  I'll take a closer look a little bit later, I had a draft for this feature somewhere on my laptop which never was put to GitHub

nbulaj:
I updated my local repo with the recent Doorkeeper changes + added missed pieces from this PR. All the work is pushed into https://github.com/doorkeeper-gem/doorkeeper-id_jag_grant so it's available for testing & improvements/fixes/etc

nbulaj:
Hello @manmohan-shaw-okta . Thanks again for your effort! :bow:  I'm gonna close this one PR in favor of https://github.com/doorkeeper-gem/doorkeeper-id_jag_grant so please if there is anything to fix/improve/etc for this grant - submit PR for the extension repo (unless core changes required). Thanks!

Issue #383, opened 2014-04-15
Multiple tokens for the same (resource_owner_id, application_id) pair?


kenn:
If my understanding is correct, a successful authentication always create a new access token.

Is there an option to return an existing token, if found for the same (resource_owner_id, application_id) pair?

Some lazy app developers (or bugs) could just ask for a new token every time, rather than correctly store and handle the token with expiration information.

The result is ever growing database records. It's like permanent memory leak, and there's no chance for garbage collection if the expiration is set at a distant future. It's so easy to accidentally attack the system.

I would like to bound the maximum number of records in `oauth_access_tokens` for predictable server sizing and provisioning, and to do so, guaranteed uniqueness of (resource_owner_id, application_id) sounds reasonable.

Was there any discussion over the matter? And would it be accepted if I create a pull request to add that option?

jiggneshhgohel:
@kenn Thanks for bringing this up and getting this reuse_access_token feature implemented. I am in 100% agreement to all your points you have and which I have quoted below:

> Some lazy app developers (or bugs) could just ask for a new token every time, rather than correctly store and handle the token with expiration information.

> In my case, I would like to return the same access token to multiple devices (e.g. iPhone and iPad) for the same user, and I wouldn't extend expiration - I'd just set expiration at a distant future and when expires / revoked, it invalidates on all devices that use the same token.

> We cannot simply delete because we set expiration at 1 year from now. Those unused tokens accumulate over time. We don't plan to use refresh token so there's no problem reusing the existing token for the same user.

> I believe reusing access token is not prohibited by the spec per se, and large scale providers like Facebook do just that. So I thought it would be great if Doorkeeper could support that use case.

And using adding `reuse_access_token` option (which by default is disabled) to `Doorkeeper.configure` block resolved the issue I had.

In case anybody ending up here wants to know about the background of the scenario I have created it as a gist at https://gist.github.com/jiggneshhgohel/193c013ea7ac8cead4c77cef1127c55c

Thanks

jiggneshhgohel:
BTW I really think there should be section in wiki about all the options supported with their usage purpose with a brief background. 

I couldn't find any detail regarding using `reuse_access_token` in the wiki and after going through this issue and the related PR #387 changes I found that there is already this option available in default generated `Doorkeeper.configure` block 

```
  # Reuse access token for the same resource owner within an application (disabled by default)
  # Rationale: https://github.com/doorkeeper-gem/doorkeeper/issues/383
[...]

PR #1862, opened 2026-07-19
[test] Document `reuse_access_token` matching scope and pin it with a regression spec [#1693]

### LINKED ISSUE #1693: Different access grants return the same access token with `reuse_access_token` enabled

55728:
Closes #1693.

### What

With `reuse_access_token` enabled, token matching considers only the **application, resource owner, scopes and custom token attributes** (plus whether a refresh token is expected, since #1850) — it does not track which authorization grant produced the token. Exchanging two independently issued authorization codes for the same combination therefore returns one shared access token.

As confirmed in the issue discussion, this is intended behavior (rationale in #383), but it surprises users running concurrent sessions of the same client: the sessions become interdependent, and refreshing the shared token in one session revokes it for the others.

This PR treats the confusion as a documentation gap:

- **Initializer template**: the `reuse_access_token` section now spells out what matching is based on, warns that separate authorization grants share a single token (with the concurrent-sessions consequence), and points to the workarounds — keep the option disabled, or differentiate sessions via `custom_access_token_attributes`.
- **Regression spec**: a new authorization_code flow scenario pins the shared-token behavior end-to-end (two separate grants → one access token), so any future change to the matching semantics is caught deliberately rather than by accident.

### Why no behavior change

`reuse_access_token` explicitly trades per-grant token isolation for fewer rows in `oauth_access_tokens`. Making it grant-aware would defeat its purpose; users who need independent token chains already have supported paths (the custom-attributes path is covered by model specs since #1853).

Docs/test-only change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ User-visible changes worth mentioning.
 - [#1855] Perform the fallback secret upgrade-on-access write (plain → hashed token or application secret) through the primary database role, so `enable_multiple_database_roles` setups no longer attempt the write on a read replica when the lookup happens in a request routed to the reading role.
 - [#1857] Pin with regression specs that a `+` between scopes in a form-encoded token request is decoded as a space (so `scope=public+write` refreshes fine), while a percent-encoded literal `+` (`%2B`) names a single scope and is rejected when unknown, per RFC 6749 §3.3. Test-only change, closes [#1686].
 - [#1859] Pin with regression specs that a refresh token bound to an expired access token can be revoked (fixed by [#1744]) and that the revoked refresh token is rejected at the token endpoint afterwards. Test-only change, closes [#1671].
+- [#1862] Document that with `reuse_access_token` enabled token matching considers only the application, resource owner, scopes and custom token attributes — separate authorization grants for the same combination intentionally share one access token — and pin the behavior with a regression spec. Docs/test-only change, closes [#1693].
 - Please add here
 
 ## 5.9.3
--- a/lib/generators/doorkeeper/templates/initializer.rb
+++ b/lib/generators/doorkeeper/templates/initializer.rb
@@ -150,6 +150,15 @@ Doorkeeper.configure do
   # token found for the application, resources owner and/or set of scopes.
   # Rationale: https://github.com/doorkeeper-gem/doorkeeper/issues/383
   #
+  # Matching considers only the application, resource owner, scopes, custom access token
+  # attributes (see +custom_access_token_attributes+) and whether a refresh token is
+  # expected — not the authorization request that produced the token. Separate authorization
+  # grants for the same combination share a single access token, so concurrent sessions of
+  # the same client become interdependent (e.g. refreshing the token in one session revokes
+  # it for the others). If you need independent tokens per session or device, keep this
+  # option disabled or differentiate the sessions with +custom_access_token_attributes+.
+  # See https://github.com/doorkeeper-gem/doorkeeper/issues/1693
+  #
   # You can not enable this option together with +hash_token_secrets+.
   #
   # reuse_access_token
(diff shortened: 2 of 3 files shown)


nbulaj (approved):
LGTM ✅

PR #1864, opened 2026-07-19
Carry custom access token attributes through the consent screen

### LINKED ISSUE #1863: `custom_access_token_attributes` are lost when the authorization goes through the consent screen
### Steps to reproduce

1. Configure a custom access token attribute:

   ```ruby
   Doorkeeper.configure do
     custom_access_token_attributes [:tenant_name]
   end
   ```

   (with `tenant_name` columns on `oauth_access_grants` / `oauth_access_tokens`)

2. Start an authorization code flow that includes the attribute, e.g.
   `GET /oauth/authorize?client_id=...&response_type=code&redirect_uri=...&tenant_name=alpha`
3. The consent screen is rendered (i.e. `skip_authorization?` is false and there is no matching token). The resource owner clicks **Authorize**.
4. Exchange the code at the token endpoint.

### Expected

The grant — and therefore the access token — carries `tenant_name = "alpha"`, the same way it does when authorization is auto-approved (skip_authorization / matching token) or when the grant is created directly.

### Actual

`grant.tenant_name` (and the token's) is `nil`.

### Cause

`app/views/doorkeeper/authorizations/new.html.erb` forwards only the standard OAuth parameters as hidden fields in the approve/deny forms. The `POST /oauth/authorize` triggered by the consent form therefore builds a `PreAuthorization` without the custom attributes, even though `AuthorizationsController#pre_auth_param_fields` already permits them.

The existing spec coverage (`spec/requests/flows/authorization_code_spec.rb`, "when custom_access_token_attributes are configured") creates the grant via the factory, which bypasses the consent screen, so the gap was never exercised.

`api_only` mode has the same problem one level up: the pre-authorization JSON that a custom consent UI is built from (`render json: pre_auth`) doesn't include the custom attributes, so the UI has nothing to send back.

55728:
Fixes #1863.

### Summary

`custom_access_token_attributes` present on the `GET /oauth/authorize` request were dropped whenever the flow went through the consent screen: the approve/deny forms in `authorizations/new.html.erb` only forward the standard OAuth parameters, so the `POST /oauth/authorize` built a fresh `PreAuthorization` without them and the grant/token stored `nil`. Auto-approved flows (`skip_authorization`, matching token) were unaffected, which is why the existing factory-based specs never caught it.

### Changes

- `app/views/doorkeeper/authorizations/new.html.erb`: render a hidden field per configured custom attribute (from `@pre_auth.custom_access_token_attributes`) in both the approve and the deny form. The controller already permits these params (`pre_auth_param_fields`), so no controller change is needed.
- `lib/doorkeeper/oauth/pre_authorization.rb`: include the custom attributes in `#as_json` (`pre_auth_hash`), so `api_only` consumers that build their own consent UI from the pre-authorization JSON can send the values back with the approval request. Reserved keys win over a colliding custom attribute name (`reverse_merge`).
- Specs:
  - feature spec driving the real consent flow (authorization request with a custom attribute → consent form → grant → token exchange), plus a no-attribute case pinning that nothing leaks when the attribute isn't supplied;
  - `PreAuthorization#as_json` spec for the new key.
- CHANGELOG entry.

The new feature spec fails without the view change (`grant.tenant_name` is `nil`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ User-visible changes worth mentioning.
 - [#1857] Pin with regression specs that a `+` between scopes in a form-encoded token request is decoded as a space (so `scope=public+write` refreshes fine), while a percent-encoded literal `+` (`%2B`) names a single scope and is rejected when unknown, per RFC 6749 §3.3. Test-only change, closes [#1686].
 - [#1859] Pin with regression specs that a refresh token bound to an expired access token can be revoked (fixed by [#1744]) and that the revoked refresh token is rejected at the token endpoint afterwards. Test-only change, closes [#1671].
 - [#1862] Document that with `reuse_access_token` enabled token matching considers only the application, resource owner, scopes and custom token attributes — separate authorization grants for the same combination intentionally share one access token — and pin the behavior with a regression spec. Docs/test-only change, closes [#1693].
+- [#1864] Fix `custom_access_token_attributes` values being dropped when the authorization goes through the consent screen: the approve/deny forms now carry the custom attributes as hidden fields, and the pre-authorization JSON (`api_only` mode) includes them so custom consent UIs can send them back.
 - Please add here
 
 ## 5.9.3
--- a/app/views/doorkeeper/authorizations/new.html.erb
+++ b/app/views/doorkeeper/authorizations/new.html.erb
@@ -29,6 +29,9 @@
       <%= hidden_field_tag :scope, @pre_auth.scope, id: nil %>
       <%= hidden_field_tag :code_challenge, @pre_auth.code_challenge, id: nil %>
       <%= hidden_field_tag :code_challenge_method, @pre_auth.code_challenge_method, id: nil %>
+      <% @pre_auth.custom_access_token_attributes.each do |attribute_name, attribute_value| %>
+        <%= hidden_field_tag attribute_name, attribute_value, id: nil %>
+      <% end %>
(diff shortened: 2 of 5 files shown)


nbulaj (approved):
👀

nbulaj:
Just one security question: doesn't it allow to hijack the custom attributes? LIke change the tenant or whatever. So end user can hack the form and put smth what is dangerous @55728

55728:
Good question! 👀✨ This PR doesn't introduce a new attack surface — the custom attributes originate from the `GET /oauth/authorize` query parameters (`PreAuthorization#initialize` slices them from `parameters`), and the authorization request is front-channel, so the end user could always supply `?tenant_name=anything` in the URL themselves. Tampering with the hidden field on the consent POST is equivalent to editing the GET URL.

The auto-approved paths (`skip_authorization` / matching token) already stored those URL-supplied values before this PR. The consent form also already round-trips `scope`, `redirect_uri`, `code_challenge`, etc. the same way, and the POST rebuilds the `PreAuthorization` and re-validates. Only attribute names whitelisted via `custom_access_token_attributes` config get through.

That said, you're right that these values are ultimately client/user-supplied input — apps must not rely on them for trust decisions without server-side validation (e.g. in the `before_successful_authorization` hook). Happy to add a note to the initializer docs in this PR or as a follow-up if you think that's worth calling out explicitly 😊

nbulaj:
Your truth @55728 , I refreshed the flow now in my :brain: . I don't think we need to document it actually, all good.  Thanks a lot!

55728:
Thank you @nbulaj 😊✨

PR #1865, opened 2026-07-20
Revoke issued tokens when an authorization code is reused [#1713]

### LINKED ISSUE #1713: Access tokens should be revoked when multiple attempts are made to exchange the same authorization code
### Steps to reproduce

I am working through a review with our security team and they requested that I revoke access tokens when there are multiple attempts to exchange the authorization code.  This is one of the security recommendations from the [rfc](https://datatracker.ietf.org/doc/html/rfc6749#section-10.5) and I went to see if there were any hooks available that would allow us to revoke the tokens, but none seem to be available.  Specifically I was hoping that I could register a hook at [this point](https://github.com/doorkeeper-gem/doorkeeper/blob/5bf9d40/lib/doorkeeper/oauth/authorization_code_request.rb#L30) so that I can revoke the access tokens.  If it would help, I can make a PR for this functionality.

>   Authorization codes MUST be short lived and single-use.  If the
>   authorization server observes multiple attempts to exchange an
>   authorization code for an access token, the authorization server
>   SHOULD attempt to revoke all access tokens already granted based on
>   the compromised authorization code.

55728:
## Summary

Fixes #1713. RFC 6749 §4.1.2 requires authorization codes to be single-use and says the authorization server SHOULD revoke (when possible) all tokens previously issued based on a code that is presented more than once; §10.5 lists the same behavior as a security consideration for compromised codes. Doorkeeper already denies the second exchange with `invalid_grant`, but the token issued by the first exchange stayed valid.

## Design

Doorkeeper had no link between a grant and the token its exchange produced, so this PR records one: a new **optional** `oauth_access_grants.access_token_id` column stores the id of the access token returned by the exchange, written inside the same locked transaction that revokes the grant.

The reference lives on the grant (not the token) deliberately:

- With `reuse_access_token`, an exchange can return a pre-existing token. A grant-side reference still captures exactly what the exchange handed out, so replaying that code revokes the right token. A token-side `grant_id` set at creation would miss this case.
- Grants are one-time-use, so one reference per grant is complete — "all tokens previously issued based on that authorization code" is always the single token (access + its refresh token, same record) the exchange returned.

Following the `previous_refresh_token` / PKCE convention, the behavior is **gated on column presence** (`AccessGrant.access_token_revoked_on_reuse?`):

- New installs get the column through the install migration template (with a comment explaining how to opt out).
- Existing apps opt in with the new generator: `rails generate doorkeeper:grant_reuse_revocation`.
- Without the column, behavior is unchanged.

Both replay-detection paths revoke the linked token:

1. The normal path — the request is denied with `invalid_grant` as before. Revocation is deferred to the last validation step (`validate_grant_accessible`, after the `redirect_uri` and PKCE checks), so a replay that cannot prove possession of the code is rejected *without* revoking anything. Only a caller that presents the correct `redirect_uri` and code verifier can trigger revocation.
2. The concurrency race — `Errors::InvalidGrantReuse` raised after `grant.lock!`. The revocation happens after the raise leaves the transaction, so the rollback cannot undo it (`lock!` has already reloaded the grant with the winning exchange's committed linkage).

Revoking an already-revoked token is a no-op (`Revocable#revoke` preserves the original `revoked_at`), so grants revoked through other flows (e.g. a user revoking an application) replay harmlessly.

### DIFF
--- a/lib/generators/doorkeeper/templates/add_access_token_to_access_grants.rb.erb
+++ b/lib/generators/doorkeeper/templates/add_access_token_to_access_grants.rb.erb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class AddAccessTokenToAccessGrants < ActiveRecord::Migration<%= migration_version %>
+  def change
+    # No index: the column is only read back from an already-loaded grant,
+    # never used to filter queries.
+    add_reference :oauth_access_grants, :access_token, index: false
+  end
+end
--- a/lib/generators/doorkeeper/templates/migration.rb.erb
+++ b/lib/generators/doorkeeper/templates/migration.rb.erb
@@ -28,6 +28,18 @@ class CreateDoorkeeperTables < ActiveRecord::Migration<%= migration_version %>
       t.string   :scopes,            null: false, default: ''
       t.datetime :created_at,        null: false
       t.datetime :revoked_at
+
+      # Authorization codes are single-use (RFC 6749 §4.1.2). Doorkeeper
+      # records the access token issued when the code is exchanged, so that
+      # a second exchange attempt revokes the previously issued tokens as
+      # recommended by RFC 6749 §10.5.
+      #
+      # Comment out this line if you don't want a reused authorization code
+      # to revoke the tokens already issued for it.
+      #
+      # No index: the column is only read back from an already-loaded grant,
+      # never used to filter queries.
+      t.references :access_token,    index: false
     end
 
     add_index :oauth_access_grants, :token, unique: true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ User-visible changes worth mentioning.
 
 ## main
 
+- [#1865] Revoke the token issued for an authorization code when the code is exchanged more than once, per RFC 6749 §4.1.2 / §10.5. Active when the `oauth_access_grants.access_token_id` column exists: new installs get it from the generated migration, existing apps can add it with `rails generate doorkeeper:grant_reuse_revocation`. Closes [#1713].
 - Please add here
 
 ## 6.0.0.beta1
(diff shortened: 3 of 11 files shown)


55728:
GuardRails flagged the ERB template as XSS but it generates a Ruby migration file, not HTML — same pattern as the existing generators.

lib/generators/doorkeeper/templates/migration.rb.erb
```diff
@@ -28,6 +28,15 @@ class CreateDoorkeeperTables < ActiveRecord::Migration<%= migration_version %>
...
       t.datetime :revoked_at
+
+      # Authorization codes are single-use (RFC 6749 §4.1.2). Doorkeeper
+      # records the access token issued when the code is exchanged, so that
+      # a second exchange attempt revokes the previously issued tokens as
+      # recommended by RFC 6749 §10.5.
+      #
+      # Comment out this line if you don't want a reused authorization code
+      # to revoke the tokens already issued for it.
+      t.bigint   :access_token_id
```
nbulaj: `t.referfences ` to be consistent with the overall design? 👀

lib/generators/doorkeeper/templates/add_access_token_to_access_grants.rb.erb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddAccessTokenToAccessGrants < ActiveRecord::Migration<%= migration_version %>
+  def change
+    add_column :oauth_access_grants, :access_token_id, :bigint
```
nbulaj: We don't add an index because we don't perform filtering by this column in any of the DB queries?

lib/generators/doorkeeper/templates/migration.rb.erb
55728: Good catch 👀✨ — switched to `t.references :access_token, index: false` (and `add_reference` in the opt-in template). `index: false` is explicit since `t.references` defaults to indexing.

lib/generators/doorkeeper/templates/add_access_token_to_access_grants.rb.erb
55728: Exactly 😊 — the column is only read back from an already-loaded grant (`find_by(id: grant.access_token_id)` is a PK lookup on the tokens table), and no query filters grants by it. Added a comment in the migration templates to document that.

nbulaj (approved):
LGTM

PR #1867, opened 2026-07-20
fix: use access_denied for resource_owner_authorize_for_client per RFC 6749 §4.1.2.1

### LINKED ISSUE #1578: Why `invalid_client` & `unauthorized_client` are non-redirectable?
In `unauthorized_client` case, the combination of client_id & redirect_uri should be valid but the client isn't allowed to use the response_type.

and `invalid_client` is error token response, not error authorization response.

Why those two are defined as "un-redirectable"?

ref.) https://github.com/doorkeeper-gem/doorkeeper/blob/15ebd3302d01bad4a408e0b10676b5f624e688ad/lib/doorkeeper/oauth/error_response.rb#L8

55728:
## Summary

Change the error returned by the `resource_owner_authorize_for_client` validation from `invalid_client` to `access_denied`, aligning with RFC 6749 Section 4.1.2.1.

Addresses the second point of #1578 (complements #1816 which addressed the first point).

## Motivation

As @nov pointed out in #1578, `invalid_client` is defined as a **token endpoint** error code (RFC 6749 [Section 5.2](https://datatracker.ietf.org/doc/html/rfc6749#section-5.2)), not an authorization endpoint error code.

When `resource_owner_authorize_for_client` fails, the resource owner is being denied access to a client application. The RFC-appropriate error code for this is `access_denied`:

> **access_denied** — The resource owner or authorization server denied the request.
> — [RFC 6749 Section 4.1.2.1](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1)

## What this changes

- `resource_owner_authorize_for_client` now returns `access_denied` instead of `invalid_client`
- Since `access_denied` is not in `NON_REDIRECTABLE_STATES`, the error is now **redirected** to the client's registered redirect URI (the redirect URI has already been validated at this point in the validation chain)
- HTTP status changes from `401 Unauthorized` to `400 Bad Request` (consistent with other redirectable authorization errors)

## How this resolves #1578

The original issue asked: *"Why are `invalid_client` & `unauthorized_client` non-redirectable?"*

| Point | Error | Resolution |
|-------|-------|------------|
| 1 | `unauthorized_client` | Made redirectable in #1816 |
| 2 | `invalid_client` | This PR. `invalid_client` now only occurs when the client itself is not found (before `redirect_uri` validation), so its presence in `NON_REDIRECTABLE_STATES` is correct per §4.1.2.1. The misplaced usage in `resource_owner_authorize_for_client` is corrected to `access_denied`. |

## Breaking change note

Applications relying on `invalid_client` being returned when `authorize_resource_owner_for_client` fails will now receive `access_denied`. This is a correction toward RFC compliance.

Fixes #1578
Closes #1816


nbulaj (approved):
LGTM

nbulaj:
@nov any chance you can check it as well as an original reporter? :pray: Thanks

nbulaj:
Sorry missclicked :|

nov:
LGTM

55728:
Thank you for the review @nbulaj, and @nov 😊

PR #1870, opened 2026-07-21
Fix NameError when `application_secret_generator` is misconfigured

55728:
## Summary

`Doorkeeper::Application#secret_generator` is supposed to raise `Doorkeeper::Errors::TokenGeneratorNotFound` when the configured `application_secret_generator` class doesn't exist, and `Doorkeeper::Errors::UnableToGenerateToken` when the class doesn't respond to `.generate`. Neither error could actually be raised: the application mixin is defined with compact module nesting (`module Doorkeeper::Orm::ActiveRecord::Mixins`), so the bare `Errors` constant in the method body never resolves to `Doorkeeper::Errors`. Both error paths instead crashed with:

```
NameError: uninitialized constant Doorkeeper::Orm::ActiveRecord::Mixins::Application::Errors
```

This turns a helpful configuration error ("`FooGenerator` not found") into a confusing constant-resolution failure pointing at Doorkeeper internals.

## Changes

- Qualify both raises as `Doorkeeper::Errors::UnableToGenerateToken` / `Doorkeeper::Errors::TokenGeneratorNotFound` in `lib/doorkeeper/orm/active_record/mixins/application.rb`.
- Add regression specs pinning both error paths (unknown generator class, generator without `.generate`), plus a spec for the `#authorized_tokens` association scope which was previously uncovered.
- CHANGELOG entry.

## Notes

- The same bare-constant pattern elsewhere in the mixins resolves fine when it's inside `included do ... end` blocks (evaluated in the model class context under `Doorkeeper` autoload paths) — only these two raises in an instance method body were affected.
- No behavior change for correctly configured generators; `Doorkeeper::SecretStoring::*` paths are untouched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ User-visible changes worth mentioning.
 - [#1862] Document that with `reuse_access_token` enabled token matching considers only the application, resource owner, scopes and custom token attributes — separate authorization grants for the same combination intentionally share one access token — and pin the behavior with a regression spec. Docs/test-only change, closes [#1693].
 - [#1864] Fix `custom_access_token_attributes` values being dropped when the authorization goes through the consent screen: the approve/deny forms now carry the custom attributes as hidden fields, and the pre-authorization JSON (`api_only` mode) includes them so custom consent UIs can send them back.
 - [#1869] Improve test coverage
+- [#1870] Fix: raise the intended `Doorkeeper::Errors::TokenGeneratorNotFound` / `UnableToGenerateToken` (instead of a confusing `NameError`) when `application_secret_generator` is misconfigured.
 - Please add here
 
 ## 5.9.3
--- a/lib/doorkeeper/orm/active_record/mixins/application.rb
+++ b/lib/doorkeeper/orm/active_record/mixins/application.rb
@@ -116,9 +116,9 @@ module Doorkeeper::Orm::ActiveRecord::Mixins
 
         return generator if generator.respond_to?(:generate)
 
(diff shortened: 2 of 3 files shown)


nbulaj:
Looks like nobody ever configured a wrong secret generator :smile:

nbulaj (commented):
BTW I'm pretty sure we have tests for these exceptions,  like [this one](https://github.com/doorkeeper-gem/doorkeeper/blob/main/spec/models/doorkeeper/access_token_spec.rb#L246-L248)

Should we remove/fix/whatever them? 🤔

55728:

Good catch! 👀✨ The existing tests (like the [one](https://github.com/doorkeeper-gem/doorkeeper/blob/b8d4c6c6995d9d6ddf8040d4b95739c4af1d2e30/spec/models/doorkeeper/access_token_spec.rb#L246-L248) you linked) cover `AccessToken#token_generator`, which lives directly under `module Doorkeeper` — so bare `Errors` resolves fine there. This bug is specific to `Application#secret_generator` in the compact-nested `Doorkeeper::Orm::ActiveRecord::Mixins` module, where the lookup path never reaches `Doorkeeper::Errors`. The existing tests are correct as-is; only this mixin needed the FQN fix.

nbulaj:
Yeah thanks @55728 , it was just a quick example, I thought we covered all the custom generator cases TBH :eyes:  But in any case I like more specific namespace resolution instead of a generic one :+1:

nbulaj (approved):

PR #1876, opened 2026-07-25
Reject a non-string `scope` parameter with `invalid_request` instead of a 500

55728:
### Problem

`Scopes.from_string` assumes a string and calls `#split`. A nested `scope` parameter — `scope[a]=b`, which Rack parses into a Hash — reaches it and raises `NoMethodError`. That is not a `DoorkeeperError`, so it bypasses `TokensController`'s `rescue_from` and surfaces as an unhandled **500** rather than RFC 6749 §5.2's 400. The crash happens *before* client authentication, so it is reachable unauthenticated. (An array `scope[]=a` coincidentally survives via `Array#split`→`Array#flatten`; only the Hash shape hits this sink. Likely pre-existing.)

### Fix

`Scopes.from_string` now raises `Errors::InvalidScopeParameter` (type `invalid_request`) for a non-string argument:

- **Token endpoint:** the controller's `rescue_from Errors::DoorkeeperError` turns it into `invalid_request`, covering every grant that parses scopes (client_credentials, password, refresh_token) through the one shared sink.
- **Authorization endpoint:** `pre_auth.authorizable?` does not rescue arbitrary errors, so raising inside validation would itself be a 500. Pre-authorization instead rejects a non-string scope up front in `validate_params` (which runs before `validate_scopes` reaches `from_string`), yielding `invalid_request`.
- The model layer always passes a string, so it is unaffected.


nbulaj (approved):
LGTM ✅

PR #1878, opened 2026-07-25
Restrict the loopback redirect_uri exception to the port (RFC 8252 §7.3)

55728:
### Problem

RFC 8252 §7.3 lets only the **port** of a loopback redirect URI vary at runtime. The check blanked the port with `URI#port=` and compared the reassembled strings, but on Ruby >= 4.0 `URI#port=` also clears the **userinfo**. So `http://attacker@127.0.0.1/cb` matched a registered `http://127.0.0.1/cb`.

**Scope:** the destination host is always the loopback interface (127.0.0.1 / ::1), so this is **not** a cross-origin open redirect. Non-loopback hosts take the exact-match branch and were never affected. Ruby-version dependent (surfaced by the `URI#port=` behavior change in Ruby 4.0).

### Fix

Compare the two URIs component by component (scheme, userinfo, host, path, query, fragment), leaving the port out. This keeps the port-only exception while no longer ignoring userinfo differences, and preserves the previous property of not letting `URI#==` normalize an empty path to `"/"`.


lib/doorkeeper/oauth/helpers/uri_checker.rb
```diff
@@ -31,16 +31,28 @@ def self.matches?(url, client_url)
...
 
+        # RFC 8252 §7.3 lets only the PORT of a loopback redirect URI vary at
+        # runtime. Compare every other component explicitly instead of blanking
+        # the port and comparing the reassembled strings: `URI#port=` also
+        # clears the userinfo on Ruby >= 4.0, which would let
+        # `http://attacker@127.0.0.1/cb` match a registered
+        # `http://127.0.0.1/cb`. Comparing components (not `URI#==`) also keeps
+        # the exception from widening past the port — e.g. an empty path is not
+        # treated as equal to "/".
+        def self.loopback_match?(url, client_url)
```
nbulaj: Can we name it like `urls_match?` or smth like that? I think `loopback_` is a bit misleading here 👀

55728: Good point 👀✨ — renamed to `urls_match?`. Thanks!

nbulaj: Much cleaner now, thanks! (since loopback is pretty specific and we're comparing more wide urls/uris instead)

nbulaj (approved):
LGTM ✅

55728:
Thank you 😊

PR #1879, opened 2026-07-25
Don't collaterally revoke a reused access token on authorization code replay

55728:
Follow-up to #1865.

### Problem

#1865 revokes the token issued for an authorization code when the code is replayed (RFC 6749 §10.5), following the grant's `access_token_id`. With `reuse_access_token` enabled, `find_or_create_for` returns a token shared across grants for the same app / owner / scope, so several grants point at one token. Replaying **one** grant's code then revoked that **shared** token, taking down another valid session that still held it.

This is an **availability** issue only — no confidentiality/integrity impact. It requires `reuse_access_token` **on** (default off) **and** the `access_token_id` column present; either flag off fully mitigates it.

### Fix (preserves #1865)

`revoke_token_issued_for_grant` now checks whether another grant still references the token before revoking. A **shared** token (referenced by another grant) is skipped — revoking it would collaterally invalidate another valid session. A token **unique to the replayed code** (no other grant references it) is revoked exactly as before, so #1865's single-use protection against a stolen code is intact.


lib/doorkeeper/oauth/authorization_code_request.rb
```diff
@@ -161,9 +161,26 @@ def revoke_token_issued_for_grant
...
+          # code and never collaterally revokes a reused, shared one.
+          next if token_shared_with_other_grant?(token)
+
+          token.revoke
         end
       end
+
+      def token_shared_with_other_grant?(token)
+        grant.class
+          .where(access_token_id: token.id)
```
nbulaj, line 188: Wont it fail on a legacy apps which didn't run the generator and haven't introduced reference column? 

I think there should be a guard (kinda `access_token_revoked_on_reuse?` call or whatever_ to be sure app uses linking between grants/tokens

55728: Good catch on raising the concern 👀✨ — the guard is already in place, but it sits above the diff hunk so it's not visible in the review:

```ruby
def revoke_token_issued_for_grant
  return unless grant.class.access_token_revoked_on_reuse?
  return if grant.access_token_id.blank?
```

`access_token_revoked_on_reuse?` checks `column_names.include?("access_token_id")`, so legacy apps that never ran `doorkeeper:grant_reuse_revocation` return early and never reach `token_shared_with_other_grant?`. The write side (`link_access_token_to_grant`) is guarded the same way.

This path is covered by the existing spec *"only denies the request when the access_token_id column is not available"*, which stubs the predicate to `false` and asserts the issued token stays unrevoked.

Let me know if you'd like me to add an inline comment on `token_shared_with_other_grant?` noting the caller-side guard — happy to do so if it improves readability 😊

nbulaj: Got it, thanks for the clarification, indeed 👍 If it's possible to add a comment to `token_shared_with_other_grant?(token)` which will mention this migration/option - will be great, thanks 🙇

55728: Added 👀✨ — `token_shared_with_other_grant?` now carries a comment pointing at the optional `oauth_access_grants.access_token_id` column (generated migration for new installs, `doorkeeper:grant_reuse_revocation` for existing apps) and noting that callers guard with `access_token_revoked_on_reuse?`, so an app that never ran the generator returns early.

nbulaj (approved):

nbulaj:
Thanks!

PR #1883, opened 2026-07-29
Merge `CHANGELOG.md` with git's `union` driver

55728:
Every changelog entry is appended in the same place, on the line above "Please add here". Two pull requests that each add one therefore touch the same line, and the second to be merged conflicts even though the changes have nothing to do with each other. Resolving it is mechanical — keep both lines — but it has to be done by hand for every pull request after the first, and again whenever a branch is brought up to date.

Git's built-in [union merge driver](https://git-scm.com/docs/gitattributes#_built_in_merge_drivers) concatenates both sides of a conflict instead of reporting one, which is exactly the desired resolution for an append-only file like this. This PR enables it for `CHANGELOG.md` via `.gitattributes`.

>[!IMPORTANT]
>Ordering: entries added in parallel end up in whatever order the merges produced, rather than by pull request number. That is a plain edit on the merged file when it matters, and the entries are already grouped by release rather than strictly ordered within one.

> [!NOTE]
>`.gitattributes` is new to this repo; the file carries a comment explaining why the driver is set, so the behaviour change is discoverable in-tree.


nbulaj (approved):
WOW I love it 👍 😍

55728:
I'm honored 😆✨

nbulaj:
This was really a noisy issue :smile:  But I never tried to check & fix that :see_no_evil:

PR #1884, opened 2026-07-29
Fall back across token types when the `token_type_hint` lookup misses

### LINKED ISSUE #1882: Introspection with `token_type_hint=refresh_token` doesn't fall back to an access token lookup (RFC 7662 §2.1 MUST)
## Description

RFC 7662 §2.1 says about `token_type_hint`:

> If the server is unable to locate the token using the given hint, it MUST extend its search across all of its supported token types.

Doorkeeper resolves the presented token in `TokensController#revocable_token` (shared by `/oauth/introspect` and `/oauth/revoke` since #1860):

```ruby
# app/controllers/doorkeeper/tokens_controller.rb
@revocable_token =
  if params[:token_type_hint] == "refresh_token"
    refresh_token
  else
    access_token || refresh_token
  end
```

With no hint (or `token_type_hint=access_token`) the search falls back across both token types, but `token_type_hint=refresh_token` searches refresh tokens only. A hint is explicitly allowed to be wrong — that's why the fallback is a MUST — so introspecting a valid access token with `token_type_hint=refresh_token` returns `active: false` for a token the server does know about.

The same asymmetry exists on `/oauth/revoke`, where RFC 7009 §2.1 has the identical MUST. There the current behavior is even codified in a spec (`spec/controllers/tokens_controller_spec.rb` — "does not revoke the access token when token_type_hint == refresh_token"): the endpoint returns 200 (as RFC 7009 requires regardless), but the token stays active even though the server could have located it. The introspection side of the wrong-hint case has no test coverage at all.

## Steps to reproduce

1. Create an access token (with or without a refresh token).
2. `POST /oauth/introspect` with `token=<the access token's token value>` and `token_type_hint=refresh_token`, authenticated as the token's client.
3. Response is `{"active": false}`; expected `active: true` after the search is extended to access tokens.

Equivalent on `/oauth/revoke`: same parameters return 200 but the access token is not revoked.

## Expected behavior

When the lookup by the hinted type finds nothing, extend the search to the other token type — for both introspection and revocation, since they share the lookup. `token_type_hint` remains a pure lookup-order optimization, which also matches how the no-hint and `access_token`-hint paths already behave.

One design point to settle: the revoke-side spec above pins the current non-compliant behavior as intended, so fixing this means changing that spec's expectation (or deliberately scoping the fix to introspection only, leaving revocation asymmetric).

55728:
Fixes #1882.

RFC 7009 §2.1 and RFC 7662 §2.1 both say that when the server cannot locate the token using the given hint, it MUST extend its search across all of its supported token types — a hint is explicitly allowed to be wrong. Doorkeeper's shared lookup honored that on the no-hint and `access_token`-hint paths, but `token_type_hint=refresh_token` searched refresh tokens only, so revoking or introspecting a valid access token under that hint reported 200-but-not-revoked / `active: false` for a token the server knows about.

The `refresh_token` path now falls back to the access token lookup, mirroring the existing fallback in the other direction. The revoke-side spec that pinned the non-compliant behavior as intended now expects the revocation, and both endpoints gain wrong-hint coverage in each direction.


nbulaj (approved):
LGTM ✅

CHANGELOG.md
```diff
@@ -13,7 +13,8 @@ User-visible changes worth mentioning.
...
 - [#1879] Fix: with `reuse_access_token` enabled, replaying an authorization code no longer revokes an access token that another grant still shares. The single-use revocation added in [#1865] followed the grant's `access_token_id`, which a reused token shares across grants, so a replay could collaterally revoke a token another valid session still held. The revocation now skips a token referenced by another grant and only reaches one unique to the replayed code; a token unique to the code is still revoked as before.
-- [#1884] Internal: merge `CHANGELOG.md` with git's `union` driver, so two pull requests that each add an entry no longer conflict on the line above "Please add here".
+- [#1883] Internal: merge `CHANGELOG.md` with git's `union` driver, so two pull requests that each add an entry no longer conflict on the line above "Please add here".
```
nbulaj, line 17: 👀  Number changed

nbulaj (approved):
Can we also please add a test for non-existing access token with `refresh_token` hint?

CHANGELOG.md
55728: Busted! 🙈

55728:
Added in 40611ba 👀✨ — a never-issued token under `token_type_hint=refresh_token` now has coverage on both endpoints: revocation answers 200 and introspection answers `active: false`.

CHANGELOG.md
nbulaj: Ah I see, [original PR](https://github.com/doorkeeper-gem/doorkeeper/pull/1883/changes#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR15) has invalid number 😃

55728: Exactly! I was mistaken 😅 and removed the old entry 👀✨

nbulaj:
Thanks!

PR #1885, opened 2026-07-30
Index `oauth_access_grants.access_token_id`, which #1879 began filtering by

55728:
When #1865 added the `access_token` reference to `oauth_access_grants`, the column was only written at code exchange and read back from an already-loaded grant, so the migration template opted out of the reference's default index — and said so: *"No index: the column is only read back from an already-loaded grant, never used to filter queries."* That was accurate at the time; I wrote it.

#1879 quietly invalidated that premise, and I missed it while reviewing my own follow-up. The shared-token guard it introduced (`token_shared_with_other_grant?`) asks, on a code replay, whether the issued token is still referenced by another grant — an `exists?` that filters `oauth_access_grants` by `access_token_id`. In the common case (no other grant references the token) the predicate matches nothing, which without an index means a scan of the whole table, and the lookup deliberately runs under `with_primary_role`, so the scan always lands on the primary. Code replays are not exotic: an interrupted network round-trip or a double submit re-sends a consumed code in ordinary operation, and grants accumulate one row per authorization on installs that don't run `doorkeeper:db:cleanup`, so the cost grows quietly with table age.

With the premise gone, the column should follow the same convention as every other reference in the schema, all of which are indexed. This PR:

- drops `index: false` from both migration templates (`add_reference`/`t.references` index by default) and replaces the stale comment with the actual access pattern,
- updates the dummy app's migration and schema so the test schema matches what the template generates.

The timing is the point of doing this now rather than later: 6.0.0.beta1 predates the column, so no released migration has created it yet and this stays a template-only change. Once 6.0.0 final ships the column unindexed, correcting it would take an additional migration plus generator support for existing installs.


nbulaj:
the-office-damn-it

55728:
You are right @nbulaj! Now I know it too 😓

nbulaj (approved):
🙇

55728:
Thank you 😊

PR #1559, opened 2022-03-02
Add resource indicators

danielcooper:
### Summary

Adds [resource indicator](https://www.rfc-editor.org/rfc/rfc8707.html) support (#1413) as an optional extension. 

### Other Information

I've got a few implementation questions:

1) Should this be an optional extension (with a migration to generate) or just added into the default setup in a disabled state? I'm not clear how best to test the apps when the new migration hasn't been run/generated?

2) The way `resource` is specified in the RFC is quite inconvenient:
```
  GET /as/authorization.oauth2?response_type=code
     &client_id=s6BhdRkqt3
     &state=tNwzQ87pC6llebpmac_IDeeq-mCR2wLDYljHUZUAWuI
     &redirect_uri=https%3A%2F%2Fclient.example.org%2Fcb
     &scope=calendar%20contacts
     &resource=https%3A%2F%2Fcal.example.com%2F
     &resource=https%3A%2F%2Fcontacts.example.com%2F HTTP/1.1
  Host: authorization-server.example.com
  ```
 
Out of the box rails params won't work in this case. You can see I've hacked together a quick fix, do you have any thoughts on a better way of doing this? Maybe it'd be better to only support one `resource`.

PR #1886, opened 2026-07-30
Add Resource Indicators for OAuth 2.0

### LINKED ISSUE #1413: Resource Indicators for OAuth 2.0
Do you have any plans to implement this rfc?
[Resource Indicators for OAuth 2.0](https://tools.ietf.org/html/rfc8707)

It can be extremely useful. Currently part of problems addressed in are solved abusing **scopes** and it is very inconvenient.

nbulaj:
Add support for Resource Indicators for OAuth 2.0 (RFC 8707). 

Clients can include a `resource` parameter in authorization and token requests to indicate the target protected resource(s). The authorization server validates resource URIs, enforces audience restriction on tokens, and includes `aud` in introspection responses. Enable by configuring `resource_indicator_validator` with a callable. Requires new `resource` columns on access grants and tokens — run `rails generate doorkeeper:resource_indicators` to add the migration.

Enable by configuring `resource_indicator_validator` with a callable. Requires new `resource` columns on access grants and tokens — run `rails generate doorkeeper:resource_indicators` to add the migration.

https://datatracker.ietf.org/doc/html/rfc8707

Closes #1413 
Reviewed version of https://github.com/doorkeeper-gem/doorkeeper/pull/1559/

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -61,6 +61,7 @@ BUNDLE_GEMFILE=gemfiles/rails_7_2.gemfile bundle exec rake spec
 
 ## Change guidance
 
+- Preserve backward compatibility.
 - Make surgical changes with specs close to the behavior you changed.
 - When changing engine behavior, check whether the impact also reaches host-app integration through `spec/dummy`.
 - When changing protocol or token behavior, look for nearby request, controller, and model specs that should move together.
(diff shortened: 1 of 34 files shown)


55728:
I wonder if the scalar `resource=` wire format gets silently dropped at the authorization endpoint 👀💭

`pre_auth_param_fields` in `authorizations_controller.rb` permits `resource: []` (array only), so the RFC 8707 §2 standard form `?resource=https://api.example.com` is filtered out by strong parameters before it reaches `PreAuthorization`. Only the Rails-specific `resource[]=` form survives.

A quick way to confirm:

```ruby
params = ActionController::Parameters.new(resource: "https://api.example.com")
params.permit(resource: [])  # => {}
```

Permitting both forms would fix it:

```ruby
permit(:resource, resource: [])
```

`PreAuthorization` already wraps the value with `Array()`, so it handles both scalar and array transparently once the param gets through.

nbulaj:
You're right @55728 , reviewed & fixed it so the standard `?resource=https://api.example.com/` wire format from RFC 8707 §2 passes through strong parameters correctly now, thanks! :bow:

55728:
Thanks for reviewing the scalar `resource=` edge case so quickly @nbulaj 😊

Looking forward to this landing — RFC 8707 support has been a long-standing request (#1413) and the design here (opt-in validator + per-grant/token storage + introspection `aud`) feels clean 🙌

PR #1676, opened 2023-11-07
Improve AuthorizationsController error response handling

camero2734:
### Summary
Hey, first time contributing, hope this is welcome

This PR has three changes to `AuthorizationsController`'s error response, explained in more detail below:
- Returns the proper status code instead of `200`
- Redirects back to the client with error (if `redirectable?`)
- Honors the `raise_on_errors?` setting

-----

I noticed that Doorkeeper returns a `200` response code when an error occurs, and also does not redirect back to the client on errors as my understanding of the spec indicates it should:

https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1

**Non-redirecting**
> If the request fails due to a missing, invalid, or mismatching
   redirection URI, or if the client identifier is missing or invalid,
   the authorization server SHOULD inform the resource owner of the
   error and MUST NOT automatically redirect the user-agent to the
   invalid redirection URI.

**Redirecting**
> If the resource owner denies the access request or if the request
   fails for reasons other than a missing or invalid redirection URI,
   the authorization server informs the client by adding the following
   parameters to the query component of the redirection URI using the
   "application/x-www-form-urlencoded" format

There is already the concept of `redirectable?` in Doorkeeper which adheres to this, but it seems like this isn't actually used for errors in `authorizations_controller`.

The status code change is applied normally, but I feel like the redirection bit is probably a breaking change since it adds redirects where there wasn't any before, so I added it as a config option via `handle_auth_errors :redirect`

Fixes #1643

Issue #947, opened 2017-03-13
Make redirect_uri optional


Linuus:
Hi!

I have a question about the `redirect_uri` being required. I've seen a few other issues about this but they are all closed and I didn't find the answers that I need :)

Let's say someone is building an iOS application. They would use the "client_credentials flow" to authenticate against doorkeeper. When they create the app in doorkeeper, what would they put in the redirect_uri field? Isn't it weird to just add some dummy uri? Since they will never use "authorization code flow" the redirect uri will never be used.

Isn't it appropriate to just make it mandatory for authorization code flow requests? So, if I don't enter a redirect_uri for my app, I can't use the "authorization code flow"? It's not like we need to redirect to _any_ uri because of this, right?

I think Twitter is doing something similar on their callback url:
```
To restrict your application from using callbacks, leave this field blank.
```

Perhaps I'm missing something obvious here so please enlighten me :D

kevinetore:
Is there any progress on this issue? I'm building a client application that implements a [Two-legged OAuth2](http://oauthbible.com/#oauth-2-two-legged) strategy. Because of the validation on the `redirect URI` I cannot do this. I could bypass this by adding a dummy URI like @Linuus mentioned but that seems a bit odd.

nbulaj:
Hi @Linuus. Does it correspond to the OAuth2 RFC ?

nbulaj:
As I can see [redirect_uri can be blank](https://tools.ietf.org/html/rfc6749#section-4.1.1). So we need to see what we can do on Doorkeeper side

Linuus:
@nbulaj Sorry for the late reply. Yes it looks like it is optional :) Thank you for taking a look at this!

nbulaj:
So I made some exploration about this issue. To implement this one it is required to:

- allow nil values for `redirect_uri` on model level
- remove `NOT NULL` constraint from `oauth_applications` table

If the first one is possible without any braking changes, the second one requires additional actions, especially for legacy projects with old versions of Doorkeeper.

Other "side-effect" is that developers can set different grant flows (client credentials, password, auth code), but create all the applications without redirect URI... So the behavior of the Doorkeeper would be abnormal.

nbulaj:
@kevinetore what kind of grant type you use?

bpieck:
[...]

PR #1890, opened 2026-07-31
[test] Pin `invalid_redirect_uri` for clients registered without a redirect URI, closes #1682

### LINKED ISSUE #1682: Doorkeeper::Errors::InvalidRedirectUri Raised When No Redirect URI Set

55728:
## Summary

Test-only change. Pins the authorization endpoint's rejection of clients that were registered without a `redirect_uri` (possible when `allow_blank_redirect_uri` permits it), so #1682 can be closed with the behavior locked in rather than just discussed.

Two altitudes:

- **`PreAuthorization` unit spec** — a client with a blank registered `redirect_uri` is not authorizable, whether the request omits `redirect_uri` or supplies one, and the error is `Errors::InvalidRedirectUri`.
- **Authorization endpoint feature spec** — both request shapes render the translated `invalid_redirect_uri` error and never reach the consent form (the existing `after` hook also asserts no grant is created).

## Why close #1682

The issue reports two things; neither marks a bug:

1. **The rejection is required by RFC 6749 §3.1.2.3** — "if no redirection URI has been registered, the client MUST include a redirection URI with the authorization request", and a provided value "MUST" be compared and matched against a registered redirection URI. A client with nothing registered therefore has no request shape the server may accept, exactly as @ThisIsMissEm concluded in the thread. `allow_blank_redirect_uri` lifts the registration-time validation for URI-less grant flows (its default already refuses blank registrations while `authorization_code` or `implicit` is enabled); it does not make redirect-based flows usable without a registered URI.

2. **"In 5.6.6 and before an authorization code was returned" does not match any release.** The permissive variant (8a2f3d60, for #1518) was reverted (aca6b47b) before ever shipping — both commits first appear in v5.5.3 — and `validate_redirect_uri` is byte-identical from v5.5.3 through current `main`. What did change shortly before the report is 5.6.7's #1676, which made the authorization endpoint honor `handle_auth_errors :raise`: for apps configured to raise, the same rejection started surfacing as a `Doorkeeper::Errors::InvalidRedirectUri` exception instead of the rendered error page. A visibility change, not an authorization-outcome change.

## Scope note

This deliberately does **not** touch the neighboring question in #1678 — whether a client that *did* register exactly one URI may omit the parameter (RFC 6749 §3.1.2.3 only mandates the parameter when registration is absent, partial, or ambiguous). That relaxation stays tracked there; these specs only pin the nothing-registered case, which must fail under any resolution of #1678.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ User-visible changes worth mentioning.
 - [#1885] Index `oauth_access_grants.access_token_id` in the migration templates: since [#1879] the code-replay revocation filters access grants by that column, the "never used to filter queries" premise behind `index: false` no longer holds.
 - [#1887] [test] Pin that a requested non-default scope reaches the authorization grant and the exchanged token, and that the authorization strategy shares the controller's pre-authorization — the mismatch reported in [#1576] does not reproduce. Test-only change, closes [#1576].
 - [#1888] Document custom grant flow registration (`Doorkeeper::GrantFlow.register`) in the README with a SAML 2.0 bearer assertion (RFC 7522) walkthrough, and pin URN-shaped custom grant types with an end-to-end request spec. Docs/test-only change, closes [#764].
+- [#1890] [test] Pin that the authorization endpoint answers `invalid_redirect_uri` for a client registered without a redirect URI (`allow_blank_redirect_uri`), whether or not the request supplies one — the behavior required by RFC 6749 §3.1.2.3. Test-only change, closes [#1682].
 - Please add here
 
 ## 6.0.0.beta1
--- a/lib/generators/doorkeeper/templates/initializer.rb
+++ b/lib/generators/doorkeeper/templates/initializer.rb
@@ -333,6 +333,9 @@ Doorkeeper.configure do
   # types, but you **need** to manually drop `NOT NULL` constraint from `redirect_uri`
   # column for `oauth_applications` database table.
   #
+  # You almost certainly do not want to change this. There are very, very few cases
+  # where you want to change this configuration value.
+  #
   # You can completely disable this feature with:
   #
   # allow_blank_redirect_uri false
--- a/spec/lib/oauth/pre_authorization_spec.rb
+++ b/spec/lib/oauth/pre_authorization_spec.rb
@@ -257,6 +257,32 @@ RSpec.describe Doorkeeper::OAuth::PreAuthorization do
     expect(pre_auth).not_to be_authorizable
   end
 
+  context "when the application has no registered redirect_uri" do
+    before do
+      allow(Doorkeeper.config).to receive(:allow_blank_redirect_uri?).and_return(true)
+    end
+
+    let(:application) { FactoryBot.create(:application, redirect_uri: "") }
+
+    # RFC 6749 §3.1.2.3: when no redirection URI has been registered, the
+    # client MUST include one in the authorization request, and a provided
+    # value must still match a registered URI — so an application registered
+    # without a redirect_uri cannot use redirect-based flows at all.
+    it "is not authorizable when the request omits the redirect_uri" do
+      attributes[:redirect_uri] = nil
+
+      expect(pre_auth).not_to be_authorizable
+      expect(pre_auth.error).to eq(Doorkeeper::Errors::InvalidRedirectUri)
+    end
+
+    it "is not authorizable when the request provides a redirect_uri" do
+      attributes[:redirect_uri] = "https://app.com/callback"
+
+      expect(pre_auth).not_to be_authorizable
+      expect(pre_auth.error).to eq(Doorkeeper::Errors::InvalidRedirectUri)
+    end
+  end
+
   context "when resource_owner cannot access client application" do
     before { allow(Doorkeeper.configuration).to receive(:authorize_resource_owner_for_client).and_return(->(*_) { false }) }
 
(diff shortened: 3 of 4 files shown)


ThisIsMissEm:
`allow_blank_redirect_uri` should probably be deprecated. The whole security model of OAuth requires a redirect URI be registered with an application for authorization code grant flow, and that simple string comparison is used in order to compare redirect URIs. This is a MUST in [OAuth 2.1](https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-15.html#section-2.3):

> The redirect URI MUST be an absolute URI as defined by [Section 4.3](https://rfc-editor.org/rfc/rfc3986#section-4.3) of [[RFC3986](https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-15.html#RFC3986)]. The redirect URI MAY include an query string component ([Appendix C.1](https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-15.html#query-string-serialization)), which MUST be retained when adding additional query parameters. The redirect URI MUST NOT include a fragment component.

> Authorization servers MUST require clients to register their complete redirect URI (including the path component). Authorization servers MUST reject authorization requests that specify a redirect URI that doesn't exactly match one that was registered, with an exception for loopback redirects, where an exact match is required except for the port URI component, see [Section 4.1.1](https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-15.html#authorization-request) for details.

For `allow_blank_redirect_uri` it should probably be removed and instead based on whether or not `grant_flows` contains `client_credentials` or not. For client credential grant flows, redirection is not used, and therefore it's not necessary for the client to have a registered URL. We probably want some guidance for the user interfaces for registering clients that has text like "A redirect URI must be provided in order to use the authorization code grant flow"

ThisIsMissEm (approved):
These tests look correct, but I would recommend removal of `allow_blank_redirect_uri` entirely and determine in based on `client_credentials` grant flow being supported or not.

55728:
Thanks for the review @ThisIsMissEm! A few notes on `allow_blank_redirect_uri`, because the current behavior is closer to what you're proposing than it looks:

The default has been grant-flow-based since 5.2.x ([#947](https://github.com/doorkeeper-gem/doorkeeper/issues/947)) — blank URIs are only allowed when neither `authorization_code` nor `implicit` is enabled:

```ruby
option :allow_blank_redirect_uri,
       default: (lambda do |grant_flows, _application|
         grant_flows.exclude?("authorization_code") &&
           grant_flows.exclude?("implicit")
       end)
```

Keying on the *absence of redirect-based flows* rather than the *presence of `client_credentials`* also covers password grant, the device authorization grant extension, and custom URI-less grants (e.g. RFC 7522 assertion grants), which don't use redirection either — a `client_credentials`-only check would break those setups.

On top of that default, the option adds two things a hardcoded rule couldn't express:

- **Per-client decisions** — the lambda receives the application, e.g. allowing a blank URI for a client-credentials-only client on a server that also enables `authorization_code` for other clients.
- **The `false` hard opt-out** — "a redirect URI is always required", which is the OAuth 2.1-style strictness, available today for those who want it.

On the security-model concern: a client registered without a redirect URI can't actually enter the redirect-based flows — any authorize attempt fails with `invalid_redirect_uri`, which is exactly the behavior this PR pins down. So a blank URI scopes a client *out of* the authorization code flow rather than punching a hole in it.

Agreed that OAuth 2.1 alignment (mandatory exact registration, etc.) is worth discussing, and the UI guidance idea is a nice touch — the bundled form already toggles the field via `allow_blank_redirect_uri?(application)`, so a hint text would slot in naturally. Both feel like a separate issue rather than part of this test-only PR though.

ThisIsMissEm:
I think this is correct. What I might err on the side of caution here is that, for the default value for `allow_blank_redirect_uri`, I would actually always enforce it, but then pass the result of that to the `allow_blank_redirect_uri`. You're passing through the grant flows the application and what that decision would have been by default, to encourage people towards using what the default is rather than getting into situations where they create clients that cannot actually do authorization flows.

The UI guidance, I think, is probably necessary here. I know that not every application is created through DoorKeeper's UIs. I know Mastodon, for instance, has its own custom UI for doing this. 

Your call on absence of redirects-based flows versus presence of client credentials: absolutely correct. That I was just trying to phrase it in a way that made somewhat sense when I was thinking about this PR and that config option.

I would probably be saying in the doc comment for `allow_blank_redirect_uri`: "You almost certainly do not want to change this. There are very, very few cases where you want to change this configuration value."

55728:
Good call — added in be76bd044043a75480855935d550631ba26df6be. I placed the note right after the paragraph explaining the default behavior, using your suggested wording verbatim, so anyone reading the initializer sees the caution before the false / custom-lambda examples.

nbulaj (approved):
LGTM , thanks ✅

PR #1891, opened 2026-07-31
Answer `invalid_redirect_uri` instead of 500 when the client registered no redirect URI

55728:
## Summary

`URIChecker.valid_for_authorization?` calls `client_url.split`, so a `nil` registered `redirect_uri` raises `NoMethodError` — an unhandled 500. One-line fix: coerce with `to_s`, so a `nil` registration behaves exactly like the empty string (nothing registered → no provided value can match → `invalid_redirect_uri`, per RFC 6749 §3.1.2.3's requirement that a provided value match a registered one).

## How the `nil` is reachable

- `allow_blank_redirect_uri` permits registering an application with a blank `redirect_uri`. The HTML form submits `""` (which already worked), but an application created over the console or an API can carry `redirect_uri = nil`.
- Any authorization request for such a client that **supplies** a `redirect_uri` parameter then crashes before client authentication — an unauthenticated 500. A request that omits the parameter is unaffected (rejected earlier by the `redirect_uri.blank?` guard).

Discovered while pinning the neighboring no-registered-URI behavior for #1682 (PR #1890); this PR is independent of that one and touches none of the same files.

## Other call site

`AuthorizationCodeRequest#validate_redirect_uri` passes `grant.redirect_uri`, which cannot be `nil` through the normal flow (pre-authorization requires a `redirect_uri` before a grant is created). For out-of-spec data the failure mode moves from a 500 to `invalid_grant` — also strictly an improvement.

### DIFF
--- a/lib/doorkeeper/oauth/helpers/uri_checker.rb
+++ b/lib/doorkeeper/oauth/helpers/uri_checker.rb
@@ -60,7 +60,10 @@ module Doorkeeper
         end
 
         def self.valid_for_authorization?(url, client_url)
-          valid?(url) && client_url.split.any? { |other_url| matches?(url, other_url) }
+          # client_url is nil when the application was registered without a
+          # redirect URI (allow_blank_redirect_uri) — no possible match, not
+          # an error.
+          valid?(url) && client_url.to_s.split.any? { |other_url| matches?(url, other_url) }
         end
 
         def self.as_uri(url)
--- a/lib/doorkeeper/oauth/helpers/uri_checker.rb
+++ b/lib/doorkeeper/oauth/helpers/uri_checker.rb
@@ -61,8 +61,8 @@ module Doorkeeper
 
         def self.valid_for_authorization?(url, client_url)
           # client_url is blank when the application was registered without a
-          # redirect URI (allow_blank_redirect_uri) — no possible match, not
-          # an error.
+          # redirect URI (allow_blank_redirect_uri) — no possible match for
+          # redirection based OAuth flows, not an error.
           return false if client_url.blank?
 
           valid?(url) && client_url.split.any? { |other_url| matches?(url, other_url) }
--- a/lib/doorkeeper/oauth/helpers/uri_checker.rb
+++ b/lib/doorkeeper/oauth/helpers/uri_checker.rb
@@ -60,10 +60,12 @@ module Doorkeeper
         end
 
         def self.valid_for_authorization?(url, client_url)
-          # client_url is nil when the application was registered without a
+          # client_url is blank when the application was registered without a
           # redirect URI (allow_blank_redirect_uri) — no possible match, not
           # an error.
-          valid?(url) && client_url.to_s.split.any? { |other_url| matches?(url, other_url) }
+          return false if client_url.blank?
+
+          valid?(url) && client_url.split.any? { |other_url| matches?(url, other_url) }
         end
 
         def self.as_uri(url)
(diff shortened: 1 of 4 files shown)


lib/doorkeeper/oauth/helpers/uri_checker.rb
```diff
@@ -60,7 +60,10 @@ def self.loopback_uri?(uri)
         end
 
         def self.valid_for_authorization?(url, client_url)
-          valid?(url) && client_url.split.any? { |other_url| matches?(url, other_url) }
+          # client_url is nil when the application was registered without a
+          # redirect URI (allow_blank_redirect_uri) — no possible match, not
+          # an error.
+          valid?(url) && client_url.to_s.split.any? { |other_url| matches?(url, other_url) }
```
ThisIsMissEm: I'm not sure I understand the change here, should it not be:
```suggestion
          return false unless valid?(url)
          return false if client_url.nil?
          return client_url.split.any? { |other_url| matches?(url, other_url) }
```

Obviously that can be written terser, but my brain doesn't want to think about that today.

ThisIsMissEm (commented):
The tests here look correct, I'm just unsure of the implementation (see comment)

lib/doorkeeper/oauth/helpers/uri_checker.rb
55728: Fair point 👀✨ — reworked it into an explicit guard:

```ruby
return false if client_url.blank?

valid?(url) && client_url.split.any? { |other_url| matches?(url, other_url) }
```

Functionally it's the same as the `to_s` coercion (`nil`/`""` → no registered URI → no match), but the early return is also how the rest of the code base treats absent values — `valid_scheme?` in this very file and `PreAuthorization#validate_redirect_uri` both use the `return false if x.blank?` shape — so this reads as the house idiom rather than a special case. Went with `blank?` over `nil?` to also cover an empty-string `redirect_uri`, which reaches this method the same way.

lib/doorkeeper/oauth/helpers/uri_checker.rb
```diff
@@ -60,6 +60,11 @@ def self.loopback_uri?(uri)
         end
 
         def self.valid_for_authorization?(url, client_url)
+          # client_url is blank when the application was registered without a
+          # redirect URI (allow_blank_redirect_uri) — no possible match, not
+          # an error.
```
ThisIsMissEm:
```suggestion
          # redirect URI (allow_blank_redirect_uri) — no possible match for redirection 
          # based OAuth flows, not an error.
```

ThisIsMissEm (approved):
LGTM, one minor suggestion on a comment that was a little short

nbulaj (approved):
❤️  nice finding, thanks!

55728:

lib/doorkeeper/oauth/helpers/uri_checker.rb
```diff
@@ -60,6 +60,11 @@ def self.loopback_uri?(uri)
         end
 
         def self.valid_for_authorization?(url, client_url)
+          # client_url is blank when the application was registered without a
+          # redirect URI (allow_blank_redirect_uri) — no possible match for redirection 
+          # based OAuth flows, not an error.
```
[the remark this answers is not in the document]
55728: Busted the ghost 👻

nbulaj (approved):
LGTM ✅  Thanks!

Issue #1894, opened 2026-07-31
Document custom client authentication methods in README


ThisIsMissEm:
In a similar vein to #1888, we should probably add documentation to the README to explain the new Client Authentication Methods registry, and how additional methods can be registered.

55728:
I can pick this up! 👀✨ The grant flow docs (#1888) worked well as a README section + matching spec, so I'd follow the same shape here.

`private_key_jwt` (#1880) seems like the natural example to walk through — it'd be the first real addition to the registry and is directly relevant to CIMDs. Since that PR is still in draft, I'd write the README example so it stands on its own as a "registering a custom method" walkthrough rather than depending on #1880 landing.

ThisIsMissEm:
Sure! That fits what I was thinking too.

Issue #1875, opened 2026-07-23
Client ID Metadata Documents (CIMDs)


ThisIsMissEm:
Hi all, so the work I'd previously started in 2024 was all in the progress of what would eventually become myself and Aaron Parecki defining the Client ID Metadata Documents internet draft, which has since been adopted by the OAuth WG at the IETF, and used in places like AT Protocol, Model Context Protocol (MCP servers), etc.

I don't think this draft could easily be implemented in user-land, but it basically solves the unbounded client proliferation problem of Dynamic Client Registration and the "pre-registration" problem in decentralized ecosystems.

Given we now have the Client Authentication Method Registry (#1772) we can now implement in user-land or in core the `private_key_jwt` client authentication method, which could be lifted up to core from the existing implementation in doorkeeper-openid which @55728 maintains. That unlocks the ability to do authenticated "more trusted" clients when using CIMDs.

Essentially what the CIMD draft specification is is an ability for a relying party (client) to say to the authorization server (IdP), "hey, here's my metadata for dynamic registration at this URL, if you don't know about me, fetch that document and use that metadata for my registration" — this is a feature that I *know* @mastodon want for their OAuth implementation, but we've never been able to sort out funding for it.

This is an issue to track CIMDs in doorkeeper, whether implemented in core or in user-land. The other work was essentially preparatory for being able to do this amongst a few other things. (I don't have time to work on this)

- Overview: https://client.dev/
- Current draft: https://www.ietf.org/archive/id/draft-ietf-oauth-client-id-metadata-document-02.html

55728:
Thanks for opening this, @ThisIsMissEm — I've read through the [draft-02](https://www.ietf.org/archive/id/draft-ietf-oauth-client-id-metadata-document-02.html) and wanted to share some thoughts on a possible roadmap.

**Phase 1 — Core (MUST-level requirements)**

- Accept `https://` URLs as `client_id` alongside existing random-string IDs (Section 3, 6.9)
- Metadata document fetcher with SSRF protection, no-redirect, 200-only (Section 4, 6.5)
- Virtual/dynamic application object from fetched metadata
- `client_id_metadata_document_supported` in AS metadata (Section 5) — minor addition to #1838
- Lift `private_key_jwt` from doorkeeper-openid_connect into core — not strictly MUST in the spec, but practically required for confidential clients since all shared-secret methods are prohibited (Section 4.1)

**Phase 2 — Hardening (SHOULD-level)**

[...]

PR #1896, opened 2026-08-01
Add built-in private_key_jwt client authentication (RFC 7523)

55728:
Part of #1875 — the first of two stacked PRs split out of #1880, as discussed there (`private_key_jwt` lands first, Client ID Metadata Documents stack on top).

Registers a `private_key_jwt` method (RFC 7523 §2.2 / OIDC Core §9) in the client authentication registry (#1840). It is **not** part of `DEFAULT_METHODS` — servers opt in through the `client_authentication` config option — and the `jwt` gem (>= 2.7) is required only when an assertion is actually verified, so it stays an optional dependency.

## What's in here

- **Assertion verification**, deliberately strict: asymmetric algorithms only (HMAC / `none` can never verify, symmetric `oct` JWKs are dropped from the key set), `iss` = `sub` = `client_id` with an optional `client_id` parameter cross-check (RFC 7521 §4.2), `aud` = configured issuer or token endpoint URL (derived from the server's own configuration — only a server that configures neither an `issuer` nor `default_url_options` falls back to the request, matching `MetadataResponse`), bounded `exp`, single-use `jti` (process-local replay guard). Malformed input — a JWT payload that isn't an object, a JWK member that isn't valid base64url, an EC point off its curve — fails authentication instead of raising out of the token endpoint.
- **Key resolution** from `jwks` / `jwks_uri` attributes you define on your Application model (Doorkeeper adds no columns itself; the dummy app defines them as the reference setup).
- **`Doorkeeper::HttpFetcher`** — an SSRF-hardened HTTP client used to fetch `jwks_uri`: HTTPS only, no redirects, 200 OK only, RFC 6890 special-use addresses refused with the connection pinned to the vetted IP (DNS-rebinding-proof), 5 KB response cap and a wall-clock deadline on the exchange. Results are memoized in **`Doorkeeper::DocumentCache`** (small fixed-TTL memo). Both are top-level classes because the stacked CIMD PR fetches metadata documents through the same client.
- **A `uses_shared_secret?` registry contract**: strategies declare whether they authenticate with a shared symmetric secret (`client_secret_basic`/`_post` and legacy callables: `true`; `none` / `private_key_jwt`: `false`), with a conservative name-based fallback for strategies that don't declare it. Callers that must refuse secret-based methods can ask the registry instead of hard-coding method names.
- The `none` method no longer matches a request carrying a `client_assertion`, so an assertion can never be downgraded to an unauthenticated public client; the refresh token grant now resolves its client through `OAuth::Client` like every other grant, which is what lets secret-less credentials work there.
- A dedicated CI job (`gemfiles/rails_8_0_jwt_2.gemfile`) exercises the jwt gem's declared lower bound, which the regular matrix never resolves.

>[!NOTE]
> 
> - This PR is intended to double as the worked example for #1894 (documenting how to add a custom client authentication method): `PrivateKeyJwt` shows the full surface — `matches_request?`, `authenticate`, `uses_shared_secret?`, registry registration and the opt-in config.
> - Full suite green (1771 examples). Every behavior above has a spec, including the negative space (oct-key downgrade attempts, replayed assertions, Host-header audiences, hostile JWK sets, transport failures at the `jwks_uri`).
> 
> The second PR in the stack adds Client ID Metadata Documents on top: URL `client_id`s, document validation and materialization, and document-based key resolution wired into this method's `KeyResolver`.

### DIFF
--- a/lib/doorkeeper/oauth/client_authentication/private_key_jwt.rb
+++ b/lib/doorkeeper/oauth/client_authentication/private_key_jwt.rb
@@ -1,0 +1,227 @@
+# frozen_string_literal: true
+
+require "uri"
+
+require "doorkeeper/oauth/client_authentication/private_key_jwt/key_resolver"
+require "doorkeeper/oauth/client_authentication/private_key_jwt/replay_guard"
+
+module Doorkeeper
+  module OAuth
+    module ClientAuthentication
+      # "private_key_jwt" client authentication (RFC 7523 / OIDC Core §9):
+      # the client authenticates with a JWT assertion signed by its private
+      # key; the server verifies it against the client's published public
+      # keys (a jwks attribute on the application model, or keys fetched
+      # from its jwks_uri). No shared secret is involved.
+      #
+      # The "jwt" gem is required only when an assertion is actually
+      # authenticated, so servers that don't enable this method don't need
+      # the dependency.
+      class PrivateKeyJwt
+        CLIENT_ASSERTION_TYPE = "urn:ietf:params:oauth:client-assertion-type:jwt-bearer"
+
+        # Assertions are verified against the client's published public
+        # keys; no shared secret is involved.
+        def self.uses_shared_secret?
+          false
+        end
+
+        # Asymmetric signature algorithms only: HMAC family (and "none") are
+        # shared-secret/unauthenticated and must never verify an assertion.
+        ALLOWED_ALGORITHMS = %w[RS256 RS384 RS512 PS256 PS384 PS512 ES256 ES384 ES512].freeze
+
+        # iss/sub identify the client, aud prevents cross-server replay,
+        # exp bounds the assertion lifetime and jti makes it single-use
+        # (OIDC Core §9 requires all of these for private_key_jwt).
+        REQUIRED_CLAIMS = %w[iss sub aud exp jti].freeze
+
+        # Upper bound on how far in the future an assertion may expire. This
+        # both rejects sloppily long-lived assertions and bounds the replay
+        # guard's memory.
+        MAX_LIFETIME = 3600
+
+        def self.matches_request?(request)
+          params = request.request_parameters.with_indifferent_access
+
+          request.post? &&
+            params[:client_assertion].present? &&
+            params[:client_assertion_type] == CLIENT_ASSERTION_TYPE
+        end
+
+        def self.authenticate(request)
+          require_jwt!
+
+          params = request.request_parameters.with_indifferent_access
+          assertion = params[:client_assertion].to_s
+
+          client_id = unverified_client_id(assertion)
+          return if client_id.blank?
+          # RFC 7521 §4.2: a client_id parameter sent alongside the assertion
+          # must agree with the assertion's issuer.
+          return if params[:client_id].present? && params[:client_id] != client_id
+
+          application = OAuth::Client.find(client_id)&.application
+          return unless application
+
+          jwk_set = KeyResolver.jwk_set_for(application)
+          return unless jwk_set
+
+          claims = verified_claims(assertion, client_id, jwk_set, request)
+          return unless claims
+          return unless ReplayGuard.instance.first_use?(
+            "#{client_id}:#{claims["jti"]}",
+            expires_at: claims["exp"].to_i,
+          )
+
+          Doorkeeper::ClientAuthentication::VerifiedCredentials.new(client_id)
+        end
+
+        # The issuer read without verifying the signature — only used to
+        # locate the client (and thereby its keys); every claim is verified
+        # against those keys before the assertion authenticates anyone.
+        def self.unverified_client_id(assertion)
+          claims, = JWT.decode(assertion, nil, false)
+          # A JWT payload is any JSON value, not necessarily an object, and
+          # nothing is verified at this point — so the decoded claims are
+          # type-checked before being indexed into.
+          return unless claims.is_a?(Hash)
+
+          issuer = claims["iss"]
+
+          issuer if issuer.is_a?(String) && issuer == claims["sub"]
+        rescue JWT::DecodeError
+          nil
+        end
+        private_class_method :unverified_client_id
+
+        def self.verified_claims(assertion, client_id, jwk_set, request)
+          claims, = JWT.decode(
+            assertion,
+            nil,
+            true,
+            algorithms: ALLOWED_ALGORITHMS,
+            jwks: jwk_set,
+            required_claims: REQUIRED_CLAIMS,
+            iss: client_id,
+            verify_iss: true,
+            sub: client_id,
+            verify_sub: true,
+            aud: acceptable_audiences(request),
+            verify_aud: true,
+          )
+
+          return unless claims["exp"].to_i <= Time.now.to_i + MAX_LIFETIME
+          return unless claims["jti"].is_a?(String) && claims["jti"].present?
+
+          claims
+        rescue JWT::DecodeError, OpenSSL::OpenSSLError
+          # A published key is only parsed far enough to be usable when it is
+          # actually needed to verify a signature, so a structurally valid but
+          # mathematically nonsensical key (an EC point that is not on the
+          # curve, say) surfaces here as a bare OpenSSL error rather than a
+          # JWT one. Both mean the same thing: this assertion does not verify.
+          nil
+        end
+        private_class_method :verified_claims
+
+        # RFC 7523bis expects the issuer identifier as the audience; older
+        # deployments use the token endpoint URL. Both are accepted.
+        #
+        # OIDC Core §9 says the audience SHOULD be the token endpoint URL, and
+        # clients follow that for every endpoint they authenticate at — so the
+        # token endpoint URL is accepted at the revocation and introspection
+        # endpoints too, not just at the endpoint being called.
+        #
+        # The endpoint URLs are built from the server's own configured
+        # identity, never from the request: aud is what keeps an assertion
+        # minted for another authorization server from being replayed here, so
+        # deriving it from the client-supplied Host header would defeat its
+        # purpose on any deployment that does not filter hosts. Only a server
+        # that identifies itself nowhere falls back to the request, which is
+        # how MetadataResponse derives its issuer as well.
+        def self.acceptable_audiences(request)
+          options = server_url_options(request)
+
+          [
+            Doorkeeper.config.issuer.presence,
+            "#{base_url(options)}#{request.path}",
+            token_endpoint_url(options),
+          ].compact.uniq
+        end
+        private_class_method :acceptable_audiences
+
+        def self.server_url_options(request)
+          configured_url_options ||
+            { protocol: request.protocol, host: request.host, port: request.optional_port }
+        end
+        private_class_method :server_url_options
+
+        # An explicitly configured canonical host wins; failing that, the
+        # issuer, when it is an absolute URL (Doorkeeper allows any string).
+        def self.configured_url_options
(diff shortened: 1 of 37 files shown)


lib/doorkeeper/oauth/client_authentication/private_key_jwt.rb
```diff
@@ -0,0 +1,227 @@
...
+            required_claims: REQUIRED_CLAIMS,
+            iss: client_id,
+            verify_iss: true,
+            sub: client_id,
+            verify_sub: true,
+            aud: acceptable_audiences(request),
+            verify_aud: true,
+          )
+
+          return unless claims["exp"].to_i <= Time.now.to_i + MAX_LIFETIME
```
nbulaj: What if attacker gonna use some weird numbers somehow? 👀  Some very big bigint or infinity float? Also I'm afraid of Ruby auto-type-casting for strings, should we validate that `exp` is a number at least?

lib/doorkeeper/oauth/client_authentication/private_key_jwt/replay_guard.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
+
+require "singleton"
+
+module Doorkeeper
+  module OAuth
+    module ClientAuthentication
+      class PrivateKeyJwt
+        # In-memory, process-local single-use guard for assertion jti values
```
nbulaj, line 9: Won't it affect multi-worker environments? If yes:

* can we explicitly document it?
* if possible - can we allow to configure own `ReplayGuard` entity? Which can use Redis or whatever

Also I think it's good to document `MAX_LIFETIME` and `MAX_ENTRIES` so everyone is aware

lib/doorkeeper/oauth/client_authentication/private_key_jwt/key_resolver.rb
```diff
@@ -0,0 +1,92 @@
...
+              parsed = JSON.parse(Doorkeeper::HttpFetcher.new.fetch(url))
+              parsed if parsed.is_a?(Hash)
+            end
+          rescue Doorkeeper::HttpFetcher::FetchError, JSON::ParserError, URI::InvalidURIError
+            nil
+          end
+          private_class_method :fetch_jwks
+
+          def self.jwks_cache
+            @jwks_cache ||= Doorkeeper::DocumentCache.new
```
nbulaj: Should we also allow this to be configurable (with `DocumentCache` by default)?  Or at least allow to configure TTL and some other important options? I'm trying to think for the future 👀

lib/doorkeeper/oauth/client_authentication/private_key_jwt.rb
```diff
@@ -0,0 +1,227 @@
...
+          )
+
+          Doorkeeper::ClientAuthentication::VerifiedCredentials.new(client_id)
+        end
+
+        # The issuer read without verifying the signature — only used to
+        # locate the client (and thereby its keys); every claim is verified
+        # against those keys before the assertion authenticates anyone.
+        def self.unverified_client_id(assertion)
+          claims, = JWT.decode(assertion, nil, false)
```
nbulaj: `JWT` allows a global config for `verify_expiration` to be set to `false`. Maybe we need to force `verify_expiration: true` option here? Or we have to follow app configuration?

nbulaj (approved):
Some minor comments to fix/discuss, otherwise looks solid (as always) 💪  Thanks!

lib/doorkeeper/oauth/client_authentication/private_key_jwt.rb
55728: Good catch 👀 I dug into this and confirmed the hole is real: the `jwt` gem validates the type on *encode* but not on *decode* — its expiration check just calls `to_i`, so a validly-signed assertion with `"exp" => "1785758290"` (a string) passes right through.

Fixed by pinning `exp` to a finite `Numeric` before comparing it or handing it to the replay guard. This rejects strings, `Infinity`, and absurdly large integers (which already fail the `MAX_LIFETIME` ceiling). I also pass `verify_expiration: true` explicitly to the decode call — see the thread on comment ④.

The string-exp spec is hand-signed because `JWT.encode` refuses to mint one, which is a nice confirmation that the gap is decode-only.

lib/doorkeeper/oauth/client_authentication/private_key_jwt/replay_guard.rb
55728: You're right 👍 — documented the process-local scope in both the class comment and the initializer template, including `MAX_ENTRIES` (10,000) and `MAX_LIFETIME` (1 hour).

Also added a `private_key_jwt_replay_guard` config option: any object answering `first_use?(key, expires_at:)` can replace the built-in guard, so a Redis-backed store (or anything shared) plugs right in:

```ruby
private_key_jwt_replay_guard MyRedisReplayGuard.new
```

lib/doorkeeper/oauth/client_authentication/private_key_jwt/key_resolver.rb
55728: Added a `private_key_jwt_jwks_cache` config option. `DocumentCache` remains the default; passing `Doorkeeper::DocumentCache.new(ttl: 300)` changes just the TTL, or any object answering `fetch(url) { ... }` can replace it entirely (e.g. a `Rails.cache`-backed wrapper):

```ruby
private_key_jwt_jwks_cache Doorkeeper::DocumentCache.new(ttl: 300)
```

lib/doorkeeper/oauth/client_authentication/private_key_jwt.rb
55728: Went with "force it" — `verify_expiration: true` is now passed explicitly to `JWT.decode`, so the host app's global `JWT.configuration.decode.verify_expiration = false` (which is typically aimed at its own tokens) cannot silently disable the only expiration check on client assertions. Pinned with a spec that globally disables it and confirms expired assertions are still rejected.

55728:
Thank you for the thorough review on this huge PR, @nbulaj 😊🙌 All four comments were spot-on 👀✨ — especially the `exp` type check, which turned out to be a real gap in the `jwt` gem's decode path. Addressed everything in the two new commits; force-pushing shortly!

lib/doorkeeper/oauth/client_authentication/private_key_jwt/key_resolver.rb
```diff
@@ -0,0 +1,100 @@
...
+            raw = raw_jwks(application)
+            return unless raw.is_a?(Hash)
+
+            keys = raw["keys"] || raw[:keys]
+            return unless keys.is_a?(Array)
+
+            asymmetric = keys.grep(Hash).reject { |key| (key["kty"] || key[:kty]).to_s == "oct" }
+            return if asymmetric.empty?
+
+            JWT::JWK::Set.new({ "keys" => asymmetric })
```
nbulaj: Won't we see namespace issues if used with [doorkeeper-jwt](https://github.com/doorkeeper-gem/doorkeeper-jwt/blob/master/lib/doorkeeper/jwt.rb#L8) lib? 👀 I **_believe_** no and constant lookup will work correctly (and Zeitwerk as well), but I have to confirm you  know 🙂

lib/doorkeeper/oauth/client_authentication/private_key_jwt.rb
```diff
@@ -0,0 +1,245 @@
...
+          # A JWT payload is any JSON value, not necessarily an object, and
+          # nothing is verified at this point — so the decoded claims are
+          # type-checked before being indexed into.
+          return unless claims.is_a?(Hash)
+
+          issuer = claims["iss"]
+
+          issuer if issuer.is_a?(String) && issuer == claims["sub"]
+        rescue JWT::DecodeError
```
nbulaj: Shouldn't it have more possible cases, just like `OpenSSL::OpenSSLError` in the `verified_claims` ? 👀  Is it possible to have a JSON error here and below as well?  Trying to save us from possible 500 server errors

nbulaj (approved):
Did one more quick look, small questions for clarifications please 🙈  And ready from my end ✅

lib/doorkeeper/oauth/client_authentication/private_key_jwt/key_resolver.rb
55728: Great catch — and it turns out the answer is the opposite of what we both hoped: **it did break**. 😅

Bare `JWT` inside `module Doorkeeper` resolves through the lexical nesting before the top level, so once doorkeeper-jwt defines `Doorkeeper::JWT`, every `JWT` reference in this class picks up that module instead of the jwt gem. Ruby dropped the top-level fallback for qualified constants back in 2.5, so `JWT::JWK::Set` becomes `Doorkeeper::JWT::JWK::Set` and raises `NameError` — and the `rescue JWT::DecodeError` clauses resolve the same way, so the NameError escaped the token endpoint as a 500. (Zeitwerk isn't involved — it's pure lexical constant lookup, and gem code isn't managed by the host app's loader anyway.)

Fixed in cfae614b by qualifying every gem reference as `::JWT`, with a regression spec that runs the happy path and the rescue path under a stubbed `Doorkeeper::JWT` (both examples fail with `NameError` before the fix).

lib/doorkeeper/oauth/client_authentication/private_key_jwt.rb
55728: Checked this against jwt 3.2.0's source and empirically — `JWT::DecodeError` alone is sufficient here, for two reasons:

1. **JSON / Base64 errors never escape the gem.** `JWT.decode` wraps `JSON::ParserError` into `JWT::DecodeError` ("Invalid segment encoding", `lib/jwt/encoded_token.rb`), and invalid base64url raises `JWT::Base64DecodeError`, which is a subclass of `JWT::DecodeError`. I fuzzed the unverified path with garbage tokens, broken JSON in header and payload, broken base64url, 500-deep nested JSON (`JSON::NestingError`), empty strings and extra segments — everything raises inside the `JWT::DecodeError` hierarchy. Non-object payloads (arrays, numbers) don't raise at all; those are handled by the `claims.is_a?(Hash)` type check.

2. **`OpenSSL::OpenSSLError` can't happen on this path.** With `verify: false` the gem skips signature verification and never touches the key or any OpenSSL API (`lib/jwt/decode.rb` — `verify_signature` is only called when `@verify` is set), so there's no OpenSSL call to fail. That's why `verified_claims` and `KeyResolver.jwk_set_for` rescue it (they parse client-supplied keys and verify signatures) but this method doesn't need to.

The same applies to `verified_claims` for JSON errors — same wrapping, same hierarchy, so `rescue JWT::DecodeError, OpenSSL::OpenSSLError` there covers it.

lib/doorkeeper/oauth/client_authentication/private_key_jwt/key_resolver.rb
nbulaj: Thanks for checking!

image

lib/doorkeeper/oauth/client_authentication/private_key_jwt.rb
nbulaj: Great then, thanks!

nbulaj:
Landed :white_check_mark:  Thanks a lot!

55728:
Thank you @nbulaj! 😊

lib/doorkeeper/oauth/client_authentication/private_key_jwt/key_resolver.rb
55728: I'm ready!

PR #1898, opened 2026-08-01
Fix token reuse for client_credentials with resource indicators

55728:
Follow-up to #1886.

`ClientCredentials::Creator` reuses an existing token whenever it matches on client and scopes, but `resource` never takes part in that decision — it's a first-class column rather than a custom attribute, so `extract_custom_attributes` drops it before the lookup.

With `reuse_access_token` enabled, that means:

- an existing unrestricted token is returned for a request asking for `resource=https://api-a.example.com/`
- a token bound to `https://api-a.example.com/` is returned for a request asking for `https://api-b.example.com/`

Either way the audience restriction the client asked for is not applied, and introspection then reports an `aud` that doesn't match what was requested.

The other two reuse paths added in #1886 already guard against this — `AccessToken.find_or_create_for` and `AuthorizationsController#matching_token?` both check `resource_indicators_match?`. The client_credentials flow doesn't go through `find_or_create_for`, so this path was easy to miss; this checks the same thing there.

The check sits on the reuse decision rather than on the lookup itself, which keeps this PR limited to the reuse bug: `revoke_previous_client_credentials_token` still finds and revokes the client's previous token whatever audience it was issued for, exactly as before. There's a spec pinning that so it's visible the behaviour didn't move.

That said — I think that revocation behaviour is worth revisiting separately. A client talking to two resources currently can't hold both tokens at once under that option, since each request revokes the other's token, which makes it hard to combine with resource indicators. RFC 9700 §4.10.2 notes a client accessing several resource servers needs a separate token per server. Happy to open a follow-up for it if you agree it's worth changing; it seemed like the wrong thing to bundle into a bug fix.

Specs cover both mismatch cases plus two positive ones (same resource still reuses, and unrestricted-with-no-resource still reuses), so ordinary client_credentials reuse stays pinned.


CHANGELOG.md
```diff
@@ -17,10 +17,10 @@ User-visible changes worth mentioning.
...
 - [#1884] Fix: `/oauth/introspect` and `/oauth/revoke` extend the token lookup across both token types when the lookup by `token_type_hint` finds nothing (RFC 7662 §2.1 / RFC 7009 §2.1), so a wrong hint no longer hides a token the server knows about ([#1882])
 - [#1885] Index `oauth_access_grants.access_token_id` in the migration templates: since [#1879] the code-replay revocation filters access grants by that column, the "never used to filter queries" premise behind `index: false` no longer holds.
+- [#1886] Add support for Resource Indicators for OAuth 2.0 (RFC 8707). Clients can include a `resource` parameter in authorization and token requests to indicate the target protected resource(s). The authorization server validates resource URIs, enforces audience restriction on tokens, and includes `aud` in introspection responses. Enable by configuring `resource_indicator_validator` with a callable. Requires new `resource` columns on access grants and tokens — run `rails generate doorkeeper:resource_indicators` to add the migration.
```
nbulaj, line 20: Should we also note current PR? 👀

55728: Thanks for the review @nbulaj! 😊 Added an entry of current PR 👀✨

nbulaj (approved):
Very thanks for that founding @55728 , appreciated 🙇

PR #1888, opened 2026-07-31
Document custom grant flow registration in the README, closes #764

55728:
## Summary

Closes #764.

The technical ask of #764 — support grant types whose names are URNs/URIs, such as the SAML 2.0 bearer assertion grant (`urn:ietf:params:oauth:grant-type:saml2-bearer`, [RFC 7522 §2.1]) — was resolved back in 5.5.0 by the grant flow registry (#1418): grant types are mapped to strategy classes through explicit registration (`grant_type_matches` accepts any `String` or `Regexp`) instead of the old camelize/constantize resolution the issue complained about.

What never happened is documentation: the issue's opening question was literally *"Is there any documentation for creating custom grant types?"*, and to this day the registry appears in neither the README nor the wiki — only in the CHANGELOG and the source. This PR closes that gap:

- **README**: new "Custom Grant Flows" section walking through `Doorkeeper::GrantFlow.register` with a SAML 2.0 bearer assertion (RFC 7522) example — flow registration, `grant_flows` configuration, the `Doorkeeper::Request::Strategy` subclass, and a `Doorkeeper::OAuth::BaseRequest` subclass with grant-specific validations. Also points at `response_type_matches`/`response_type_strategy` and `register_alias` for authorization-endpoint flows and extension aliases.
- **Spec**: an end-to-end request spec mirroring the README example, pinning URN-shaped grant types at the `/oauth/token` level (token issuance for a valid assertion, `invalid_grant` for an unverifiable one, `unsupported_grant_type` for unregistered types). The registry so far had unit specs only.

The README example is verified by the added spec. One subtlety it documents: `BaseRequest#authorize` reads `access_token` but doesn't define the reader, so subclasses must declare `attr_reader :access_token` themselves (as `PasswordAccessTokenRequest` does).

Docs/test-only change — no behavior changes.

[RFC 7522 §2.1]: https://www.rfc-editor.org/rfc/rfc7522#section-2.1

PR #1899, opened 2026-08-01
Document custom client authentication methods in the README

### LINKED ISSUE #1894: Document custom client authentication methods in README

55728:
Closes #1894.

Same shape as #1888: a README section plus an end-to-end request spec that mirrors the example, so the documented walkthrough is actually exercised rather than just written down.

The example registers a method that reads credentials a partner integration sends in its own headers. I deliberately kept it independent of #1896, so it reads as a "registering a custom method" walkthrough on its own and doesn't need that PR to land first.

Alongside the `matches_request?` / `authenticate` contract and the fact that `client_authentication` is ordered, the section calls out two things that seemed easy to get wrong while writing it:

- `matches_request?` is evaluated across the **whole registry** for the RFC 6749 §2.3 "no more than one method" rule, not just the enabled methods — so a matcher that is too broad collides with a built-in method and the request comes back `invalid_request`.
- The returned credentials are resolved with `by_uid_and_secret`, so a blank secret only ever resolves a public client. A method that proves identity some other way still has to produce the registered secret for a confidential client. (Verified: confidential + blank secret resolves to `nil`.)

The spec covers the happy path, a wrong secret, the built-in methods still working alongside it, the §2.3 collision, and the method showing up in `token_endpoint_auth_methods_supported`.

Happy to swap the example for something else if you'd prefer a different one — a standard method like mTLS felt misleading given the `by_uid_and_secret` constraint above.


nbulaj (approved):
Great work! 👍

55728:
Thank you for the review @nbulaj! 😊

Heads-up: it looks like the `merge=union` driver doesn't kick in for merges performed on GitHub's side (Squash, Rebase, or even Create a merge commit) — `.gitattributes` merge drivers are only honoured by local `git merge`.

So CHANGELOG conflicts will still pop up for PRs merged via the web UI, unfortunately 😓

I'm going to look into an alternative approach to keep the friction low. Will follow up! 👀✨

nbulaj:
Yeah I also noticed that from the GitHub perspective nothing changed, unfortunately.  Thanks anyway! :bow:

Issue #1600, opened 2022-11-08
Is there a way to store Doorkeeper::AccessToken data inside Doorkeeper::JWT (and not use db-backed tokens)?


NeilSlater-Clixifix:
### Steps to reproduce

Not a bug report, but a feature request for relying on token data storage *inside* issued JWT

I *think* this is either an undocumented existing feature (in which case I would appreciate a "how to" link), or more likely something that Doorkeeper 5.5 does not support.

### Expected behavior

There should be a way to inform Doorkeeper / Doorkeeper::JWT to not use the`Doorkeeper::AccessToken` database model, and that instead the details such as client id, approved scopes etc should be stored and read direct from the JWT. This is a common use-case for JWTs as an authentication mechanism, that they *embed* data for access control.

### Actual behavior

As far as I can tell from a quick browse of the code and local testing, Doorkeeper always uses database-backed access tokens, and JWTs are not exceptions - the whole serialised JWT is used as an index string to lookup server-side credentials. That means for busy API services, we cannot explore the possibility of skipping the db I/O for granting and reading access tokens.

### System configuration

**Doorkeeper initializer**:

```ruby
# config/initializers/doorkeeper.rb
Doorkeeper.configure do
  access_token_generator '::Doorkeeper::JWT'
end
```

**Doorkeeper JWT initializer**:

```ruby
# config/initializers/doorkeeper_jwt.rb

Doorkeeper::JWT.configure do
  token_payload do |opts|
    {
      iss: SITE,
      iat: Time.now.utc.to_i,

      # see JWT reserved claims - https://tools.ietf.org/html/draft-jones-json-web-token-07#page-7
      jti: SecureRandom.uuid,

      # Illustrative example (client credentials in this case, so no User information)
      doorkeeper_access: {
        application_id: 2,
        expires_in: 7200,
        scopes: "core_api read write"
      }
    }
  end
end
```

User and Doorkeeper::Application data could be similarly embedded on granting the token. This will make the token longer, but allow clients to skip more database reads on auth checks. Usually this would be combined with shorter token expiry times, and matches patterns of use for client apps that repeatedly make "one shot" requests and don't store access tokens.

nbulaj:
Currently there is now way to use just raw JWT tokens without database (and it's not possible for Doorkeeper by design as we can't revoke such tokens and do some other things which is a core of the gem)

nbulaj:
[...]

Issue #1719, opened 2024-07-25
Refresh Tokens as-implemented are susceptible to Refresh Token Reuse Attacks


ThisIsMissEm:
Currently there is no chain-of-trust relationship between access tokens and refresh tokens, this means that if an attacker gains access to a refresh token, then they may still be able to attempt to use it, without all access tokens stemming from that stolen refresh token being revoked.

Auth0 has an extremely [good write up](https://auth0.com/blog/refresh-tokens-what-are-they-and-when-to-use-them/#Refresh-Token-Automatic-Reuse-Detection) on this which explains the attack vector.

The solution here is to create a "token family", where refresh tokens are a separate record to access tokens, and they are linked together in a multiple directional way; e.g., you can start at one access token and revoke all subsequent access tokens stemming from that access token.

Having a separate refresh token model would also allow for expiration of refresh tokens independent of access tokens (which I [don't believe](https://github.com/doorkeeper-gem/doorkeeper/blob/main/lib/doorkeeper/oauth/refresh_token_request.rb) is currently possible)

ransombriggs:
I hacked "chain-of-trust" into our implementation and was wondering what level of interest there is in this from the maintainers of doorkeeper.  If there is interest I can share the data model changes and if those look good, I can put together some pull requests.

ThisIsMissEm:
I would be very interested in this for Mastodon

55728:
Related: #1932 (scope separation on refresh tokens — same structural area)

PR #1901, opened 2026-08-04
Tests only: Add spec coverage to pin answered issue behaviors

### LINKED ISSUE #984: Is there Digest support for Doorkeeper Client Credentials Authentication

### LINKED ISSUE #1291: Add IndieAuth support
spec: https://www.w3.org/TR/indieauth/

Basically, all this would require for minimal support is:

- support `id` as a `response_type`
- don't look in the database for client info for `id` requests
- support the `me` JSON response from the authorization endpoint

/cc @aaronpk

### LINKED ISSUE #1554: Response header too long for default nginx configuration
### Steps to reproduce

Doorkeeper is used in Mastodon. An app developer encounters error when trying to authenticate a user through OAuth with a long URL or a long state param, which causes Doorkeeper to return a response with a header so long that it causes nginx to return HTTP 502 when using default proxy buffer values.

Please see linked issue mastodon/mastodon#12915.

### Expected behavior

Either:

- OAuth response is returned that can be processed by nginx correctly
- An error is raised in Ruby because user input is too long
- Documentation is updated to advise about min. required proxy buffer size

### Actual behavior

- OAuth response contains such a long header it causes a HTTP 502 `upstream sent too big header while reading response header from upstream` error in nginx with default proxy buffer sizes

### System configuration

**Doorkeeper initializer**:

https://github.com/mastodon/mastodon/blob/50ab3f3dcb6b00109fa1462a5ca0228563abb99b/config/initializers/doorkeeper.rb

**Ruby version**: `3.0.3`

**Gemfile.lock**:

https://github.com/mastodon/mastodon/blob/50ab3f3dcb6b00109fa1462a5ca0228563abb99b/Gemfile.lock

### LINKED ISSUE #1663: reuse_access_token config should be used when refreshing a token
Maybe I am wrong, but I feel like `reuse_access_token` should apply when refreshing a token that has not expired yet. 

An improper use of the `refesh_token` grant type could lead to the creation of an unnecessary number of access tokens and bloat the underlying database.

I believe the only change needed to make this possible would be to replace `create_for` [here](https://github.com/doorkeeper-gem/doorkeeper/blob/b3d6787341561e5bc51336e3e04622ca131d6445/lib/doorkeeper/oauth/refresh_token_request.rb#L78) by the `find_or_create_for` method of the access token mixin.

### LINKED ISSUE #1756: How to implement using multi-database like Makara
### Steps to reproduce
I'm using Makara to scale my Database, but currently, due to replication lag, I have encountered an error 401 even though I already inserted/created oauth_access_tokens.

Is there a way in the configuration to route all token related query to the Main database?

### LINKED ISSUE #1759: Introspecting a token should behave the same wrt "revoke on use"
### Steps to reproduce

Our ResourceServer is a separate application from the AuthorizationServer, so we can't modify our controllers to invoke `doorkeeper_authorize!`. Instead, we call `token/introspect` on provided tokens. The RS is itself a client application; it uses `client_credentials` and has a special `introspect` scope to allow it to examine tokens issued by any other client. We noticed that revoke-on-use does not happen as it would if we could use the provided `doorkeeper_authorize!` action helper. 

### Expected behavior

Calling `token/introspect` on a token should have the same behavior as `doorkeeper_authorize!` has on it.

### Actual behavior

The tokens aren't revoked, and we don't know what's the appropriate path to create portable behavior that is identical to `doorkeeper_authorize!` 

### System configuration

Here's a diff that adds a new example for the TokensController

```
diff --git a/spec/controllers/tokens_controller_spec.rb b/spec/controllers/tokens_controller_spec.rb
index de0088b6..eb55fb1f 100644
--- a/spec/controllers/tokens_controller_spec.rb
+++ b/spec/controllers/tokens_controller_spec.rb
@@ -496,6 +496,18 @@ RSpec.describe Doorkeeper::TokensController, type: :controller do
         )
       end
 
+      it "revokes the previous refresh_token of the token being introspected" do
+        previous_token = FactoryBot.create(:access_token, refresh_token: "refresh_token")
+        token_for_introspection.previous_refresh_token = previous_token.refresh_token
+        token_for_introspection.save!
+
+        request.headers["Authorization"] = "Bearer #{access_token.token}"
+
+        post :introspect, params: { token: token_for_introspection.token }
+
+        expect(previous_token.reload).to be_revoked
+      end
+
       it "responds with invalid_token error if authorized token doesn't have introspection scope" do
         access_token.update(scopes: "read write")
 
```

### LINKED ISSUE #1764: Issue: Configuring Active Record to Use Different Base Classes for Sharded and Non-Sharded Models
**Background**
In our legacy Rails application, we have been using the active_record_shards gem to handle database sharding. Prior to upgrading to Rails 7.2, we patched certain models (e.g., Doorkeeper models) using not_sharded, ensuring they always connected to the primary database:
```
Doorkeeper::Application.class_eval do
  not_sharded
  ...
end
```
With the upgrade, we introduced two base classes to handle this explicitly:
```
class GlobalRecord < ActiveRecord::Base
  self.abstract_class = true
  connects_to database: { writing: :primary }
end
```
```
class ApplicationRecord < ActiveRecord::Base
  self.abstract_class = true
  ...
  # Defaults to shards
end
```
**Problem**
Instead of manually replacing every model’s superclass (e.g., Doorkeeper::Application < GlobalRecord), we are looking for a configuration-based solution that allows certain models to automatically connect to the primary database while keeping the rest on sharded connections.

Is there a recommended approach in Rails 7.2 to handle this use case without modifying every model explicitly?

**Expected Behavior**
A way to configure specific models (e.g., Doorkeeper::Application) to always use GlobalRecord or connect to :primary without manually changing every occurrence.
Ideally, this would be a simple Active Record configuration instead of modifying all model definitions.
Any guidance or best practices for handling this in Rails 7.2 would be greatly appreciated.

### LINKED ISSUE #1787: Refresh Token Rotation + Expiry questions
I have a couple of questions regarding refresh tokens:

1) **Refresh Token Rotation** 
understand Doorkeeper only revokes a previously-used refresh token **if** the generated access token is used once. Why is that? Shouldn't two requests to refresh an access token with the same refresh token mean the refresh token should be immediately revoked?

FYI I am using doorkeeper-mongodb as well, not sure if this is a bug in that repository. I thought that once a refresh token is used to generate an access token it should immediately be revoked. 

2) **Refresh Token Expiry**
I notice that refresh tokens don't have any expiry on them, why not?

55728:
>[!TIP]
>Happy to drop any `Closes` if you'd prefer to keep specific issues open.

This PR closes #984, closes #1291, closes #1554, closes #1663, closes #1756, closes #1759, closes #1764, and closes #1787.

---

Several open issues are questions that have been answered — by design decisions, RFC requirements, or behavior that already exists — but where nothing pins the answer down. This PR adds minimal regression specs for six of them so the answers stay true, and closes three more whose resolution needs no new spec.

Test-only change — no behavior is modified.

## Pinned with specs

**#984 — Digest client authentication** (`spec/lib/client_authentication_spec.rb`)
The client authentication registry ships exactly `client_secret_basic`, `client_secret_post` and `none`. The IANA "OAuth Token Endpoint Authentication Methods" registry defines no digest-based method, so Doorkeeper provides none; deployments needing another scheme can register one via `Doorkeeper::ClientAuthentication.register`.

**#1554 — state parameter length** (`spec/requests/flows/authorization_code_spec.rb`)
The authorization response echoes `state` back byte-for-byte however long it is, as RFC 6749 §4.1.2 requires. An oversized `Location` header is a reverse-proxy buffer concern (e.g. nginx `large_client_header_buffers`), not something the authorization server may fix by altering the response.

**#1663 — `reuse_access_token` does not apply to the refresh grant** (already closed; spec added here) (`spec/requests/flows/refresh_token_spec.rb`)
Refreshing always issues a new access token, even with `reuse_access_token` enabled. Reusing a matching live token would return the soon-to-expire token the client is refreshing away from (making the grant a no-op), and a matching token from another rotation chain — another device's session — would hand this client that chain's refresh token, so revoking one session would collaterally break the other.

**#1759 — introspection vs. revoke-on-use** (`spec/requests/endpoints/introspection_spec.rb`)
Both sides of the asymmetry: introspecting a rotated access token does not revoke the previous refresh token (RFC 7662 introspection reports state, it is not "use"), while authenticating a protected resource request with that token does. If introspection finalized rotation, a resource server would destroy the lost-response retry grace merely by asking.

**#1787 — refresh token rotation and expiry** (`spec/requests/flows/refresh_token_spec.rb`)
With `refresh_token_revoked_on_use?`, the same refresh token is accepted again while the rotated access token is unused — the deliberate grace period that lets a client retry a refresh whose response was lost. And refresh tokens have no expiry of their own: `expires_in` bounds only the access token. A configurable refresh-token TTL would need refresh tokens to have their own lifecycle and belongs to the redesign discussion in #1719/#1730/#1771.

## Closed without new specs

**#1291 — IndieAuth support**
Superseded: Client ID Metadata Documents (#1875, implementation in #1897) cover the URL-client model IndieAuth was asking for, and IndieAuth itself plans to adopt CIMD. Remaining IndieAuth-specific pieces belong in an extension gem.

**#1756 — multi-database setups (Makara)**
Supported via the configurable model classes (`access_token_class`, `access_grant_class`, `application_class`): point them at your own models and put the connection handling there. The options are already pinned by `spec/lib/config_spec.rb`.

**#1764 — different base classes for sharded models**
Same answer as #1756 — the concrete model classes are configurable, so each can inherit from whatever base its shard needs.

### DIFF
--- a/spec/lib/client_authentication_spec.rb
+++ b/spec/lib/client_authentication_spec.rb
@@ -62,4 +62,22 @@ RSpec.describe Doorkeeper::ClientAuthentication do
       expect(described_class.get(nil)).to be_nil
     end
   end
+
+  # Regression specs for https://github.com/doorkeeper-gem/doorkeeper/issues/984
+  #
+  # The IANA "OAuth Token Endpoint Authentication Methods" registry defines no
+  # HTTP Digest based method, so Doorkeeper ships without one. A deployment
+  # that needs another scheme registers it through ClientAuthentication.register.
+  describe "built-in methods" do
+    it "enables only the RFC 6749 secret methods and none by default" do
+      expect(described_class::DEFAULT_METHODS).to eq(%i[client_secret_basic client_secret_post none])
+      expect(Doorkeeper::ClientAuthentication::Registry.registered_methods.keys)
+        .to include(*described_class::DEFAULT_METHODS)
+    end
+
+    it "does not include a digest authentication method" do
+      expect(described_class.get(:digest)).to be_nil
+      expect(described_class.get(:client_secret_digest)).to be_nil
+    end
+  end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ User-visible changes worth mentioning.
 - [#1899] Document the client authentication methods registry (`Doorkeeper::ClientAuthentication.register`) in the README with a walkthrough for registering a custom method, and pin it with an end-to-end request spec. Docs/test-only change, closes [#1894].
 - [#1891] Fix: an authorization request carrying a `redirect_uri` for a client registered without one (`redirect_uri` is `nil` under `allow_blank_redirect_uri`) now answers `invalid_redirect_uri` instead of crashing with an unhandled 500.
 - [#1896] Add an opt-in `private_key_jwt` client authentication method (RFC 7523 / OIDC Core §9, requires the `jwt` gem >= 2.7) that verifies assertions against the client's published public keys — `jwks` / `jwks_uri` attributes you define on your Application model, the latter fetched with an SSRF-hardened HTTP client. The jti replay guard and the fetched-JWKS cache are process-local by default and can be replaced with shared stores via the `private_key_jwt_replay_guard` / `private_key_jwt_jwks_cache` config options. Part of [#1875].
+- [#1901] [test] Pin the answered behaviors behind [#984], [#1554], [#1600], [#1663], [#1759] and [#1787] with regression specs — the built-in client authentication methods, the unmodified echo of long `state` values, DB persistence with custom token generators, refresh token rotation/expiry semantics and the introspection asymmetry. Test-only change, closes those issues along with [#1291], [#1756] and [#1764].
 - Please add here
 
 ## 6.0.0.beta1
--- a/spec/requests/endpoints/introspection_spec.rb
+++ b/spec/requests/endpoints/introspection_spec.rb
@@ -75,6 +75,58 @@ RSpec.describe "Introspection endpoint" do
     end
   end
 
+  # Regression specs for https://github.com/doorkeeper-gem/doorkeeper/issues/1759
+  #
+  # With refresh_token_revoked_on_use? the previous refresh token is revoked
+  # when the rotated access token is first *used* — i.e. when it authenticates
+  # a protected resource request (Doorkeeper::OAuth::Token.authenticate).
+  # Introspection (RFC 7662) reports token state; it is not token use, so it
+  # must not finalize the rotation. Both sides of the asymmetry are pinned.
+  context "when the introspected token still references a previous refresh token" do
+    before do
+      Doorkeeper.configure do
+        orm DOORKEEPER_ORM
+        use_refresh_token
+      end
+    end
+
+    let!(:old_token) do
+      FactoryBot.create(:access_token, application: client, use_refresh_token: true)
+    end
+
+    let!(:rotated_token) do
+      FactoryBot.create(
+        :access_token,
+        application: client,
+        use_refresh_token: true,
+        previous_refresh_token: old_token.refresh_token,
+      )
+    end
+
+    it "does not revoke the previous refresh token" do
+      post introspection_endpoint_url,
+           params: {
+             client_id: client.uid,
+             client_secret: client.secret,
+             token: rotated_token.token,
+           }
+
+      expect(response).to be_successful
+      expect(json_response).to include("active" => true)
+      expect(old_token.reload).not_to be_revoked
+      expect(rotated_token.reload.previous_refresh_token).to eq(old_token.refresh_token)
+    end
+
+    it "revokes the previous refresh token when the token authenticates a protected resource request" do
+      get "/full_protected_resources",
+          headers: { "HTTP_AUTHORIZATION" => "Bearer #{rotated_token.token}" }
+
+      expect(response).to be_successful
+      expect(old_token.reload).to be_revoked
+      expect(rotated_token.reload.previous_refresh_token).to be_blank
+    end
+  end
+
   it "rejects the request when it uses more than one client authentication method (RFC 6749 §2.3)" do
     post introspection_endpoint_url,
          params: {
--- a/spec/requests/flows/authorization_code_spec.rb
+++ b/spec/requests/flows/authorization_code_spec.rb
@@ -148,6 +148,20 @@ feature "Authorization Code Flow" do
     url_should_not_have_param("code_challenge_method")
   end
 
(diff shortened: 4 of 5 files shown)


spec/lib/client_authentication_spec.rb
```diff
@@ -62,4 +62,21 @@
...
+  # Regression specs for https://github.com/doorkeeper-gem/doorkeeper/issues/984
+  #
+  # The IANA "OAuth Token Endpoint Authentication Methods" registry defines no
+  # HTTP Digest based method, so Doorkeeper ships without one. A deployment
+  # that needs another scheme registers it through ClientAuthentication.register.
+  describe "built-in methods" do
+    it "registers exactly the default methods" do
+      expect(Doorkeeper::ClientAuthentication::Registry.registered_methods.keys)
+        .to match_array(described_class::DEFAULT_METHODS)
+    end
```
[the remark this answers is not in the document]
55728: Good catch — the registry does ship `:private_key_jwt` as a registered (but not default-enabled) method since #1896. Reworked the example to pin the `DEFAULT_METHODS` list itself and that each default is registered, instead of an exact match over the registry. Fixed in ebabf9c8.

spec/models/doorkeeper/access_token_spec.rb
```diff
@@ -138,6 +138,41 @@ def self.generate(opts = {})
       expect(token.token).to match(/custom_generator_token_\d+/)
     end
 
+    # Regression spec for https://github.com/doorkeeper-gem/doorkeeper/issues/1600
+    #
+    # Access tokens are DB-backed by design: a custom generator (e.g.
+    # Doorkeeper::JWT) only customizes the token value, it does not make
+    # Doorkeeper stateless. The generated value is persisted and is what
+    # subsequent lookups resolve.
+    it "persists the custom generated value as a database record" do
```
nbulaj: Well actually for this issue I even had a PoC with a stateless token entity which allows to bypass DB operations (I just never opened a PR 😄 ). So theoretically it should be possible (maybe with restrictions, IDK). I'd say we don't need to add a test for it (looks like generator already covered, right?). And maybe not to close the issue itself

55728: Makes sense — I'll drop the spec and the `Closes #1600`. Good to know there's a stateless PoC out there 👀✨

nbulaj (approved):
The rest looks good :+1:  Thanks!

55728:
Thanks for the review @nbulaj! 😊 Removed the spec and the `Closes #1600`! 👀✨

ThisIsMissEm:
This all looks good to me.

55728:
Thanks @ThisIsMissEm 😊

PR #1902, opened 2026-08-04
Unify the default-scope calculation across the authorization and token endpoints, closes #1889

### LINKED ISSUE #1889: Default-scope calculation diverges between the authorization and token endpoints when dynamic scopes are enabled
### Context

Splitting this out of #1576 (as suggested by the discussion there — see https://github.com/doorkeeper-gem/doorkeeper/issues/1576#issuecomment-2824959754): there are two independent implementations of the default-scope calculation used when a request omits the `scope` parameter, and they disagree:

- `Doorkeeper::OAuth::PreAuthorization#build_scopes` (authorization endpoint):

  ```ruby
  server.default_scopes.allowed(client_scopes)
```

- `Doorkeeper::OAuth::BaseRequest#build_scopes` (token endpoint — `password`, `client_credentials`):

  ```ruby
  client_scopes.allowed(default_scopes)
  ```

`Scopes#allowed` is not symmetric: only the receiver's scopes act as dynamic scope patterns (`user:*`), the argument is treated as concrete values. So the two implementations put the wildcard on opposite sides.

### Actual behavior (measured on `main`, 6.0.0.beta1)

Without dynamic scopes both compute a plain intersection and only the scope *order* differs (`"profile read"` vs `"read profile"`) — cosmetic.

With `enable_dynamic_scopes`, the same server + application configuration yields three different outcomes depending on the endpoint:

| Configuration | Authorization endpoint (`PreAuthorization`) | Token endpoint (`BaseRequest`) |
| :--- | :--- | :--- |
| `default_scopes :"user:*"`, app scopes `user:1` | grants `user:1` | grants **empty scopes** |
| `default_scopes :"user:1"`, app scopes `user:*` | **`invalid_scope` error** | grants `user:1` |

Repro script (run against the dummy app / rails console):

```ruby
Doorkeeper.configure do
  enable_dynamic_scopes
  default_scopes :"user:*"
end
app = Doorkeeper::Application.create!(name: "t", redirect_uri: "https://app.test/cb", scopes: "user:1")

pre_auth = Doorkeeper::OAuth::PreAuthorization.new(
  Doorkeeper.config,
  { client_id: app.uid, response_type: "code", redirect_uri: app.redirect_uri },
)
pre_auth.authorizable?
pre_auth.scopes.to_s
# => "user:1"

Doorkeeper::OAuth::ClientCredentialsRequest.new(Doorkeeper.config, app, {}).scopes.to_s
# => ""
```

Swap the two scope strings (`default_scopes :"user:1"`, app scopes `user:*`) and the results invert: the pre-authorization fails validation with `invalid_scope` while the client-credentials request grants `user:1`.

### Expected behavior

A given server + application configuration should compute the same default scope set regardless of which endpoint performs the calculation.

### Discussion

Both current outcomes are "fail-safe" (the divergence grants fewer scopes or errors, never more), so this is a consistency bug rather than a privilege issue — but unifying the two requires deciding which side of `allowed` should hold the patterns:

- `client_scopes.allowed(default_scopes)` treats the application's registered scopes as the allowed patterns, matching how `ScopeChecker` validates an explicitly requested scope against `app_scopes`.
- `default_scopes.allowed(client_scopes)` treats the server defaults as the patterns, which supports configurations where the wildcard lives in `default_scopes`.

Neither direction handles both configurations above; a unified implementation may want to accept a scope when it matches in either direction. Happy to send a PR once there's maintainer guidance on the intended semantics.

55728:
When a request omits the `scope` parameter, the default scopes are computed by two independent implementations that disagree once dynamic scopes are enabled:

- `PreAuthorization#build_scopes` (authorization endpoint) uses `default_scopes.allowed(client_scopes)`
- `BaseRequest#build_scopes` (token endpoint) uses `client_scopes.allowed(default_scopes)`

`Scopes#allowed` is not symmetric — only the *receiver* acts as dynamic scope patterns — so the same server + application configuration grants different scopes depending on which endpoint the client happens to hit:

| `default_scopes` | application scopes | authorization endpoint | token endpoint |
|---|---|---|---|
| `user:*` | `user:1` | grants `user:1` | grants an **empty** scope set |
| `user:1` | `user:*` | fails with `invalid_scope` | grants `user:1` |

## The fix

This PR introduces `Scopes#common`: a symmetric operation that matches a dynamic scope pattern on *either* side against the concrete scope on the other and returns the concrete scope. Both `build_scopes` implementations now delegate to it, so the two configurations above both grant `user:1` at both endpoints.

Neither existing direction could serve as the canonical one — each handles only one row of the table, so picking one would regress configurations that currently work through the other endpoint. The symmetric operation is the only option that keeps both rows working, which is why the PR proposes it rather than a coin-flip between the two `allowed` orders.

**This does not grow the granted authority.** Every scope `common` grants is one the client could already obtain by requesting it explicitly: a concrete `scope=user:1` request is validated by `ScopeChecker` against the application's registered patterns, which accepts exactly these combinations today. The change only makes the *implicit* default consistent with the *explicit* request path.

Without dynamic scopes, `common` degenerates to the plain intersection both endpoints already computed, in each endpoint's historical scope order — the existing flow specs pass unmodified.

## Out of scope

With **disjoint** scope sets (no overlap at all), the endpoints also diverge — the token endpoint issues a token with an empty scope set while the authorization endpoint answers `invalid_scope`. That asymmetry predates dynamic scopes and is left as is here; happy to look at it separately if there's interest.

Closes #1889.


lib/doorkeeper/oauth/scopes.rb
```diff
@@ -100,6 +100,21 @@ def allowed(other)
...
+      # Returns the scopes present in both sets, taking dynamic scopes
+      # into account: a dynamic scope pattern (e.g. `user:*`) on either
+      # side matches the concrete scope (e.g. `user:1`) on the other,
+      # and the concrete scope is what ends up in the result. Unlike
+      # #allowed, the same scopes are returned regardless of which set is
+      # the receiver; only the order differs (the argument's order first).
+      #
+      # @param other The set of scopes to intersect with
+      def common(other)
+        other = self.class.from_array(to_array(other))
```
nbulaj, line 114: doesn't it the same as `&` (intersection) ? 🤔  BTW maybe we had to fix `allowed` implementation ?

55728: Not quite — today `&` delegates to `allowed` when dynamic scopes are on, so it inherits the same asymmetry:

```ruby
Scopes.from_string("user:*") & Scopes.from_string("user:1")  # => ["user:1"]
Scopes.from_string("user:1") & Scopes.from_string("user:*")  # => []
```

`common` restores the symmetric intersection `&` had before e2b8dabb.

As for fixing `allowed` directly — it's intentionally directional (same direction as `exists?` / `ScopeChecker`). Making it symmetric would let a wildcard on the filtered side match concrete scopes on the allowed side, which is a live path in doorkeeper-openid_connect's dynamic client registration. The `build_scopes` call sites are safe because the argument is `default_scopes` (server config, not client input).

Happy to fold this into `&` instead of adding a method if you'd prefer 😊 — the tradeoff is that `&` is also public API, so dynamic-scope users using it as a filter would pick up the widening too.

lib/doorkeeper/oauth/scopes.rb
```diff
@@ -100,6 +100,21 @@ def allowed(other)
         self.class.from_array(filtered_scopes)
       end
 
+      # Returns the scopes present in both sets, taking dynamic scopes
+      # into account: a dynamic scope pattern (e.g. `user:*`) on either
+      # side matches the concrete scope (e.g. `user:1`) on the other,
+      # and the concrete scope is what ends up in the result. Unlike
+      # #allowed, the same scopes are returned regardless of which set is
+      # the receiver; only the order differs (the argument's order first).
+      #
```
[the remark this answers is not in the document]
55728: Good catch, the docstring was wrong. When the *argument* holds the pattern, the concrete scope is contributed by the receiver instead, so the result is not in the argument's order:

```
receiver="user:* public"  argument="user:1 public"  => ["user:1", "public"]
receiver="user:1 public"  argument="user:* public"  => ["public", "user:1"]
```

Reworded to describe what actually happens — scopes matched from the argument first, then any scope only the receiver contributes.

spec/lib/oauth/scopes_spec.rb
```diff
@@ -252,6 +266,23 @@
             expect(scopes.all).to eq(%w[public])
           end
         end
+
+        describe "#common" do
+          it "allows user:1 when the receiver holds the pattern" do
+            scopes = described_class.from_string("public user:*").common(described_class.from_string("public user:1"))
+            expect(scopes.all).to eq(%w[public user:1])
+          end
```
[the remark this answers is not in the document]
55728: `common` has no delimiter-dependent logic of its own: it goes through `exists?`, which is where `dynamic_scope_match?` and the configured delimiter live. The existing custom-delimiter context covers `has_scopes?` for exactly that reason, and neither `#&` nor `#allowed` carries a duplicate of it. Adding one only under `#common` would test `exists?` a third time rather than anything new about this method.

lib/doorkeeper/oauth/scopes.rb
nbulaj: And we didn't break what was assumed to be fixed in #1754 ? :eyes: I'm just trying not to break scopes behavior, especially for the projects which somehow can rely on `scopes.rb` (and use these modified API methods)

55728: Fair concern — I checked it from both ends 👀✨

**No existing method in `scopes.rb` is modified.** The diff is 17 insertions, 0 deletions: `#&`, `#allowed`, `#exists?` and the deprecation note are byte-for-byte what #1754 left behind. `#common` is purely additive.

**#1754's specs pass unmodified** (including dynamic scopes and `#&` / `#allowed` blocks):

```
$ bundle exec rspec spec/lib/oauth/refresh_token_request_spec.rb spec/lib/oauth/scopes_spec.rb \
                    spec/lib/oauth/base_request_spec.rb spec/lib/oauth/pre_authorization_spec.rb
149 examples, 0 failures
```

**And structurally it can't regress #1754.** `common` is `allowed` plus the mirror direction, so `a.common(b)` is always a superset of `a.allowed(b)` — it can add a scope, never drop one. #1754 was about scopes being *dropped*; since `common` never drops anything `allowed` returned, that class of bug can't come back through this change.

nbulaj: ok now I'm calm :smiley:  Thanks a lot! :bow:

nbulaj (approved):
LGTM as well ✅

PR #1903, opened 2026-08-05
Stop revoking client_credentials tokens across resource audiences

55728:
## Problem

`revoke_previous_client_credentials_token` keeps one access token per client, but until now it did so regardless of the audience each token was issued for: requesting a token for one resource revoked the client's live token for another.

A client talking to two resource servers could not hold both tokens at once, which makes the option hard to combine with resource indicators. RFC 9700 §4.10.2 notes that a client accessing several resource servers needs a separate token per server.

## Solution

Treat tokens bound to different audiences as different tokens: move the resource match into the shared `find_active_existing_token_for` lookup, where both the reuse path and the revocation path pick it up.

- **Same resource** → previous token is still revoked (unchanged)
- **Different resource** → previous token is left alone (new)
- **No resource indicators** → behaviour is unchanged


nbulaj (approved):
Makes sense, missed that 👀  Thanks

Issue #1768, opened 2025-04-09
Support OAuth RFC -  JWT for Client Authentication and Authorization Grants


hagaivcita:
Are there any plans to support JSON Web Token (JWT) Profile for OAuth 2.0 Client Authentication and Authorization Grants?
[RFC 7523](https://datatracker.ietf.org/doc/html/rfc7523)

From the Abstract:
> Abstract
>
> This specification defines the use of a JSON Web Token (JWT) Bearer
   Token as a means for requesting an OAuth 2.0 access token as well as
   for client authentication.

ThisIsMissEm:
The "correct" place for this would be in doorkeeper-openid_connect, but it doesn't yet support them: https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/v1.8.11/app/controllers/doorkeeper/openid_connect/discovery_controller.rb

ThisIsMissEm:
Oh, actually, that's the client assertions spec, `private_key_jwt` is a token endpoint auth method, which is what I was thinking of, which allows a public client to use client credentials via JWTs https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication

hagaivcita:
> Oh, actually, that's the client assertions spec, `private_key_jwt` is a token endpoint auth method, which is what I was thinking of, which allows a public client to use client credentials via JWTs https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication

Yeah, that's in the token endpoint. `private_key_jwt` is a client authentication method defined in [RFC 7523](https://datatracker.ietf.org/doc/html/rfc7523), instead of authenticating with the `client_secret`.

ThisIsMissEm:
We'd need support somewhere for OAuth Assertion Framework [RFC7521](https://datatracker.ietf.org/doc/html/rfc7521) for this to be possible — I don't think this spec is currently implemented in any maintained doorkeeper code or plugins. There was this & its forks, but they haven't been maintained in a decade: https://github.com/kiorux/doorkeeper-jwt_assertion/

ThisIsMissEm:
This would be enabled by #1772, which allows registering additional client authentication methods.

avner-vcita:
@ThisIsMissEm Hello! Has there been any recent progress regarding this issue?

ThisIsMissEm:
I think #1772 is still awaiting review, but the person to review is busy with life things at the moment, so we all just need to be patient

[...]

PR #1905, opened 2026-08-07
Tests Only: Cover private_key_jwt on the client_credentials grant

55728:
#1768 asked for [RFC 7523](https://www.rfc-editor.org/rfc/rfc7523), "JSON Web Token (JWT) Profile for OAuth 2.0 Client Authentication **and** Authorization Grants". The RFC defines two independent things, and both are reachable on Doorkeeper today, by different routes. This PR pins that with end-to-end specs.

No library code changes. Full suite green, mutation-checked.

## Where each half stands

**§2.2, client authentication.** Built in as the `private_key_jwt` method (#1896), opt-in through `client_authentication`.

**§2.1, authorization grant.** Not built in — under §2.1 the assertion's issuer speaks *for* the subject, so which issuers may do that is a trust decision only the host application can make. `Doorkeeper::GrantFlow` is the documented mechanism for this, and registering `urn:ietf:params:oauth:grant-type:jwt-bearer` through it works today.

## What this adds

One new spec (`spec/requests/flows/jwt_profile_rfc7523_spec.rb`), three groups:

1. **§2.2 via `client_credentials`** — assertion-only auth, no client secret; bad key → `invalid_client`.
2. **§2.1 via `GrantFlow`** — jwt-bearer grant issues a token for the assertion's subject; untrusted issuer → `invalid_grant`; unregistered → `unsupported_grant_type`.
3. **Both composing** — `private_key_jwt` authentication + jwt-bearer assertion in one request.

> [!TIP]
> Happy to split the issue (close §2.2 here, open a focused issue for a built-in §2.1 grant) if you'd prefer to keep that door open. Also happy to trim the §2.1 example down to "the grant type can be registered" if you'd rather not carry a worked example.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ User-visible changes worth mentioning.
 - [#1901] [test] Pin the answered behaviors behind [#984], [#1554], [#1600], [#1663], [#1759] and [#1787] with regression specs — the built-in client authentication methods, the unmodified echo of long `state` values, DB persistence with custom token generators, refresh token rotation/expiry semantics and the introspection asymmetry. Test-only change, closes those issues along with [#1291], [#1756] and [#1764].
 - [#1902] Fix: requests that omit `scope` now compute the same default scopes at the authorization and token endpoints (`Scopes#common`, symmetric). With dynamic scopes enabled, a scope pattern in either `default_scopes` or the application's scopes grants the matching concrete scope at both endpoints, closes [#1889].
 - [#1903] Fix: `revoke_previous_client_credentials_token` no longer revokes a client's live access token issued for a different `resource` ([#1886]), so a client can keep one audience-restricted token per resource server.
+- [#1905] [test] Cover `private_key_jwt` client authentication on the `client_credentials` grant, the one flow where an assertion is the client's only credential end to end. Test-only change.
 - Please add here
 
 ## 6.0.0.beta1
--- a/spec/requests/flows/private_key_jwt_spec.rb
+++ b/spec/requests/flows/private_key_jwt_spec.rb
@@ -77,6 +77,41 @@ feature "private_key_jwt client authentication" do
     expect(Doorkeeper::AccessToken.first.application.uid).to eq(@client.uid)
   end
 
+  # RFC 7523 §2.2 is what private_key_jwt implements, and client_credentials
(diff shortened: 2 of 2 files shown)


nbulaj:
hey @55728 , thanks for this one. I see the spec but it looks to me that reaching requested (in the original ticket) behavior today requires some non-trivial coding, right? And it looks to me like related to a https://github.com/doorkeeper-gem/doorkeeper-id_jag_grant which already adds RFC 7523 :eyes:

55728:
Both fair 😊

The §2.1 half was ~130 lines of host-app code, so it showed the extension point works rather than that Doorkeeper supports the grant — I oversold that.

On `doorkeeper-id_jag_grant`: it adds a `jwt-bearer` grant for the ID-JAG profile rather than plain RFC 7523 §2.1 — it pins `typ: "oauth-id-jag+jwt"` (rejecting a generic JWT as type confusion) and requires `client_id`, `jti` and `iat` beyond RFC 7523 §3's `iss`/`sub`/`aud`/`exp`, so a plain RFC 7523 assertion doesn't verify there.

Trimmed the PR to the part that stands alone: `private_key_jwt` had no `client_credentials` coverage — the one built-in grant where the assertion is the client's only credential end to end. Two scenarios folded into the existing flow spec; the jwt-bearer example and the RFC 7523 framing are gone, and it no longer claims to close #1768.

For the issue I'd answer rather than close by PR: §2.2 built in via #1896, §2.1 via a custom `GrantFlow` or `doorkeeper-id_jag_grant` for the ID-JAG profile. Or I can drop this PR entirely — let me know which you'd prefer!

nbulaj:
> For the issue I'd answer rather than close by PR: §2.2 built in via https://github.com/doorkeeper-gem/doorkeeper/pull/1896

That sounds like a better option to me :+1: It will be helpful for anyone else who will be looking into it 

I'd prefer to have specs covering not GitHub issues, but code & some basic/specific cases, following RFC and so on. Balance between what we _wanna_ to test and what we _actually have_ to test. If this PR covers such - I'm happy to have it!

55728:
It does 👀✨ — the trimmed version is just the `client_credentials` coverage gap for `private_key_jwt`. Thanks! 😊

nbulaj (approved):
Thanks, LGTM ✅  🙇

PR #1910, opened 2026-08-13
Add opt-in `validate_client_before_resource_owner_authentication` to validate the client before resource owner authentication

### LINKED ISSUE #1711: Doorkeeper appears to be missing a way to validate client configuration before redirecting to the authentication page
### Steps to reproduce

We are using a DIY implementation but it basically boils down to the following configuration.

```
Doorkeeper.configure do
    current_user || redirect_to(sign_in_url)
  end
end
```

When there are no valid application clients go to the `/oauth/authorize` endpoint with an invalid `client_id`.

http://localhost:3000/oauth/authorize?client_id=invalid-client-id

### Expected behavior

I was expecting to get an error that the `client_id` was not configured properly.  I double checked the spec and it does not say if it should prevent authorization, it just says that it should not be redirected back to the invalid redirection uri.  From a usability standpoint, it feels preferable to prevent this sort of behavior earlier rather than after the redirect.

>   If the request fails due to a missing, invalid, or mismatching redirection URI, or if the client identifier is missing or invalid, the authorization server SHOULD inform the resource owner of the error and MUST NOT automatically redirect the user-agent to the invalid redirection URI.

I did look at a workaround for this from `resource_owner_authenticator` itself, but it was really awkward because `pre_auth` calls this method as well.  I think that in order to get this to behave how I would expect, the `PreAuthorization` would need to be split so that we could validate before we have an identity, then authorize after that the identity is valid for the client.

### Actual behavior

I was redirected to `sign_in_url` and did not get an error until authenticated.

### System Information

**Gemfile.lock**:

<details>
  <summary>Gemfile.lock content</summary>
  
```
    doorkeeper (5.7.0)
      railties (>= 5)
```
</details>

55728:
## Summary

Fixes #1711.

The authorization endpoint runs `authenticate_resource_owner!` before looking at the request, so a request with an unknown `client_id` or an invalid `redirect_uri` sends the user through the whole login flow only to be shown an error that was already determined before they typed a password. RFC 6749 §4.1.2.1 (§4.2.2.1 for the implicit flow) asks for the resource owner to be informed when the `client_id` is missing or invalid, and §3.1.2.4 asks the same for the redirect URI; running these checks first is what lets the endpoint inform the user without a detour through the login form.

This PR adds an opt-in configuration option:

```ruby
Doorkeeper.configure do
  validate_client_before_resource_owner_authentication
end
```

When enabled, the authorization endpoint's `new` action validates the resource-owner-independent client settings (`client_id` present, client registered, `redirect_uri` matching) **before** authenticating the resource owner, and renders the error page without asking the user to log in. Valid requests keep the exact current behavior.

> [!NOTE]
>
> ### Cost when enabled
>
> `OAuth::Client.find` runs twice for a valid `new` request (once pre-authentication, once in the memoized `pre_auth`). Accepted to keep the nil-owner instance strictly out of the request's later lifecycle.

<details>

<summary>Design notes</summary>

- **Opt-in, disabled by default.** Enabling it changes when the configured `resource_owner_authenticator` runs, and lets an unauthenticated party learn whether a `client_id` is registered. Client identifiers are not confidential (RFC 6749 §2.2), but the default stays as-is for backward compatibility. The option could become the default in a future major release if desired.
- **Only the leading owner-independent validations run early.** `PreAuthorization#client_settings_valid?` runs `client_id` / `client` / `redirect_uri` in their declared order, so error precedence matches a full validation run. `authorize_resource_owner_for_client` and everything after it still run post-authentication, keeping the existing precedence (e.g. `access_denied` before `invalid_scope`) intact.
- **The nil-owner `PreAuthorization` is never memoized.** The pre-authentication check builds its own instance and discards it when the request proceeds, so a configured `authorize_resource_owner_for_client` hook keeps seeing the real resource owner (naively reusing the instance would answer `access_denied` to every logged-out user on such apps — this is the bug in the workaround code posted on the issue).
- **Pre-authentication errors are rendered, never redirected**, even under `handle_auth_errors :redirect`: with a missing `client_id` the `redirect_uri` cannot be validated, and RFC 6749 §3.1.2.4 forbids redirecting the user-agent to an invalid redirection URI (`invalid_client` / `invalid_redirect_uri` were already non-redirectable). `handle_auth_errors :raise` keeps raising the same exceptions as before.
- Compatible with doorkeeper-openid_connect: its `authenticate_resource_owner!` override (prepended on `Doorkeeper::Helpers::Controller`) never runs when the filter chain halts here, which only happens for requests that could not produce a redirectable OIDC response anyway.

</details>

### DIFF
--- a/app/controllers/doorkeeper/authorizations_controller.rb
+++ b/app/controllers/doorkeeper/authorizations_controller.rb
@@ -2,6 +2,9 @@
 
 module Doorkeeper
   class AuthorizationsController < Doorkeeper::ApplicationController
+    before_action :validate_client,
+                  only: :new,
+                  if: -> { Doorkeeper.config.validate_client_before_resource_owner_authentication? }
     before_action :authenticate_resource_owner!
 
     def new
@@ -45,11 +48,57 @@ module Doorkeeper
 
       if Doorkeeper.configuration.redirect_on_errors? && pre_auth.error_response.redirectable?
         redirect_or_render(pre_auth.error_response)
-      elsif Doorkeeper.configuration.api_only
-        render json: pre_auth.error_response.body, status: pre_auth.error_response.status
       else
-        render :error, locals: { error_response: pre_auth.error_response }, status: pre_auth.error_response.status
+        render_error_response(pre_auth.error_response)
       end
+    end
+
+    def render_error_response(error_response)
+      if Doorkeeper.configuration.api_only
+        render json: error_response.body, status: error_response.status
+      else
+        render :error, locals: { error_response: error_response }, status: error_response.status
+      end
+    end
+
+    # Refuses the request before resource owner authentication when the
+    # client_id or redirect_uri is missing or invalid. Errors are rendered,
+    # never redirected, regardless of `handle_auth_errors :redirect`: no
+    # failure this check can produce leaves a redirect target worth trusting.
+    # Either the client failed first, so the redirect URI was never reached,
+    # or the redirect URI is itself the invalid one — and RFC 6749
+    # Section 3.1.2.4 forbids redirecting the user-agent to an invalid
+    # redirection URI.
+    #
+    # Host applications that need to refuse a client on their own terms (an
+    # allow-list, a registry lookup) can override this and call +super+ first:
+    #
+    #   class AuthorizationsController < Doorkeeper::AuthorizationsController
+    #     private
+    #
+    #     def validate_client
+    #       super
+    #       return if performed?
+    #       return if MyRegistry.allowed?(client_pre_auth.client.application)
+    #
+    #       head :forbidden
+    #     end
+    #   end
+    def validate_client
+      return if client_pre_auth.client_valid?
+
+      error_response = client_pre_auth.error_response
+      error_response.raise_exception! if Doorkeeper.config.raise_on_errors?
+
+      render_error_response(error_response)
+    end
+
+    # The pre-authentication view of the request. It carries no resource owner
+    # (nobody is authenticated yet), so it is kept apart from #pre_auth:
+    # owner-dependent validations (authorize_resource_owner_for_client) and the
+    # views need the one built with current_resource_owner.
+    def client_pre_auth
+      @client_pre_auth ||= OAuth::PreAuthorization.new(Doorkeeper.configuration, pre_auth_params)
     end
 
     def can_authorize_response?
--- a/lib/doorkeeper/oauth/pre_authorization.rb
+++ b/lib/doorkeeper/oauth/pre_authorization.rb
@@ -5,6 +5,16 @@ module Doorkeeper
     class PreAuthorization
       include Validations
 
+      # The validations that identify the client and its redirect URI. None of
+      # them depend on the resource owner, so the authorization endpoint can
+      # run them before authenticating anyone and spare the user a login that
+      # can only end on an error page. RFC 6749 Section 4.1.2.1 (Section 4.2.2.1
+      # for the implicit flow) asks for the resource owner to be informed when
+      # the client_id is missing or invalid, and Section 3.1.2.4 asks the same
+      # for the redirect URI; running these first is what lets the endpoint
+      # inform them without a detour through the login form.
+      CLIENT_VALIDATIONS = %i[client_id client redirect_uri].freeze
+
       validate :client_id, error: Errors::InvalidRequest
       validate :client, error: Errors::InvalidClient
       validate :redirect_uri, error: Errors::InvalidRedirectUri
@@ -43,6 +53,24 @@ module Doorkeeper
 
       def authorizable?
         valid?
+      end
+
+      # Runs only CLIENT_VALIDATIONS, in declared order, so a request
+      # from an unknown client or with an invalid redirect URI can be refused
+      # without a resource owner. Error precedence matches a full #validate
+      # run because these are the first validations declared.
+      def client_valid?
+        @error = nil
+        @missing_param = nil
+
+        self.class.validations.each do |validation|
+          next unless CLIENT_VALIDATIONS.include?(validation[:attribute])
+
+          @error = validation[:options][:error] unless send("validate_#{validation[:attribute]}")
+          break if @error
+        end
+
+        @error.nil?
       end
 
       def scopes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ User-visible changes worth mentioning.
 - Fix: the `client_secret_basic` strategy now requires a `client_id` sent in the request body to name the same client as the `Authorization: Basic` header — the RFC 7521 §4.2 agreement check `private_key_jwt` already applies to an assertion's issuer. A request presenting Basic credentials for one client and a `client_id` for another was authenticated as the Basic client, silently discarding the other identity. A bare `client_id` is not a client authentication method of its own, so the RFC 6749 §2.3 multiple-methods check does not (and should not) count it.
 - [#1907] Fix: a `resource` parameter no longer produces a 500 at the authorization endpoint when `resource_indicator_validator` is configured without the `doorkeeper:resource_indicators` migration. Such a request is now answered with `server_error`, as the token endpoint already did, and the missing migration is warned about at boot.
 - [#1909] Add Rails 8.1 to CI test matrix.
+- [#1910] Add opt-in `validate_client_before_resource_owner_authentication` configuration option: the authorization endpoint validates `client_id` and `redirect_uri` before authenticating the resource owner, so users are not sent through login for a request that can only fail.
 - [#PR ID] Description of the change.
 
 ## 6.0.0.beta2
--- a/lib/doorkeeper/config.rb
+++ b/lib/doorkeeper/config.rb
@@ -182,6 +182,13 @@ module Doorkeeper
         @config.instance_variable_set(:@force_pkce, true)
       end
 
+      # Validate the authorization request's client_id and redirect_uri before
+      # authenticating the resource owner, so users are not sent through login
+      # for a request that can only fail (disabled by default)
+      def validate_client_before_resource_owner_authentication
+        @config.instance_variable_set(:@validate_client_before_resource_owner_authentication, true)
+      end
+
       # Use an API mode for applications generated with --api argument
       # It will skip applications controller, disable forgery protection
       def api_only
@@ -627,6 +634,10 @@ module Doorkeeper
 
     def force_pkce?
       option_set? :force_pkce
+    end
+
+    def validate_client_before_resource_owner_authentication?
+      option_set? :validate_client_before_resource_owner_authentication
     end
 
     def enforce_configured_scopes?
--- a/lib/generators/doorkeeper/templates/initializer.rb
+++ b/lib/generators/doorkeeper/templates/initializer.rb
@@ -202,6 +202,18 @@ Doorkeeper.configure do
   #
   # force_pkce
 
+  # Validate the authorization request's client_id and redirect_uri before
+  # authenticating the resource owner, so users are not sent through login for
+  # a request that can only fail (RFC 6749 Section 4.1.2.1 asks for the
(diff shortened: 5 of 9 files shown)


55728:
This is a false positive — `validation[:attribute]` is drawn from `CLIENT_SETTINGS_VALIDATIONS = %i[client_id client redirect_uri].freeze`, not from user input. The `send("validate_#{validation[:attribute]}")` pattern is the same one `Doorkeeper::Validations#validate` already uses (validations.rb L14).

nbulaj (commented):
Hello. It looks to me like something which is easily achievable by overriding/extending the authorizations controller with the same custom `before_action` hook which gonna validate whatever is required. Why do we need to have this specific case in the core? 👀

55728:
Hi @nbulaj, thanks for taking a look! 😊

You're right that a `before_action` can work — the reporter of #1711 did exactly that ([this comment](https://github.com/doorkeeper-gem/doorkeeper/issues/1711#issuecomment-2239543075)). The catch is that doing it correctly from outside turns out to be tricky: the workaround in the issue actually has a subtle bug — it runs `authorizable?` with `resource_owner = nil`, which fires `authorize_resource_owner_for_client` against `nil` and rejects valid requests with `access_denied` if the app configures that hook. It only works for the reporter because they use the default `->(_c, _o) { true }`.

The fix requires splitting the validation chain into client-settings-only vs. resource-owner-dependent parts — which is what `PreAuthorization#client_settings_valid?` in this PR does. A host-side copy of that split would silently drift whenever validations change, and the private APIs it would depend on (`pre_auth`, `pre_auth_params`, `render_error`) have no compatibility guarantees.

That said, if the controller wiring feels too specific for core, I'd be happy to slim this down to just the `client_settings_valid?` primitive so host apps can build their own hook on a supported API. Let me know which direction you prefer 👀✨

app/controllers/doorkeeper/authorizations_controller.rb
```diff
@@ -2,6 +2,9 @@
 
 module Doorkeeper
   class AuthorizationsController < Doorkeeper::ApplicationController
+    before_action :validate_client_settings!,
+                  only: :new,
```
nbulaj, line 6: Can it be on a class level, smth like

```ruby
if Doorkeeper.config.validate_client_before_resource_owner_authentication?
  before_action :validate_client_settings!,
                  only: :new,
end
```

To avoid unnecessary additional items in the before_actions chain?

app/controllers/doorkeeper/authorizations_controller.rb
```diff
@@ -45,13 +48,37 @@ def render_error
...
 
+    # Refuses the request before resource owner authentication when the
+    # client_id or redirect_uri is missing or invalid. The PreAuthorization
+    # built here carries no resource owner, so it must not be memoized as
+    # @pre_auth: owner-dependent validations (authorize_resource_owner_for_client)
+    # and the views need the one built with current_resource_owner. Errors are
+    # rendered, never redirected, regardless of `handle_auth_errors :redirect` —
+    # the redirect URI is not validated at this point, and RFC 6749
+    # Section 3.1.2.4 forbids redirecting to an invalid redirection URI.
+    def validate_client_settings!
```
nbulaj: Hm, I won't call it "client setting", I\'d better name it just `validate_client` , WDYT?

lib/doorkeeper/oauth/pre_authorization.rb
```diff
@@ -45,6 +55,24 @@ def authorizable?
         valid?
       end
 
+      # Runs only CLIENT_SETTINGS_VALIDATIONS, in declared order, so a request
+      # from an unknown client or with an invalid redirect URI can be refused
+      # without a resource owner. Error precedence matches a full #validate
+      # run because these are the first validations declared.
+      def client_settings_valid?
```
nbulaj: Same here and for the constant above, just `client_valid?`

nbulaj (commented):
OK, I see. I just have a feeling that we will find somebody who gonna need to add a custom logic around it ("_validate in my custom Redis registry that client is in the allow-list_") and they still need to extend the controller. So maybe we should add this option proactively? Like make `validate_client_before_resource_owner_authentication` to accept a block and pass `pre_auth` object there? Just not sure how to track custom error then 🤔  I don't want to make/add any hacky additionals to support this feature.

lib/doorkeeper/oauth/pre_authorization.rb
55728: Agreed — renamed to `client_valid?` and `CLIENT_VALIDATIONS`.

app/controllers/doorkeeper/authorizations_controller.rb
55728: Done — `validate_client` (dropped the bang too). No collision: `PreAuthorization#validate_client` is a different object, and nothing in `Helpers::Controller` or the other Doorkeeper controllers uses that name.

app/controllers/doorkeeper/authorizations_controller.rb
55728: I tried it — I'd rather keep the runtime `if:`, for two reasons.

**It turns the option into a boot-time-only setting, and the specs stop being able to see the feature.** The class body is evaluated once when the controller is autoloaded, so `Doorkeeper.config` has to be final by then. With the snippet applied, 5 of this PR's examples fail — and 2 others keep passing vacuously: the request still ends in `invalid_client`, just after the login detour, so a green suite no longer distinguishes "refused before authentication" from "refused after authentication".

**The codebase already gates `before_action`s at request time:**

```ruby
# applications_controller.rb:8
before_action :force_json_format, if: :api_only?

# application_metal_controller.rb:8
before_action :enforce_content_type,
              if: -> { Doorkeeper.config.enforce_content_type }
```

The cost is one predicate call, only on `GET /oauth/authorize`.

Happy to switch if you'd rather it be explicitly boot-time-only — just say so and I'll adapt the specs.

55728:
I share the worry about a second callback interface. I think subclassing already covers the Redis-allow-list case, as long as the method is named and the object is reachable:

```ruby
class AuthorizationsController < Doorkeeper::AuthorizationsController
  private

  def validate_client
    super
    return if performed?
    return if MyRedisRegistry.allowed?(client_pre_auth.client.application)

    head :forbidden
  end
end
```

Two things I changed to make that work:

**`client_pre_auth`** — a memoized private reader for the owner-less `PreAuthorization`. Without it, an override reaching for `pre_auth` triggers `current_resource_owner` → the authenticator runs right there, defeating the point. (With a redirect-style authenticator, `pre_auth.resource_owner` memoizes as `302`.)

**`return if performed?`** after `super` — when `super` has already rendered, continuing raises `DoubleRenderError`.

This also answers the "not sure how to track custom error" question: the override renders whatever it wants using the host app's own error vocabulary, so Doorkeeper doesn't need to carry a custom error out of a block.

I added a controller spec pinning the seam. If you'd still prefer the block form, it can be layered on later without changing anything here — but I'd rather not ship both.

app/controllers/doorkeeper/authorizations_controller.rb
nbulaj: Boot time is OK actually (I mean we don't need to change the code to satisfy tests :) ), but agree to follow the format we already use in the code base 👍

nbulaj:
I'm currently thinking if we should make it a default behavior. I checked the RFC , [4.1.2.1](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1).  Error Response

> If the request fails due to a missing, invalid, or mismatching
>    redirection URI, or if the client identifier is missing or invalid,
>    the authorization server SHOULD inform the resource owner of the
>    error and MUST NOT automatically redirect the user-agent to the
>    invalid redirection URI.
> 
>    If the resource owner denies the access request or if the request
>    fails for reasons other than a missing or invalid redirection URI,
>    the authorization server informs the client by adding the following
>    parameters to the query component of the redirection URI using the
>    "application/x-www-form-urlencoded" format, per [Appendix B](https://datatracker.ietf.org/doc/html/rfc6749#appendix-B):

But it's fully about the case when **we have** a `resource owner` already. I wasn't able to find direct instructions around that specific case covered by this PR.

BTW what is the behavior of the major providers like Google, GitHub, Microsoft? Do they validate client first without touching resource owner?

55728:
> what is the behavior of the major providers like Google, GitHub, Microsoft? Do they validate client first without touching resource owner?

Good questions 👀✨ — I measured the providers rather than guessing. Short version: **only one of the three validates the client before touching the resource owner**, so I'd keep this opt-in.

### What the providers do (logged out, unknown `client_id`)

| Provider | Result |
| --- | --- |
| **Google** | `302` → `accounts.google.com/signin/oauth/error?authError=…`, which decodes to `invalid_client` / "The OAuth client was not found." **No login.** |
| **GitHub** | `302` → `github.com/login?client_id=…&return_to=%2Flogin%2Foauth%2Fauthorize…` — **login first**, error only afterwards. |
| **Microsoft Entra ID** | The sign-in page (`sErrorCode: 50058`, "a silent sign-in request was sent but no user is signed in"), on `/common` and on a tenant-specific endpoint alike. `AADSTS700016` shows up only after sign-in. **Login first.** |

For what it's worth, GitLab — which uses Doorkeeper — `302`s to `/users/sign_in`, i.e. exactly today's behavior.

Reproducible in any logged-out browser:

```
https://accounts.google.com/o/oauth2/v2/auth?client_id=nonexistent.apps.googleusercontent.com&redirect_uri=https%3A%2F%2Fexample.com%2Fcb&response_type=code&scope=openid
https://github.com/login/oauth/authorize?client_id=nonexistent&redirect_uri=https%3A%2F%2Fexample.com%2Fcb&response_type=code
```

### On the RFC

You're right, and I leaned on it harder than it supports. §4.1.2.1 says *what* to do ("inform the resource owner"), not *when* — and Doorkeeper does inform them, just after the login detour. §3.1 arguably cuts the other way:

> The authorization server MUST first verify the identity of the resource owner.

I read that as "before issuing a grant" rather than "before parsing the request", but it is certainly not an argument *for* this PR.

The nearest normative precedent is OIDC Core, which separates the two steps and orders them explicitly: §3.1.2.2 *Authentication Request Validation* ("The Authorization Server MUST validate the request received as follows"), then §3.1.2.3 *Authorization Server Authenticates End-User* ("**If the request is valid**, the Authorization Server attempts to Authenticate the End-User…"). That is relevant here because doorkeeper-openid_connect prepends onto this very controller. Still, I'd rather present this PR as what it is — a UX fix with a spec precedent, not a compliance fix — and I'll adjust the description accordingly.

### Default or opt-in

I'd keep it opt-in here, given that two of the three references behave like Doorkeeper does today. Three things I'd want resolved before flipping the default:

1. **Cost paid by everyone.** `OAuth::Client.find` runs twice per valid `GET /oauth/authorize` (the note in the description). Fine for people who opt in, less so as a default.
2. **Unauthenticated oracle.** With it on, anyone can distinguish a registered `client_id` (`invalid_client`) from a registered client with a wrong redirect URI (`invalid_redirect_uri`) without an account. Client identifiers are not confidential (§2.2) and redirect URIs generally are not either, but on a private or internal AS that is a change of posture that deserves to be a deliberate choice.
3. **`validate_client` goes live.** A host app that already defines that method on a subclass of `Doorkeeper::AuthorizationsController` would silently have it promoted to a `before_action`. Harmless while the option is off.

If you want it on by default eventually, say so and I'll change the option shape now so the flip is a one-liner and hosts keep an escape hatch:

```ruby
option :validate_client_before_resource_owner_authentication, default: false
```

(the same shape as `force_ssl_in_redirect_uri`) instead of the current `force_pkce`-style enable-only builder method, which has no way to express "off". Then `default: true` in a future major is the entire change.

app/controllers/doorkeeper/authorizations_controller.rb
55728: Ha, fair — I motivated it backwards. The consistency argument was the real one; I should have led with that 😅

nbulaj (approved):
Agree, let's go with an option. We can always make it default if required

55728:
Thank you @nbulaj! 😊

PR #1912, opened 2026-08-17
[ci skip] Add AI skills to cover Doorkeeper rules & RFC compliance

nbulaj:
You know

image

### DIFF
--- a/.agents/skills/antipattern-prevention/SKILL.md
+++ b/.agents/skills/antipattern-prevention/SKILL.md
@@ -1,0 +1,204 @@
+---
+name: antipattern-prevention
+description: Avoid common Ruby and Rails antipatterns that degrade maintainability and performance. Use when writing new code, reviewing PRs, or refactoring existing code in Doorkeeper.
+---
+
+# Antipattern Prevention
+
+When writing or reviewing code in Doorkeeper, use this skill to avoid common antipatterns that degrade maintainability and performance.
+
+## 1. Ruby Iteration Instead of SQL
+
+**Severity:** High
+
+```ruby
+# BAD — loads all tokens into memory
+AccessToken.all.select { |t| t.expired? }
+
+# GOOD — push to database
+# The codebase uses expiration_time_sql for this:
+# lib/doorkeeper/models/concerns/expiration_time_sql_math.rb
+```
+
+```ruby
+# BAD
+ids = AccessToken.pluck(:id).select { |id| id > 100 }
+
+# GOOD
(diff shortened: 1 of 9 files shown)


.agents/skills/rfc-compliance/SKILL.md
```diff
@@ -0,0 +1,149 @@
...
+- `code_verifier` is 43-128 characters from `[A-Z] / [a-z] / [0-9] / "-" / "." / "_" / "~"`
+
+## Token Introspection (RFC 7662)
+
+- MUST require authentication of the requesting party
+- Response for inactive/invalid tokens: `{"active": false}` — no other fields
+- Response for active tokens includes: `active`, `scope`, `client_id`, `token_type`, `exp`, `iat`, `sub`, `aud`, `iss`
+- Refresh tokens MUST NOT include `token_type` or `exp` in introspection response
+
+**Reference:** `lib/doorkeeper/oauth/token_introspection.rb`
```
55728, line 80: Doorkeeper does exclude `token_type`/`exp` for refresh tokens, so the *behavior* described here is accurate — but RFC 7662 §2.2 lists all introspection response members as OPTIONAL, not MUST NOT. Might be worth softening to something like "Doorkeeper omits…" so an agent doesn't cite it as an RFC requirement elsewhere.

.agents/skills/antipattern-prevention/SKILL.md
```diff
@@ -0,0 +1,204 @@
...
+rescue
+  nil
+end
+
+# GOOD — use the existing HttpFetcher which handles timeouts and errors
+# Reference: lib/doorkeeper/http_fetcher.rb
+def fetch_jwks(uri)
+  response = http_fetcher.fetch(uri)
+  JSON.parse(response)
+rescue HttpFetcher::Error => e
```
55728: The "GOOD" example references `HttpFetcher::Error`, but the actual exception class is `HttpFetcher::FetchError` (`lib/doorkeeper/http_fetcher.rb:88`) — an agent following this as-is would hit a `NameError` 🙈

```suggestion
rescue HttpFetcher::FetchError => e
```

nbulaj: Nice catch, fixed! :+1:  Thanks

.agents/skills/rfc-compliance/SKILL.md
nbulaj: Agree, rephrased, thanks

55728 (approved):
Thank you so much! 😊

PR #1917, opened 2026-08-22
Make the fallback secret upgrade conditional on the value it matched

55728:
Split out of #1904 as offered [there](https://github.com/doorkeeper-gem/doorkeeper/pull/1904#issuecomment-5288370039) — this commit fixes a pre-existing race on `main` and is independent of the secret rotation feature that PR adds.

## The problem

When a fallback secret strategy is configured, `SecretStorable.find_by_fallback_token` re-stores a credential in the current format as soon as the fallback strategy matches it:

```ruby
def upgrade_fallback_value(instance, attr, plain_secret)
  upgraded = secret_strategy.store_secret(instance, attr, plain_secret)
  instance.update(attr => upgraded)
end
```

The upgrade is a write on what is otherwise a read path (finding a record by its secret), and it writes **by id alone**, from a value that was read before the comparison that decided to upgrade it. Nothing stops the row moving between the match and the write, and when it does, the matched value is written back over whatever replaced it.

Concretely, this races `#renew_secret`: an authentication that started before the renewal can finish after it and put the superseded secret back. The credential the application believes it has replaced keeps authenticating the client, and the one it just handed out is gone. The write has had this shape since fallback secrets were introduced in a00bcd1a (v5.1.0), so every release since is affected.

The window is narrow — it needs a fallback strategy configured, the column still holding a value in the fallback format, and a concurrent write in that moment — and it closes for good the first time an upgrade succeeds, since the stored value is no longer in the fallback format. It is not something an outside party can drive.

### Reproducing it

No threads needed — a stale instance is enough, because the read that decides the write happened when the instance was loaded:

```ruby
in_flight = Doorkeeper::Application.find(app.id)   # loaded with the old secret

app.renew_secret
app.save!
new_secret = app.plaintext_secret

in_flight.secret_matches?(plain_secret)            # => true, and it writes

app.reload
app.secret_matches?(new_secret)   # => false  (on main)
app.secret_matches?(plain_secret) # => true   (on main)
```

## The fix

Write only while the column still holds the value that matched:

```ruby
scope = where(id: instance.id, attr => matched)
```

Losing the race now means no upgrade rather than a stale write. The credential stays in the fallback format and is upgraded by whichever authentication comes next — which is the state it was already in a moment earlier, so nothing is lost by skipping it.

## What changes along with it

- **`#update` → a conditional `#update_all`.** `update_all` does not maintain timestamps, so `updated_at` is set explicitly where the model has it (applications; access tokens and grants have no such column). Validations no longer gate the upgrade, which I took as the right trade — the storage format of a secret is not something an unrelated invalid attribute should be able to hold back — and the write no longer carries along whatever else was pending on the instance.
- **ORM extensions.** This concern is included by the ORM extensions too, so the conditional write goes through a small `update_matched_rows` helper: `update_all` on an Active Record relation, `update` on a Sequel dataset. Both answer how many rows were written, which is the value the fix needs.
- **Dirty tracking.** Since the write goes past the instance, the columns it wrote are marked clean afterwards (`clear_attribute_changes`, guarded by `respond_to?` for ORMs without dirty tracking), and the in-memory assignment `store_secret` had already made is undone when the write finds the row moved. A caller that goes on to lock or reload the record is then not refused over a change that is either already persisted or was never made. Deliberately not a `reload`: this is nominally a read path, and a `reload` would discard every unrelated unsaved change the caller was holding.
- **Return value.** `upgrade_fallback_value` now documents a boolean — whether the stored value was upgraded. It previously returned whatever `#update` did, so this is the same shape.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ User-visible changes worth mentioning.
 - [#1909] Add Rails 8.1 to CI test matrix.
 - [#1910] Add opt-in `validate_client_before_resource_owner_authentication` configuration option: the authorization endpoint validates `client_id` and `redirect_uri` before authenticating the resource owner, so users are not sent through login for a request that can only fail.
 - [#1916] Fix broken Coveralls coverage reporting.
+- [#1917] Fix: the fallback secret upgrade no longer writes the matched secret back over a value stored in the meantime, which could undo a concurrent `#renew_secret` and leave the superseded secret valid. The upgrade is now conditional on the column still holding the value that matched.
 - [#1918] The api_only controller specs no longer `load` the real controller sources, which detached the coverage of every other example that ran them and made the reported coverage depend on the random example order.
 - [#PR ID] Description of the change.
 
--- a/lib/doorkeeper/models/concerns/secret_storable.rb
+++ b/lib/doorkeeper/models/concerns/secret_storable.rb
@@ -82,18 +82,153 @@ module Doorkeeper
         # @param plain_secret
         #   The plain secret to upgrade.
         #
+        # @return [Boolean]
+        #   Whether the stored value was upgraded. False means the row no
+        #   longer holds the value that was matched.
+        #
         def upgrade_fallback_value(instance, attr, plain_secret)
+          # The value the fallback strategy matched against. The upgrade is a
+          # write on what is otherwise a read path, so the row can move between
+          # the match and this write — another request renewing the secret, or
+          # the application replacing it. Writing by id alone would put the
+          # matched value back over whatever replaced it, so the write is
+          # conditional on the column still holding it.
+          matched = instance.public_send(attr)
           upgraded = secret_strategy.store_secret(instance, attr, plain_secret)
 
-          # The upgrade is a write on what is otherwise a read path (finding a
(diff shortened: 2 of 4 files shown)


55728:
Both findings are false positives — the `GitlabSecurity/PublicSend` rule flagging `public_send` regardless of where the argument comes from. `attr` is always a literal (`:token`, `:refresh_token`, `:secret`), never user-supplied, and both sends already exist on `main`; the scan only reports lines the diff touches.

nbulaj (commented):
Before you add more changes - I don't think we have to support all the ORMs here. I can suggest to move some general logic into separate concern/mixin methods so any ORM can implement them. 

Currently implementation looks complex, I need time to understand if it can be simplified 👀

55728:
You were right on both counts — it can be simplified, and along exactly the line you suggested. I've opened #1923 with that restructure, and left this PR untouched so the two shapes can be compared.

Happy to close this one in favour of #1923 if you prefer that direction.

55728:
Closing in favour of #1923, which restructures this along the lines @nbulaj suggested — the shared concern stays slim and each ORM extension owns its conditional write.

PR #1918, opened 2026-08-22
Keep the api_only controller specs from resetting the controllers' coverage

55728:
Coveralls has started reporting per-job coverage today (#1915/#1916 follow-ups), and #1917 immediately got "Coverage decreased (-0.02%)" on a random subset of jobs — 2 of 20 on one push, 7 of 20 on the next, with the set of jobs changing each time — for a change that touches nothing near the lines in question.

## What is going on

The lines that flip are in `app/controllers/doorkeeper/applications_controller.rb`, and the cause is `spec/requests/applications/api_only_spec.rb` (mine, #1887): it re-evaluates the real controller sources into stubbed `ActionController::API` subclasses with `load`, in a `before` that runs for every example.

Ruby records coverage per file name and attaches the counters to the code *as compiled*. Each `load` compiles the file again and points the file's entry in the coverage result at a fresh array. The original controllers — the ones every other example runs — keep counting into the array the report no longer reads. So after this spec has run, nothing the original controllers do is reported any more, and the controllers' coverage depends on which examples happened to run after it. Locally, three seeds on `main` give three different sets of uncovered lines in that file; the total moves between 98.38% and 98.41%.

## The fix

Evaluate the sources under a file name of their own (`eval(File.read(path), TOPLEVEL_BINDING, "#{path} (api_only)")`) instead of `load`ing them. The stubbed constants are reopened exactly as before; the eval'd copy is not part of the report (eval'd code is not covered by default), and the original controllers keep their counters.

After the change, the suite's line coverage is identical on every seed tried (1, 3, 7), and about a percentage point higher than the best order used to report — 99.39% vs 98.41% on `main` — since the controllers' coverage is now counted in full.

No behaviour change in the specs themselves: same stubs, same examples, all green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,12 +7,13 @@ User-visible changes worth mentioning.
 
 ## main
 
-- [#1906] Internal: exempt `Doorkeeper::Config` from `Metrics/ClassLength` with a directive on the class itself instead of raising the cop's global ceiling, so adding a configuration option no longer trips the limit.
 - Fix: the `client_secret_basic` strategy now requires a `client_id` sent in the request body to name the same client as the `Authorization: Basic` header — the RFC 7521 §4.2 agreement check `private_key_jwt` already applies to an assertion's issuer. A request presenting Basic credentials for one client and a `client_id` for another was authenticated as the Basic client, silently discarding the other identity. A bare `client_id` is not a client authentication method of its own, so the RFC 6749 §2.3 multiple-methods check does not (and should not) count it.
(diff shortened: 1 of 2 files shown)


nbulaj:
do we still need it after [upgrade of simplecov](https://github.com/doorkeeper-gem/doorkeeper/pull/1919)? I see coverage is increased per each job for ~0.2%

image

55728:
Yes, still needed 👀✨ — the reset happens in Ruby's `Coverage` module, below simplecov, so no simplecov version can fix it. The same report was closed as a Ruby limitation just recently: simplecov-ruby/simplecov#389 ("`Coverage.result` replaces per-line hits when a file is `load`ed again, and SimpleCov consumes that snapshot as-is at exit time").

nbulaj (approved):
Makes sense 👍

PR #1920, opened 2026-08-24
Require Ruby >= 3.2 in the gemspec, matching the CI matrix

55728:
Since 37034516 updated the CI matrix to 3.2 / 3.3 / 3.4 / 4.0 (dropping 3.1 as EOL), the gemspec still has `required_ruby_version = ">= 2.7"`. This brings it in line with the matrix so the two stay in sync.

Ruby 2.7, 3.0 and 3.1 have all reached end-of-life, and 6.0.0 hasn't shipped yet, so a major version feels like a natural moment to tighten the floor. 3.2 is also the minimum Rails 8.0 requires.

>[!NOTE]
>
> ### Relation to #1837
> 
> I opened #1837 for the same line back in June and closed it myself once ruby-i18n/i18n#739 resolved the 3.1 breakage that motivated it. This time the reasoning is simpler — just aligning the gemspec with where the CI matrix already is after 37034516.
> 
> Happy to close this if you'd rather keep the wider bound — just wanted to flag the gap 😊

### Changes

- `doorkeeper.gemspec`: `required_ruby_version` `>= 2.7` → `>= 3.2`
- `CHANGELOG.md`: entry under `## main`

No CI or source changes — the matrix already reflects the new floor.

### Verification

```
$ bundle exec rake spec
1937 examples, 0 failures
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ User-visible changes worth mentioning.
 
 ## main
 
+- Require Ruby >= 3.2 in the gemspec, matching the CI matrix (3.2 / 3.3 / 3.4 / 4.0). Ruby 2.7, 3.0 and 3.1 have reached end-of-life.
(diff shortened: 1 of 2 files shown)


nbulaj:
I always didn't want to restrict Ruby version to support legacy applications which still didn't migrate to a newer Ruby/Rails... So we guarantee compatibility with a up-to-date Ruby/Rails from the CI matrix, but use with older versions as well but at one's own discretion. But with the 6.0.0 release I think we can move ahead and bump this restriction since all the Rails >=7.2  also require Ruby 3.x

nbulaj (approved):

55728:
Thank you so much @nbulaj! 😊

PR #1923, opened 2026-08-30
Fix the fallback secret upgrade race through an ORM write hook

55728:
### Summary

Supersedes #1917, restructured per the review feedback there: the shared concern no longer tries to absorb every ORM's way of writing conditionally.

The bug is unchanged from #1917. `upgrade_fallback_value` writes the upgraded secret by id alone, from a value read before the comparison that decided to upgrade it. The write sits on what is otherwise a read path, so the row can move in between — most notably racing `#renew_secret`: an authentication that started before the renewal can finish after it and put the superseded secret back, leaving a credential the application believes it has replaced still able to authenticate the client, and the one it just handed out gone. Pre-existing since fallback secrets were introduced in a00bcd1a (v5.1.0).

### Structure

- `Doorkeeper::Models::SecretStorable` keeps only what every ORM shares: read the matched value before `store_secret` assigns the upgraded one over it, hand both to a single hook — `write_upgraded_secret(instance, attr, matched, upgraded)` — and put the instance back when the hook answers that nothing was written. The default implementation of the hook is the historical unconditional `#update`, so the ORM extensions (doorkeeper-mongodb, doorkeeper-sequel) behave exactly as they do on every released version until they implement a conditional write in their own terms.
- `Doorkeeper::Orm::ActiveRecord::Mixins::SecretStorable` (new) implements the conditional write for the three Active Record models, relying on what Active Record guarantees instead of feature-sniffing for it: `update_all` answering a row count, the configured `primary_key` (scalar, composite, or absent), `record_timestamps`, optimistic locking, `clear_attribute_changes`, `with_primary_role`.

### Behaviour (Active Record)

- The upgrade writes only while the column still holds the value that matched. Losing the race means no upgrade rather than a stale write: the credential stays in the fallback format and is upgraded by whichever authentication comes next.
- Row identity comes from every column the configured primary key names; a model with no primary key is left to the value condition alone.
- `update_all` does not maintain timestamps, so `updated_at` is stamped explicitly, deferring to `record_timestamps` the way `#update` did.
- Under optimistic locking the version joins the condition and is bumped explicitly, and the instance is brought in step afterwards, so a subsequent save is not refused over a change that is already persisted.
- The lookup still returns the resource either way — the fallback value did match at read time — but a lost race leaves the instance holding the value that is actually stored, with unrelated unsaved changes intact (no `reload`).

Three deliberate behaviour notes: validations no longer gate the upgrade on this path (the storage format of a secret should not be held back by an unrelated invalid attribute), the write no longer carries along other pending changes on the instance (both same as #1917), and model callbacks do not run on the upgrade write — it is a single conditional statement, and what it writes is the value the lookup already matched, re-encoded into the current storage format.

### DIFF
--- a/lib/doorkeeper/models/concerns/secret_storable.rb
+++ b/lib/doorkeeper/models/concerns/secret_storable.rb
@@ -70,8 +70,9 @@ module Doorkeeper
           end
         end
 
-        # Allow implementations in ORMs to replace a plain
-        # value falling back to to avoid it remaining as plain text.
+        # Replaces a value found through the fallback strategy with its
+        # upgraded form, so it does not remain stored under the old strategy
+        # (e.g. as plain text).
         #
         # @param instance
         #   An instance of this model with a plain value token.
@@ -82,13 +83,54 @@ module Doorkeeper
         # @param plain_secret
         #   The plain secret to upgrade.
         #
+        # @return [Boolean]
+        #   Whether the stored value was upgraded.
+        #
         def upgrade_fallback_value(instance, attr, plain_secret)
+          # The value the fallback strategy matched against, read before
+          # `store_secret` assigns the upgraded one over it.
+          matched = instance.public_send(attr)
           upgraded = secret_strategy.store_secret(instance, attr, plain_secret)
 
-          # The upgrade is a write on what is otherwise a read path (finding a
-          # record by its secret), so it must reach the primary database when
-          # automatic role switching would route the surrounding request to a
-          # read replica.
+          return true if write_upgraded_secret(instance, attr, matched, upgraded)
+
+          # Nothing was written: put the instance back rather than leave it
+          # carrying a secret that was never stored.
+          instance.public_send(:"#{attr}=", matched)
+          false
+        end
+
+        # ORM hook: writes the upgraded value to storage and answers whether
+        # it was written.
+        #
+        # The upgrade is a write on what is otherwise a read path (finding a
+        # record by its secret), so the row can move between the match and
+        # this write — another request renewing the secret, or the
+        # application replacing it. An ORM can therefore make the write
+        # conditional on +attr+ still holding +matched+, answering false when
+        # it no longer does, as the Active Record implementation does
+        # (Doorkeeper::Orm::ActiveRecord::Mixins::SecretStorable). This
+        # default keeps the historical unconditional write for ORMs that have
+        # not implemented a conditional one.
+        #
+        # @param instance
+        #   The instance whose row to write, already carrying +upgraded+.
+        #
+        # @param attr
+        #   The secret attribute name being upgraded.
+        #
+        # @param matched
+        #   The stored value the fallback lookup matched.
+        #
+        # @param upgraded
+        #   The upgraded value to store.
+        #
+        # @return [Boolean]
+        #   Whether the value was written.
+        #
+        def write_upgraded_secret(instance, attr, _matched, upgraded)
+          # The write must reach the primary database when automatic role
+          # switching would route the surrounding request to a read replica.
           if respond_to?(:with_primary_role)
             with_primary_role { instance.update(attr => upgraded) }
           else
--- a/lib/doorkeeper/models/concerns/secret_storable.rb
+++ b/lib/doorkeeper/models/concerns/secret_storable.rb
@@ -96,8 +96,29 @@ module Doorkeeper
 
           # Nothing was written: put the instance back rather than leave it
           # carrying a secret that was never stored.
-          instance.public_send(:"#{attr}=", matched)
+          restore_matched_secret(instance, attr, matched)
           false
+        end
+
+        # ORM hook: puts +matched+ back on +instance+ after nothing was
+        # written. This default assigns through the attribute writer, which
+        # an ORM can bypass the way the Active Record implementation does:
+        # a custom writer can transform +matched+ on the way in, leaving the
+        # attribute dirty with a value the row never held, and a later save
+        # would write that over whatever replaced the matched secret — the
+        # stale write the conditional upgrade exists to prevent.
+        #
+        # @param instance
+        #   The instance to restore, still carrying the unwritten upgrade.
+        #
+        # @param attr
+        #   The secret attribute name being restored.
+        #
+        # @param matched
+        #   The stored value the fallback lookup matched.
+        #
+        def restore_matched_secret(instance, attr, matched)
+          instance.public_send(:"#{attr}=", matched)
         end
 
         # ORM hook: writes the upgraded value to storage and answers whether
@@ -123,7 +144,10 @@ module Doorkeeper
         #   The stored value the fallback lookup matched.
         #
         # @param upgraded
-        #   The upgraded value to store.
+        #   The upgraded value to store, as returned by the strategy. The
+        #   instance carries it too, assigned through the attribute writer —
+        #   an implementation that writes past the writer should persist the
+        #   value the writer left on the instance rather than this one.
         #
         # @return [Boolean]
         #   Whether the value was written.
--- a/lib/doorkeeper/orm/active_record/mixins/secret_storable.rb
+++ b/lib/doorkeeper/orm/active_record/mixins/secret_storable.rb
@@ -1,0 +1,113 @@
+# frozen_string_literal: true
+
+module Doorkeeper::Orm::ActiveRecord::Mixins
+  # Active Record implementation of the SecretStorable write hook: the
+  # fallback secret upgrade only writes while the column still holds the
+  # value the lookup matched, so losing a race against a concurrent write —
+  # another request renewing the secret, or the application replacing it —
+  # cannot put the superseded secret back.
+  #
+  # The write is a single conditional statement, so unlike the historical
+  # `#update` it runs no model callbacks: what it writes is the value the
+  # lookup already matched, re-encoded into the current storage format.
+  module SecretStorable
+    extend ActiveSupport::Concern
+
+    class_methods do
+      # Writes the upgraded secret over +matched+ on +instance+'s row,
+      # conditional on +attr+ still holding +matched+, and answers whether
+      # the row was written.
+      def write_upgraded_secret(instance, attr, matched, _upgraded)
+        # The value written is the one `store_secret` left on the instance:
+        # it assigns through the attribute writer, which `#update` then ran
+        # again on assignment, so a custom writer override reached storage.
+        # `update_all` does not go through writers, so writing the strategy's
+        # pre-writer return value would both skip the override and leave the
+        # instance out of step with its row.
+        upgraded = instance.read_attribute(attr)
+
+        # `update_all` does not maintain timestamps, which `#update` did —
+        # see `upgrade_timestamps` for the terms it did that on.
+        changes = { attr => upgraded }.merge!(upgrade_timestamps(instance))
+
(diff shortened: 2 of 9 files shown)


lib/doorkeeper/orm/active_record/mixins/secret_storable.rb
```diff
@@ -0,0 +1,79 @@
...
+    class_methods do
+      # Writes +upgraded+ over +matched+ on +instance+'s row, conditional on
+      # +attr+ still holding +matched+, and answers whether the row was
+      # written.
+      def write_upgraded_secret(instance, attr, matched, upgraded)
+        changes = { attr => upgraded }
+        # `update_all` does not maintain timestamps, which `#update` did —
+        # including its deference to `record_timestamps`, which a model that
+        # stamps its own timestamps turns off.
+        changes[:updated_at] = Time.now.utc if stamps_updated_at?(instance)
```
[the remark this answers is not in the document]
55728: Good catch — all three gaps were real, and they are now closed by deriving the timestamps through Active Record's own helpers instead of hand-rolling them:

- The stamped changes come from `touch_attributes_with_time`, so the columns are the model's actual update timestamp attributes (`updated_on` and aliased names as well as `updated_at`) and the time respects the connection's configured timezone. The helper is public on every Rails version the gemspec supports (7.0 through 8.1) and is the same one `Relation#touch_all` pairs with `update_all`.
- `record_timestamps` is now read on the instance, so a per-record `record.record_timestamps = false` is honored the way `#update` honored it.
- A timestamp the caller had already changed is left out of the write entirely, matching `#update`'s `will_save_change_to_attribute?` skip — so `sync_upgraded_instance` can no longer overwrite it and mark the caller's pending change clean. The instance sync now covers every timestamp column actually written, not just a literal `updated_at`.

Covered by three new specs: an instance with timestamps turned off, a pending `updated_at` change surviving the upgrade (still dirty, row untouched), and an `updated_on`-only model being stamped and brought in step.

lib/doorkeeper/orm/active_record/mixins/secret_storable.rb
```diff
@@ -0,0 +1,90 @@
...
+      # The columns identifying +instance+'s row and the values it holds in
+      # them, taken from the configured primary key: it may be named
+      # something other than `id`, or be composed of several columns. A
+      # model with no primary key at all is left to the value condition
+      # alone, which reaches every row still holding the matched secret.
+      def primary_key_conditions(instance)
+        # Spelled without `index_with`, which the gemspec's Rails floor
+        # predates.
+        Array(primary_key).to_h { |key| [key, instance.public_send(key)] } # rubocop:disable Rails/IndexWith
+      end
```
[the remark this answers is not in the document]
55728: Not taking this one. Three reasons:

The hook has exactly two call sites, and both run on an instance that was just materialized from a query: `find_by_fallback_token` calls it inside the `tap` on the `find_by` result, and `Application#secret_matches?` calls it right after comparing the stored value on a record the authentication flow just loaded. There is no point in either path where a pending primary-key change can exist on the instance, so the in-memory values *are* the database values and `attribute_in_database` would read the same thing.

The row-identity condition is also a narrowing, not the guard. The guard is `attr => matched` — that is why a model with no primary key at all is already left to the value condition alone (see the comment on `primary_key_conditions`). For a `query_constraints :tenant_id, :id` model with a non-unique `id`, a cross-tenant row would have to hold the same `id` *and* the same stored secret to be touched, and even then the write would re-encode that row's own matched value into the current storage format — functionally a no-op for it. And if the identity condition ever did miss the intended row, the failure mode is the safe one: the scope matches nothing, the hook answers false, and the credential is upgraded by the next authentication instead.

Finally, the query-constraints reflection is Rails 7.1+, above even the oldest CI leg (`gemfiles/rails_7_0.gemfile`). Reaching it would need exactly the `respond_to?` feature-sniffing this restructure was asked to remove (see the review on #1917), for a scenario the call sites cannot produce.

lib/doorkeeper/orm/active_record/mixins/secret_storable.rb
```diff
@@ -0,0 +1,90 @@
...
+        # longer holds would set it backwards.
+        if locking_enabled?
+          version = instance.public_send(locking_column) || 0
+          scope = scope.where(locking_column => version)
+          changes[locking_column] = version + 1
+        end
+
+        # The write must reach the primary database when automatic role
+        # switching would route the surrounding request to a read replica.
+        written = with_primary_role { scope.update_all(changes) }
```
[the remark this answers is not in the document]
55728: Half taken. The factual half is right: `update_all` runs no model callbacks, where the historical `instance.update` did, and the PR description only listed the validation and pending-attribute differences. I've added it as a third deliberate behaviour note in the description and in the mixin's header comment (e0117653).

Not taking the lock/recheck-and-save shape, though — holding a row lock and re-issuing a callback-bearing `save` on what is otherwise a read path is the structure #1917 had, and this PR exists because the review there asked for it to go away. The conditional single-statement write is the point of the restructure.

On impact: an application auditing secret changes through model callbacks still sees every write that changes what a secret *is* — `renew_secret`, regeneration, revocation all go through the normal persistence path. The only write callbacks now miss is this one-time, in-place re-encoding of a value the lookup already matched, which does not change the secret itself.

lib/doorkeeper/orm/active_record/mixins/secret_storable.rb
```diff
@@ -0,0 +1,94 @@
...
+    extend ActiveSupport::Concern
+
+    class_methods do
+      # Writes +upgraded+ over +matched+ on +instance+'s row, conditional on
+      # +attr+ still holding +matched+, and answers whether the row was
+      # written.
+      def write_upgraded_secret(instance, attr, matched, upgraded)
+        # `update_all` does not maintain timestamps, which `#update` did —
+        # see `upgrade_timestamps` for the terms it did that on.
+        changes = { attr => upgraded }.merge!(upgrade_timestamps(instance))
```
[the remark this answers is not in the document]
55728: Good catch — taken. `store_secret` assigns through the attribute writer and returns the pre-writer transformed value, and the historical `instance.update(attr => upgraded)` ran the writer again on assignment, so a custom writer override always reached storage. Writing the strategy's return value through `update_all` skipped the override *and* left the instance out of step with its row after `clear_attribute_changes`.

Fixed by persisting the value the writer left on the instance (`read_attribute`), which is exactly what `#update` persisted for any writer: both paths store `writer(transformed)`. Added a regression spec with a model that overrides the secret writer, asserting the stored value carries the writer's transformation and the instance stays in step.

lib/doorkeeper/orm/active_record/mixins/secret_storable.rb
```diff
@@ -0,0 +1,102 @@
...
+      end
+
+      private
+
+      # The columns identifying +instance+'s row and the values it holds in
+      # them, taken from the configured primary key: it may be named
+      # something other than `id`, or be composed of several columns. A
+      # model with no primary key at all is left to the value condition
+      # alone, which reaches every row still holding the matched secret.
+      def primary_key_conditions(instance)
```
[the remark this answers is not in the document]
55728: Taken. Added a spec with a `primary_key = nil` model and two rows sharing the matched plain secret: the upgrade rewrites both rows (secret and timestamp) and leaves the looked-up instance clean and in step, pinning the documented value-condition-only behavior.

lib/doorkeeper/models/concerns/secret_storable.rb
```diff
@@ -82,13 +83,57 @@ def find_by_fallback_token(attr, plain_secret)
...
 
-          # The upgrade is a write on what is otherwise a read path (finding a
-          # record by its secret), so it must reach the primary database when
-          # automatic role switching would route the surrounding request to a
-          # read replica.
+          return true if write_upgraded_secret(instance, attr, matched, upgraded)
+
+          # Nothing was written: put the instance back rather than leave it
+          # carrying a secret that was never stored.
+          instance.public_send(:"#{attr}=", matched)
```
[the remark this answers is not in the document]
55728: Taken. The restore path had the same writer hole the write path had: assigning `matched` back through a transforming writer leaves the attribute dirty with a value the row never held, and a later save writes it over whatever replaced the matched secret — the stale write this PR exists to prevent, reintroduced through the failure path.

The restoration is now its own small hook: the shared default keeps the writer assignment (for the default unconditional `#update`, where a failed write means a validation failure, not a lost race), and the Active Record implementation restores the raw attribute (`write_attribute`) and clears only that attribute's change — the attribute then holds exactly what it held when the row was read, so there is nothing left to save. Added the custom-writer lost-race spec, asserting the instance comes back clean and a later save does not touch the concurrently renewed secret; it fails on the previous head.

lib/doorkeeper/orm/active_record/mixins/secret_storable.rb
```diff
@@ -0,0 +1,113 @@
...
+
+module Doorkeeper::Orm::ActiveRecord::Mixins
+  # Active Record implementation of the SecretStorable write hook: the
+  # fallback secret upgrade only writes while the column still holds the
+  # value the lookup matched, so losing a race against a concurrent write —
+  # another request renewing the secret, or the application replacing it —
+  # cannot put the superseded secret back.
+  #
+  # The write is a single conditional statement, so unlike the historical
+  # `#update` it runs no model callbacks: what it writes is the value the
```
nbulaj, line 11: Can we please put a short note in the changelog for that? I don't think anyone rely on callbaks here, but just to be on a safe side. GitLab uses encrypted secrets, hope we don't break anything

nbulaj (approved):
Solid ✅ 

I believe we can close #1917, right?

55728:
Will do — I'll add a changelog note about the upgrade write no longer running model callbacks before merging.

And yes, happy to close #1917 now 👍

Thanks for the review that shaped this direction, @nbulaj 👀✨

lib/doorkeeper/orm/active_record/mixins/secret_storable.rb
55728: Done 😊 — the changelog entry now notes that the Active Record upgrade write is a single `update_all` and no longer runs model callbacks or validations.

I checked GitLab's usage against their `master` — the upgrade path reaches tokens but not application secrets there, and the token models define no callbacks on this column, so this should be a safe change for them 👀✨

nbulaj: Thanks!

nbulaj: I'd also add a note in upgrade guides as well, just in case :eyes:  :bow:

55728: Good idea 👀✨ I'll add a note to the upgrade guides as a follow-up 🚀

55728: Done 😊 — added upgrade notes to both places:

- Wiki
  - [Migration from old versions § Fallback secret upgrade race fix](https://github.com/doorkeeper-gem/doorkeeper/wiki/Migration-from-old-versions#fallback-secret-upgrade-race-fix-1923)
- Guides repo
  - https://github.com/doorkeeper-gem/guides/pull/21

nbulaj: Thanks a lot! :bow:

PR #1924, opened 2026-09-07
Build the private_key_jwt audience only from the server's own identity

55728:
## What

`private_key_jwt` client authentication verifies an assertion's `aud` claim against a list of values that identify this authorization server. Until now that list was built from the request when the server had no identity of its own configured — neither `issuer` nor Rails' `default_url_options[:host]` — so the value an assertion was checked against came from the `Host` header the caller sends.

This PR derives the acceptable audiences only from the server's own configuration. A server that identifies itself nowhere now has no acceptable audience at all and refuses every assertion, and `Doorkeeper::Config` says so at boot rather than leaving it to be discovered one request at a time.

## Why

`aud` is the only claim that ties an assertion to *this* server rather than to another one. RFC 7523 §3 requires the audience to identify the authorization server, and RFC 7523 §3 (and the Security Considerations in §8) treat that check as what stops an assertion issued for one server from being presented to another.

When the audience is derived from a request header, the caller picks the value it will be compared against — so an assertion minted for a *different* authorization server can be presented here and pass the audience check. The assertion still has to verify against the client's published keys, so this affects a client the server already knows; but the audience check exists precisely so that an assertion scoped to another server is not usable here, and it was not doing that job.

Deployments that configure `issuer` or `default_url_options[:host]` — which the initializer already told operators to do for this method — were never affected: the request was only consulted as a fallback.

## Changes

- `Doorkeeper::OAuth::ClientAuthentication::PrivateKeyJwt` builds the acceptable audiences from `issuer` and `default_url_options` only. The private `server_url_options` helper, which held the request fallback, is removed.
- An assertion is refused outright when no audience is acceptable, rather than being handed an empty audience list to verify against.
- New `Doorkeeper::Config` validation: enabling `private_key_jwt` without an `issuer` and without `default_url_options[:host]` logs an error at boot explaining that every assertion will be refused.
- The initializer template's note on audiences is updated to match.

## Breaking change

A deployment that enables `private_key_jwt` while configuring neither `issuer` nor `default_url_options[:host]` authenticated clients against a request-derived audience before, and now authenticates none until it is given an identity. The boot-time error names both options.

Configuring either one restores authentication; no client-side change is required for a client that already sends the issuer identifier or the token endpoint URL as `aud`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ User-visible changes worth mentioning.
 - [#1916] Fix broken Coveralls coverage reporting.
 - [#1918] The api_only controller specs no longer `load` the real controller sources, which detached the coverage of every other example that ran them and made the reported coverage depend on the random example order.
 - [#1923] Fix: the fallback secret upgrade no longer writes the matched secret back over a value stored in the meantime, which could undo a concurrent `#renew_secret` and leave the superseded secret valid. Active Record writes the upgrade conditionally on the column still holding the value that matched; other ORMs can implement the new `write_upgraded_secret` hook. The Active Record write is a single `update_all` statement, so model callbacks and validations no longer run on this upgrade (timestamps and optimistic locking are still maintained).
+- **[BREAKING]** Fix: `private_key_jwt` client authentication no longer accepts an audience derived from the request's `Host` header, which let a client assertion minted for another authorization server be replayed here. A server that configures neither `issuer` nor Rails' `default_url_options[:host]` now has no acceptable audience and refuses every assertion, and is warned about it at boot.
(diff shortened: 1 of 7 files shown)


55728:
Hello @nbulaj 👋

The CI failures are unrelated to this PR — I ran main on my fork and it fails with the same `unknown keyword: quirks_mode` error (a json gem incompatibility in the request specs): https://github.com/55728/doorkeeper/actions/runs/33759168292/job/101716766271

I'll look into it separately 👀✨

55728:
Superseded by #1926, which rebases this on top of #1925 (the `json < 3` pin) so CI can run cleanly 👀✨

PR #1932, opened 2026-09-11
Keep the granted scope on refresh tokens across narrowed refreshes

### LINKED ISSUE #1771: Refresh tokens cannot reduce scopes correctly
In OAuth 2.0, when using refresh tokens via the `refresh_token` grant flow, it is possible to reduce the scope of the new access token, and then to restore the access token back to the originally authorized scopes.

From: https://www.rfc-editor.org/rfc/rfc6749#section-6
> The authorization server MAY issue a new refresh token, in which case the client MUST discard the old refresh token and replace it with the new refresh token.  The authorization server MAY revoke the old refresh token after issuing a new refresh token to the client.  If a new refresh token is issued, the refresh token scope MUST be identical to that of the refresh token included by the client in the request.

What this essentially means is that the refresh token always has the full authorized scopes available, but the access token it generates may have reduced scopes.

So the flow goes as follows:

1. authorization code grant flow is authorized with scopes `read write`
2. authorization code is exchanged for the first access token and first refresh token.
3. first access token has `read write` scopes
4. first refresh token used with `read` (by passing `scope` parameter when refreshing)
5. second access token is issued and should have `read` scope, the second refresh token should still have `read write` scope.
6. second refresh token is used with `read write` (expanding the scope back to originally authorized scopes)
7. third access token should have `read write` scope, since it was originally authorized by the access grant.

You can see a rspec request test for this in: https://github.com/doorkeeper-gem/doorkeeper/compare/main...ThisIsMissEm:doorkeeper:feat/refresh-token-scope-changing

At this time, I think this ability is **not** supported by doorkeeper, and supporting it would mean having to change how all access tokens and refresh tokens persisted. Instead of the refresh token just being a column on `oauth_access_tokens` we either need a table `oauth_refresh_tokens` OR we need to make the `oauth_access_tokens` table support both access tokens and refresh tokens via a `token_type` column, and then some sort of relationship between the two would need to be established (I think)

This is also related to #1719 and the whole concept of "token families" basically the access grant specifies the start of the token family, then each access token and refresh token form a somewhat tree structure.

55728:
Closes #1771

## Problem

[RFC 6749 §6](https://www.rfc-editor.org/rfc/rfc6749#section-6) says two things about scope on the refresh token grant:

> `scope` ... if omitted is treated as equal to the scope originally granted by the resource owner.

> If a new refresh token is issued, the refresh token scope MUST be identical to that of the refresh token included by the client in the request.

Doorkeeper derived both from the scope of the access token the refresh token is stored on. So once a client narrowed a refresh (`scope=read` after being granted `read write`), the refresh token it got back was itself narrowed to `read`: omitting `scope` on the next refresh stayed at `read`, and asking for `read write` again was refused with `invalid_scope`. The chain could never return to the granted scope, which is the scenario @ThisIsMissEm laid out in #1771.

## Change

The granted scope gets its own column, `refresh_token_scopes` on `oauth_access_tokens`, following the `previous_refresh_token` / `resource` pattern (additive column, generator, `column_names` presence check). No separate refresh token table is needed.

- `AccessToken#generate_refresh_token` records the access token's scope as the granted scope when a refresh token is first issued (authorization code, password, or a host app creating the record itself), so every chain starts from what the resource owner granted.
- `RefreshTokenRequest` validates a requested scope against, and defaults an omitted scope to, the presented refresh token's granted scope, and carries that scope unchanged onto the refresh token it issues. The access token still gets the (possibly narrowed) requested scope, as before.
- New `doorkeeper:refresh_token_scopes` generator for existing installs; the install migration includes the column for new ones.

## Backward compatibility

- Without the column, `AccessToken#refresh_token_scopes` reports the access token's scope, so the refresh grant behaves exactly as before. Nothing changes for a host app until it runs the generator.
- Rows that predate the migration have a `NULL` column and get the same fallback until they rotate, at which point the new row records the granted scope.
- Token responses, revocation and the revoke-on-use rotation are untouched. `reuse_access_token` gains one more matching condition: a candidate's refresh token scope must equal the scopes the request grants, so a token narrowed on refresh (access token `read`, refresh token granted `read write`) is not reused for a grant of `read`. Rows without a stored granted scope match on the access token scope as before.
- The ORM extensions (mongodb / sequel) are unaffected until they add the field, since the presence check reads `column_names` the same way `resource_indicators_supported?` does.

## Introspection

Introspecting a refresh token (`token_type_hint=refresh_token`, or a refresh token found by the fallback lookup) now reports the granted scope, per RFC 7662 §2.2 (`scope` describes the presented token). The access token still reports its own, possibly narrowed, scope. Without the column, or for rows that predate it, the access token scope is reported as before.


lib/doorkeeper/models/access_token_mixin.rb
```diff
@@ -429,6 +472,50 @@ def use_refresh_token?
...
+
+      OAuth::Scopes.from_string(stored)
+    end
+
+    # @param value [String, Array, Doorkeeper::OAuth::Scopes, nil]
+    #   scope to store; normalized the same way `scopes=` is. A blank value
+    #   means "no explicit granted scope", so `generate_refresh_token` derives
+    #   it from the access token scope as if nothing had been assigned.
+    def refresh_token_scopes=(value)
+      normalized =
```
nbulaj, line 502: Should we add a `refresh_token_scopes_supported?` guard here as well? For the apps who don't have a column, but try to write it

nbulaj (approved):
Solid 👍

lib/doorkeeper/models/access_token_mixin.rb
55728: Yes, good catch 👀✨ — added.

Without the column the reader already fell back silently, but the writer still went through `super`, so a host app assigning `refresh_token_scopes` before running the migration hit `NoMethodError: super: no superclass method 'refresh_token_scopes='`. The refresh grant itself was guarded at the call site, so only host-app code could reach it.

The setter now returns early when `refresh_token_scopes_supported?` is false, mirroring `refresh_token_scopes_string`. Went with ignoring rather than raising so the column-absent behavior is symmetric; happy to make it a descriptive error instead if you'd prefer the misconfiguration to be loud.

PR #1933, opened 2026-09-11
Warn at boot when a grant flow deprecated by RFC 9700 is enabled

### LINKED ISSUE #1892: Deprecation or removal of insecure grant flows
As of the publication of [RFC 9700](https://www.rfc-editor.org/info/rfc9700/): [BCP 240:](https://www.rfc-editor.org/info/bcp240/) Best Current Practice for OAuth 2.0 Security, several grant flows have noted attacks against them or security issues, leading to their deprecation. This then becomes [complete removal in OAuth 2.1](https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-15.html#name-differences-from-oauth-20)

The grant flows that should be removed or deprecated are:
- resource owner password grant flow
- implicit grant flow

I would lean towards removal, given we are working on 6.x with major breaking changes, I could potentially see the case for supporting them but with heavy deprecation notices (e.g., require a `enable_deprecated_grant_flows` type option that needs to be explicitly opted into to keep them until 7.x.

55728:
RFC 9700 (OAuth 2.0 Security Best Current Practice, BCP 240) deprecates the implicit grant ([§2.1.2](https://www.rfc-editor.org/rfc/rfc9700#section-2.1.2)) and the resource owner password credentials grant ([§2.4](https://www.rfc-editor.org/rfc/rfc9700#section-2.4)), and OAuth 2.1 removes both. Applications that still enable them in `grant_flows` currently get no signal from Doorkeeper.

This PR adds `validate_deprecated_grant_flows` to `Doorkeeper::Config::Validations`. When `implicit` or `password` is enabled (directly, or through an alias that expands to one of them), a `[DOORKEEPER]` warning is logged through `Rails.logger` at configuration time, once per deprecated flow:

```
[DOORKEEPER] The implicit grant flow is deprecated by RFC 9700 (OAuth 2.0 Security Best Current Practice) and removed from OAuth 2.1. It is enabled via grant_flows and may be removed in a future Doorkeeper release.
```

> [!NOTE]
>
> - The warning is emitted at boot (in `validate!`) rather than on first use, matching the existing preference in `Config::Validations` for surfacing configuration issues during boot instead of on the first request (see the comment on `validate_client_authentication_conflict`).
> - No state is kept and no behaviour changes: both flows keep working exactly as before. This is a notice only.
> - Not in scope, left for follow-ups under #1892: an explicit opt-in such as `enable_deprecated_grant_flows`, and the eventual removal. 

Closes #1892


nbulaj (approved):
LGTM ✅  👍

nbulaj:
Thanks!

Issue #1895, opened 2026-08-01
Public Clients and security exposure (OAuth 2.1)


ThisIsMissEm:
In draft -15 of OAuth 2.1, there is a line in [Client Authentication](https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-15.html#name-client-authentication) that says:

> The authorization server MUST consider the security implications of interacting with unauthenticated clients and take measures to limit the potential exposure of tokens issued to such clients, (e.g., limiting the lifetime of refresh tokens).

My interpretation of this is that to be fully compliant with OAuth 2.1, an authorization server MUST NOT issue a long-lived access token to a public client (unauthenticated client), e.g., limiting them to a max of N weeks or something. This would be a stark divergence from Doorkeeper in the wild today, and a significant ecosystem security improvement, since it would act as a forcing factor to get people to upgrade from unauthenticated/public clients to confidential clients. 

(The weird awkward exception to this is Mastodon where confidential clients are effectively used as public clients due to Mastodon not supporting public client registration via its' proprietary dynamic client registration mechanism).

We should perhaps consider for 6.x what exactly this would mean for Doorkeeper, and whether Doorkeeper would try to keep you safe by default, with the acknowledgement that it would be a breaking change, or whether we continue as existing today, where access tokens can be configured to never expire regardless of client authentication method.

Issue #1364, opened 2020-02-26
Wrong behavior in refreshing a token


rishabhsairawat:
The behavior of refreshing a token is not valid.

## Doorkeeper Configuration:
```
# config/initializers/doorkeeper.rb

Doorkeeper.configure do
  # ...
  access_token_expires_in 2.hours

  custom_access_token_expires_in do |context|
     1.hour if context.grant_type == Doorkeeper::OAuth::PASSWORD
  end

  use_refresh_token do |context|
     true if context.grant_type == Doorkeeper::OAuth::PASSWORD
  end
  
# ...
end
```

## Steps to reproduce
1. Create a token using `password` grant_type.
2. The response of Step 1 will be having a token with expiry as **1 hour.**
3. Now create a token using `refresh_token` from the last step.
4. The response of Step 3 will be having a token with expiry as  **2 hours**.

## Expected behavior
Token expiry in both cases should be the same as I haven't defined token expiry for refresh_token grant type explicitly.

## Actual behavior
The expiry of the token received after refreshing the original token isn't equal to the expiry of the original token. It violates the [OAuth2.0 RFC:](https://tools.ietf.org/html/rfc6749#section-1.5)

> Refresh tokens are issued to the client by the authorization server and are
   used to obtain a new access token when the current access token
   becomes invalid or expires, or to obtain additional access tokens
   with identical or narrower scope (access tokens may have a shorter
   lifetime and fewer permissions than authorized by the resource
   owner).  

I am using doorkeeper version 5.0.2. But the same problem exists with the master.

I think by default, the expiry of token received after refreshing should be the same as that of the original token. @nbulaj I can prepare a PR to fix this. But first let me know your thoughts on this.

nbulaj:
I didn't figure out why it violates the OAuth2.0 RFC (just reading the text you posted above), but I definitely agree on that: "_Token expiry in both cases should be the same_"  :+1: 

So it would be great to see a PR, thanks!

nbulaj:
I didn't figure out why it violates the OAuth2.0 RFC (just reading the text you posted above), but I definitely agree on that: "_Token expiry in both cases should be the same_"  :+1: 

So it would be great to see a PR, thanks!

rishabhsairawat:
@nbulaj As RFC mentions that token received after refreshing the original token may have a shorter expiry. But here I am able to get token with more expiry than the original token.

nbulaj:
Let's do it!

PR #1934, opened 2026-09-13
Consult `custom_access_token_expires_in` in the refresh_token grant

55728:
Refs #1895, refs #1366.

### Background

Since #1366 (#1364) the refresh_token grant gives the new access token the TTL of the token being refreshed, without consulting `custom_access_token_expires_in`. That fixed a real problem: a token issued with a grant-specific lifetime (`1.hour if context.grant_type == PASSWORD`) lost it on refresh, because the callable saw `refresh_token` as the grant type, returned `nil`, and the default `access_token_expires_in` took over. Inheriting the original TTL was the right call for that problem, and this PR keeps it.

What inheriting does not cover is letting the host decide the lifetime of refreshed tokens. The callable is never asked, so a policy such as "tokens obtained through refresh live for 10 minutes", or a policy introduced after a refresh chain started, never reaches that chain. OAuth 2.1 §2.4 (#1895) makes this matter: an authorization server is expected to limit the exposure of tokens issued to public clients, and today a never-expiring token refreshed by a public client stays never-expiring for as long as the chain is refreshed. This PR is the first of two; the second adds a `public_client_access_token_expires_in` ceiling on top of it.

### Change

- `Authorization::Token.access_token_expires_in` takes an optional block that supplies the TTL to fall back to when no custom expiration applies (no callable, or the callable returned `nil`). Without a block it is `access_token_expires_in`, exactly as before; the existing callers are unchanged.
- `RefreshTokenRequest#create_access_token` goes through `access_token_expires_in` again, passing the TTL of the token being refreshed as that fallback. The callable is called with `Doorkeeper::OAuth::REFRESH_TOKEN` as the grant type, the refresh token's application as `client`, and the resolved scopes.

So:

| `custom_access_token_expires_in`    | TTL of the refreshed token                                   |
| ----------------------------------- | ------------------------------------------------------------ |
| not configured                      | the original token's (unchanged)                             |
| returns `nil` for `refresh_token`   | the original token's (unchanged, #1364)                      |
| returns a value for `refresh_token` | that value (new)                                             |
| returns a value unconditionally     | that value — the inherited TTL is replaced (new; most common when the callable does not branch on grant type) |
| returns `Float::INFINITY`           | never expires (new)                                          |

### Compatibility

- Hosts without a callable, and hosts whose callable does not return a value for the `refresh_token` grant type, see no change. The #1364 scenario is pinned as a request spec.
- A callable that already branches on `context.grant_type == REFRESH_TOKEN` (written when the pre-#1366 behavior was current) starts taking effect again. That is the intended enhancement.
- A callable that returns a value unconditionally (without branching on `context.grant_type`) now applies to refreshes too. If the value matches what the token was originally issued with, the result is the same; if it differs, the refreshed token gets the callable's value instead of the inherited one.
- A callable that assumes `context.resource_owner` is present is now also called for refreshes. `resource_owner` is the record when `use_polymorphic_resource_owner` is on; otherwise the refresh token only stores the owner's id and `resource_owner` is `nil`, as it already is for the client_credentials grant. Callables that are safe for client_credentials are safe here.

### Docs

- The initializer template notes that the block is called for the refresh_token grant and that returning `nil` keeps the TTL of the token being refreshed.


nbulaj (approved):
LGTM ✅

55728:
Thank you @nbulaj! 😊

PR #1935, opened 2026-09-13
Add `public_client_access_token_expires_in` configuration option (OAuth 2.1 §2.4)

### LINKED ISSUE #1895: Public Clients and security exposure (OAuth 2.1)

55728:
Closes #1895. Builds on #1934.

### Why

OAuth 2.1 (draft-ietf-oauth-v2-1-15) §2.4:

> The authorization server MUST consider the security implications of interacting with unauthenticated clients and take measures to limit the potential exposure of tokens issued to such clients, (e.g., limiting the lifetime of refresh tokens).

The requirement is a policy one (what the measures are is left to the authorization server), so Doorkeeper can meet it by giving hosts a switch rather than by changing defaults. A host can already give public clients shorter tokens with `custom_access_token_expires_in` and withhold refresh tokens with `use_refresh_token`, for the grants that issue a token from scratch. Two things stop that from being a guarantee today:

- The refresh_token grant inherits the TTL of the token being refreshed (#1366). A never-expiring token issued before such a policy kept minting never-expiring tokens through the refresh_token grant. #1934 lets the callable take part again, but a callable is still opt-in logic the host has to write correctly.
- The callable has to remember `confidential?` in every branch. Forgetting it in one branch silently exempts public clients from the policy.

### What

A new, opt-in configuration option:

```ruby
Doorkeeper.configure do
  # Cap the lifetime of access tokens issued to public (non-confidential) clients.
  public_client_access_token_expires_in 1.hour
end
```

- It is a ceiling, applied in `Authorization::Token.access_token_expires_in`, which every grant goes through: authorization_code, password, client_credentials, implicit and, since #1934, refresh_token. The cap applies on top of whatever `access_token_expires_in`, `custom_access_token_expires_in` or the inherited TTL of a refreshed token would have given (`[ttl, cap].compact.min`). A never-expiring TTL (`nil`, including one from `Float::INFINITY`) is capped as well.
- Public means `!application.confidential?`. A request that carries no application at all (`skip_client_authentication_for_password_grant`, or a refresh token without an application) is treated as a public client: no client authenticated, so nothing to exempt.
- Confidential clients are not affected. The default is `nil`, which changes nothing for existing installations.
- The helper `Authorization::Token.cap_for_public_client(configuration, application, expires_in)` is public so that code issuing tokens outside these grants (custom grant flows, extensions) can apply the same ceiling.

### What it does not do

- It does not touch refresh tokens themselves (they carry no expiry in Doorkeeper; #1787 left that to hosts) and it does not add the §4.3.1 refresh token reuse detection, which is #1719.
- It does not retroactively shorten tokens that already exist. With `reuse_access_token` enabled, a still-valid token issued to a public client before the cap was configured is handed back as-is until it expires or is revoked; a never-expiring one is reused indefinitely.
- It does not change the `confidential` default or the behavior of any grant for confidential clients. An installation whose applications are all confidential (Mastodon registers every app as confidential) sees no effect from this option; that is by design, since the option keys off the client type the host registered.

### Docs

- Initializer template: the new option is documented next to `custom_access_token_expires_in`. The template also described `context.client` as `Doorkeeper::OAuth::Client`; it is the `Doorkeeper::Application` record (or `nil` when client authentication was skipped), which is what a `confidential?` check in the callable relies on, so the two mentions are corrected.

### Specs

- `spec/lib/oauth/authorization/token_spec.rb`: the cap applied to the callable's value, to the fallback, to a never-expiring TTL, to a request without a client; not applied to a confidential client or when the option is unset.
- `spec/lib/oauth/refresh_token_request_spec.rb`: a never-expiring token refreshed by a public client is capped; by a confidential client it keeps its expiry.
- `spec/requests/flows/{authorization_code,client_credentials,refresh_token}_spec.rb`: end to end, public vs confidential.
- `spec/lib/config_spec.rb`: default and setter.
- The `server` doubles in the existing request specs gained `public_client_access_token_expires_in: nil`, since `access_token_expires_in` now reads it.


nbulaj (approved):
👍

55728:
Thank you! 👀✨

PR #1939, opened 2026-09-14
Release 5.9.8

### LINKED ISSUE #1937: 5.9.7: `Token.from_request` raises `ActionDispatch::Http::Parameters::ParseError` when the body is malformed JSON, even with a valid Bearer header
Since 5.9.7, the "RFC 6750 §2 multi-method check in `Doorkeeper::OAuth::Token.from_request`" reads `request.parameters` and `request.POST` on every request. When the request body is not parseable (for example, malformed JSON with `Content-Type: application/json`), Rails raises `ActionDispatch::Http::Parameters::ParseError` from inside `from_request`. In 5.9.6 the header extractor matched first and the body was never parsed.

`doorkeeper_token` is documented as token-or-nil. It now raises instead.

## Steps to reproduce

Standalone script, no database needed. It loads `lib/doorkeeper/oauth/token.rb` from a tag and sends one request with a Bearer header and a malformed JSON body.

```ruby
# repro.rb — run with: ruby repro.rb 5.9.6 ; ruby repro.rb 5.9.7
# (download lib/doorkeeper/oauth/token.rb from each tag as token-<version>.rb first)
require "action_controller/railtie"
require "base64"

module Doorkeeper; module OAuth; end; end
load "./token-#{ARGV[0]}.rb"

def request_for(body)
  env = Rack::MockRequest.env_for(
    "/protected", method: "POST", input: body,
    "CONTENT_TYPE" => "application/json",
    "HTTP_AUTHORIZATION" => "Bearer valid-token",
  )
  ActionDispatch::Request.new(env)
end

methods = %i[from_bearer_authorization from_access_token_param from_bearer_param]
puts "token.rb from v#{ARGV[0]}"
puts "  well-formed body: #{Doorkeeper::OAuth::Token.from_request(request_for('{"a":1}'), *methods).inspect}"
begin
  puts "  malformed body:   #{Doorkeeper::OAuth::Token.from_request(request_for("not json"), *methods).inspect}"
rescue StandardError => e
  puts "  malformed body:   raised #{e.class}: #{e.message}"
end
```

Output:

```
token.rb from v5.9.6
  well-formed body: "valid-token"
  malformed body:   "valid-token"

token.rb from v5.9.7
  well-formed body: "valid-token"
  malformed body:   raised ActionDispatch::Http::Parameters::ParseError: Error occurred while parsing request parameters
```

Environment: doorkeeper 5.9.7, actionpack 8.1.3.1, rack 3.2.7, ruby 4.0.6. Default `access_token_methods` (`:from_bearer_authorization, :from_access_token_param, :from_bearer_param`).

## Where it raises

```
actionpack-8.1.3.1/lib/action_dispatch/http/parameters.rb:98:in 'parse_formatted_parameters'
actionpack-8.1.3.1/lib/action_dispatch/http/request.rb:414:in 'ActionDispatch::Request#POST'
actionpack-8.1.3.1/lib/action_dispatch/http/parameters.rb:57:in 'ActionDispatch::Http::Parameters#parameters'
doorkeeper-5.9.7/lib/doorkeeper/oauth/token.rb:72:in 'Token.from_access_token_param'
doorkeeper-5.9.7/lib/doorkeeper/oauth/token.rb:102:in 'Token.transmission_methods_used'
doorkeeper-5.9.7/lib/doorkeeper/oauth/token.rb:48:in 'Token.from_request'
doorkeeper-5.9.7/lib/doorkeeper/oauth/token.rb:62:in 'Token.authenticate'
doorkeeper-5.9.7/lib/doorkeeper/rails/helpers.rb:75:in 'Helpers#doorkeeper_token'
```

Two call sites parse the body: `from_access_token_param` / `from_bearer_param` (`request.parameters`) and `parameter_sources` (`request.POST`). Guarding only the extractors is not enough; `parameter_sources` raises next.

## Impact in a real Rails app

Our app returns 500 instead of 400 for any authenticated API request with an unparseable body:

1. The controller rescues `ParseError` via `rescue_from` and renders 400.
2. `ActionController::Instrumentation#process_action` then calls `append_info_to_payload` in an `ensure` block. Our override calls `doorkeeper_token` to log the token id (a common pattern).
3. `doorkeeper_token` re-raises `ParseError`. This happens after `rescue_from` already ran, so the exception reaches the `exceptions_app`, which raises again, and Rails logs `Error during failsafe response: Error occurred while parsing request parameters` and returns 500.

Before 5.9.7 the same requests returned 400. We had to pin `doorkeeper != 5.9.7`.

## Suggested fix

Rails guards its own logging path against exactly this case:

```ruby
# actionpack/lib/action_dispatch/http/filter_parameters.rb
def filtered_parameters
  @filtered_parameters ||= parameter_filter.filter(parameters)
rescue ActionDispatch::Http::Parameters::ParseError
  @filtered_parameters = {}
end
```

55728:
Backport of #1938 to the 5.9 maintenance branch, plus the version bump for the 5.9.8 release. Fixes #1937 on the 5.9 line.

Special thanks to @luizkowalski 👀✨


nbulaj (approved):
🥇

nbulaj:
Thanks!

Issue #96, opened 2012-06-11
Test User / Dev Mode Support


jsmestad:
It would be great to have some sort of 'test user' account for development apps or maybe a read-only setup. Thoughts?

PR #1941, opened 2026-09-16
Add experimental Client ID Metadata Documents support

### LINKED ISSUE #1875: Client ID Metadata Documents (CIMDs)

55728:
Closes #1875. Supersedes #1880 — this is the second of the two stacked PRs the prototype was split into. The first one (#1896, `private_key_jwt`) has since merged, and so has #1926 (audience from the server's own identity only), which was carved out of this branch. This now sits directly on `main`, rebased on top of both, and is **split into 54 commits ordered for review**: hardening of the code that already ships first, then the new modules one layer at a time, then the wiring, then documentation. Each commit carries its own specs and message; `git log --reverse` reads as a narrative.

Implements the MUSTs of [draft-ietf-oauth-client-id-metadata-document](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document) behind an opt-in `use_client_id_metadata_documents` config option: clients can identify themselves with an `https://` `client_id` whose metadata document Doorkeeper fetches, validates and materializes as an application row, instead of pre-registering. Section numbers below follow **-02**.

## How the commits are organised

| Commits | What they do | Reaches |
|---|---|---|
| 1–10 | `HttpFetcher` hardening: IPv6 literals vetted through `URI#hostname`, out-of-range ports refused, the post-RFC 6890 special-use ranges, a wall clock over the whole exchange (and no `Net::HTTP` retry), a socket-level byte budget that counts headers, UTF-8 / BOM handling | registered `private_key_jwt` clients' `jwks_uri` too |
| 11–13 | `KeyResolver`: JWK Sets built one key at a time (RFC 7517 §5), private/symmetric key material dropped, `use` / `key_ops` honoured | registered clients too |
| 14–20 | `PrivateKeyJwt`: NumericDate pinning, malformed JOSE headers and non-UTF-8 claims refused without raising, `jti` bounded, `nbf` always verified, global JWT leeway extends the replay guard, length-prefixed replay key | registered clients too |
| 21–22 | `ReplayGuard`: per-client accounting, then a pool of its own for URL clients | registered clients too |
| 23–29 | Client authentication plumbing: the RFC 6749 §2.3 check counts strategies not keys, `uses_shared_secret?` fails closed, strategies declare `auth_method_name`, the server records it on the credentials, `token_endpoint_auth_signing_alg_values_supported` is published (with a boot warning for a strategy that declares none) | every host; metadata output changes |
| 30–31 | Audience specs exercise the request path; the token endpoint URL is accepted under *every* configured identity (issuer and Rails host both) | registered clients too |
| 32–39 | The option, the dummy-app column, and the new modules: `ClientIdMetadata`, `UrlValidator`, `Document`, the blank-redirect exemption, `ApplicationFactory`, `resolve` | new code only |
| 40–46 | Wiring: `Client.find` / `Client.authenticate`, a document client held to the method its document names, the metadata flag, `private_key_jwt` for document clients, the URL-pool mark, provenance carried on the credentials | opt-in |
| 47–52 | The grants and endpoints a public document client is refused, consent on every authorization, the host on the consent screen and the authorized applications list, two boot warnings | opt-in |
| 53–54 | README, the initializer notes, CHANGELOG, and the adoption note for the Sequel / Mongoid extensions | docs |

Reviewers short on time: commits 1–31 are the ones that change behaviour for servers that never enable the option, and are each small; 32–54 are the feature.

## One thing to know before reading the rest

The feature needs a datetime column Doorkeeper's own migrations do not add:

```ruby
add_column :oauth_applications, :client_id_metadata_materialized_at, :datetime
```

(a declared field with the Mongoid extension, a Sequel migration with the Sequel one). It records which rows were materialized from a fetched document, which is the only thing that tells a document client apart from an application someone pre-registered under a URL-shaped uid — Section 7.1 says the `https://` prefix cannot, and Section 7.2 permits such registrations. Without the column every document client is refused. The generated initializer explains it, and the boot logs say so if the option is on without it.

## What's in here

- **URL `client_id`s** (Section 3 constraints: https, path required, no dot segments / fragment / userinfo, port in range, length-bounded; a query component is only SHOULD NOT in -02 and is tolerated). Opaque client_ids keep resolving against registered applications, so both kinds coexist (Section 7.1).
- **Document fetching** through the SSRF-hardened `Doorkeeper::HttpFetcher`: no redirects, 200 only, RFC 6890 special-use addresses refused with the connection pinned to the vetted IP, a declared non-JSON media type refused before parsing, a **5 KB body cap backed by a 13 KB socket budget that counts the status line and headers too**, a non-UTF-8 body refused, and a 10-second ceiling over the whole exchange.
- **Document validation** (Section 4 / 4.1): simple-string `client_id` match, `client_secret*` properties forbidden, shared-secret `token_endpoint_auth_method`s rejected — the draft's own names outright, and any configured method that does not declare itself secret-free through the registry's `uses_shared_secret?` contract — and the method must be one this server has configured. Public keys only: a `jwks` carrying private or symmetric key material is refused with the document, and such keys arriving from a `jwks_uri` are dropped when the keys are read. A secret presented alongside a URL client_id is refused outright.
- **Materialization**: a validated client is upserted as an application row (`uid` = the URL, stamped with `client_id_metadata_materialized_at`). Dirty tracking means an unchanged document issues no UPDATE. The insert is confined to a savepoint so a lost uid race can be recovered from inside a caller's transaction, and the provenance is re-checked under a row lock inside the write (Active Record; the Sequel and Mongoid extensions keep their plain save, which the initializer now says). For a `private_key_jwt` document client the row write is deferred until its assertion verifies, so an unauthenticated token request persists nothing.
- **Confidential document clients**: keys come from the document — inline `jwks` or a fetched `jwks_uri` — wired into `KeyResolver`, on a cache of their own so unauthenticated traffic cannot evict what registered clients depend on.
- **Consent** (Sections 8.3 / 8.4 / 8.5): the consent screen and the authorized applications list name the client_id URL's host beside whatever `client_name` the document declares, since nothing in the document is verified and the host is the one part of the identity the client demonstrably controls (new `doorkeeper.authorizations.new.client_id_host_html` translation). A document client is asked for consent on **every** authorization — an earlier authorization is not reused, since the document's redirect URIs and keys may have changed since.
- **Grants a document client cannot use**: a public (`none`) document client is refused the `client_credentials` and `password` grants and the introspection endpoint.
- **AS metadata** advertises `client_id_metadata_document_supported` (Section 6).

## Hardening that reaches registered clients too

Sharing the `private_key_jwt` and remote-fetch paths meant tightening them for everyone, so these also change for registered `private_key_jwt` clients from 6.0.0.beta2 (commits 1–31):

- `exp` / `nbf` / `iat` must be finite JSON numbers, `jti` at most 255 characters, `nbf` always verified, and a host-configured global JWT leeway extends the replay guard's memory of a `jti` accordingly.
- An assertion's audience may name the token endpoint URL built from either `issuer` or Rails' `default_url_options[:host]` when both are configured. (The removal of the `Host`-header fallback itself shipped in #1926 and is not part of this PR.)
- The replay guard remembers `jti` per client. A custom `private_key_jwt_replay_guard` now receives keys shaped `"<length>:<client_id>:<jti>"`, marked `url:` for a document client — the mark says which pool the built-in guard accounts the entry in, not which assertion it is, so single use is decided on the key without it. **Entries stored before the upgrade no longer match.**
- The keys an assertion was verified against travel with the credentials it produces (`VerifiedCredentials#from_metadata_document?`), so a client resolved after verification cannot have become the other kind in between.
- Four more IANA special-purpose address ranges are blocked.

The metadata endpoint also changes: each client authentication method is advertised under the IANA name its strategy declares (`auth_method_name`) rather than its registration key — **hosts running doorkeeper-openid_connect need its matching change so `/.well-known/openid-configuration` does not disagree** — with the `token_endpoint_auth_signing_alg_values_supported` RFC 8414 requires alongside an assertion-based method, gathered from the strategies' new `auth_signing_alg_values` declaration. A strategy that does not declare `uses_shared_secret?` is now treated as secret-based. And the RFC 6749 §2.3 multiple-methods check counts strategies rather than registration keys, so one strategy registered under several names is a single method rather than a request refused as `invalid_request`.

## Using `private_key_jwt` with document clients

Since #1926 every `private_key_jwt` assertion is audience-checked against the server's own identity (`issuer` or Rails' `default_url_options[:host]`), never against the request's `Host` header, and a server that configures neither is warned about it at boot. This PR adds nothing to that requirement, but document clients are the reason it matters twice over: a document `client_id` resolves to the same client, and the same keys, at every server implementing the draft, so an audience taken from the request would make an assertion sent to one such server replayable at all the others by whoever received it. A server that identifies itself nowhere refuses a document client's assertion before fetching anything, so an unauthenticated request cannot trigger an outbound fetch on the way to a refusal.

## Operational notes worth reading before enabling

- **What triggers a fetch**, and therefore a row: an authorization request as soon as its resource owner is signed in (and before any sign-in with `validate_client_before_resource_owner_authentication` on), and any endpoint that authenticates the client — token, revocation, introspection — for a public document client. Each such request holds its Rails thread for the length of the fetch. **Rate limiting is left to the host application, and this feature needs it.**
- **Views**: if you copied Doorkeeper's views or overrode the authorization screen or its translations, port the `client_id_host_html` line before enabling this, or your users see only the unverified `client_name`. The translation ships in English only. In `api_only` mode, derive the host with `Doorkeeper::ClientIdMetadata.display_host`.
- **Disabling later** also disables every client the feature materialized: a stamped row is refused as a client while the option is off. Already-issued access tokens stay valid until they expire. The host is still displayed for such rows, so a resource owner reviewing what to revoke still sees it.
- **Not compatible with `enable_application_owner confirmation: true`** — a document client is registered by no one, so its row has no owner. A warning is logged at boot when both are set.
- A registered application holding a URL as its uid pre-empts the document client at that URL, so if your host lets users choose uids, make sure they cannot choose `https://` URLs.
- **Adopting a materialized row** by clearing its stamp: with Active Record the write re-checks the stamp under a row lock, so an adoption racing a resolution wins; with the Sequel and Mongoid extensions clear it while no resolution of that client is in flight.

## Intentionally deferred

- A user-configurable size cap (5 KB / 13 KB are constants)
- `last_fetched_at` on the materialized row for staleness detection
- `token_endpoint_auth_methods_supported` (plural) client metadata
- A `client_metadata_changed` hook with token revocation (including JWKS key rotation)
- Enforcing "fetch only after resource-owner authentication" as a feature guarantee, and a loopback exception for development
- A config seam to swap the metadata fetcher or its cache, HTTP cache semantics, `logo_uri` storage

## Tests

Full stack green at the tip: **2445 examples, 0 failures** (58 files changed), and green at the two checkpoints between the hardening and the feature (after commit 22 and after commit 39). Every MUST has a spec, including negative-space assertions — redirect target never requested, no connection attempted on special-use resolution, shared-secret downgrade attempts, a secret presented for a URL client, malformed documents, hostile JWK sets, uid races recovered through the winner, and provenance changing between the two lookups an authentication makes.


55728:
@ThisIsMissEm Heads up — this PR supersedes #1897. The branch is rebased onto the current `main` and split into 54 commits ordered for review. Your earlier feedback is appreciated! 👀✨

erikaxel:
We are about to run CIMD in production today for a ChatGPT connector, and we’re watching this PR with a view to migrating onto it — so a data point on the a thing that would currently break for us.

ChatGPT’s document (https://chatgpt.com/oauth/client.json) declares:
```
"token_endpoint_auth_method": "private_key_jwt",
"token_endpoint_auth_methods_supported": ["none", "private_key_jwt"]
```
`Document#token_endpoint_auth_method` reads the singular property only, so this document yields `confidential? == true` and a `private_key_jwt assertion` is then required at the token endpoint. Our server offers `authorization_code` with forced PKCE and no `private_key_jwt` at all, so we hold the client to the none it advertises in the plural list, and the flow completes as a public client.

I saw plural `token_endpoint_auth_methods_supported` on the deferred list, and in #1875 the derivation was sketched as “confidential derived as any allowed method besides none”. I’d like to make the case that this is the wrong default for exactly this shape of document: the client is advertising a set of methods it can do, and `none` is in it. Deriving `confidential` from the singular preferred method makes an authorization server reject a client that is perfectly able to authenticate the way that server actually supports.

A suggestion: when the document advertises several methods, pick the intersection with what the server has configured, and treat the client as public when that intersection is `{none}`. A document naming only `private_key_jwt` stays confidential and is correctly refused by a server that doesn’t offer it. The singular property becomes the client’s preference among the ones it supports, which reads closer to what RFC 7591 registers than treating it as the sole declaration.

ThisIsMissEm:
@erikaxel yeah, so `token_endpoint_auth_methods_supported` is technically not yet in the CIMD draft specification: https://github.com/oauth-wg/draft-ietf-oauth-client-id-metadata-document/issues/87 + https://github.com/oauth-wg/draft-ietf-oauth-client-id-metadata-document/issues/96 — we were blocked on an OIDC spec which defined `token_endpoint_auth_methods_supported` to be correctly added to the IANA registry.

So that will naturally require changes to implementations of the draft specification once that lands for this is updated.

ThisIsMissEm:
@erikaxel also, have you seen the rc/beta release notes?
- https://github.com/doorkeeper-gem/doorkeeper/releases/tag/v6.0.0.rc1 
- https://github.com/doorkeeper-gem/doorkeeper/releases#release-v6.0.0.beta2
- https://github.com/doorkeeper-gem/doorkeeper/releases#release-v6.0.0.beta1

We've added `private_key_jwt` to doorkeeper core, and provided a standard pathway for implementing additional client authentication methods, as well as supporting RFC 8414 for clients to discover OAuth server configuration, and a heap of other big changes.

erikaxel:
Thanks for the quick reply! 

Yes, we are using 6.0.0rc1, and I saw the support for `private_key_jwt`.

However we went with the the 'none' strategy because that is the only one supported by both ChatGPT and Claude. Actually, from my understanding Claude only supports `none` at the moment.  https://claude.ai/oauth/claude-code-client-metadata (Although they don't use `token_endpoint_auth_methods_supported` at all, they just have `token_endpoint_auth_method`.

PS: I will admit I'm fairly new to these parts of oauth, so I might have gotten something wrong. But I have tested now that it works without  `private_key_jwt` for both Claude and ChatGPT. 

Let me know if I can be of further help.

ThisIsMissEm:
@55728 we do actually have a security problem here: right now we infer the `token_endpoint_auth_methods` for a client based on `confidential` and presence of `secret` on the `oauth_application` — we don't actually persist the client authentication methods. Here's my read on what we need to do:

In order to support multiple client authentication methods at application registration, we'd need to persist an array of `token_endpoint_auth_methods` on `oauth_applications`, that would be a migration to add the column (nullable, string set like `scopes`), if the column value is `null` then that means infer the value as we do today based on `secret` and `confidential`. 

We would then deprecate inference of `token_endpoint_auth_methods` in Doorkeeper 7.0.0 making it `NOT NULL`, which gives a timeframe for doing the backfill. Each consumer of doorkeeper will have its own logic for backfill based on `client_authentication` in `Doorkeeper.config` and what the inferred `token_endpoint_auth_methods` are for each client. We would provide a method to get the inferred client authentication methods for a client.

That backfill can be pretty expensive to perform, e.g., it's pretty common for Mastodon servers to have hundreds of thousands of registered clients due to dynamic client registration. However, Mastodon today only supports confidential clients with a secret and does not support public clients, so they could backfill with the following:

```sql
UPDATE oauth_applications
SET token_endpoint_auth_methods = 'client_secret_basic client_secret_post'
WHERE token_endpoint_auth_methods IS NULL;
```

For an authorization server that supports public clients (non-confidential, no secret) and confidential clients using the default value for `client_authentication` in `Doorkeeper.config`, then you could backfill with something like:

```sql
UPDATE oauth_applications
SET token_endpoint_auth_methods = CASE
  WHEN confidential THEN 'client_secret_basic client_secret_post'
  ELSE 'none'
END
WHERE token_endpoint_auth_methods IS NULL;
```

If `client_authentication` narrows from the default, or supports other methods, then you'd need to negotiate the correct methods supported by the authorization server for the given client, which means you can't do a bulk SQL `UPDATE` operation.

---

If an application creation presents just `token_endpoint_auth_method` then the `token_endpoint_auth_methods` on the `oauth_application` model would be a set with a single member (the value of `token_endpoint_auth_method`), if `token_endpoint_auth_methods` is present, then you use that value for the `token_endpoint_auth_methods` stored against the `oauth_application` record. This works for both Dynamic Client Registration and CIMD. 

If an application creation does not present `token_endpoint_auth_method` or `token_endpoint_auth_methods` then the model's `token_endpoint_auth_methods` must be set to NULL until Doorkeeper 7.0.0, when the application creation would be changed to fail with a validation error (missing `token_endpoint_auth_method` or `token_endpoint_auth_methods`) — this is because the model cannot infer the correct method when that becomes required.

_Technically_, if `token_endpoint_auth_method` is not present in DCR, then the default value of `client_secret_basic` is used, even though most implementations treat `client_secret_basic` and `client_secret_post` as interchangeable (and I'd discourage using `client_secret_basic` because it's not actually in conformance with HTTP Basic Authentication specification: it performs the `application/x-www-form-urlencoded` routine from WHATWG on the `username` and `password` values before concatenating them with `:` and doing base64 encoding of the concatenated string — regular HTTP Basic Authentication doesn't do the encoding over `username` and `password`).

---

During a request to the authorize endpoint, PAR endpoint, or DCR registration endpoint, if none of the `token_endpoint_auth_methods` of an `oauth_application` are supported by the authorization server then the authorization server should return the `invalid_client` error code — for the authorize endpoint request, if the `client_id` and `redirect_uri` are both valid, then the authorization server can redirect back to the `redirect_uri` with the `error` code of `invalid_client` (client authentication failed), if the `client_id` or `redirect_uri` are not valid, the authorization server must not redirect back to the `redirect_uri` and instead show an error page (we already support this negotiation for errors).

_Note: currently `invalid_client_metadata` is an error code *only* supported on the DCR registration endpoint. Also, don't confuse `invalid_client_metadata` with `invalid_metadata` which is only valid for the authorization endpoint. Currently CIMD does not have a change to `invalid_client_metadata` to make it supported for both authorization endpoint, token endpoint, AND registration endpoint, this could be a good issue to file in the CIMD issue tracker._

---

During a request to the token endpoint, introspection endpoint or others that may require client authentication, then the authorization server must enforce client authentication in accordance with the `token_endpoint_auth_methods` on the `oauth_application` model, and the `error` returned would be `invalid_client` per rfc6749 section 3.2.1 and 5.2.

---

(n.b., there are `revocation_endpoint_auth_methods_supported` and `introspection_endpoint_auth_methods_supported` in RFC 8414's Authorization Server Metadata, however, I'd agree with the OpenID Connect working group that it was a mistake to create those separate fields, per [OpenID Connect Relying Party Metadata Choices 1.0](https://openid.net/specs/openid-connect-rp-metadata-choices-1_0-final.html#name-client-authentication-metho))

ThisIsMissEm:
@erikaxel okay, hopefully my above reply makes sense, but the specification for `token_endpoint_auth_methods_supported` is clear: if `token_endpoint_auth_methods_supported` and `token_endpoint_auth_method` are both supported, use `token_endpoint_auth_methods_supported`'s value, if `token_endpoint_auth_methods_supported` isn't supported ignore it, and use `token_endpoint_auth_method`. As a client, you can check what token endpoint auth methods are supported by requesting the OAuth Authorization Server Metadata (RFC 8414) and ensure that your values intersect with the authorization server's values.

ThisIsMissEm:
@55728 oh, also the model should probably have a TEXT column which contains the full client metadata since other values within it may help extensions or be used for negotiating say private_key_jwt signing algorithms

55728:
@ThisIsMissEm thank you 😊 — I reproduced it, and I agree. Below: what I measured, why I'd rather land the fix outside this branch, a couple of implementation details that affect how this lands in Doorkeeper, and then some questions I've been sitting on that you're the right person to ask.

## Reproduced

With `client_authentication %i[client_secret_basic client_secret_post none private_key_jwt]`, a registered confidential application that published a `jwks` — one set up for `private_key_jwt` — authenticates the token request with `client_secret_post` and is issued a token. Same on `main` (6.0.0.rc1) and on this branch. Nothing pins a registered client to the method it was registered for.

The one thing I'd separate out: this is core behaviour that shipped in rc1, and it became reachable when #1896 put `private_key_jwt` in core and made `client_authentication` a list. Document clients are the part that's already pinned — `Doorkeeper::OAuth::Client` holds a document client to the single method its document names on every path into `.authenticate`, and refuses a presented `client_secret` for a URL `client_id` before fetching anything. So the exposure is registered applications, and this branch doesn't widen it.

## Why not in this branch

A `token_endpoint_auth_methods` column on `oauth_applications` is a core model change with a generator migration, a deprecation window and a backfill that reaches every Doorkeeper host, most of whom will never enable CIMD. This branch is 54 commits behind an opt-in flag and currently conflicting with `main`.

It's also a release call I shouldn't make on my own: rc1 is out, so whether this goes into 6.0, 6.1 or 7.0 is @nbulaj's. I've opened #1948 with the design and open questions for him. Happy to write the PR.

## Implementation details that affect how this lands

**1. `NULL` inference needs to include `private_key_jwt`, or it breaks rc1 clients.**

You describe `NULL` as inferring "as we do today based on `secret` and `confidential`", which leaves out assertion-based clients. `jwks` / `jwks_uri` aren't in the core generator migration — they only exist in `spec/dummy/db/schema.rb`, so hosts add them. A registered `private_key_jwt` client under rc1 is therefore a row with `confidential = true` and a secret it never uses, and inferring `client_secret_basic client_secret_post` for it means every such client breaks the moment enforcement lands unless the host backfilled exactly right.

For `NULL` to be a real no-op it should infer every configured method the row could satisfy: secret present → the shared-secret methods; `jwks` / `jwks_uri` present (`respond_to?`-guarded, since they're host-added) → `private_key_jwt`; `confidential == false` → `none`. That's today's behaviour exactly, which is what makes it safe to ship ahead of the backfill — and it makes the "method to get the inferred methods" you propose the same code path enforcement uses, so the two can't drift.

**2. `invalid_client` isn't in the authorization endpoint's error set.**

Agreed on token and introspection — RFC 6749 §5.2 names "unsupported authentication method" in the definition of `invalid_client`, so that's this case precisely.

At the authorization endpoint it isn't in the set: §4.1.2.1 and §4.2.2.1 both enumerate `invalid_request`, `unauthorized_client`, `access_denied`, `unsupported_response_type`, `invalid_scope`, `server_error`, `temporarily_unavailable`. Doorkeeper already assumes this — `OAuth::ErrorResponse::NON_REDIRECTABLE_STATES` contains `:invalid_client` and `#status` answers 401 for it, so it's never redirected back. `unauthorized_client` is the code that fits "this client may not use this flow here", and the redirect negotiation you describe then applies to it unchanged.

**One scope note:** doorkeeper core has no DCR registration endpoint and no PAR endpoint, so in this gem the enforcement points are authorization, token, introspection and revocation. I'd still design the column and its writer so a DCR extension can set it, which covers what you describe.

And agreed on the RFC 8414 aside — splitting `revocation_endpoint_auth_methods_supported` / `introspection_endpoint_auth_methods_supported` out of the token endpoint's list was unfortunate, and the RP Metadata Choices position is the one to follow.

## The TEXT column

For document clients I'd rather not, and I want to check my reasoning against yours.

The validated document is already available and fresher than a column: `Doorkeeper::ClientIdMetadata.document_for(client_id)` returns the parsed `Document`, memoized with a 60-second TTL, and `private_key_jwt` key resolution reads it there rather than off the row — so signing-algorithm negotiation already has access.

I've been treating §8.3 / §8.4 as meaning a document client is whatever its URL serves *now*, which is why `ApplicationFactory` re-applies the document on every resolution instead of trusting the row. A stored copy would be a second source of truth that's stale by construction. If that reading is stricter than you intended, say so — it's load-bearing in a few places here.

There's an operational cost as well. `ApplicationFactory` deliberately doesn't `save` when nothing changed, because resolution runs on unauthenticated traffic and Sequel writes every column on save, so a write per request would have concurrent resolutions of one popular document client serialize on its row. Raw JSON isn't byte-stable across fetches, so storing it turns nearly every resolution back into a write. It also puts up to 5 KiB of unvalidated remote JSON into every host's database, which host apps will render.

Where I think a stored copy does earn its place is **registered** clients under DCR — you want what was actually submitted, and there's no URL to re-fetch. That's the same core issue as the column above, so I'd fold it in there.

## Questions for you, as an author of the draft

This implementation has three places where I had to pick a reading, and they're all still open upstream. I'd rather ask than keep guessing:

**1. Omitted `token_endpoint_auth_method` (draft issue `#92`).** Doorkeeper treats an omitted value as `none`. My reasoning: §4 says the values are the ones RFC 7591 registers but not that RFC 7591's *default* comes with them, and that default — `client_secret_basic` — is a method §4.1 forbids these clients outright, so it can't be the intended reading. I settled on `none` because the AT Protocol profile makes the property optional and asks it only of confidential clients, so a public client omitting it is the ordinary case there, and refusing would turn all of those into `invalid_client`. Is that the direction `#92` is heading, or is refusing an omitted value the preferred outcome?

**2. A method the authorization server doesn't support.** Right now Doorkeeper fails document *validation* when `token_endpoint_auth_method` names a method this server isn't configured for, so the client is refused as `invalid_client` at resolution. That's exactly what @erikaxel hit: their server offers no `private_key_jwt`, ChatGPT's document names it, document rejected. Is refusing at resolution the behaviour you'd want, or should the server be expected to negotiate? I ask because the answer shapes what `token_endpoint_auth_methods_supported` should do when `#87` lands — if negotiation is the model, then the plural is the mechanism and the singular is a preference, and `confidential?` has to derive from the intersection rather than from the singular.

**3. `private_key_jwt` with no keys.** Doorkeeper refuses a document naming `private_key_jwt` that publishes neither `jwks` nor `jwks_uri`, on the grounds that it describes a confidential client which can never authenticate, yet would still materialize rows and grants at the authorization endpoint. I have this filed as hygiene rather than compliance, since RFC 7591 §2 makes both optional and the draft states no requirement. Worth a MUST in the draft, or deliberately left open?

On @erikaxel's original report: understood that the plural isn't in the draft yet (`#87`, `#96`), so I'm not going to guess at its semantics. When it lands it's a contained change here — `Document#token_endpoint_auth_method` and `#confidential?` become a set plus a negotiation against `Doorkeeper.config.client_authentication_methods`.

---

The design is in #1948 — would welcome your thoughts on the notes above and the three questions.

ThisIsMissEm:
@55728 I would argue that #1948 is a blocker for shipping CIMD support, so needs to be done before release. It's fine to do it separately, but without it we could get issues where the incorrect authentication method is used for a given client.

ThisIsMissEm:
## On "Implementation details that affect how this lands":

1. Yes, private_key_jwt has landed, however, the initializer does not enable it by default, hence my set of `client_secret_basic client_secret_post none` — arguably the initializer probably needs to case on whether non-confidential clients are supported and whether private_key_jwt is supported.
2. `invalid_client` seems like it should be the correct value based on its description, however, it is indeed absent from the error codes for the authorization endpoint, despite the IANA registry claiming otherwise; have [emailed the mailing list](https://mailarchive.ietf.org/arch/msg/oauth/GlT6Ue2LIOqi_JSdyL-6zKT8F2w/) to seek clarification.

## For your other questions:

1. I think you are correct, however, we should be explicit in CIMD, and this is my attempt at making it explicit: https://github.com/oauth-wg/draft-ietf-oauth-client-id-metadata-document/pull/99
2. So building on #1948, if the CIMD does not carry any of the client authentication methods supported by the authorization server, reject, otherwise, use only the supported methods. For the ChatGPT case, it declares `token_endpoint_auth_methods_supported` so the `Doorkeeper::Application`'s token endpoint authorization methods would be `"none private_key_jwt`, and an authorization server not supporting `private_key_jwt` would mean only `none` is usable. That also allows an authorization server to start supporting `private_key_jwt` with that client in the future, if they add support for `private_key_jwt`.
3. `private_key_jwt` without `jwks` or `jwks_uri` should abort the authorization request, however, I'm not sure with which error. I cannot find a defined error for what happens under this scenario, so have opened https://github.com/oauth-wg/draft-ietf-oauth-client-id-metadata-document/issues/100

Hopefully my changes in https://github.com/oauth-wg/draft-ietf-oauth-client-id-metadata-document/pull/99 will help us land both #1948 and this PR.

55728:
@ThisIsMissEm Thank you 😊 — the draft PR (oauth-wg/draft-ietf-oauth-client-id-metadata-document#99), the new issue (oauth-wg/draft-ietf-oauth-client-id-metadata-document#100) and the mailing list thread are exactly the kind of upstream clarity this implementation needs. I'll track all three and update #1948 and this branch as they land. 👀✨

nbulaj:
Hello all. @55728  I believe this one should be a real draft, right? Because there are some open questions for the CIMD standard and the related implementation.

55728:
Good point — converted to draft. The spec itself still has open questions (like [draft issue #92](https://github.com/ietf-wg-oauth/draft-ietf-oauth-client-id-metadata-document/issues/92) on the default `token_endpoint_auth_method`), and the "Intentionally deferred" section lists what this PR deliberately leaves for later. Happy to promote back to ready once the picture is clearer.

nbulaj:
BTW I see an increasing interest in CIMD being included to the Doorkeeper so this is a candidate for 6.0 as well. Solid work @55728  :muscle: :+1:

Issue #792, opened 2016-02-25
:token added to filter_parameters breaks logging of the application


dup2:
The gem uses

```ruby
module Doorkeeper
  class Engine < Rails::Engine
    initializer "doorkeeper.params.filter" do |app|
      app.config.filter_parameters += [:client_secret, :code, :token]
    end
   # .. omitted
  end
end
```
to prevent pot. sensitive information being logged anywhere. 

These parameters are greedy, they will also affect any other parameters in a request containing these words, i.e. with the default configuration the paramter `custom_token` will also get logged as FILTERED which affects our application.

With a quick code check, I found `:access_token` and `:refresh_token` being used, I'm not sure what `:code` is meant for.

I'd propose to use more specific matchers instead of the generic `:token` to be added in the initializer:

`[:client_secret, :access_token, :refresh_token]`

What do you think? Shall I create  a PR for this?

Cheers

tute:
Fixed. Thanks for reporting!

PR #1942, opened 2026-09-16
Add opt-in client secret rotation with a grace period

### LINKED ISSUE #1675: Better support for credential rotation
Right now credential rotation while using Doorkeeper seems a little cumbersome. From my POV, there are currently 2 major approaches that can be taken:

1. Only the secret is rotated. This approach seems more elegant, but the major drawback here is that client applications cannot authenticate until they have updated their credentials in the code.
2. The entire OAuth application is rotated. This allows for a grace period during which time new credentials can be added/tested while the old credentials remain active. However, this method is not without consequences. Assuming we do not purge all tokens during the rotation, if the client wants to revoke a token during the transition period, they may end up attempting to revoke a token created by a different OAuth application (not allowed). Also if anyone is using the oauth_application_id or uid for metrics they have to take this transition into account.

My suggestion/feature request here would be to add additional support for approach 1. Perhaps we could have an "old_secret" field that remains active for a period of time after the secret field is updated. Thoughts?

55728:
Closes #1675.

Rotating a client secret today means downtime: the moment a `renew_secret` is saved, every client still holding the old secret is rejected, and it stays rejected until it has been redeployed. The issue lays out the two workarounds people reach for — rotate only the secret and accept the outage, or rotate the whole OAuth application and inherit a second `client_id` that breaks revocation and metrics — and asks for a third option.

This implements the shape @ClaytonPassmore proposed in the thread: keep the superseded secret in an `old_secret` column, let both authenticate the client, and let the application decide when the old one stops working rather than expiring it on a timer. The flow is the one described there:

```ruby
secret = application.rotate_secret!  # both secrets now authenticate
# ... hand `secret` to the client, let it deploy ...
application.clear_old_secret!        # only the new secret authenticates
```

The change is 29 commits, each with its own specs, ordered roughly writer → reader → hardening → docs: generator and columns, the option and its availability check, `rotate_secret!` and its rollback repair, `clear_old_secret!` and `retained_at:`, the comparison (refactor, honour the retained secret, deadline, constant-cost comparison, `after_old_secret_used`), the boot warnings, `revoke_tokens:`, serialization and `#inspect`, the strategy predicate and re-derivation, then README/initializer and the changelog. Seven further commits harden the edges: `Array()` for serialization inputs, reading the commit state off the row for `ConnectionFailed`, the connection-pool path for the joinable-transaction guard, caller-specific error messages, doc consistency, spec-gap closures, and parameter-based hook validation. Every commit passes the rotation specs on its own.

## What's in it

- **`enable_secret_rotation`** — off by default. While it is off, nothing in the secret comparison changes.
- **`rails generate doorkeeper:secret_rotation`** — adds `old_secret` and `old_secret_created_at` to `oauth_applications`.
- **`Application#rotate_secret!(revoke_old: false, revoke_tokens: false)`** — returns the new plain text secret, which is the only chance to read it when secrets are hashed.
- **`Application#clear_old_secret!`** — ends the grace period; returns whether there was anything to clear.
- **`after_old_secret_used`** — called with the application whenever a client authenticates with the superseded secret rather than the current one. Validated at configuration time against its parameter signature: a non-callable, a required keyword, or a positional count that cannot accept one argument raises `ArgumentError` at boot instead of failing the first authentication with the old secret.
- **`secret_rotation_grace_period`** — optional deadline after which the old secret stops authenticating. Nil by default, which is the "clear it yourself" model above.
- **`Application#old_secret_matches?` / `#old_secret_expired?`** — the old-secret half of `#secret_matches?`, exposed for applications that want to know which secret a client presented and whether its grace period has run out.

`#renew_secret` is untouched: it writes the current secret alone and only assigns it, so with the feature on it leaves whatever an earlier rotation retained still authenticating. To replace a secret with no grace period — or to end one already under way — use `rotate_secret!(revoke_old: true)`.

## Knowing when to clear

Deciding that the grace period is over means knowing whether anybody still depends on the old secret, and an implementation that only made the old secret *work* would leave that question unanswerable. `after_old_secret_used` closes it:

```ruby
after_old_secret_used ->(application) {
  StatsD.increment("oauth.old_secret_used", tags: ["client:#{application.uid}"])
}
```

It fires only when the old secret is what actually authenticated the client, so a quiet counter means the rotation is complete. It runs on successful authentications only — a caller that can produce neither secret never reaches it — so it adds nothing an attacker can measure.

`secret_rotation_grace_period` is the belt to that braces, for operators who would rather not depend on remembering. Past the deadline the old secret stops authenticating; it stays in the column until `#clear_old_secret!` or `#rotate_secret!(revoke_old: true)` removes it, so an operator can still see that a rotation happened. An old secret with no `old_secret_created_at` is treated as expired rather than as ageless — honouring a deadline nobody can date would leave exactly the indefinitely-valid secret the option was configured to prevent.

## Why not build this on the client authentication registry

The registry landed in #1839–#1841 and would be the natural home for a general "this client has a previous credential" notion, so it is worth saying why this deliberately isn't that.

The other registered methods don't rotate through a second stored credential. `private_key_jwt` rotates by publishing a second JWK — the client adds a key to its JWKS, moves over, removes the old one, and the server needs no new concept because a JWK set is already a set. mTLS rotates by registering a second certificate, with the same property. `client_secret_basic` and `client_secret_post` are the only methods where the credential is a single opaque value the server stores, and they already share one: `Application#secret_matches?`. So an abstraction at the registry level would today have exactly one implementation and two callers that are the same code path.

Scoping to secret-to-secret rotation keeps the change to the one place that needs it. The registry is still there when a second caller shows up, and nothing here forecloses moving the logic behind a method contract later.

## Timing

The feature adds a fact about a client that wasn't previously observable: whether it is midway through a rotation. Comparing the old secret only when one exists would put that fact in the response time, and under bcrypt the gap would be the dominant cost of the request rather than a subtle one.

So while the option is on, `#secret_matches?` evaluates both halves and combines them afterwards — no early return when the current secret matches — and `#old_secret_matches?` performs a comparison even when there is no old secret, against the current secret, discarding the result. Grace-period expiry is likewise decided after the comparison has run, so an expired old secret costs exactly what a live one does.

The dummy comparison is made against a real stored secret rather than a constant, so that it costs what a genuine comparison costs *whenever both columns were written by the same strategy* — which is every application Doorkeeper rotated itself. It does not equalise a `secret` and an `old_secret` held in different formats, which the fallback strategy allows: measured at 200 attempts, a bcrypt `secret` with a plain `old_secret` runs 2× faster than one with a bcrypt `old_secret`, because the plain value fails the bcrypt comparison before any work factor is applied. That is the fallback strategy's existing property — it already leaks the same bit about `secret` — rather than something introduced here, and it disappears once the old secret is upgraded on its first successful use.

Two further things this does not claim to fix, both pre-existing and both larger signals:

- An unknown `client_id` performs no comparison at all, because there is no application to compare against.
- With a fallback strategy configured, a non-matching secret costs one more comparison than a matching one, since the fallback is only consulted after the active strategy has failed. That makes failures slower, not faster.

## Concurrency

Both `#rotate_secret!` and `#clear_old_secret!` are read-modify-writes on the row — one retains the secret currently stored, the other decides whether there is a grace period to end — so both take the row lock, as `AuthorizationCodeRequest` and `RefreshTokenRequest` do.

Without it, two rotations racing (an admin double-click, a rake task overlapping a manual rotation) would let the later one overwrite a secret the earlier one had already handed to a client, which would then never authenticate, with nothing in the data to show why. `#clear_old_secret!` reads its "is there anything to clear?" answer under the lock for the same reason: read outside it, the answer describes whatever the instance was loaded with, which a rotation committed since then has already made wrong in both directions.

Taking the lock reloads the row, which Active Record refuses to do over unsaved changes rather than discard them silently, so both methods require a clean record. That is pinned in the specs.

When either write fails, the row is rolled back but Active Record leaves the in-memory attributes as the failed write left them — an instance still holding a secret that was never stored, or claiming a grace period it no longer has, is a trap for any caller that rescues and carries on. Each method restores the columns it writes (and `#rotate_secret!` the volatile plaintext with them); nothing else the caller had changed is touched.

## Issued tokens

Rotation on its own leaves tokens alone: a routine rotation is not a revocation, and the tokens were issued to the same client.

For a leaked secret, `rotate_secret!(revoke_old: true, revoke_tokens: true)` revokes the access tokens already issued **and** the application's unredeemed authorization codes. The codes are included because an authorization code is exchanged with the client secret (RFC 6749 §4.1.3) — leaving them alive would hand back an access token minted after the revocation. Pass `revoke_tokens:` whenever the secret may already have been used: the rotation stops the secret from authenticating, but any access token or refresh token minted with it before the rotation stays valid until it expires, and whoever holds one no longer needs the secret. A secret does not hand out a token issued to someone else — except under `reuse_access_token`, where a `client_credentials` request made with the leaked secret is answered with the token that grant already holds. Flag if you'd rather `revoke_tokens:` mean access tokens only and grants get their own keyword.

`rotate_secret!(revoke_tokens: true)` revokes only for a rotation the row holds. Whether the rotation committed is read off the row, with one SELECT on the failure path: since Rails 7.1 a COMMIT lost to `ActiveRecord::ConnectionFailed` invalidates the transaction without rolling the records back, so the instance comes out looking exactly like a committed one while the row most likely still holds the previous secret. Where the COMMIT did reach the server and only its acknowledgement was lost, the row says so and the sweep runs as it should. A host `after_commit` callback raising is the same: the rotation is committed, the row confirms it, and the sweep runs.

`rotate_secret!(revoke_tokens: true)` refuses to run inside an open, joinable transaction (`SecretRotationInTransaction`): the revocation has to wait for the rotation's row lock to be released, and a transaction it joined holds that lock until its own commit — which would recreate the lock-order inversion the sweep is kept out of the lock to avoid. Rotate without `revoke_tokens:`, then call `#revoke_issued_credentials!` once the transaction has committed. A non-joinable transaction is let through: Rails' transactional tests and DatabaseCleaner wrap every example in one, so refusing it would raise in every host application's test suite. The joinable check goes through the connection pool rather than `connection`, the permanent-checkout API Rails 7.2 soft-deprecated: under `permanent_connection_checkout = :disallowed` the old path raises whenever nothing has leased the thread's connection yet, which is how every request and job starts.

## Calling `rotate_secret!` twice

Only one generation is retained, so a second rotation ends the first one's grace period early: the secret it retained is replaced by the one the second rotation supersedes, and any client that hadn't yet moved off it stops authenticating.

Treating that as the specified behaviour rather than guarding against it, for two reasons. It matches `previous_refresh_token`, which keeps exactly one generation of refresh tokens for the same reason. And the alternatives are worse: refusing to rotate while an old secret is pending would block the compromise case, which is precisely when you want to rotate immediately and repeatedly; retaining N generations turns one column into a table and multiplies the comparisons the timing defence has to keep constant.

`#rotate_secret!` raising when the feature isn't available covers the neighbouring foot-gun — calling it without the migration would otherwise drop the secret it was asked to retain, cutting off every client the caller meant to protect.

## Notes

- Rotation is active only when the option is set **and** both the `old_secret` and `old_secret_created_at` columns exist, following `pkce_supported?`. Enabling the option without running the generator leaves authentication behaving exactly as before rather than raising on every token request — and says so once per process, the first time the columns are looked for. That is deliberately not a boot hook: reading the columns is a database read, and a boot hook asking for them would open a connection during tasks that have none (`assets:precompile` and friends) and wait through its retries to find that out.
- With `enable_secret_rotation` set under an ORM other than Active Record, Doorkeeper warns at boot that the option has no effect. Everything the option changes lives in `Doorkeeper::ApplicationMixin` and in the Active Record application mixin; the other ORM adapters ship application models that include neither, so the option is a no-op, and one nothing else would report — the warning the Active Record model gives for missing columns never fires where there is no column to miss.
- A secret the active strategy recognises as its own is carried over as *stored*, not re-derived — hashing strategies draw a fresh salt on every write, so there is no plaintext left to hash again. A secret the **`fallback:` strategy** wrote is re-derived under the active one as it is retained. Carrying that one over as stored is what an earlier revision did, and it was wrong twice: the plaintext of a `fallback: :plain` secret outlived the very rotation that replaced it, sitting in `old_secret` for the length of the grace period on a row whose `secret` is now hashed; and the two columns then cost different comparisons, so a rotated application answered measurably faster than an unrotated one — to anyone, holding no secret at all. Re-deriving needs the plaintext, which only a restorable fallback has (`:plain` is the one Doorkeeper ships); the re-derived value is written through the active strategy's `store_secret`, as `renew_secret` and the fallback upgrade do, so that a strategy whose `store_secret` wraps what it stores keeps authenticating the retained secret. Against a fallback that hashes there is nothing to re-derive from and the value is carried over as stored, as it always was. What decides is `SecretStoring::Base.recognizes_stored_secret?`, new here and answering `true` by default, so a custom strategy that does not override it retains as stored as well — misreading a value it *did* write would replace a working retained secret with the hash of its own hash. `Sha256Hash` with `default_generator_method :hex` is the one combination that cannot tell the two apart (both are 64 hex characters); README says so and says what to do about it. The fallback strategy is still honoured when the old secret authenticates, upgrade-on-match side effect included.
- `old_secret` is not indexed: it is only ever compared against after the client has been found by `uid`, never looked up by value.
- Neither new column is serialized by default. The public `as_json` whitelist excludes them structurally; the owner/admin view, which otherwise dumps every attribute, has them withheld explicitly. The retained secret authenticates the client exactly as `secret` does, and unlike `secret` — surfaced deliberately through `#read_attribute_for_serialization` — nothing needs to hand it out. `old_secret_created_at` is withheld with it, since on its own it still reports that a client is midway through a rotation. The exclusion is expressed against `only` as well as `except`, because ActiveModel honours one or the other and never both; both are converted with `Array()`, matching how ActiveModel itself converts them, so a Set, an Enumerator, or any object responding to `to_a` is narrowed the same way. It is a default and not a guarantee: ActiveModel appends a `methods:` reader after that filtering, so `serializable_hash(methods: [:old_secret])` still answers a caller that names it.
- Both columns are named on the application model's own `filter_attributes`, which is what Active Record matches against column names, so they read as `[FILTERED]` from `#inspect` and from the bind parameters Active Record logs. Deliberately *not* added to `config.filter_parameters`, which reaches every model and every log line in the host application: neither is the name of a request parameter Doorkeeper accepts, and `token` was in that list once and was taken back out for exactly that reason (#792). Unconditionally, not gated on the option: turning rotation off does not clear the column, so a server that rotated once and then disabled the option still holds a live credential there. Because the columns are declared on the model, `Doorkeeper::Application.filter_attributes` is composed from the base class once, when the model loads: entries a host adds to `ActiveRecord::Base.filter_attributes` afterwards do not reach the application model's `#inspect`, though they keep reaching the token and grant models, which declare no list of their own. Rails' generated `:secret` filter already matches `old_secret` and `old_secret_created_at` as a substring, so both columns read as `[FILTERED]` in SQL bind logs on a default host; the README says so next to the `config.filter_parameters` note.
- `SecretRotationNotEnabled` gives a different message for `#clear_old_secret!` than for `#rotate_secret!`: that path needs the columns alone and is meant to work with the option off, so telling its caller to add the option would send them the wrong way.
- The columns are **not** in the install migration, since the feature is opt-in and off by default — same as resource indicators and PKCE. Say the word if you'd rather new installs carry them.

## Future work this leaves open

- Moving the "previous credential" notion behind the client authentication registry, if a second method ever needs it.
- Exposing rotation in the applications admin UI. Left out here to keep the change to the model and the comparison.

## Tests

`spec/models/doorkeeper/secret_rotation_spec.rb` covers the model API, both hashing strategies, the fallback upgrade path, the column-missing and option-off states, the lock and rollback behaviour on both writers (including a COMMIT lost to `ConnectionFailed` and `permanent_connection_checkout = :disallowed`), serialization on both owner paths (with `Set` inputs pinned), the hook and its parameter-signature validation, the grace period, the comparison-count assertions that pin the timing behaviour, the `filter_attributes` composition, and the regression specs for the COMMIT-failure sweep and the `store_secret` re-derivation. `spec/requests/flows/secret_rotation_spec.rb` drives the token endpoint end to end over both `client_secret_basic` and `client_secret_post`. Full suite passes with 0 failures.


nbulaj:
**_Side note for us_** (no need to do anything right now):  we have to consider refactoring validations to their own concerns, maybe with a registry, with a single call like `ConfigValidatorRegistry.validate_all!` instead of this increasing number of procedural calls:

image

Smth like separating entity/domain related validations into own modules/classes which register them-self in a registry which later run all of them at once. Our grant flows registry as an example

PR #1943, opened 2026-09-19
[test] Cover stateless JWT token edge cases

55728:
Hey @nbulaj 👋

I took your `stateless-jwt-token` branch for a spin and added spec coverage for the parts that weren't pinned down yet. All spec-only, no changes to lib/.

What's covered:

- **Token.authenticate branching** — decoder not configured / raising / returning non-Hash, opaque tokens falling through to DB, 2-segment and 4-segment tokens taking the database path
- **Application resolving** — matching / unknown / blank / missing `client_id`, and a failing `find_by` (DB fault tolerance)
- **find_or_create_for** — no record persisted, claims content (owner, scope, client_id, iat, exp, expires_in), custom expires_in, polymorphic resource owner, custom attributes, generator argument forwarding, generator errors
- **Config validations** — each warning independently, all four at once, no warnings when properly configured, block-form `use_refresh_token`
- **Edge cases** — non-expiring tokens (no `exp`), nil/blank scope, aud variants

46 → 126 examples, full suite stays green (2044 examples, 0 failures).

Two things I noticed while writing specs (info-level, not blocking):

1. With `polymorphic_resource_owner?` enabled and a nil resource owner (client_credentials), `build_stateless_token` sets `resource_owner_type` to `"NilClass"` — the DB path leaves it `nil`.
2. `jwt?` splits on `"."`, so `"a.b."` (empty signature, think `alg: none`) is treated as 2 segments and falls through to DB lookup. No real harm, but worth noting.

Let me know what you think! 🙌


55728:
Hey @nbulaj 👋

Converting this to draft to lighten your review queue — I know there are a few PRs from me stacked up right now and I don't want to overwhelm you. The code is ready whenever you have bandwidth; I'll mark it ready for review again when the timing feels right.

No rush at all! 😊

nbulaj (approved):
👍

55728:
Thank you for the review! @nbulaj 😊

PR #1944, opened 2026-09-19
Add opt-in to keep the previous access token usable after a refresh

### LINKED ISSUE #1730: There is no way to refresh an access token without revoking the previous access token
### Steps to reproduce

Right now there appears to be two possible behaviors when a refresh token is refreshed.

1. If the `previous_refresh_token` column exists then the refresh token / access token are both revoked when the `TokenInfoController::show` action is requested for the first time.
   - It would help if the comment describing this behavior were more specific about what ["used"](https://github.com/doorkeeper-gem/doorkeeper/blob/e7f88ae87e821baa878a8653249829664acdbd21/lib/generators/doorkeeper/templates/migration.rb.erb#L67-L70) means.
2. If it does not exist both the refresh and access token are revoked immediately.

### Desired behavior

I was wondering why we revoke both the refresh and access token at the same time when revoked immediately.  I would have expected that the refresh token would be revoked, but that the old access token would remain valid until it expired.  I had assumed I missed something in the [RFC](https://datatracker.ietf.org/doc/html/rfc6749) about this but could not find anything.  I would try and hack this in, but since refresh tokens and access tokens use the same model + column for revocation it was not straight forward.

55728:
## Summary

Closes #1730.

A refresh token lives on the same `oauth_access_tokens` record as the access token it was issued with, and both share `revoked_at`. Exchanging the refresh token therefore also revokes the previous access token (immediately, or on the first use of the new access token when the `previous_refresh_token` column exists). RFC 6749 §6 only speaks of revoking the old *refresh* token, and clients that refresh ahead of expiry, or from several processes, get requests that are still in flight with the previous access token rejected.

This PR lets the refresh token be revoked on its own, following the same additive-column pattern as `previous_refresh_token`, `resource` and `refresh_token_scopes` (#1932).

## Two-step opt-in

1. `rails generate doorkeeper:refresh_token_revoked_at` adds a nullable `refresh_token_revoked_at` datetime (new installs get it from the install migration).
2. `revoke_previous_access_token_on_refresh false` in the initializer (default `true`).

Either one alone keeps the current behavior, so running the migration never changes semantics by itself.

## How it works

- `AccessToken#refresh_token_revoked?` is `revoked? || refresh_token_revoked_at` in the past. Records revoked before the column existed already count as having their refresh token revoked, so **no backfill and no second column** are needed (the two-column concern raised in the issue thread does not apply here).
- The reader consults the column whenever it exists, whatever the option currently says: flipping the option back to `true` must not resurrect refresh tokens revoked under `false`. Only the writer (`AccessToken#revoke_refresh_token`) depends on the option.
- `RefreshTokenRequest` (immediate mode) and `AccessToken#revoke_previous_refresh_token!` (revoke-on-use mode) go through `revoke_refresh_token`. The #1787 grace period is untouched.
- `RefreshTokenRequest`, `TokenIntrospection` and `RevocableRefreshToken` ask `refresh_token_revoked?` for a presented refresh token:
  - introspection reports the refresh token inactive and the access token of the same record active (RFC 7662 §2.2 describes the presented token);
  - the revocation endpoint treats a refresh token revoked on its own as an invalid token (RFC 7009 §2.2) instead of revoking the live access token with it. Revocation itself stays per record (RFC 7009 §2.1).
- `reuse_access_token`: a record whose refresh token was revoked on its own still has `revoked_at: nil`, so it stayed among the reuse candidates and the token response would have handed the client a dead refresh token. `find_or_create_for` now skips it.
- `doorkeeper:db:cleanup:expired_tokens` also removes expired records whose refresh token was revoked on its own. With `revoked_at` empty and a `refresh_token` present, these records were not covered by either cleanup task.

> [!IMPORTANT]
>
> ## Compatibility
> 
> - Default behavior is unchanged with or without the column.
> - `#revoked?`, `#accessible?`, `#revoke` and `.revoke_all_for` keep their meaning (the whole record / the access token).
> - The Sequel and MongoDB extensions ship their own access token mixins; callers fall back to `revoked?` / `revoke` for models that do not implement the new methods, as #1932 did.

## Known limitation (documented in the README)

Doorkeeper does not link the records of one refresh chain, so revoking the *current* refresh token does not reach a previous access token that was kept alive; it stays usable until it expires. RFC 7009 §2.1 says the server SHOULD invalidate access tokens based on the same grant, which would need a chain/family identifier (#1719). The README recommends a short `access_token_expires_in` when opting in. Dropping the column forgets refresh-only revocations; the README says to revoke those records first.


55728:
Hey @nbulaj 👋

Converting this to draft to lighten your review queue — I know there are a few PRs from me stacked up right now and I don't want to overwhelm you. The code is ready whenever you have bandwidth; I'll mark it ready for review again when the timing feels right.

No rush at all! 😊

PR #1946, opened 2026-09-19
Track refresh token families to revoke a whole refresh chain at once

### LINKED ISSUE #1945: Add refresh token chain tracking to support grant-level revocation (RFC 7009 §2.1)
## Background

#1944 will add an opt-in to keep the previous access token usable after a refresh. This means multiple access tokens from the same refresh chain can now be alive at the same time, which is the intended behavior.

However, Doorkeeper currently has no way to link the records of one refresh chain together. When a refresh token is revoked via the revocation endpoint, the previous access tokens that were kept alive are not reachable — they remain usable until they expire.

RFC 7009 §2.1 says the authorization server SHOULD invalidate access tokens based on the same authorization grant.

## Proposal

Add a nullable `refresh_token_family_id` column to `oauth_access_tokens`, following the same additive-column pattern as `previous_refresh_token`, `refresh_token_scopes` (#1932) and `refresh_token_revoked_at` (#1944).

- Assign a family ID when a token is first issued.
- Carry it forward on each refresh.
- On revocation of a refresh token, revoke all records sharing the same family ID.

This is an additive change — without the column, the current behavior is preserved.

Related: #1719 (refresh token reuse detection builds on the same family ID infrastructure)
Related: #1730

55728:
## Summary

Closes #1945.

Every refresh creates a new `oauth_access_tokens` record, and nothing links the records of one refresh chain together. Revoking a refresh token at the revocation endpoint therefore revokes the record of the presented token only, while RFC 7009 §2.1 says:

> If the particular token is a refresh token and the authorization server supports the revocation of access tokens, then the authorization server SHOULD also invalidate all access tokens based on the same authorization grant.

During the #1787 grace period (`previous_refresh_token` column), the previous record lives until the new access token is first used. With #1944 (`revoke_previous_access_token_on_refresh false`), every previous access token that has not expired yet remains usable, which is the limitation #1944 documents.

## What changes

A nullable `refresh_token_family_id` column on `oauth_access_tokens`, following the additive-column pattern of `previous_refresh_token`, `resource`, `refresh_token_scopes` (#1932) and `refresh_token_revoked_at` (#1944): part of the install migration, added to existing installs by `rails generate doorkeeper:refresh_token_family_id`, detected with `column_names.include?`. The generated migration also adds a (non-unique) index, since revocation looks records up by this column.

- **Assignment**: `AccessToken#generate_refresh_token` starts a family (`SecureRandom.uuid`) when a refresh token is issued outside the refresh grant. Doing it there, like `refresh_token_scopes`, covers the authorization code and password grants as well as host apps that create the record themselves. It is kept across validations of the same new record, and an explicitly assigned value wins.
- **Carry-over**: `RefreshTokenRequest` copies the family of the presented refresh token onto the record it creates.
- **No backfill**: a refresh token that predates the column joins a family the next time it is refreshed (`AccessToken#ensure_refresh_token_family_id!`). The identifier is written with a conditional `UPDATE ... WHERE refresh_token_family_id IS NULL` and read back, so concurrent refreshes of the same token during the grace period all continue one family.
- **Revocation**: `RevocableRefreshToken#revoke` goes through the new `AccessToken#revoke_refresh_token_family`, which revokes the record and every unrevoked record of its family, whichever refresh token of the chain is presented. `application_id` is part of the condition, so a family identifier can never reach another client's tokens, whatever a host app stores in the column.

> [!IMPORTANT]
>
> ## Compatibility
> 
> - Without the column nothing changes. Records without a family revoke alone, as before.
> - Revoking an *access* token, `AccessToken#revoke` and `AccessToken.revoke_all_for` are unchanged (RFC 7009 §2.1 makes the access token case a MAY).
> - The reader and writer of the attribute ignore a missing column (same request as on #1932 for `refresh_token_scopes=`).
> - The Sequel and MongoDB extensions ship their own access token mixins; callers fall back to `#revoke` for models that do not implement the new methods.
> - `RefreshTokenRequest#create_access_token` was at the `Metrics/AbcSize` limit, so the two attributes a refresh carries along the chain (`refresh_token_scopes` from #1932 and the family) moved into one private `refresh_chain_attributes`. No behavior change.

## Relation to other work

- #1944: independent, based on `main`. Once #1944 is merged this branch needs a rebase (conflicts are all "both sides appended at the same anchor"); I checked the combination locally: with `revoke_previous_access_token_on_refresh false`, revoking the current refresh token now reaches the previous access tokens that were kept alive. The "Things to keep in mind" bullet in #1944's README section should then point to this feature.
- #1719: refresh token reuse detection needs the same family identifier; this PR only adds the infrastructure and the revocation endpoint behavior.


55728:
Closing to reopen from an upstream branch (instead of a fork) for easier branch management.

PR #1950, opened 2026-09-21
Clarify that `fallback: :plain` is a migration-period setting

55728:
The `fallback:` option of `hash_token_secrets` / `hash_application_secrets` keeps values written before hashing was enabled valid while they migrate, which makes it a setting for the migration period. Nothing currently says so.

The generated initializer tells users upgrading that they "will probably want to enable the fallback to plain tokens", and then never mentions taking it out again. Nothing expires it either, so an installation that finished migrating years ago keeps it configured without anything pointing that out.

That matters most for `:plain`. `find_by_fallback_token` looks the column up by `fallback_secret_strategy.transform_secret(input)`, which for `Plain` is the input unchanged — so while the fallback is configured, whatever the credential column holds is itself a valid credential. Against anyone who can read the column (a database dump, a read-only injection, a query written to a log), which is the threat hashing is there for, the hashing protects nothing until the fallback is removed.

## Changes

**Generated initializer** (`lib/generators/doorkeeper/templates/initializer.rb`) — the fallback section now states that the option is for the migration period and should be removed once every row is hashed; that a value found through the fallback is rewritten in the new format as it is used, so rows migrate themselves as their tokens and secrets are presented, and rows never presented again keep their old format; and that while a `:plain` fallback is configured the stored values must be protected as carefully as plaintext credentials. The `hash_token_secrets` note that suggests adding `fallback: :plain` now points at that section.

**Boot-time warning** (`Doorkeeper::Config::Validations#validate_secret_fallback_strategies`) — warns for as long as either fallback strategy is configured, following the existing `validate_*` pattern:

```
[DOORKEEPER] hash_token_secrets is configured with a fallback strategy
(Doorkeeper::SecretStoring::Plain), which is intended for the migration period only.
Remove the fallback: option once every row has been migrated to the hashed format.
While it is configured, a value stored in the token credential column is itself a
valid credential, so that column needs to be protected as carefully as a plaintext
credential.
```

Two deliberate choices:

- **The warning fires on the setting, not on the presence of unhashed rows.** Answering "are there rows left?" means querying every credential table at boot, and the point is to surface the setting itself.
- **A fallback to another hashing strategy gets the first half only.** `hash_application_secrets using: BCrypt, fallback: Sha256Hash` still transforms what it stores, so its stored values are not credentials on their own — it gets the reminder to remove the option, without the claim about the column.


nbulaj (approved):
LGTM ✅

55728:
Thank you! 👀✨

